PR #10, opened 2020-03-02
Flat type structure

### LINKED ISSUE #8: Type naming and decorations
Currently, we are trying to represent types by value. For nested types, such as pointers/structures/arrays, we have a redirection using `Storage`s. We have two problems with this:
  - only the nested types get decorated/named (since they have separate structs)
  - the decision of whether, say a pointer class should be in `Type::Pointer { class, .. }` or in the `struct PointerDeclaraion` field, is arbitrary. If we move the pointer class and array length attributes to the respective `enum Type` variants, we basically end up with `Storage<Type>` for dependent types. And if we have `Storage<Type>`, we might as well store all types in it, not just the dependent ones.

One interesting idea that I thought worth considering is changing the representation a bit based on whether the name or decorations are present. I.e. these become the signs for the IR to make it a separate entry somewhere, which leads to a separate typedef when generating MSL and such.

kvark:
Fixes #8

### DIFF
--- a/src/back/msl.rs
+++ b/src/back/msl.rs
@@ -20,7 +20,10 @@ use std::{
     },
 };
 
-use crate::{FastHashMap, FastHashSet};
+use crate::{
+    storage::Token,
+    FastHashMap, FastHashSet
+};
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct BindTarget {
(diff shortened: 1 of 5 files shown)


kvark:
TL;DR: it's a bit more verbose, but much more elegant/uniform. Also, properly supports names of constants and types. I think it's the only way to go.

PR #11, opened 2020-03-05
Fix newline handling

msiglreith:
Support also non-LF variants


kvark:
I'm stopping pushing here and moving to the PR workflow :)

PR #12, opened 2020-03-05
Require hex prefix

msiglreith:
Includes stuff from https://github.com/gfx-rs/naga/pull/11

Still not happy with that as handling of int/floats still clashes with hex values ):
0xF will be interpreted as [float]xF and stuck, not sure how to encode this information in pest


kvark:
@msiglreith thank you for the PR!
For some reason, I'm not getting a notification anywhere, and I didn't see this. Trying to figure out what's not configured correctly. In the meantime I pushed ee7882723a8b499e615bbb03387ad07d08b15d00 that addresses this. Sorry!
Also, let's talk on "#naga:matrix.org" :)

PR #16, opened 2020-03-10
[wgsl] rewrite with our own parser

### LINKED ISSUE #14: Going lower with WGSL parsing
Do we really need a parser? Especially one with compile-time code generation from an external grammar, like Pest (that we are currently using for WGSL). Technically, all we care about is working with tokens. And we also need recursive-descent parsing.

Since we are trying to gracefully handle errors, we could get the recursive descent by just trying to parse things that make sense in a context.

Problems:
  - semantics of the language separated between the Pest file and the code. We are still matching over rules, we still have to have a catch-all case, even if `unreachable!`. We might as well do it on tokens then.
  - the Pest file has different grammar from the upstream spec, and it's an extra thing to maintain
  - extra dependency, slower compiles

After speaking with @jrmuizel, I'm thinking that we could just write a simple tokenizer. Possibly using the perfect hash tables of [phf](https://github.com/sfackler/rust-phf).

kvark:
Fixes #14

Goods of this PR:
-  less dependencies (less risk, better compile times, easier to understand etc)
-  less heap allocations (everything is sub-slicing the input)
-  proper error handling instead of panics, better error messages


src/front/wgsl.rs
```diff
@@ -71,563 +201,743 @@ impl<'a> ExpressionContext<'a> {
 
     fn parse_binary(
```
jrmuizel: parse_binary_op is clearer

jrmuizel:
Mostly these seems fine and is probably better to get in sooner and iterate rather than worrying too much before landing.

kvark:
Thank you for the review, @jrmuizel !

PR #18, opened 2020-03-12
Singular, intrinsic, and derivative expressions

kvark:
This PR adds all the missing expression types to the IR. It also upgrades the WGSL lexer and expression parser to handle them.

### DIFF
--- a/src/front/wgsl.rs
+++ b/src/front/wgsl.rs
@@ -26,6 +26,8 @@ pub enum Token<'a> {
     Word(&'a str),
     Operation(char),
     LogicalOperation(char),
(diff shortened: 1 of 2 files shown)


jrmuizel:
I think it would be better to resolve the builtin functions during typecheck/name resolution and not handle them explicitly during parsing.

Other than that the stuff here seems fine.

kvark:
Thanks for the review!
That suggestion makes sense, although we aren't ready to roll it out yet (we don't parse function calls). I'll be thinking about this more. The PR has a lot of good stuff that would be great to get in now.

PR #19, opened 2020-03-12
[wgsl] Add imported function calls

kvark:


kvark:
I think this is straightforward, we should proceed. (names are subject to revise, but logic seems fine).

PR #22, opened 2020-03-14
[wgsl] control flow parsing

kvark:
This PR gets us the remaining bits required to parse "boids.wgsl" 🎉 
It also improves IR a bit:
  - adds notion of local variables to function, instead of the variable statements
  - adds loops

We can add "test-data/*.wgsl" to automated testing now.

It will probably needs some work before we are able to generate a sensible SPIR-V, but architecturally we are almost there.


kvark:
Alright, I'm going to go ahead here.

PR #24, opened 2020-03-14
Typifier skeleton

kvark:
There is a [small TODO](https://github.com/gfx-rs/naga/blob/76eb307d350e63073bd378e0d2d4fc7d2e5676f9/src/front/wgsl.rs#L526) in the current WGSL parser:
```rust
                Token::Separator('.') => {
                    let _ = lexer.next();
                    let _name = lexer.next_ident()?;
                    let expr = crate::Expression::AccessIndex {
                        base: handle,
                        index: 0, //TODO: compute from `name`
                    };
                    handle = ctx.expressions.append(expr);
                }
```

We have a base expression, and we need to access a sub-field. The IR doesn't care about names, so in case of a structure it just needs an index into the member list. The problem, however, is realizing that the base type *is* a structure. In order to compute them, we essentially need to know all the types of all the expressions.

This is where `Typifier` comes in. It's a generic processor of IR that helps to build that type information for a growing list of expressions.

### DIFF
--- a/src/front/spirv.rs
+++ b/src/front/spirv.rs
@@ -1135,6 +1135,7 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                     let type_id = self.next()?;
                     let id = self.next()?;
                     let type_lookup = self.lookup_type.lookup(type_id)?;
+                    let ty = type_lookup.handle;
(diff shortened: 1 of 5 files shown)


kvark:
Partially reviewed by @jrmuizel. Merging.

PR #26, opened 2020-03-14
General cleanup: rustfmt, clippy, unsafe

Skepfyr:
Cleaned up all the clippy lints (bar one), removed all the unsafe (!), and ran rustfmt.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ license = "MPL-2.0"
 bitflags = "1"
 fxhash = "0.2"
 log = "0.4"
+num-traits = "0.2"
 spirv = { package = "spirv_headers", version = "1" }
 
 [dev-dependencies]
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -58,7 +58,7 @@ impl<T> Handle<T> {
         }
     }
 
-    pub fn index(&self) -> usize {
+    pub fn index(self) -> usize {
         self.index as usize
     }
 }
@@ -71,6 +71,12 @@ pub struct Arena<T> {
     data: Vec<T>,
 }
 
+impl<T> Default for Arena<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl<T> Arena<T> {
     pub fn new() -> Self {
         Arena {
@@ -78,7 +84,7 @@ impl<T> Arena<T> {
         }
     }
 
-    pub fn iter<'a>(&'a self) -> impl Iterator<Item = (Handle<T>, &'a T)> {
+    pub fn iter(&self) -> impl Iterator<Item = (Handle<T>, &T)> {
         self.data
             .iter()
             .enumerate()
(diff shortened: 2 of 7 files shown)


src/back/msl.rs
```diff
@@ -74,12 +70,17 @@ enum LocationMode {
     Uniform,
 }
 
+#[derive(Debug, Clone, Copy)]
```
Skepfyr: Not entirely sure about this, it also causes clippy to get angry with `resolve_binding` as it's cheaper to take `self` rather than `&self` there now that it is copy.

src/front/spirv.rs
```diff
@@ -1344,14 +801,739 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+            } => {
+                let low = self.next()?;
+                let extended = match width {
+                    32 => f64::from(f32::from_bits(low)),
+                    64 => {
+                        inst.expect(4)?;
+                        let high = self.next()?;
+                        f64::from_bits((u64::from(high) << 32) | u64::from(low))
+                    }
+                    _ => return Err(Error::InvalidTypeWidth(u32::from(width))),
```
Skepfyr: Note this is a change in behaviour.

Skepfyr:
Removed the rustfmt changes.

src/arena.rs
```diff
@@ -119,36 +125,36 @@ mod tests {
...
         assert!(t1 == t2);
         assert!(arena[t1] == arena[t2])
     }
 
     #[test]
     fn fetch_or_append_unique() {
-        let mut arena: Arena<f64> = Arena::new();
-        let t1 = arena.fetch_or_append(std::f64::NAN);
-        let t2 = arena.fetch_or_append(std::f64::NAN);
+        let mut arena: Arena<u8> = Arena::new();
```
kvark: what's the motivation for these test changes?

src/front/spirv.rs
```diff
@@ -128,16 +120,12 @@ fn map_vector_size(word: spirv::Word) -> Result<crate::VectorSize, Error> {
...
     }
 }
 
 fn map_storage_class(word: spirv::Word) -> Result<spirv::StorageClass, Error> {
-    if word > LAST_KNOWN_STORAGE_CLASS as u32 {
-        Err(Error::UnsupportedStorageClass(word))
-    } else {
-        Ok(unsafe { std::mem::transmute(word) })
-    }
+    spirv::StorageClass::from_u32(word).ok_or(Error::UnsupportedStorageClass(word))
```
kvark: woot! that's a great way to convert enums :) 👍

src/front/spirv.rs
```diff
@@ -707,618 +681,36 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
-                    // done
-                    let handle = module.functions.append(fun);
-                    self.lookup_function.insert(fun_id, handle);
-                    self.lookup_expression.clear();
-                    self.lookup_sampled_image.clear();
-                }
-                _ => return Err(Error::UnsupportedInstruction(self.state, inst.op))
-                //TODO
-            }
+                Op::Capability => self.parse_capability(inst),
```
kvark: is that something Clippy complained about?

src/front/wgsl.rs
```diff
@@ -340,6 +338,7 @@ pub struct ParseError<'a> {
     pub pos: (usize, usize),
 }
 
+#[derive(Debug, Default)]
```
kvark: why do we need default here?

src/arena.rs
Skepfyr: Float equality checxking is a thorny topic and it didn't seem to benefit these tests (I'll happily be corrected on that point though). The reason I noticed was because clippy was angry (although it can just be silenced which would be perfectly valid here).

src/front/spirv.rs
Skepfyr: It complained this function was too complicated, this seemed like the most sensible transformation to fix that.

src/front/wgsl.rs
Skepfyr: So this is clippy's new_without_default lint, while there is a new method that takes no args I don't think there's any harm in having a Default impl, and it does make some things slightly nicer.

kvark: hm, I'm not certain that this is a good idea. `Default` should be used for things that are truly "default", not just *any* thing with this format of constructor. I.e. the arguments of a constructor don't imply a specific semantics that `Default` does.
Can we have that change reverted and the `new_without_default` lint disabled?

src/arena.rs
kvark: makes sense 👍

src/front/wgsl.rs
Skepfyr: Yeah, I've always been slightly dubious of that lint myself.

Cargo.toml
```diff
@@ -13,6 +13,7 @@ license = "MPL-2.0"
 bitflags = "1"
 fxhash = "0.2"
 log = "0.4"
+num-traits = "0.2"
```
kvark: where is this used?

kvark: nvm, see it now :)

PR #28, opened 2020-03-15
Remove Expression::Mul

kvark:
We now have a more powerful `Binary` expression type.

This PR also adds quad -> MSL conversion to the tests. This is strictly better than running it in the example, since the example isn't supposed to have any shader-specific information. Yet, the binding mapping is required. So the example would need to support some sort of a RON input with the binding map, but the tests can just have all of it specified in the code.


kvark:
I think this is straightforward, please shout out otherwise!

PR #30, opened 2020-03-16
[msl] operators, array access, and built-in calls

kvark:
This PR upgrades the Metal backend to the point of producing *a* result for "boids" without panicing. There is still work to be done before this result is valid and can be consumed by Metal SDK. In the meantime, the tests now do the check for "boids" to ensure we can always convert.


kvark:
I don't think this is controversial: no architectural decisions taken, just expanding the current support.

PR #36, opened 2020-03-20
Boids running on MSL

kvark:
Made more progress in MSL backend, to the point of successfully generating the [boids.metal](https://github.com/gfx-rs/naga/files/4357652/boids.metal.txt) without warnings or errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .PHONY: all clean
+.SECONDARY: boids.metal quad.metal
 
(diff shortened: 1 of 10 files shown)


kvark:
This is ready for review. I want to refactor the interface module a bit more before merging.

kvark:
I think that looks great?
Would be even better to have more eyes on this.
Even after landing :)

PR #37, opened 2020-03-20
Move the global usage into the IR function

### LINKED ISSUE #35: Explicit interface of entry points
SPIR-V lists all the inputs and outputs for each entry point. The question is whether it's a good idea to represent this somehow in the IR. See https://github.com/gpuweb/gpuweb/issues/592 for WGSL discussion.

Pros for specifying:
  - match SPIR-V
  - easier to translate from IR to anything, since we know the interface

Cons for specifying:
  - more verbose and redundant
  - incomplete - stuff like uniform or storage buffers are also technically the interface, so we still need to go and scan the uses of the globals

Another thought I had is that it doesn't make sense to specify that interface *per entry point*. It seems like it would make more sense to have for the function.

kvark:
Fixes #35

### DIFF
--- a/src/back/msl.rs
+++ b/src/back/msl.rs
@@ -22,8 +22,7 @@ use std::{
 
 use crate::{
     arena::Handle,
-    proc::GlobalUse,
-    FastHashMap, FastHashSet
+    FastHashMap,
 };
 
 /// Expect all the global variables to have a pointer type,
@@ -226,7 +225,7 @@ impl AsName for Option<String> {
 struct TypedGlobalVariable<'a> {
     module: &'a crate::Module,
     handle: crate::Handle<crate::GlobalVariable>,
-    usage: GlobalUse,
+    usage: crate::GlobalUse,
 }
 impl Display for TypedGlobalVariable<'_> {
     fn fmt(&self, formatter: &mut Formatter<'_>) -> Result<(), FmtError> {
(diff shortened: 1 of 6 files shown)


src/back/msl.rs
```diff
@@ -819,8 +813,10 @@ impl<W: Write> Writer<W> {
 
                 if em != spirv::ExecutionModel::GLCompute {
                     writeln!(self.out, "struct {} {{", location_input_name)?;
-                    for &handle in var_inputs.iter() {
-                        let var = &module.global_variables[handle];
+                    for ((handle, var), &usage) in module.global_variables.iter().zip(&fun.global_usage) {
```
Napokue: `module.global_variables.iter().zip(&fun.global_usage)` is used four times in this function. Isn't it possible to do this once and put it into a variable, to reduce iterations through `global_variables` and prevent to zip four times?

kvark: we could do it, yeah, although it's not going to save us any lines :)

Napokue: Less lines are not always better :) I was pointing it out more as a performance optimisation. I am not sure if there is a need for it though.

kvark: ah ok. There isn't any performance impact here. It's just an iteration over a few dozen of globals, super light.

Napokue: Sure, then it is all fine by me, rest looks good 👍

kvark:
Appears that both @jrmuizel and @Napokue looked at this. Thank you!

PR #38, opened 2020-03-21
Use `NonZeroU32` for handle index

### LINKED ISSUE #21: Implement NonZero for Handle
That would make all the `Option<Handle<T>>` to be free and not pollute the data cache.

GabrielMajeri:
This PR makes a `Handle`'s index to be stored using `NonZeroU32`, by using the 1-based position in the `Arena`. Fixes #21.

The `Handle.index(self)` function has been changed to return the 0-based index, otherwise it would break the code with look-up tables for indexes.

I also ran `cargo fmt` on the file, which changed some of the formatting.

### DIFF
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -1,8 +1,4 @@
-use std::{
-    fmt,
-    hash,
-    marker::PhantomData,
-};
+use std::{fmt, hash, marker::PhantomData};
 
 /// An unique index in the arena array that a handle points to.
 ///
@@ -79,9 +75,7 @@ impl<T> Default for Arena<T> {
 
 impl<T> Arena<T> {
     pub fn new() -> Self {
-        Arena {
-            data: Vec::new(),
-        }
+        Arena { data: Vec::new() }
     }
 
     pub fn len(&self) -> usize {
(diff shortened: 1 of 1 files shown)


kvark:
Thank you for the well-needed feature!
Please don't run cargo fmt yet, we haven't figured out the right options.

src/arena.rs
```diff
@@ -58,8 +54,10 @@ impl<T> Handle<T> {
         }
     }
 
+    /// Returns the zero-based index of this handle.
     pub fn index(self) -> usize {
-        self.index as usize
+        let index = self.index.get() - 1;
```
kvark: would it be simpler to just start an `Arena` with on dummy element all the time?

GabrielMajeri: It could work, but since `Arena` is generic I'm not sure what value to give to the dummy element?

kvark: could require `T: Default`

GabrielMajeri: I've tried, started adding `derive(Default)` to various types in this crate, but ended up requiring that some `spirv_headers` types implement `Default`.

kvark: If it's too difficult, it may be ok to just land what you had originally.
Technically speaking, you can always `impl Default` manually for our structs in order to be able to use that approach. Please make a judgement call for whether there is too much boilerplace, and we'll go from there.

GabrielMajeri:
I've looked into implementing `Default` for most of the crates types, and for many types it seems a bit forced.

From a simplicity standpoint, having a few `+/- 1` in `Arena` to keep the indexes non-zero is simpler.

PR #40, opened 2020-03-26
Add support for ConstantComposite

### LINKED ISSUE #39: Add support for ConstantComposite
As I was testing to run this following GLSL code in SPIR-V format using the SPIR-V front-end I got an UnsupportedInstruction error for OpConstantComposite.

This is the fragment part of the quad.wgsl part in GLSL
```
#version 450
layout(location = 0) in vec2 a_uv;
layout(location = 0) out vec4 o_color;

void main()
{
    o_color = vec4(1, 0, 0, 1);
    return;
}
```

Eventually the o_color OpStore instruction is `OpStore %o_color %12` where %12 is the OpConstComposite: `%12 = OpConstantComposite %v4float %float_1 %float_0 %float_0 %float_1`.

Napokue:
Fix #39

### DIFF
--- a/src/back/msl.rs
+++ b/src/back/msl.rs
@@ -403,29 +403,7 @@ impl<W: Write> Writer<W> {
                 }
             }
             crate::Expression::Constant(handle) => {
-                let kind = match module.constants[handle].inner {
-                    crate::ConstantInner::Sint(value) => {
-                        write!(self.out, "{}", value)?;
-                        crate::ScalarKind::Sint
-                    }
-                    crate::ConstantInner::Uint(value) => {
-                        write!(self.out, "{}", value)?;
-                        crate::ScalarKind::Uint
-                    }
-                    crate::ConstantInner::Float(value) => {
-                        write!(self.out, "{}", value)?;
-                        if value.fract() == 0.0 {
-                            self.out.write_str(".0")?;
-                        }
-                        crate::ScalarKind::Float
-                    }
-                    crate::ConstantInner::Bool(value) => {
-                        write!(self.out, "{}", value)?;
-                        crate::ScalarKind::Bool
-                    }
-                };
-                let width = 32; //TODO: not sure how to get that...
-                Ok(MaybeOwned::Owned(crate::TypeInner::Scalar { kind, width }))
+                self.put_constant(handle, module)
             }
             crate::Expression::Compose { ty, ref components } => {
                 let inner = &module.types[ty].inner;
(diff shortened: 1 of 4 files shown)


src/back/msl.rs
```diff
@@ -403,29 +403,52 @@ impl<W: Write> Writer<W> {
                 }
             }
             crate::Expression::Constant(handle) => {
-                let kind = match module.constants[handle].inner {
+                let width = 32; //TODO: not sure how to get that...
+                match module.constants[handle].inner {
                     crate::ConstantInner::Sint(value) => {
                         write!(self.out, "{}", value)?;
-                        crate::ScalarKind::Sint
+                        Ok(MaybeOwned::Owned(
```
kvark: We don't need to construct this type and return it from the `match`. We only need to do `write!()` stuff in the match arms. After the match, we could get the type by looking at `module.constants[handle].ty` without guessing.

src/back/msl.rs
```diff
@@ -403,29 +403,52 @@ impl<W: Write> Writer<W> {
...
+                                kind: crate::ScalarKind::Bool,
+                                width,
+                            }
+                        ))
                     }
-                };
-                let width = 32; //TODO: not sure how to get that...
-                Ok(MaybeOwned::Owned(crate::TypeInner::Scalar { kind, width }))
+                    crate::ConstantInner::Composite(handle, _) => {
+                        let inner = &module.types[handle].inner;
```
kvark: we need to actually write down the construction of a type, i.e. something like that:
```rust
crate::ConstantInner::Composite(ref members) => {
  let ty_name = module.types[const.ty].name.or_index(const.ty);
  write!(self.out, "{}("), ty_name)?;
  for (i, &member) in members.iter().enumerate() {
    if i != 0 {
      write!(self.out, ",")?;
    }
    self.put_constant(member); // there isn't a method like this, we'd need to move the relevant part of `put_expression` into it.
  }
  wrote!(self.out, ")")?;
}
```

src/lib.rs
```diff
@@ -97,17 +97,7 @@ pub enum ConstantInner {
...
-impl ConstantInner {
-    pub fn scalar_kind(&self) -> ScalarKind {
-        match *self {
-            ConstantInner::Sint(_) => ScalarKind::Sint,
-            ConstantInner::Uint(_) => ScalarKind::Uint,
-            ConstantInner::Float(_) => ScalarKind::Float,
-            ConstantInner::Bool(_) => ScalarKind::Bool,
-        }
-    }
+    Composite(Handle<Type>, Vec<Handle<Constant>>),
```
kvark: we don't need the type here, it's already stored in `Constant::ty`

src/front/spirv.rs
```diff
@@ -1305,6 +1306,41 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+    ) -> Result<(), Error> {
+        self.switch(ModuleState::Type, inst.op)?;
+        inst.expect_at_least(3)?;
+        let type_id = self.next()?;
+        let type_lookup = self.lookup_type.lookup(type_id)?;
+        let ty = type_lookup.handle;
+
+        let id = self.next()?;
+
+        let mut constituents: Vec<Handle<crate::Constant>> = vec![];
```
kvark: shouldn't need to specify the type here

kvark: also, let's use `with_capacity` constructor

src/back/msl.rs
```diff
@@ -403,29 +403,52 @@ impl<W: Write> Writer<W> {
...
+                        Ok(MaybeOwned::Borrowed(inner))
```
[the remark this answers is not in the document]
kvark: not at all

src/back/msl.rs
```diff
@@ -601,6 +579,46 @@ impl<W: Write> Writer<W> {
...
 
+    fn put_constant<'a>(
+        &mut self,
+        handle: Handle<crate::Constant>,
+        module: &'a crate::Module,
+    ) -> Result<MaybeOwned<'a, crate::TypeInner>, Error> {
+        let constant = &module.constants[handle];
+        let ty = &module.types[constant.ty];
+
+        match &constant.inner {
```
kvark: please match without a reference here, instead put `ref` in non-movable fields binding, e.g. `constituents`

PR #46, opened 2020-04-13
Spirv backend

Napokue:
- Quad example compiles from both a SPIR-V and WGSL front-end

### DIFF
--- a/examples/convert.rs
+++ b/examples/convert.rs
@@ -30,7 +30,7 @@ fn main() {
         return;
     } else if args[1].ends_with(".spv") {
         let input = fs::read(&args[1]).unwrap();
-        naga::front::spirv::parse_u8_slice(&input).unwrap()
+        naga::front::spv::parse_u8_slice(&input).unwrap()
     } else if args[1].ends_with(".wgsl") {
         let input = fs::read_to_string(&args[1]).unwrap();
         naga::front::wgsl::parse_str(&input).unwrap()
@@ -71,6 +71,27 @@ fn main() {
         };
         let msl = msl::write_string(&module, options).unwrap();
         fs::write(&args[2], msl).unwrap();
+    } else if args[2].ends_with(".spv") {
+        use naga::back::spv;
+
+        let debug_flag = args.get(3).map_or(spv::WriterFlags::DEBUG, |arg| {
+            if arg.parse().unwrap() {
+                spv::WriterFlags::DEBUG
+            } else {
+                spv::WriterFlags::NONE
+            }
+        });
+
+        let spv = spv::writer::Writer::new(&module.header, debug_flag).write(&module);
+
+        let bytes = spv
+            .iter()
+            .fold(Vec::with_capacity(spv.len() * 4), |mut v, w| {
+                v.extend_from_slice(&w.to_le_bytes());
+                v
+            });
+
+        fs::write(&args[2], bytes.as_slice()).unwrap();
     } else {
         panic!("Unknown output: {:?}", args[2]);
     }
--- a/src/back/mod.rs
+++ b/src/back/mod.rs
@@ -1,1 +1,2 @@
 pub mod msl;
+pub mod spv;
--- a/src/back/spv/mod.rs
+++ b/src/back/spv/mod.rs
@@ -1,0 +1,42 @@
+mod layout;
+pub mod writer;
+
+use spirv::*;
+
+bitflags::bitflags! {
+    pub struct WriterFlags: u32 {
+        const NONE = 0x0;
+        const DEBUG = 0x1;
+    }
+}
+
+struct PhysicalLayout {
+    magic_number: Word,
+    version: Word,
+    generator: Word,
+    bound: Word,
+    instruction_schema: Word,
+}
+
+#[derive(Default)]
+pub(crate) struct LogicalLayout {
+    capabilities: Vec<Word>,
+    extensions: Vec<Word>,
+    ext_inst_imports: Vec<Word>,
+    memory_model: Vec<Word>,
+    entry_points: Vec<Word>,
+    execution_modes: Vec<Word>,
+    debugs: Vec<Word>,
+    annotations: Vec<Word>,
+    declarations: Vec<Word>,
+    function_declarations: Vec<Word>,
+    function_definitions: Vec<Word>,
+}
+
+struct Instruction {
+    op: Op,
+    wc: u32,
+    type_id: Option<Word>,
+    result_id: Option<Word>,
+    operands: Vec<Word>,
+}
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 **/*.rs.bk
 Cargo.lock
 .DS_Store
+.idea
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Rust            |                    |       |
 
 Back-end        |       Status       | Notes |
 --------------- | ------------------ | ----- |
-SPIR-V (binary) |                    |       |
+SPIR-V (binary) | :construction:     |       |
 WGSL            |                    |       |
 Metal           | :construction:     |       |
 HLSL            |                    |       |
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -2,7 +2,7 @@ use std::{fmt, hash, marker::PhantomData, num::NonZeroU32};
 
 /// An unique index in the arena array that a handle points to.
 ///
-/// This type is independent of `spirv::Word`. `spirv::Word` is used in data
+/// This type is independent of `spv::Word`. `spv::Word` is used in data
 /// representation. It holds a SPIR-V and refers to that instruction. In
 /// structured representation, we use Handle to refer to an SPIR-V instruction.
 /// `Index` is an implementation detail to `Handle`.
--- a/src/back/spv/layout.rs
+++ b/src/back/spv/layout.rs
@@ -1,0 +1,85 @@
+use crate::back::spv::{Instruction, LogicalLayout, PhysicalLayout};
+use spirv::*;
+use std::iter;
+
+impl PhysicalLayout {
+    pub(crate) fn new(header: &crate::Header) -> Self {
+        let version: Word = ((header.version.0 as u32) << 16)
+            | ((header.version.1 as u32) << 8)
+            | header.version.2 as u32;
+
+        PhysicalLayout {
+            magic_number: MAGIC_NUMBER,
+            version,
+            generator: header.generator,
+            bound: 0,
+            instruction_schema: 0x0u32,
+        }
+    }
+
+    pub(crate) fn in_words(&self, sink: &mut impl Extend<Word>) {
+        sink.extend(iter::once(self.magic_number));
+        sink.extend(iter::once(self.version));
+        sink.extend(iter::once(self.generator));
+        sink.extend(iter::once(self.bound));
+        sink.extend(iter::once(self.instruction_schema));
+    }
+}
+
+impl LogicalLayout {
+    pub(crate) fn in_words(&self, sink: &mut impl Extend<Word>) {
+        sink.extend(self.capabilities.iter().cloned());
+        sink.extend(self.extensions.iter().cloned());
+        sink.extend(self.ext_inst_imports.iter().cloned());
+        sink.extend(self.memory_model.iter().cloned());
+        sink.extend(self.entry_points.iter().cloned());
+        sink.extend(self.execution_modes.iter().cloned());
+        sink.extend(self.debugs.iter().cloned());
+        sink.extend(self.annotations.iter().cloned());
+        sink.extend(self.declarations.iter().cloned());
+        sink.extend(self.function_declarations.iter().cloned());
+        sink.extend(self.function_definitions.iter().cloned());
+    }
+}
+
+impl Instruction {
+    pub(crate) fn new(op: Op) -> Self {
+        Instruction {
+            op,
+            wc: 1, // Always start at 1 for the first word (OP + WC),
+            type_id: None,
+            result_id: None,
+            operands: vec![],
+        }
+    }
+
+    pub(crate) fn set_type(&mut self, id: Word) {
+        assert!(self.type_id.is_none(), "Type can only be set once");
+        self.type_id = Some(id);
+        self.wc += 1;
+    }
+
+    pub(crate) fn set_result(&mut self, id: Word) {
+        assert!(self.result_id.is_none(), "Result can only be set once");
+        self.result_id = Some(id);
+        self.wc += 1;
+    }
+
+    pub(crate) fn add_operand(&mut self, operand: Word) {
+        self.operands.push(operand);
+        self.wc += 1;
+    }
+
+    pub(crate) fn add_operands(&mut self, operands: Vec<Word>) {
+        for operand in operands.into_iter() {
+            self.add_operand(operand)
+        }
+    }
+
+    pub(crate) fn to_words(&self, sink: &mut impl Extend<Word>) {
+        sink.extend(Some((self.wc << 16 | self.op as u32) as u32));
+        sink.extend(self.type_id);
+        sink.extend(self.result_id);
+        sink.extend(self.operands.iter().cloned());
+    }
+}
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1,0 +1,1013 @@
+/*! Standard Portable Intermediate Representation (SPIR-V) backend !*/
+use crate::back::spv::{Instruction, LogicalLayout, PhysicalLayout, WriterFlags};
+use crate::{FastHashMap, FastHashSet};
+use spirv::*;
+
+trait LookupHelper<T> {
+    type Target;
+    fn lookup_id(&self, handle: crate::Handle<T>) -> Option<Word>;
+    fn lookup_handle(&self, word: Word) -> Option<crate::Handle<T>>;
+}
+
+impl<T> LookupHelper<T> for FastHashMap<Word, crate::Handle<T>> {
+    type Target = T;
+
+    fn lookup_id(&self, handle: crate::Handle<T>) -> Option<Word> {
+        let mut word = None;
+        for (k, v) in self.iter() {
+            if *v == handle {
+                word = Some(*k);
+                break;
+            }
+        }
+        word
+    }
+
+    fn lookup_handle(&self, word: u32) -> Option<crate::Handle<T>> {
+        let mut handle = None;
+        for (k, v) in self.iter() {
+            if *k == word {
+                handle = Some(*v);
+                break;
+            }
+        }
+        handle
+    }
+}
+
+#[derive(Debug, PartialEq)]
+struct LookupFunctionType {
+    parameter_type_ids: Vec<Word>,
+    return_type_id: Word,
+}
+
+pub struct Writer {
+    physical_layout: PhysicalLayout,
+    logical_layout: LogicalLayout,
+    id_count: u32,
+    capabilities: FastHashSet<Capability>,
+    debugs: Vec<Instruction>,
+    annotations: Vec<Instruction>,
+    writer_flags: WriterFlags,
+    void_type: Option<u32>,
+    lookup_type: FastHashMap<Word, crate::Handle<crate::Type>>,
+    lookup_function: FastHashMap<Word, crate::Handle<crate::Function>>,
+    lookup_function_type: FastHashMap<Word, LookupFunctionType>,
+    lookup_constant: FastHashMap<Word, crate::Handle<crate::Constant>>,
+    lookup_global_variable: FastHashMap<Word, crate::Handle<crate::GlobalVariable>>,
+}
+
+impl Writer {
+    pub fn new(header: &crate::Header, writer_flags: WriterFlags) -> Self {
+        Writer {
+            physical_layout: PhysicalLayout::new(header),
+            logical_layout: LogicalLayout::default(),
+            id_count: 0,
+            capabilities: FastHashSet::default(),
+            debugs: vec![],
+            annotations: vec![],
+            writer_flags,
+            void_type: None,
+            lookup_type: FastHashMap::default(),
+            lookup_function: FastHashMap::default(),
+            lookup_function_type: FastHashMap::default(),
+            lookup_constant: FastHashMap::default(),
+            lookup_global_variable: FastHashMap::default(),
+        }
+    }
+
+    fn generate_id(&mut self) -> Word {
+        self.id_count += 1;
+        self.id_count
+    }
+
+    fn bytes_to_words(&self, bytes: &[u8]) -> Vec<Word> {
+        bytes
+            .chunks(4)
+            .map(|chars| chars.iter().rev().fold(0u32, |u, c| (u << 8) | *c as u32))
+            .collect()
+    }
+
+    fn string_to_words(&self, input: &str) -> Vec<Word> {
+        let bytes = input.as_bytes();
+        let mut words = self.bytes_to_words(bytes);
+
+        if bytes.len() % 4 == 0 {
+            // nul-termination
+            words.push(0x0u32);
+        }
+
+        words
+    }
+
+    fn try_add_capabilities(&mut self, capabilities: &[Capability]) {
+        for capability in capabilities.iter() {
+            self.capabilities.insert(*capability);
+        }
+    }
+
+    fn instruction_capability(&self, capability: Capability) -> Instruction {
+        let mut instruction = Instruction::new(Op::Capability);
+        instruction.add_operand(capability as u32);
+        instruction
+    }
+
+    fn instruction_ext_inst_import(&mut self, name: &str) -> Instruction {
+        let mut instruction = Instruction::new(Op::ExtInstImport);
+        let id = self.generate_id();
+        instruction.set_result(id);
+        instruction.add_operands(self.string_to_words(name));
+        instruction
+    }
+
+    fn instruction_memory_model(&mut self) -> Instruction {
+        let mut instruction = Instruction::new(Op::MemoryModel);
+        let addressing_model = AddressingModel::Logical;
+        let memory_model = MemoryModel::GLSL450;
+        self.try_add_capabilities(addressing_model.required_capabilities());
+        self.try_add_capabilities(memory_model.required_capabilities());
+
+        instruction.add_operand(addressing_model as u32);
+        instruction.add_operand(memory_model as u32);
+        instruction
+    }
+
+    fn instruction_entry_point(
+        &mut self,
+        entry_point: &crate::EntryPoint,
+        ir_module: &crate::Module,
+    ) -> Instruction {
+        let mut instruction = Instruction::new(Op::EntryPoint);
+        let function_id = self
+            .lookup_function
+            .lookup_id(entry_point.function)
+            .unwrap();
+
+        instruction.add_operand(entry_point.exec_model as u32);
+        instruction.add_operand(function_id);
+
+        if self.writer_flags.contains(WriterFlags::DEBUG) {
+            let mut debug_instruction = Instruction::new(Op::Name);
+            debug_instruction.set_result(function_id);
+            debug_instruction.add_operands(self.string_to_words(entry_point.name.as_str()));
+            self.debugs.push(debug_instruction);
+        }
+
+        instruction.add_operands(self.string_to_words(entry_point.name.as_str()));
+
+        let function = &ir_module.functions[entry_point.function];
+        for ((handle, _), &usage) in ir_module
+            .global_variables
+            .iter()
+            .zip(&function.global_usage)
+        {
+            if usage.contains(crate::GlobalUse::STORE) || usage.contains(crate::GlobalUse::LOAD) {
+                let id = self.get_global_variable_id(
+                    &ir_module.types,
+                    &ir_module.global_variables,
+                    handle,
+                );
+                instruction.add_operand(id);
+            }
+        }
+
+        self.try_add_capabilities(entry_point.exec_model.required_capabilities());
+        match entry_point.exec_model {
+            ExecutionModel::Vertex => {}
+            ExecutionModel::Fragment => {
+                let execution_mode = ExecutionMode::OriginUpperLeft;
+                self.try_add_capabilities(execution_mode.required_capabilities());
+
+                let mut execution_mode_instruction = Instruction::new(Op::ExecutionMode);
+                execution_mode_instruction.add_operand(function_id);
+                execution_mode_instruction.add_operand(execution_mode as u32);
+                execution_mode_instruction.to_words(&mut self.logical_layout.execution_modes);
+            }
+            _ => unimplemented!(),
+        }
+
+        instruction
+    }
+
+    fn get_type_id(
+        &mut self,
+        arena: &crate::Arena<crate::Type>,
+        handle: crate::Handle<crate::Type>,
+    ) -> Word {
+        match self.lookup_type.lookup_id(handle) {
+            Some(word) => word,
+            None => {
+                let (instruction, id) = self.instruction_type_declaration(arena, handle);
+                instruction.to_words(&mut self.logical_layout.declarations);
+                id
+            }
+        }
+    }
+
+    fn get_constant_id(
+        &mut self,
+        handle: crate::Handle<crate::Constant>,
+        ir_module: &crate::Module,
+    ) -> Word {
+        match self.lookup_constant.lookup_id(handle) {
+            Some(word) => word,
+            None => {
+                let (instruction, id) = self.instruction_constant_type(handle, ir_module);
+                instruction.to_words(&mut self.logical_layout.declarations);
+                id
+            }
+        }
+    }
+
+    fn get_global_variable_id(
+        &mut self,
+        arena: &crate::Arena<crate::Type>,
+        global_arena: &crate::Arena<crate::GlobalVariable>,
+        handle: crate::Handle<crate::GlobalVariable>,
+    ) -> Word {
+        match self.lookup_global_variable.lookup_id(handle) {
+            Some(word) => word,
+            None => {
+                let global_variable = &global_arena[handle];
+                let (instruction, id) =
+                    self.instruction_global_variable(arena, global_variable, handle);
+                instruction.to_words(&mut self.logical_layout.declarations);
+                id
+            }
+        }
+    }
+
+    fn get_function_type(
+        &mut self,
+        ty: Option<crate::Handle<crate::Type>>,
+        arena: &crate::Arena<crate::Type>,
+    ) -> Word {
+        match ty {
+            Some(handle) => self.get_type_id(arena, handle),
+            None => match self.void_type {
+                Some(id) => id,
+                None => {
+                    let id = self.generate_id();
+
+                    let mut instruction = Instruction::new(Op::TypeVoid);
+                    instruction.set_result(id);
+
+                    self.void_type = Some(id);
+                    instruction.to_words(&mut self.logical_layout.declarations);
+                    id
+                }
+            },
+        }
+    }
+
+    fn find_scalar_handle(
+        &self,
+        arena: &crate::Arena<crate::Type>,
+        kind: crate::ScalarKind,
+        width: u8,
+    ) -> crate::Handle<crate::Type> {
+        let mut scalar_handle = None;
+        for (handle, ty) in arena.iter() {
+            match ty.inner {
+                crate::TypeInner::Scalar {
+                    kind: _kind,
+                    width: _width,
+                } => {
+                    if kind == _kind && width == _width {
+                        scalar_handle = Some(handle);
+                        break;
+                    }
+                }
+                _ => continue,
+            }
+        }
+        scalar_handle.unwrap()
+    }
+
+    fn instruction_type_declaration(
+        &mut self,
+        arena: &crate::Arena<crate::Type>,
+        handle: crate::Handle<crate::Type>,
+    ) -> (Instruction, Word) {
+        let ty = &arena[handle];
+        let id = self.generate_id();
+        let mut instruction;
+
+        match ty.inner {
+            crate::TypeInner::Scalar { kind, width } => {
+                match kind {
+                    crate::ScalarKind::Sint => {
+                        instruction = Instruction::new(Op::TypeInt);
+                        instruction.set_result(id);
+                        instruction.add_operand(width as u32);
+                        instruction.add_operand(0x1u32);
+                    }
+                    crate::ScalarKind::Uint => {
+                        instruction = Instruction::new(Op::TypeInt);
+                        instruction.set_result(id);
+                        instruction.add_operand(width as u32);
+                        instruction.add_operand(0x0u32);
+                    }
+                    crate::ScalarKind::Float => {
+                        instruction = Instruction::new(Op::TypeFloat);
+                        instruction.set_result(id);
+                        instruction.add_operand(width as u32);
+                    }
+                    crate::ScalarKind::Bool => {
+                        instruction = Instruction::new(Op::TypeBool);
+                        instruction.set_result(id);
+                    }
+                }
+                self.lookup_type.insert(id, handle);
+            }
+            crate::TypeInner::Vector { size, kind, width } => {
+                let scalar_handle = self.find_scalar_handle(arena, kind, width);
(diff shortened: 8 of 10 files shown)


src/back/spirv/layout.rs
```diff
@@ -0,0 +1,216 @@
...
+        for function_definition in self.function_definitions.iter() {
+            words.push(*function_definition);
+        }
+
+        words
+    }
+}
+
+pub(crate) struct Instruction {
+    op: u32,
```
kvark, line 153: any reason this isn't `Op`?

src/back/spirv/layout.rs
```diff
@@ -0,0 +1,216 @@
...
+            op: op as u32,
+            wc: 1, // Always start at 1 for the first word (OP + WC),
+            type_id: None,
+            result_id: None,
+            operands: vec![],
+        }
+    }
+
+    pub(crate) fn set_type(&mut self, id: Word) {
+        if self.type_id.is_none() {
```
kvark, line 172: Can we instead assert that the type id is None? I don't think we expect ourselves to call this multiple times on the same instruction.

src/back/spirv/layout.rs
```diff
@@ -0,0 +1,216 @@
...
+            self.wc += 1;
+        }
+        self.type_id = Some(id);
+    }
+
+    pub(crate) fn set_result(&mut self, id: Word) {
+        if self.result_id.is_none() {
```
kvark, line 179: same here

src/back/spirv/layout.rs
```diff
@@ -0,0 +1,216 @@
...
+        }
+        self.result_id = Some(id);
+    }
+
+    pub(crate) fn add_operand(&mut self, operand: Word) {
+        self.operands.push(operand);
+        self.wc += 1;
+    }
+
+    pub(crate) fn add_operands(&mut self, operands: Vec<Word>) {
```
kvark, line 190: should be `&[Word]`

src/back/spirv/layout.rs
```diff
@@ -0,0 +1,216 @@
...
+        for operand in operands {
+            self.add_operand(operand);
+        }
+    }
+
+    pub(crate) fn to_words(&self) -> Vec<Word> {
+        let mut words = Vec::with_capacity(self.wc as usize);
+        let wc_op = (self.wc << 16 | self.op) as u32;
+
+        words.push(wc_op);
```
kvark, line 200: putting the whole problem of heap allocations aside, all of this can be simplified like:
```rust
{
iter::once((self.wc << 16 | self.op) as u32)
  .chain(self.type_id)
  .chain(self.result_id)
  .chain(self.operands)
}
```

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
+/*! Standard Portable Intermediate Representation (SPIR-V) backend !*/
+use super::layout::{Instruction, Module};
+use crate::FastHashMap;
+use spirv::*;
+
+trait LookupHelper<T> {
+    type Target;
+    fn lookup_id(&self, handle: &crate::Handle<T>) -> Option<Word>;
+    fn lookup_handle(&self, word: &Word) -> Option<crate::Handle<T>>;
```
kvark, line 9: any reason the `&Word` here is by reference? and same for `&crate::Handle`

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                word = Some(*k);
+                break;
+            }
+        }
+        word
+    }
+
+    fn lookup_handle(&self, word: &u32) -> Option<crate::Handle<T>> {
+        let mut handle = None;
+        for (k, v) in self.iter() {
```
kvark, line 28: why not just `self.get(word)`?

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+}
+
+impl<T> LookupHelper<T> for FastHashMap<Word, crate::Handle<T>> {
+    type Target = T;
+
+    fn lookup_id(&self, handle: &crate::Handle<T>) -> Option<Word> {
+        let mut word = None;
+        for (k, v) in self.iter() {
+            if v.eq(handle) {
+                word = Some(*k);
```
kvark, line 19: nit: why not just `return`?

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+    fn lookup_id(&self, handle: &crate::Handle<T>) -> Option<Word>;
+    fn lookup_handle(&self, word: &Word) -> Option<crate::Handle<T>>;
+}
+
+impl<T> LookupHelper<T> for FastHashMap<Word, crate::Handle<T>> {
+    type Target = T;
+
+    fn lookup_id(&self, handle: &crate::Handle<T>) -> Option<Word> {
+        let mut word = None;
+        for (k, v) in self.iter() {
```
kvark, line 17: functional style would be better here, i.e. `self.iter().find(...).map(...)`

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+}
+
+pub struct Parser {
+    module: Module,
+    id_count: u32,
+    capabilities: Vec<Capability>,
+    debugs: Vec<Instruction>,
+    annotations: Vec<Instruction>,
+    debug_enabled: bool,
+    void_type: Option<u32>,
```
kvark, line 51: should this use `Word`

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+    lookup_function_type: FastHashMap<Word, LookupFunctionType>,
+    lookup_constant: FastHashMap<Word, crate::Handle<crate::Constant>>,
+    lookup_global_variable: FastHashMap<Word, crate::Handle<crate::GlobalVariable>>,
+    lookup_import: FastHashMap<Word, String>,
+    lookup_label: Vec<Word>,
+    current_label: Word,
+}
+
+impl Parser {
+    pub fn new(module: &crate::Module, debug_enabled: bool) -> Self {
```
kvark, line 63: Boolean arguments are an anti-pattern. Perhaps, it could be just a public field?

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+        self.id_count += 1;
+        self.id_count
+    }
+
+    fn bytes_to_words(&self, bytes: &[u8]) -> Vec<Word> {
+        let words: Vec<Word> = bytes
+            .chunks(4)
+            .map(|chars| match chars.len() {
+                4 => {
+                    (chars[3] as u32) << 24
```
kvark, line 93: it looks like the whole `match` could be replaced by
```rust
chars.iter().rev().fold(0u32, |u, c| (u << 8) | c as u32)
```

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+
+    fn string_to_words(&self, input: &str) -> Vec<Word> {
+        let mut words: Vec<Word> = self.bytes_to_words(input.as_bytes());
+
+        let last_word = words.last().unwrap();
+
+        let last_character = last_word.to_le_bytes()[3];
+
+        if last_character != 0x0 {
+            // nul-termination
```
kvark, line 121: is this required by the spec?

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+    }
+
+    fn string_to_words(&self, input: &str) -> Vec<Word> {
+        let mut words: Vec<Word> = self.bytes_to_words(input.as_bytes());
+
+        let last_word = words.last().unwrap();
+
+        let last_character = last_word.to_le_bytes()[3];
+
+        if last_character != 0x0 {
```
kvark, line 120: alternatively, could check if `input.as_bytes() % 4 == 0`

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+
+    fn instruction_capability(&self, capability: &Capability) -> Instruction {
+        let mut instruction = Instruction::new(Op::Capability);
+        instruction.add_operand(*capability as u32);
+        instruction
+    }
+
+    fn try_add_capabilities(&mut self, capabilities: &[Capability]) {
+        for capability in capabilities.iter() {
+            if !self.capabilities.contains(capability) {
```
kvark, line 136: nit: use `FastHashSet` with a single command for checking and inserting, see [insert](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert) that returns a boolean

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+        self.try_add_capabilities(entry_point.exec_model.required_capabilities());
+        match entry_point.exec_model {
+            ExecutionModel::Vertex | ExecutionModel::GLCompute => {}
+            ExecutionModel::Fragment => {
+                let execution_mode = ExecutionMode::OriginUpperLeft;
+                self.try_add_capabilities(execution_mode.required_capabilities());
+                let mut execution_mode_instruction = Instruction::new(Op::ExecutionMode);
+                execution_mode_instruction.add_operand(function_id);
+                execution_mode_instruction.add_operand(execution_mode as u32);
+                for word in execution_mode_instruction.to_words() {
```
kvark, line 215: In most, if not all, of these places, the heap allocation in `to_words` isn't needed. One way to get around is by using the `Extend<Word>` generic

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+
+    fn get_type_id(
+        &mut self,
+        arena: &crate::Arena<crate::Type>,
+        handle: &crate::Handle<crate::Type>,
+    ) -> Word {
+        match self.lookup_type.lookup_id(&handle) {
+            Some(word) => word,
+            None => {
+                let (instruction, id) = self.instruction_type_declaration(arena, *handle);
```
kvark, line 234: shouldn't we add it to the lookup table?

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+        &self,
+        arena: &crate::Arena<crate::Type>,
+        kind: &crate::ScalarKind,
+        width: &u8,
+    ) -> crate::Handle<crate::Type> {
+        let mut scalar_handle = None;
+        for (handle, ty) in arena.iter() {
+            match ty.inner {
+                crate::TypeInner::Scalar {
+                    kind: _kind,
```
kvark, line 318: would be better to rename function arguments to avoid collision with these
underscores should be left for things that are not used in Rust

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                    width: _width,
+                } => {
+                    if kind == &_kind && width == &_width {
+                        scalar_handle = Some(handle);
+                        break;
+                    }
+                }
+                _ => continue,
```
kvark, line 326: could as well be `=> {}`?

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                    id
+                }
+            },
+        }
+    }
+
+    fn find_scalar_handle(
+        &self,
+        arena: &crate::Arena<crate::Type>,
+        kind: &crate::ScalarKind,
```
kvark, line 311: we should just pass things by value

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+        width: &u8,
```
kvark, line 312: and here

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+    ) -> crate::Handle<crate::Type> {
+        let mut scalar_handle = None;
+        for (handle, ty) in arena.iter() {
+            match ty.inner {
+                crate::TypeInner::Scalar {
+                    kind: _kind,
+                    width: _width,
+                } => {
```
kvark, line 320: condition could be a part of the match, i.e. `if find_kind == kind && find_width == width => {..}`

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+        let ty = &arena[handle];
+        let id = self.generate_id();
+        let mut instruction;
+
+        match ty.inner {
+            crate::TypeInner::Scalar { kind, width } => {
+                match kind {
+                    crate::ScalarKind::Sint => {
+                        instruction = Instruction::new(Op::TypeInt);
+                        instruction.set_result(id);
```
kvark, line 346: should we move `set_result()` call outside of the match?

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                instruction = Instruction::new(Op::TypeVector);
+                instruction.set_result(id);
+                instruction.add_operand(scalar_id);
+                instruction.add_operand(size as u32);
+
+                self.lookup_type.insert(id, handle);
+            }
+            crate::TypeInner::Matrix {
+                columns,
+                rows: _,
```
kvark, line 381: neat trick!

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                kind,
+                width,
+            } => {
+                let scalar_handle = self.find_scalar_handle(arena, &kind, &width);
+                let scalar_id = self.get_type_id(arena, &scalar_handle);
+
+                instruction = Instruction::new(Op::TypeMatrix);
+                instruction.set_result(id);
+                instruction.add_operand(scalar_id);
+                instruction.add_operand(columns as u32);
```
kvark, line 391: why are we not adding this to `lookup_type`?

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                instruction.add_operand(type_id);
+
+                match size {
+                    crate::ArraySize::Static(word) => {
+                        instruction.add_operand(word);
+                    }
+                    _ => panic!("Array size {:?} unsupported", size),
+                }
+
+                self.lookup_type.insert(id, base);
```
kvark, line 416: why are we registering id with the `base`?

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                instruction = Instruction::new(Op::TypeImage);
+                instruction.set_result(id);
+                instruction.add_operand(type_id);
+                instruction.add_operand(dim as u32);
+
+                // TODO Add Depth, but how to determine? Not yet in the WGSL spec
+                instruction.add_operand(1);
+
+                if flags.contains(crate::ImageFlags::ARRAYED) {
+                    instruction.add_operand(1);
```
kvark, line 442: please prefer to use if as an expression. i.e. `instruction.add_operand(if .. else ..);`

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                } else {
+                    instruction.add_operand(0);
+                }
+
+                if flags.contains(crate::ImageFlags::MULTISAMPLED) {
+                    instruction.add_operand(1);
```
kvark, line 448: same here

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                }
+
+                (instruction, id)
+            }
+            crate::ConstantInner::Bool(val) => {
+                let type_id = self.get_type_id(arena, &constant.ty);
+                let mut instruction;
+
+                if val {
+                    instruction = Instruction::new(Op::ConstantTrue);
```
kvark, line 588: similarly, please use if as an expression here, will avoid repeating code

src/back/spirv/parser.rs
```diff
@@ -0,0 +1,1545 @@
...
+                (instruction, id)
+            }
+        }
+    }
+
+    fn get_pointer_id(
+        &mut self,
+        arena: &crate::Arena<crate::Type>,
+        handle: &crate::Handle<crate::Type>,
+        class: &StorageClass,
```
kvark, line 619: all of these should be passed by value
only bigger than word sizes should be passed by reference

src/back/mod.rs
```diff
@@ -1 +1,2 @@
-pub mod msl;
\ No newline at end of file
+pub mod msl;
+pub mod spirv;
```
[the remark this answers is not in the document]
Napokue: Will also rename the front-end of SPIR-V. That was also named spirv.rs, that is the reason I did so too.

src/back/spirv/layout.rs
```diff
@@ -0,0 +1,216 @@
+use spirv::*;
+
+pub(crate) struct Module {
```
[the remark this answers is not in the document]
Napokue: Yes it is. As far as I understand from the spec, you have a Module which consists of two layouts: the Physical Layout and the Logical Layout. The Physical Layout is indeed also called Header, but I haven't found it called that in the spec instead of Physical Layout.

src/back/spirv/layout.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    pub(crate) fn in_words(&self) -> Vec<Word> {
+        let mut words: Vec<Word> = vec![];
+
+        words.push(self.magic_number);
+        words.push(self.version);
+        words.push(self.generator);
+
+        match self.bound {
+            0 => panic!("Bound cannot be 0"),
```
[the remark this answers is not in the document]
Napokue: Spec: 
```
0 < id < Bound

Bound should be small, smaller is better, with all <id> in a module being densely packed and near 0.
```
By this I concluded that a Bound never is allowed to be zero. I haven't found this explicitly stated in the spec, so I could be wrong.

src/back/spirv/layout.rs
```diff
@@ -0,0 +1,216 @@
...
+            _ => words.push(self.bound),
+        }
+
+        words.push(self.instruction_schema);
+        words
+    }
+}
+
+pub(crate) struct LogicalLayout {
+    pub(crate) capabilities: Vec<Word>,
```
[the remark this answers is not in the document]
Napokue: Well that depends on how we want to deal with the parsing. My philosophy is doing it in steps and doing it like so:
1) Parse the IR module, start by going through everything in a AST-like order, so we would reduce the amount of iterations over the IR. Now it doesn't have to comply to the order of instructions that SPIR-V expects. That is done in the `write_logical_layout` in `parser.rs`
2) After everything is parsed, loop through all those Instructions, but now order them in the way SPIR-V expects. That is done here in the `in_words`.

Benefit of doing this is that we are free in what order we process things, instead of immediately adding it to the output stream, and we are able to separate the logic of parsing with the actual structure of a SPIR-V module.

I am not yet sure how we could improve this without having these vectors at the moment.

src/back/spirv/layout.rs
Napokue: Nope no reason, will change it to `Op` and cast it later in the `to_words`.

src/back/spirv/layout.rs
Napokue: Yes that is true, I can do something like: 
```rust
assert!(self.type_id.is_some(), "Type can only be set once");
self.type_id = Some(id);
self.wc += 1;
```

src/back/spirv/layout.rs
Napokue: Is there a reason to want an array and why would we want it to be reference? As `self.operands` is a `Vec<u32>` I cannot push a reference. I can dereference it with `*`, but I am not sure if that is a correct way.

src/back/spirv/parser.rs
Napokue: Could you elaborate why it is an anti-pattern and how you would like to see it as public field?

src/back/spirv/parser.rs
Napokue: I am not yet sure, but it seems to work. It reverses all the characters in the iteration, then does `fold` which is something completely new for me and it produces one single output.

src/back/spirv/layout.rs
Napokue: I can do this:
```rust
std::iter::once((self.wc << 16 | self.op as u32) as u32)
    .chain(self.type_id)
    .chain(self.result_id)
    .chain(self.operands)
    .collect::<Vec<_>>()
```
But I get the error that self.operands cannot be moved (because Vec does not have the Copy trait). Is there an easy way to fix this. If that works, then I really like this solution.

src/back/spirv/layout.rs
```diff
@@ -0,0 +1,216 @@
...
+        PhysicalLayout {
+            magic_number: MAGIC_NUMBER,
+            version,
+            generator,
+            bound: 0,
+            instruction_schema,
+        }
+    }
+
+    pub(crate) fn in_words(&self) -> Vec<Word> {
```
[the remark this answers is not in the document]
Napokue: Could you elaborate on that? As far as I understand vectors their values are stored on the heap and the references stored on the stack? 

I never worked with the case you suggested: `W: impl Extend<Word>`. What is the benefit and how would I implement it.

src/back/spirv/layout.rs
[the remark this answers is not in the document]
Napokue: I guess your suggestions means that all content of this file has to be moved to `mod.rs` and removing the pub(crate), so it will still work?

I tried it, and that works, but wouldn't that remove the benefit of having multiple files to split the code? As I am used to OOP programming, there it is usually standard practice to split all clases to their own files. Is this different in Rust?

src/back/spirv/parser.rs
Napokue: Agreed, but this part is still a wonky way to retrieve these values. I am already planning of having other lookups. I will discuss this with you soon.

src/back/spirv/parser.rs
Napokue: Yes it should, and I will change it, but this should not be a field on its own.

src/back/spirv/parser.rs
Napokue: In the definition of a `Literal String` in the spec it says the following: 

> The final word contains the string’s null-termination character (0), and all contents past the end of the string in the final word are padded with 0.

I have found out through the weeks that when a `Literal String`'s last character is no null termination, that `spirv-val` does not validate the spirv file correctly.

src/back/spirv/parser.rs
Napokue: Okay, yes that sounds good. So this will become:
```rust
fn try_add_capabilities(&mut self, capabilities: &[Capability]) {
    for capability in capabilities.iter() {
        self.capabilities.insert(*capability);
    }
}
```

src/back/spirv/parser.rs
Napokue: It is inserted, but not at this place. It is getting inserted in the `instruction_type_declaration` function, but I completely agree that it will be far better to insert here, for both constantly having to think about inserting and for duplicate code.

src/back/spirv/parser.rs
Napokue: Sure, always wanting to adapt more to the idiomatic rust ways. Do you mean something like this:
```rust
instruction.add_operand(if flags.contains(crate::ImageFlags::ARRAYED) {
    1
} else {
    0
});
```
Or how would you want to format this?

src/back/spirv/layout.rs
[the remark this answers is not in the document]
kvark: Would be good to avoid collision with `crate::Module` semantics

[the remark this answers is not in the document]
kvark: Similar to how the IR is set up: you don't need to move the logic into `mod.rs`, just the definitions that are used across different sub-modules. The logic itself can be in sub-modules, e.g. `impl super::MyStruct { ... }`. You still have multiple sub-modules with this approach.
The benefit of this is that these types are not visible to other front-ends and backends.

src/back/spirv/layout.rs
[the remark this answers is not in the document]
kvark: You'd do something like
```rust
fn fill_words(&self, sink: &mut impl Extend<Word>) {
  sink.extend(iter::once(some_word));
  sink.extend(some_other_words);
}
```
The benefit here is that you don't force the heap allocation upon the caller. They control the storage and just provide it to you in a safe and abstract way, to populate.

src/back/spirv/layout.rs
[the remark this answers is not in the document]
kvark: What happens if an empty module is saved?
I think we should just write 1 instead of panicking

src/back/spirv/layout.rs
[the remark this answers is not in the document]
kvark: Ok, the problem I see with this approach is that it's basically introducing another IR layer. And given that this is SPIRV, which is semantically close to WGSL and Naga IR, having *another* layer here is akin to admitting that Naga IR failed.

We should try to write things down on the fly with a single pass over the module (plus whatever the `GlobalUsage` processor does). If that is not possible for some thing, then we should start asking ourselves if IR needs to be changed. And only then we'd fall back to having multiple passes.

src/back/spirv/layout.rs
kvark: I assume the caller has a list of things to pass here. If the caller doesn't know the length statically, we could have an `Iterator<Item = Word>` here.
The problem with `Vec` in the argument is that it forces a heap allocation, while it doesn't care really. If it's `&[Word]` you can do `operands.iter().cloned()` to produce an iterator over `Word`, which is what you populate `self.operands` with.

src/back/spirv/layout.rs
kvark: right, in order to avoid moving `operands` out, we need to iterate it and clone the items: `self.operands.iter().cloned()`

src/back/spirv/parser.rs
kvark: Simply because when you see a function call `foo(something, true)`, it doesn't give you any idea on what that `true` means. You could say the same thing about integers, but the problem with `bool` is that it's notoriously general and used often.

The idiomatic way to do it would be either having
```rust
enum Debug {
  On,
  Off,
}
```
so that the call spot always sees the semantics.
Or having it a bitflag: `ParserFlags::DEBUG`
Or having it the builder pattern (or an init struct): `Parser::new(module).with_debug()`.

Basically, any API where the meaning of this is obvious at the call site :)

src/back/spirv/parser.rs
kvark: Yeah, fold is a classy pure function, which can be used to implement a lot of different passes over data. As we are shifting the result to the left on each element, the first byte will end up in the high spot, hence I reversed the iterator to accomodate.

src/back/spirv/parser.rs
kvark: Yes, that's cool

src/back/spirv/layout.rs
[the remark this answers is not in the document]
Napokue: Done that, looking good!

src/back/spirv/layout.rs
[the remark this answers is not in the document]
Napokue: That would go against the SPIR-V spec in my opinion. As with `1` the program is saying that it does contain an instruction with a result id. 

Are there any cases where an user passes in an empty module? If so, maybe we can abort the parsing silently, and hint in the console, something like that?

src/back/spirv/layout.rs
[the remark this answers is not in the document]
Napokue: Great point made, I think we can definitely do one pass directly translating the IR to the SPIR-V. I am not yet sure how, but we could discuss that.

src/back/spirv/layout.rs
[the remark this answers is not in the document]
kvark: No, that's not contradicting the spec. The bound is literally a higher bound. The spec even says "bound *should* be small". It is valid for it to be anything as long as all the IDs are lower.

src/back/spv/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+            0 => panic!("Bound cannot be 0"),
+            _ => words.push(self.bound),
+        }
+
+        words.push(self.instruction_schema);
+        words
+    }
+}
+
+pub(crate) struct LogicalLayout {
```
kvark, line 60: no need for `pub(crate)`

src/back/spv/parser.rs
```diff
@@ -0,0 +1,1521 @@
...
+    lookup_function_type: FastHashMap<Word, LookupFunctionType>,
+    lookup_constant: FastHashMap<Word, crate::Handle<crate::Constant>>,
+    lookup_global_variable: FastHashMap<Word, crate::Handle<crate::GlobalVariable>>,
+    lookup_import: FastHashMap<Word, String>,
+    lookup_label: Vec<Word>,
+    current_label: Word,
+}
+
+impl Parser {
+    pub fn new(module: &crate::Module, debug_enabled: bool) -> Self {
```
kvark, line 63: there is still a problem here with the `bool` argument

src/back/spv/parser.rs
```diff
@@ -0,0 +1,1521 @@
...
+                word = Some(*k);
+                break;
+            }
+        }
+        word
+    }
+
+    fn lookup_handle(&self, word: &u32) -> Option<crate::Handle<T>> {
+        let mut handle = None;
+        for (k, v) in self.iter() {
```
kvark, line 28: this is still not just a hashmap access? IIRC you wanted to rewrite this entirely, but I think it's fine to just fix and land, then improve later.

src/back/spv/parser.rs
```diff
@@ -0,0 +1,1521 @@
...
+    }
+
+    fn string_to_words(&self, input: &str) -> Vec<Word> {
+        let mut words: Vec<Word> = self.bytes_to_words(input.as_bytes());
+
+        let last_word = words.last().unwrap();
+
+        let last_character = last_word.to_le_bytes()[3];
+
+        if last_character != 0x0 {
```
kvark, line 104: would be easier to check for `input.as_bytes().len() % 4 == 0`, since you wouldn't need to get the `last_word` and `last_character` at all.

Napokue:
I think they aren't showing because I already fixed some comments and pushed, which makes it outdated.

src/back/spirv/layout.rs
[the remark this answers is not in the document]
Napokue: True, this `Module` struct is redundant and is only a wrapper around Physical- and LogicalLayout. I remove this and instead directly use the `PhysicalLayout`  and `LogicalLayout`.

src/back/spirv/parser.rs
Napokue: Not exactly sure, because that will only tell if the input is correctly aligned, or am I missing something?

src/back/spirv/parser.rs
Napokue: Could you explain maybe why? Because as far as I understood by reading through the Rust documentation, using references does not transfer ownership, and we wouldn't want to have ownership of these values in this function, or do we? Maybe I am not completely understanding this concept.

examples/convert.rs
```diff
@@ -72,6 +72,30 @@ fn main() {
...
+            debug_enabled = args[3].parse().unwrap();
+        } else {
+            debug_enabled = true;
+        }
+
+        let mut parser = spirv::parser::Parser::new(&module, debug_enabled);
+        let spirv = parser.parse(&module);
+
+        let mut bytes: Vec<u8> = vec![];
+        for byte in spirv.iter() {
```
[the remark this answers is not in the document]
Napokue: @kvark came with another suggestion, because this one doesn't work for fixed-sized arrays:
`let bytes = spirv.iter().fold(Vec::with_capacity(spv.len() * 4), |mut v, w| { v.extend_from_slice(&w.to_le_bytes()); v })`

src/back/spirv/parser.rs
kvark: If the input is exactly 4 byte aligned, then last character is not 0.

src/back/spirv/parser.rs
kvark: Down to the hardware, the fastest way to call this function is via registers. Register needs to essentially store a copy of the data. If you are using a reference, you are telling it to work with a pointer, hoping that it will optimize this down to a register. This hope doesn't always materialize, and it's simpler to just tell it that you don't care about the original, aliasing, etc, you only need the value.

src/back/spirv/parser.rs
Napokue: Thanks that is logical indeed. I made a small change, but I guess you meant this:
```
if input.as_bytes().len() %4 == 0
```

src/back/spirv/parser.rs
Napokue: Okay, that makes sense. You said somewhere in this PR that you only want to use references whenever the value will be bigger than 32 bits. So that would mean those optimization are greater than using references for values smaller than 32 bits?

Could I use this principle as guideline, so I won't put references everywhere? 😅

src/back/spirv/parser.rs
Napokue: Didn't know that!

src/back/spirv/parser.rs
kvark: A general principle is: only put stuff behind a ref if its size is greater than the size of a pointer (i.e. 64 bit).

Sometimes, we make exception from this. For example, all descriptors in wgpu-rs are passed by reference, regardless of their actual byte size.

Napokue: Thanks for all that information, that helps a lot!

src/back/spirv/layout.rs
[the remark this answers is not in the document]
Napokue: Okay, so I think I get what you mean. Bound is always: [total_id] + 1. So if we don't have any ids (aka instructions), that means that bound = 0 + 1 = 1, right? So not having instructions results in a Bound of 1.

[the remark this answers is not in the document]
kvark: That's not what I'm saying:)
Bound is at least as high as the highest used ID, +1.
So if you have IDs 0, 1, 3, then bound can be 4, or it can be 10. Both are valid choices.
If you have no IDs, 0 is not a valid choice by the spec, but 1 is, still. I.e. 1 is still higher than "any used IDs" :)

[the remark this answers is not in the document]
Napokue: However, in the Validation rules of SPIR-V it says: "There is at least one OpEntryPoint instruction, unless the Linkage capability is being used". I'm not familiar with "Linkage", but for the rest of the capabilities, it is actually necessary to have at least one instruction. 

That being said, I actually think we don't need to check here for the Bound, but we need to check if there is one Entry Point available.

For the time being, I think the best we can do is just returning `1` when the match matches `0`.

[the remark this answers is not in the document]
Napokue: I gave this some more though. and the "0" match is even unreachable, as the bound is now always set at `self.id_count + 1`. I will remove the match.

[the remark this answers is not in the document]
kvark: yeah, that makes sense :+1:

src/back/spirv/parser.rs
Napokue: Code is removed

src/back/spirv/parser.rs
Napokue: Code is removed

src/back/spirv/parser.rs
Napokue: Code is removed

src/back/spv/parser.rs
Napokue: Yes, this code is removed :)

src/back/spirv/parser.rs
Napokue: Code is removed

src/back/spirv/layout.rs
Napokue: I'm not sure if I'm exactly following your thoughts. I changed the function now to:
```rust
pub(crate) fn add_operands(&mut self, operands: Vec<Word>) {
    self.operands.extend(operands.iter().cloned());
}
```
Would this suffice to circumvent the heap allocation, or does the argument needs to be refactored to either an `Iterator` or `&[ ]`?

**Side-note:** `add_operands` is only used for [`string_to_words`](https://github.com/Napokue/naga/blob/spirv-backend/src/back/spv/parser.rs#L97). Maybe there is some potential to refactor that one too and return something else than a `Vec`? I'm not sure yet though, as I've tried, but failed to do so up until now 😅

kvark: It looks like this should be `fn add_operands(&mut self, operands: impl Iterator<Item = Word>)`, and `string_to_words` should return `impl Iterator<Item = Word>`

src/back/spirv/layout.rs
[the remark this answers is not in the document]
Napokue: Gave this idea some time to grow, but I'm not sure if there is a chance to have it single passed. 

So we are back at the topic why it is designed like this: the Functions arena is parsed first, due to loop efficiency. The vision I had, was that when looping through Functions and encountering Types, Constants, Variables etc. it would be checked if it is already parsed, if not, parse it. So we are removing the need to loop through each individual arena.

There are variables like Pointers, and maybe some other instructions that are generated on runtime, but need to be inserted in the Type Declarations. If we do it single pass, it wouldn't be possible as we don't know where the Type Declarations are in the one stream of Words.

I'm thinking about if the IR can be improved to help this case, but I am not sure at the moment. Do you think it should be possible to do it in one pass?

src/back/spirv/parser.rs
Napokue: I've tried to remove the unnecessary references, but stumbled upon many move errors, especially for the `arena` in [`instruction_type_declaration`](https://github.com/Napokue/naga/blob/spirv-backend/src/back/spv/parser.rs#L351). The value is moved here, because I would give ownership to the `find_scalar_handle` function.

examples/convert.rs
```diff
@@ -72,6 +72,30 @@ fn main() {
         };
         let msl = msl::write_string(&module, options).unwrap();
         fs::write(&args[2], msl).unwrap();
+    } else if args[2].ends_with(".spv") {
+        use naga::back::spirv;
+
+        let debug_enabled;
+
+        if args.len() == 4 {
+            debug_enabled = args[3].parse().unwrap();
```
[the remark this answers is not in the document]
kvark: can we follow the suggestion?

examples/convert.rs
```diff
@@ -72,6 +72,30 @@ fn main() {
...
+            debug_enabled = true;
+        }
+
+        let mut parser = spirv::parser::Parser::new(&module, debug_enabled);
+        let spirv = parser.parse(&module);
+
+        let mut bytes: Vec<u8> = vec![];
+        for byte in spirv.iter() {
+            let bits = byte.to_le_bytes();
+            bytes.push(bits[0]);
```
kvark, line 92: let's do `extend_from_slice` (like you suggested earlier, but without fancy functional style)

src/back/spirv/layout.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub(crate) debugs: Vec<Word>,
+    pub(crate) annotations: Vec<Word>,
+    pub(crate) type_declarations: Vec<Word>,
+    pub(crate) constants: Vec<Word>,
+    pub(crate) global_variables: Vec<Word>,
+    function_declarations: Vec<Word>,
+    pub(crate) function_definitions: Vec<Word>,
+}
+
+impl LogicalLayout {
```
kvark, line 82: should this just `derive(Default)`?

src/back/spirv/layout.rs
kvark: would this still make sense to do here?

src/back/spirv/mod.rs
```diff
@@ -0,0 +1,2 @@
+mod layout;
+pub mod parser;
```
kvark, line 2: missing newline?

src/back/spirv/layout.rs
Napokue: Yes, I will change it back to `Op`

src/back/spirv/parser.rs
kvark: Keep the references for stuff that is non-movable and isn't meant to be moved
My recommendation was simply not referencing trivial types like `ScalarKind`, since they are smaller than their references :)

src/back/spirv/layout.rs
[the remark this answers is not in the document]
kvark: let's talk about this on the chat

kvark:
Landing checklist:
- [ ] rebase on top of master
- [ ] cargo fmt (the rest of the code should already be formatted)
- [ ] make it so every commit is testable (e.g. squash everything, or push one by one to let CI kick in)

PR #47, opened 2020-04-17
Reduce duplicate values in IR module

### LINKED ISSUE #42: Multiple float constants with the same value
Having the following WGSL input:

```
[[location 0]] var<in> a_uv : vec2<f32>;
[[location 0]] var<out> o_color : vec4<f32>;

fn main() -> void {
  o_color = vec4<f32>(1.0, 0.0, 0.0, 1.0);
  return;
}
entry_point fragment as "main" = main;
```

The vector float 4 is parsing this to the following expression:
```
Constant(
    Handle(1),
),
Constant(
    Handle(2),
),
Constant(
    Handle(3),
),
Constant(
    Handle(4),
),
Compose {
    ty: Handle(2),
    components: [
        Handle(3),
        Handle(4),
        Handle(5),
        Handle(6),
    ],
},
```

You would expect that two constant handles would point to the value "1.0" constant, and two constant handles point to the value "0.0" constant. Each float is parsed as a separate float constant, despite having the same value.

Napokue:
Fix #42 

Reduces the amount of duplicate information in the IR module. Constants and Global Variable arena now only contains unique values.

Also did this for expression to reduce duplicate expressions, but I am not sure if we would want that. Could there be a case where duplicate expressions are acceptable?

Note: Boids IR has been decreased by about ~300 lines.

### DIFF
--- a/src/front/wgsl.rs
+++ b/src/front/wgsl.rs
@@ -467,7 +467,7 @@ impl Parser {
                 return Ok(expr);
             }
             Token::Word("true") => {
-                let handle = ctx.constants.append(crate::Constant {
+                let handle = ctx.constants.fetch_or_append(crate::Constant {
                     name: None,
                     specialization: None,
                     inner: crate::ConstantInner::Bool(true),
@@ -482,7 +482,7 @@ impl Parser {
                 crate::Expression::Constant(handle)
             }
             Token::Word("false") => {
-                let handle = ctx.constants.append(crate::Constant {
+                let handle = ctx.constants.fetch_or_append(crate::Constant {
                     name: None,
                     specialization: None,
                     inner: crate::ConstantInner::Bool(false),
@@ -498,7 +498,7 @@ impl Parser {
             }
             Token::Number(word) => {
                 let (inner, kind) = Self::get_constant_inner(word)?;
-                let handle = ctx.constants.append(crate::Constant {
+                let handle = ctx.constants.fetch_or_append(crate::Constant {
                     name: None,
                     specialization: None,
                     inner,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,7 +83,7 @@ pub enum TypeInner {
     Sampler,
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct Constant {
     pub name: Option<String>,
     pub specialization: Option<spirv::Word>,
@@ -91,7 +91,7 @@ pub struct Constant {
     pub ty: Handle<Type>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum ConstantInner {
     Sint(i64),
     Uint(u64),


Napokue:
There are some direct append usages left, but I don't think they need to be changed.

src/front/wgsl.rs
```diff
@@ -548,7 +548,7 @@ impl Parser {
             other => return Err(Error::Unexpected(other)),
         };
         self.scopes.pop();
-        Ok(ctx.expressions.append(expression))
+        Ok(ctx.expressions.fetch_or_append(expression))
```
kvark, line 551: Not sure this is a good idea. There is a cost to `fetch_or_append` proportional to the number of elements, and it's not optimized yet. There may be many many expressions, so checking for duplication on each one may be expensive.

src/front/wgsl.rs
```diff
@@ -1300,7 +1300,7 @@ impl Parser {
                 lexer.expect(Token::Operation('='))?;
                 let inner = self.parse_const_expression(lexer, &mut module.types, &mut module.constants)?;
                 lexer.expect(Token::Separator(';'))?;
-                let const_handle = module.constants.append(crate::Constant {
+                let const_handle = module.constants.fetch_or_append(crate::Constant {
```
kvark, line 1303: We are walking over the top level declarations in WGSL here. Why would there be redundant constants or variables?

Napokue: Yes you are right about this one. As the names always have to be unique variables, read [here](https://gpuweb.github.io/gpuweb/wgsl.html#validation). However, there is not a mention about constants not being able to be duplicated, but I guess that should also be the case. I am removing this.

src/front/wgsl.rs
```diff
@@ -1310,7 +1310,7 @@ impl Parser {
             }
             Token::Word("var") => {
                 let (name, class, ty) = self.parse_variable_decl(lexer, &mut module.types, &mut module.constants)?;
-                let var_handle = module.global_variables.append(crate::GlobalVariable {
+                let var_handle = module.global_variables.fetch_or_append(crate::GlobalVariable {
```
Napokue, line 1313: Removing this one too.

src/front/wgsl.rs
Napokue: Yes that is true, but then there will be duplicate expressions in this Arena. If I use this simple WGSL:
```
fn main() -> void {
  var a : vec4<f32> = vec4<f32>(1.0, 0.0, 0.0, 1.0);
  return;
}
entry_point fragment as "main" = main;
```

With only `append` the IR looks like this:
```
expressions: Arena {
	data: [
		Constant(
			Handle(1),
		),
		Constant(
			Handle(2),
		),
		Constant(
			Handle(2),
		),
		Constant(
			Handle(1),
		),
		Compose {
			ty: Handle(1),
			components: [
				Handle(1),
				Handle(2),
				Handle(3),
				Handle(4),
			],
		},
		LocalVariable(
			Handle(1),
		),
	],
},
```
As you can see you would get double constant expressions refering to the same Handle.

While using it with `fetch_or_append`, would result in this:
```
expressions: Arena {
	data: [
		Constant(
			Handle(1),
		),
		Constant(
			Handle(2),
		),
		Compose {
			ty: Handle(1),
			components: [
				Handle(1),
				Handle(2),
				Handle(2),
				Handle(1),
			],
		},
		LocalVariable(
			Handle(1),
		),
	],
},
```
This is only a single line of WGSL and in the future it would also exponentially increase the amount of expressions in the IR.

Now I understand the performance issue, and it will be a trade-off:
1 ) Cluttering the IR with duplicate expressions and let the back-end take care of the duplicates by not including them. There still needs to be a check in the back-end for duplicates. I am not sure if that is less, even or more expensive.
2 ) Checking in the front-end and prevent from ever inserting duplicates in the IR

kvark: Thank you for the detailed comparison, it's very helpful for the discussion!
It comes down to what we call an expression in the IR. If it's an expression in the AST sense, then it makes total sense to deduplicate them (at least, if it costed us nothing, but I think it does).
If it's an expression in the SPIR-V sense (i.e. an SSA value), then we shouldn't be deduplicating at least some of the expressions (different loads from a variable, for example).

It's not clear to me where to draw the line at this point, but it's a safe bet to not deduplicate the expressions at least today. Let's keep it, at least until we settle on a stronger meaning of an "expression", as well as have a faster way (lookup hashmap?) to deduplicate them (current one is too naive).

Napokue: Agreed, I think trying to "deduplicate" the expressions here are as of yet an unknown cost. 

Lets revert this one too and only keep the constants as `fetch_or_append`. That is my use-case for the SPIR-V back-end at the moment: it is generating a lot of OpTypeInt's and OpTypeFloats with the same value.

In the meantime, while Naga is maturing we can always resume this discussion. :)

Issue #32, opened 2020-03-16
Serialization of Module


kvark:
I was thinking that it would be great to be able to save (and load?) [RON](https://github.com/ron-rs/ron) files by the converter. The joys of RON is that it's very nice to inspect visually (by humans):
  - array elements have index numbers, which is critical for associating them with `Handle(xxx)` values through the rest of the module
  - nice representation of maps and enums
  - otherwise similar to pretty `Debug` implementation
  - ability to just tweak it by hand

To get this, we need a `serde` optional feature, implementation of `Serialize`+`Deserialize` for all IR objects, and also https://github.com/gfx-rs/rspirv/issues/123 resolved.

paulkernfeld:
Should this have separate `serialize` and `deserialize` features as in gfx-rs/rspirv#123, or is it fine to just have a single `serde` feature for both?

kvark:
In general, it would be better to separate serialization from deserialization, since the latter is taking way more compile time. This would also let us avoid conflicting the `serde` feature with optional dependency.

PR #48, opened 2020-05-01
Derive Serialize and Deserialize for all IR types

paulkernfeld:
Add serialize and deserialize Cargo features, a step towards #32

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,11 +9,16 @@ repository = "https://github.com/gfx-rs/naga"
 keywords = ["shader", "SPIR-V"]
 license = "MPL-2.0"
 
+[features]
+serialize = ["serde", "spirv/serialize"]
+deserialize = ["serde", "spirv/deserialize"]
+
 [dependencies]
 bitflags = "1"
 fxhash = "0.2"
 log = "0.4"
 num-traits = "0.2"
+serde = { version = "1", optional = true, features = ["serde_derive"] }
 spirv = { package = "spirv_headers", version = "1" }
 
 [dev-dependencies]
--- a/tests/convert.rs
+++ b/tests/convert.rs
@@ -49,3 +49,13 @@ fn convert_boids() {
         msl::write_string(&module, options).unwrap();
     }
 }
+
+#[test]
+#[cfg(all(feature = "serialize", feature = "deserialize"))]
+fn boids_serde_round_trip() {
+    let module = load_wgsl("boids");
+    let module_ser = ron::ser::to_string(&module).unwrap();
+    let module_de = ron::de::from_str::<naga::Module>(&module_ser).unwrap();
+    let module_ser_ser = ron::ser::to_string(&module_de).unwrap();
+    assert_eq!(module_ser, module_ser_ser);
+}
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -9,11 +9,38 @@ use std::{fmt, hash, marker::PhantomData, num::NonZeroU32};
 type Index = NonZeroU32;
 
 /// A strongly typed reference to a SPIR-V element.
-#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
-#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+#[repr(transparent)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize), serde(into = "SerHandle"))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize), serde(from = "SerHandle"))]
 pub struct Handle<T> {
     index: Index,
     marker: PhantomData<T>,
+}
+
+/// This type allows us to make the serialized representation of a Handle more concise
+#[allow(dead_code)]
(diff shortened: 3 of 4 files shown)


paulkernfeld:
The build should pass as soon as a version of `rspirv` containing gfx-rs/rspirv#123 is released.

paulkernfeld:
I'm not sure of the best way to handle the Monocodus fmt errors; the commit suggestions from the Monocodus bot produced invalid code. Possible solutions:

1. Ignore the Monocodus bot and merge the code
2. Try to fix only the fmt errors that Monocodus detected
3. Run `cargo fmt` on all of `src/lib.rs`
4. Run `cargo fmt` on the whole project (probably in a separate PR)

kvark:
Sorry about that! Please don't mix formatting into the PR, we'll do it separately. Monocodus is a fairly new thing we are testing, and it just got an ability to be configured. I disabled the formatting check for now.

Cargo.toml
```diff
@@ -9,14 +9,19 @@ repository = "https://github.com/gfx-rs/naga"
...
 log = "0.4"
 num-traits = "0.2"
+serde = { version = "1", optional = true, features = ["derive"] }
 spirv = { package = "spirv_headers", version = "1" }
 
 [dev-dependencies]
 env_logger = "0.6"
 ron = "0.5"
-serde = { version = "1", features = ["serde_derive"] }
+serde = { version = "1", features = ["derive"] }
```
kvark, line 27: this feature doesn't exist in 1.0.0: https://crates.io/crates/serde/1.0.0
let's use `serde_derive` for compatibility, or put a specific serde version here that has that

paulkernfeld: Got it; I went with the `serde_derive` feature

paulkernfeld:
Yeah, good point! I added two tests.

test-data/boids-module.ron
```diff
@@ -0,0 +1,2325 @@
...
+                                index: 4,
+                                marker: (),
+                            ),
+                            index: 1,
+                        ),
+                        Call(
+                            name: "atan2",
+                            arguments: [
+                                (
+                                    index: 5,
```
kvark, line 397: It looks to me that most of the space in this RON is dedicated to `Handle<>` serialization.
Would you be open to augment it a bit in the same way `wgpu` deals with Ids?
See https://github.com/gfx-rs/wgpu/blob/0a0ef9a1000a4d327209570dd2606008b133a2e3/wgpu-core/src/id.rs#L22-L29

In this case, it would need to look like `Id(index)`

tests/convert.rs
```diff
@@ -49,3 +49,16 @@ fn convert_boids() {
         msl::write_string(&module, options).unwrap();
     }
 }
+
+/// This is a "poor man's snapshot test." The idea is to eventually use insta (#41).
```
kvark, line 53: What is the goal of this snapshot test? Ensure that the serialized form is unchanged?
It could be a bigger promise to make than we are ready to, at this point.
Although it wouldn't be as much of a problem if the saved RON was smaller :) So maybe with the `Handle<>` serialization change it's going to be ok.

paulkernfeld: Fair enough. I replaced this with a round-trip test, which makes a more reasonable promise.

kvark:
Let's squash some more, given one of the commits is "Add poor man's snapshot test for serialization". Maybe even squash everything? Would be easier to confirm CI is happy with each commit, this way

paulkernfeld:
All right, I squashed into a single commit. BTW in case you missed it due to the long PR history, the build is currently failing because it's waiting on a deploy of gfx-rs/rspirv#123.

test-data/boids-module.ron
paulkernfeld: I gave this a shot but I had trouble getting `serde_derive` to work with the `PhantomData`. Do you think it should be possible with `serde_derive`? Or possibly I could try implementing `Serialize` and `Deserialize` by hand for `Handle<T>`?

kvark: The code I linked to does almost precisely this, it also has `PhantomData`. Implementing the traits manually is also OK to me, but I think the same approach as linked is slightly easier.

paulkernfeld: Yeah, you're right. I'll take another look to see what I am getting wrong.

paulkernfeld: All right, I implemented and tested the concise handle serialization, then pushed it onto this PR. Fun trick 😀

kvark:
@paulkernfeld just checking, this is still blocked?

paulkernfeld:
@kvark thanks for checking in! It looks like this is still blocked on a release of `spirv_headers` that includes the `serde` feature that was merged in gfx-rs/rspirv#125.

paulkernfeld:
Ah, it's actually now separate `serialize` and `deserialize` features but I think otherwise my comment is true.

lachlansneff:
@paulkernfeld Are you still working on this? If not, I'd be happy to take over.

kvark:
Replaced by #125

paulkernfeld:
@lachlansneff thanks for picking up the torch on this! I lost track of it.

Issue #31, opened 2020-03-16
Avoid panics in the Metal backend


kvark:
We have more panics than needed. The corners were cut when writing the backend as a PoC. However, for any serious use, it should return an error instead of panicing. This applies to all the front-ends and backends. Panic is only allowed where our internal consistency is shattered, and even then it would work best if we just error out.

PR #51, opened 2020-05-18
Avoid panics in the Metal backend

paulkernfeld:
This is a draft implementation for #31. I think there are a lot of different ways to structure this code, so let me know if you think this is a good strategy. If so, I can replace more panics.

### DIFF
--- a/src/back/msl.rs
+++ b/src/back/msl.rs
@@ -76,6 +76,7 @@ pub enum Error {
     InvalidImageFlags(crate::ImageFlags),
     MutabilityViolation(crate::Handle<crate::GlobalVariable>),
     BadName(String),
+    UnexpectedGlobalType(crate::Handle<crate::Type>),
 }
 
 impl From<FmtError> for Error {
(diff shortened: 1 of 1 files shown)


src/back/msl.rs
```diff
@@ -227,8 +228,9 @@ struct TypedGlobalVariable<'a> {
     handle: crate::Handle<crate::GlobalVariable>,
     usage: crate::GlobalUse,
 }
-impl Display for TypedGlobalVariable<'_> {
-    fn fmt(&self, formatter: &mut Formatter<'_>) -> Result<(), FmtError> {
+
+impl<'a> TypedGlobalVariable<'a> {
+    fn try_fmt<W: Write>(&self, formatter: &mut W) -> Result<(), Error> {
```
kvark, line 233: niice! I was missing exactly this

paulkernfeld:
👍 awesome! I just squashed the two commits.

kvark:
Thanks!

PR #52, opened 2020-05-19
Replace panics with errors in Metal backend

### LINKED ISSUE #31: Avoid panics in the Metal backend

paulkernfeld:
Fixes #31

### DIFF
--- a/src/back/msl.rs
+++ b/src/back/msl.rs
@@ -77,6 +77,22 @@ pub enum Error {
     MutabilityViolation(crate::Handle<crate::GlobalVariable>),
     BadName(String),
     UnexpectedGlobalType(crate::Handle<crate::Type>),
+    UnimplementedBuiltIn(spirv::BuiltIn),
+    UnimplementedBindTarget(BindTarget),
+    UnexpectedIndexing(crate::TypeInner),
+    UnsupportedCompose(crate::Handle<crate::Type>),
+    UnexpectedLoadPointer(crate::TypeInner),
+    UnsupportedBinaryOp(crate::BinaryOperator),
+    UnableToInferBinaryOpOutput(crate::TypeInner, crate::BinaryOperator, crate::TypeInner),
+    UnexpectedImageType(crate::TypeInner),
+    UnexpectedDistanceArgument(crate::TypeInner),
+    UnsupportedCall(String),
+    UnsupportedExpression(crate::Expression),
+    UnableToReturnValue(crate::Handle<crate::Expression>),
+    UnsupportedStatement(crate::Statement),
+    UnsupportedDim(spirv::Dim),
+    ExecutionModelReturnType(spirv::ExecutionModel, crate::Handle<crate::Type>),
+    AccessIndexExceedsStaticLength(u32, u32),
 }
 
 impl From<FmtError> for Error {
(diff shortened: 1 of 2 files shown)


src/back/msl.rs
```diff
@@ -269,41 +285,49 @@ impl<'a> TypedGlobalVariable<'a> {
...
                 let name = match built_in {
                     spirv::BuiltIn::ClipDistance => "clip_distance",
                     spirv::BuiltIn::GlobalInvocationId => "thread_position_in_grid",
                     spirv::BuiltIn::PointSize => "point_size",
                     spirv::BuiltIn::Position => "position",
-                    _ => panic!("Built in {:?} is not implemented", built_in),
+                    _ => return Err(Error::UnimplementedBuiltIn(built_in)),
                 };
-                formatter.write_str(name)
+                Ok(formatter.write_str(name)?)
```
kvark, line 299: we shouldn't need that syntax: `Ok(something?)`, unless it's deliberate

src/lib.rs
```diff
@@ -228,10 +228,10 @@ pub enum Expression {
 }
 
 pub type Block = Vec<Statement>;
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct FallThrough;
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
kvark, line 234: It seems to me that we could consistently return handles to things in the error instead of their copies. If true, that would allow us to not derive `Clone` for these types, while the users would still be able to get the actual data from the module arenas. Would you be open to explore this direction?

src/back/msl.rs
paulkernfeld: Ah, okay. How would you prefer this line be written?

kvark: Actually, I was wrong here. Sorry!
The `?` operator can convert between error types, but just calling `write_str` would not.

src/lib.rs
paulkernfeld: I just took a look at what this would involve. I think it would require a decent amount of refactoring, because we'd need to change a lot of functions from taking `&X` to `Handle<X>`.

Just for my own understanding, how bad is it for these types to implement `Clone`? My reasoning was that cloning the types to return errors would be fine because it wouldn't slow down the hot path.

paulkernfeld:
I am happy to squash down to a single commit after we get the other requested changes sorted out. I mainly kept the individual commits separate in order to be able to revert them more easily if anything needed to be changed during the review process.

src/lib.rs
kvark: Yeah, it's not a concern about performance. Rather, if this is meant to only be used by error reporting, I don't want users of Naga to consider this API to be public, unless there is a strong case for it.
Given your findings, we can go ahead with this. Let's please add `//Note` comments above these structs saying that `Clone` is only used by the error reporting.

src/back/msl.rs
paulkernfeld: Yeah, that tripped me up at first as well! 😂

src/lib.rs
paulkernfeld: Is the following note okay?

```
/// Clone is used only for error reporting and is not intended for end users
```

Should I add this note for any data struct that implements `Clone`? Or just stuff that I am adding in this PR? E.g. `ScalarKind` previously derived `Clone`.

kvark: I wasn't considering to make this a doc comment. Rather - a note for us internally.
It only needs to be on the stuff that you added.

paulkernfeld: All right, I added this note for each new derived clone.

paulkernfeld:
I have now squashed down to a single commit so I think this should be ready to merge 😀

Issue #43, opened 2020-04-08
WGSL doesn't type check values for vectors


Napokue:
Having the following WGSL:
```
[[location 0]] var<in> a_uv : vec2<f32>;
[[location 0]] var<out> o_color : vec4<f32>;

fn main_frag() -> void {
  o_color = vec4<f32>(1, 0, 0, 1);
  return;
}
entry_point fragment as "main" = main_frag;
```

The values provided into the vector float 4 with the 4 integer values will be correctly parsed to the IR model:
```
Constant {
    name: None,
    specialization: None,
    inner: Sint(
        1,
    ),
    ty: Handle(4),
},
Constant {
    name: None,
    specialization: None,
    inner: Sint(
        0,
    ),
    ty: Handle(4),
},
```

The back-ends would now have to check if the vector where these constants are provided for is able to use them. I think it would be better to let the WGSL front-end check if the values in a vector are of the same Type as the composite they are in.

This is not something only applicable to vectors, and should have a general type check to ensure a correct IR model is delivered to the back-end.

kvark:
Closed as a part of #59

PR #54, opened 2020-05-20
Check constant types

paulkernfeld:
A baby step towards #43

### DIFF
--- a/src/front/wgsl.rs
+++ b/src/front/wgsl.rs
@@ -3,7 +3,6 @@ use crate::{
     proc::{Typifier, ResolveError},
     FastHashMap,
 };
-use crate::proc::UnexpectedConstantTypeError;
 
 
 #[derive(Debug, PartialEq)]
@@ -161,7 +160,7 @@ pub enum Error<'a> {
 }
 
 impl<'a> From<crate::proc::UnexpectedConstantTypeError> for Error<'a> {
-    fn from(error: UnexpectedConstantTypeError) -> Self {
+    fn from(error: crate::proc::UnexpectedConstantTypeError) -> Self {
         Error::UnexpectedConstantType(error)
     }
 }
--- a/src/front/wgsl.rs
+++ b/src/front/wgsl.rs
@@ -3,7 +3,7 @@ use crate::{
     proc::{Typifier, ResolveError},
     FastHashMap,
 };
-use crate::proc::typifier::UnexpectedConstantTypeError;
+use crate::proc::UnexpectedConstantTypeError;
 
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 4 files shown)


paulkernfeld:
Let me know if I'm on the right track with this! I think there are a lot of ways to structure this code.

Since this doesn't actually depend on any WGSL-specific concepts, we should be able to move this code into `lib.rs` or maybe `typifier`. That way, other front ends could benefit from the same check.

src/front/wgsl.rs
```diff
@@ -354,6 +355,38 @@ pub enum Scope {
...
+        ) => Ok(()),
+        (
+            crate::ConstantInner::Bool(_),
+            crate::TypeInner::Scalar { kind: crate::ScalarKind::Bool, width: _ },
+        ) => Ok(()),
+        (
+            crate::ConstantInner::Composite(inner),
+            _
+        ) => Ok(()), // TODO recursively check composite types
+        (other_inner, other_type_inner) => return Err(Error::UnexpectedConstantType(
```
kvark, line 383: `return` isn't needed here

src/front/wgsl.rs
```diff
@@ -354,6 +355,38 @@ pub enum Scope {
     GeneralExpr,
 }
 
+fn check_constant_types<'a>(
+    inner: &crate::ConstantInner,
+    type_inner: &crate::TypeInner
+) -> Result<(), Error<'a>> {
+    match (inner, type_inner) {
```
kvark, line 362: another way to write this could be:
```rust
let expected_kind = match inner {
  crate::ConstantInner::Sint(_) => crate::ScalarKind::Sint,
  ...
  crate::ConstantInner::Composite(inner) => {
     return Something;
  }
};
match type_inner {
  crate::TypeInner::Scalar {  kind, width: _ } if kind == expected_kind => Ok(()),
  _=> Err(...),
}
```

It's up to you to pick a preference here.

paulkernfeld: I'd like to leave it as-is for now, since I'm guessing that the code will need to change substantially to incorporate type checking for composite types anyways.

paulkernfeld:
Okay, I moved the function to `src/proc/typifier`. Can you check to see if the errors are structured in a reasonable way?

src/proc/mod.rs
```diff
@@ -1,4 +1,4 @@
 mod interface;
-mod typifier;
+pub(crate) mod typifier;
```
kvark, line 2: let's be consistent here and just add the types you need in the following `pub use` statement instead of switching the visibility of this module

paulkernfeld: 👍

PR #55, opened 2020-05-25
Add initial wasm build

pjoe:
Adds support for building to wasm, using wasm-pack.

Includes size optimization settings for release build in Cargo.toml

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,23 +19,26 @@ log = "0.4"
 num-traits = "0.2"
 spirv = { package = "spirv_headers", version = "1" }
 
-wasm-bindgen = { version = "0.2", optional = true }
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasm-bindgen = "0.2"
 # The `console_error_panic_hook` crate provides better debugging of panics by
 # logging them with `console.error`. This is great for development, but requires
 # all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
 # code size when deploying.
-console_error_panic_hook = { version = "0.1.1", optional = true }
+console_error_panic_hook = "0.1.1"
 # `wee_alloc` is a tiny allocator for wasm that is only ~1K in code size
 # compared to the default allocator's ~10K. It is slower than the default
 # allocator, however.
 #
 # Unfortunately, `wee_alloc` requires nightly Rust when targeting wasm for now.
-wee_alloc = { version = "0.4.2", optional = true }
+wee_alloc = "0.4.2"
 
 [dev-dependencies]
 env_logger = "0.6"
 ron = "0.5"
 serde = { version = "1", features = ["serde_derive"] }
+
+[target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 wasm-bindgen-test = "0.2"
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ log = "0.4"
 num-traits = "0.2"
 spirv = { package = "spirv_headers", version = "1" }
 
-wasm-bindgen = "0.2"
+wasm-bindgen = { version = "0.2", optional = true }
 # The `console_error_panic_hook` crate provides better debugging of panics by
 # logging them with `console.error`. This is great for development, but requires
 # all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,9 @@ repository = "https://github.com/gfx-rs/naga"
 keywords = ["shader", "SPIR-V"]
 license = "MPL-2.0"
 
+[lib]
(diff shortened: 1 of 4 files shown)


src/wasm.rs
```diff
@@ -0,0 +1,40 @@
...
+    // we will get better error messages if our code ever panics.
+    //
+    // For more details see
+    // https://github.com/rustwasm/console_error_panic_hook#readme
+    #[cfg(feature = "console_error_panic_hook")]
+    console_error_panic_hook::set_once();
+}
+
+#[wasm_bindgen]
+pub fn wgsl2msl(input: &str) -> String {
```
kvark, line 21: What happens when we add any functions to this module? If I understand correctly, these are public extern functions, so nothing is going to strip unused logic parts. If we add `wgsl2hlsl` for example, the produced WASM module would have to contain both.

Cargo.toml
```diff
@@ -9,14 +9,43 @@ repository = "https://github.com/gfx-rs/naga"
 keywords = ["shader", "SPIR-V"]
 license = "MPL-2.0"
 
+[lib]
+crate-type = ["cdylib", "rlib"]
```
kvark, line 13: Unfortunately that would make the `cdylib` target to be generated for *every* build of naga as a dependency, even of other Rust code. See https://github.com/rust-lang/cargo/issues/4280 .

Cargo.toml
```diff
@@ -9,14 +9,43 @@ repository = "https://github.com/gfx-rs/naga"
...
+
 [dependencies]
 bitflags = "1"
 fxhash = "0.2"
 log = "0.4"
 num-traits = "0.2"
 spirv = { package = "spirv_headers", version = "1" }
 
+wasm-bindgen = "0.2"
```
kvark, line 22: this should be wasm-target-only, see [wgpu-rs](https://github.com/gfx-rs/wgpu-rs/blob/4b16959e3e5e35178ba7b22cf0ad182344615da6/Cargo.toml#L79) as an example

src/wasm.rs
pjoe: Yeah this isn't really the best solution, just something to get started. The wasm interface should also be split in front-end returning AST and backend taking AST and returning text(or bin) output. But that requires someway to pass the AST efficiently.

Maybe could use features to flag individual front/back-ends?

Cargo.toml
pjoe: Hmmm, not sure how to solve that, wasm-pack told me this is needed for being able to compile to wasm 😟

Cargo.toml
pjoe: I briefly tried making wasm-bindgen optional ... but couldn't quite make it work. Maybe a question of the right cmd line incantation

pjoe:
NP, at least I learned a bit more about how Rust works :)

For reference I originally did this in separate repo: https://github.com/pjoe/wasm-naga

In terms of use case, I think the most usable ones will be glsl2spirv and glsl2wgsl, I just added wgsl2msl, as only msl backend was available, and wgsl seemed easier for testing than spirv as frontend

kvark:
> NP, at least I learned a bit more about how Rust works :)

Hopefully, that's not too painful :)

For reference I originally did this in separate repo: https://github.com/pjoe/wasm-naga

Let's point to it from our README!

> I think the most usable ones will be glsl2spirv and glsl2wgsl,

Yes, that sounds great. Also, you are right that the client could control the WASM binary size by changing the requested features, so this concern is addressable. (better than having multiple library targets in a repo)

Finally, the concern about `crate-type` is not a showstopper. It was a big problem for spirv-cross, because it has a huge C++ dependency that simply failed to link in Gecko as a separate target. This isn't affecting Naga, it's just that the compile times would suffer a bit.

What is a showstopper, however, if is the WASM dependencies with regards to cargo-vendor. That's annoying, but the only workaround would be for us to maintain a fork with them removed. It seems cleaner to instead officially maintain `wasm-naga` on top of the unchanged `naga`.

pjoe:
@kvark: got stubborn 😁, so managed to get the wasm-bingen stuff optional, not quite sure about the crate-type, found this open rust issue: https://github.com/rust-lang/cargo/issues/4881

Cargo.toml
pjoe: This is now flagged by target_arch

kvark: This is better, but unfortunately it's still a subject to the problem with cargo-vendor, which will happily consider these dependencies to be needed for Gecko.
To make it clear, I think we should keep the wasm-naga as a separate project for now.

pjoe:
For reference: cargo vendor issue about cargo vendor for single platform: https://github.com/rust-lang/cargo/issues/7058

Cargo.toml
pjoe: agreed

pjoe:
Closing this PR. For now wasm work continues in https://github.com/pjoe/wasm-naga

PR #68, opened 2020-06-17
Minor cleanup pass, consistent line endings

kvark:
We are exposing arena types to avoid hitting https://github.com/rust-lang/rust/issues/73410
We are converting all line endings to Unix style. Previously GLSL had Windows style.

cc @pjoe this will likely introduce rebase problems, sorry!


pjoe:
@kvark no worries haven't gotten too far yet, but thanks for the heads up. I'd better hurry up and rebase 🎨

PR #69, opened 2020-06-17
Add glsl handling to convert sample

pjoe:
- Also fis extension check for spv and wgsl (missing .)


Napokue:
Shouldn't it be better to use the [Path struct](https://doc.rust-lang.org/std/path/struct.Path.html), so we can use the [Extension](https://doc.rust-lang.org/std/path/struct.Path.html#method.extension) function?

pjoe:
Slowly improving my noob rust skills :)

PR #77, opened 2020-06-19
Add member offsets and array strides

kvark:
Related to:
- https://github.com/gpuweb/gpuweb/issues/561
- https://github.com/gpuweb/gpuweb/issues/773

### DIFF
--- a/src/back/msl.rs
+++ b/src/back/msl.rs
@@ -451,7 +451,11 @@ impl<W: Write> Writer<W> {
                         write!(self.out, ".{}", COMPONENTS[index as usize])?;
                         Ok(MaybeOwned::Owned(crate::TypeInner::Scalar { kind, width }))
                     }
-                    crate::TypeInner::Array { base, size } => {
+                    crate::TypeInner::Array {
+                        base,
+                        size,
+                        stride: _,
+                    } => {
                         if let crate::ArraySize::Static(length) = size {
                             return Err(Error::AccessIndexExceedsStaticLength(index, length));
                         }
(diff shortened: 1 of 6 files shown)


src/front/spv.rs
```diff
@@ -1208,14 +1237,23 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+                    .ok_or(Error::MissingDecoration(spirv::Decoration::Offset))?,
             });
         }
         let inner = crate::TypeInner::Struct { members };
+        let decor = self.future_decor.remove(&id);
+        if let Some(ref decor) = decor {
+            if decor.block.is_some() {
+                // do nothing
+            }
+        }
```
Napokue, line 1251: What is the reason for doing this, as the code does not do anything with it?

src/back/msl.rs
```diff
@@ -451,7 +451,11 @@ impl<W: Write> Writer<W> {
                         write!(self.out, ".{}", COMPONENTS[index as usize])?;
                         Ok(MaybeOwned::Owned(crate::TypeInner::Scalar { kind, width }))
                     }
-                    crate::TypeInner::Array { base, size } => {
+                    crate::TypeInner::Array {
+                        base,
+                        size,
+                        stride: _,
```
Napokue, line 457: Does Metal have a stride or is it just that it is not planned yet to support strides for our Metal back-end?

src/front/spv.rs
kvark: the decoration applies to that type, so it needs to be used here, somehow
I just haven't figured out how exactly :)

src/back/msl.rs
kvark: we'll need to insert padding ourselves, that's a big TODO for now

Issue #59, opened 2020-06-08
Validation function for the IR


kvark:
The `Module` struct that our front-ends produce generally has some level of type-level encoding of the constraints, however many properties can only be checked at run-time. One big area of validation is type checking:
  - builtins have the expected types
  - constructors are using consistent values, e.g. #43
  - types generally agree between producer and consumer

Another area is access checking:
  - all the handles are actually valid
  - input globals can only be read
  - can't linearly sample integer textures
  - depth images can only be used with comparison samplers
  - and more!

Finally we need to enforce some of the conventions. For example, variables can either be stored as pointers (that are loaded/stored), or directly. There is currently no consistency, and whatever we choose would need to be enforced.

We have a [proc](https://github.com/gfx-rs/naga/tree/63ab925f9c0e5e39466863ce70ac002fe4a5c7c8/src/proc) module for end-point-independent processing. I think the `validator` sub-module could live there. The main task of it would be getting a module and making sure that it's valid. We'd be running it in CI tests to make sure that the module contents produced by the front-ends, or by different processors/transformers, are all valid.

PR #79, opened 2020-06-21
Add support for return value to SPIR-V backend

GabrielMajeri:
Adds support for `Statement::Return { Some(value) }` in SPIR-V backend.

I've added a test to `convert.rs`, which tries to parse the `glsl_phong_lighting.frag` file (with the `glsl` feature) and then convert it to SPIR-V.

After the changes, the test no longer panics on return with value statement. It does fail on `not implemented: Divide`, so I've set it to be ignored.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -885,10 +885,24 @@ impl Writer {
         output: &mut Vec<Instruction>,
     ) -> Instruction {
         match statement {
-            crate::Statement::Return { value: _ } => match function.return_type {
-                Some(_) => unimplemented!(),
-                None => Instruction::new(Op::Return),
-            },
+            crate::Statement::Return { value } => {
+                match function.return_type {
+                    Some(_) => {
+                        let value = value.unwrap();
+
+                        // Parse the expression
+                        let value_expression = &function.expressions[value];
+                        let (value_id, _) =
+                            self.parse_expression(ir_module, function, value_expression, output);
+
+                        // Construct the return value instruction
+                        let mut instruction = Instruction::new(Op::ReturnValue);
+                        instruction.add_operand(value_id);
+                        instruction
+                    }
+                    None => Instruction::new(Op::Return),
+                }
+            }
             crate::Statement::Store { pointer, value } => {
                 let mut instruction = Instruction::new(Op::Store);
 
--- a/tests/convert.rs
+++ b/tests/convert.rs
@@ -1,12 +1,25 @@
+#[cfg(feature = "glsl")]
+use spirv::ExecutionModel;
+
+fn load_test_data(name: &str) -> String {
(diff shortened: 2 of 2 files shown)


src/back/spv/writer.rs
```diff
@@ -885,10 +885,36 @@ impl Writer {
...
+                        let (value_id, value_ty) =
+                            self.parse_expression(ir_module, function, value_expression, output);
+
+                        // Check that the actual returned type matches the expected one
+                        let expected_ty = &ir_module.types[expected_ty].inner;
+                        assert_eq!(expected_ty, value_ty, "Type of returned expression does not match declared function return type");
+
+                        // Construct the return value instruction
+                        let mut instruction = Instruction::new(Op::ReturnValue);
+                        instruction.set_result(value_id);
```
GabrielMajeri, line 906: I'm not sure if `set_result` is correct here or it should be `add_operand`

Napokue: It should be set to `add_operand` indeed. 

The way I determine is by looking at the spec. The spec will say something like: `Result<id>`, so then you will use `set_result` function. When the spec mentions: `<id> Result Type`, you will use `set_type`. When both are not mentioned, it will be `set_operand`.

src/back/spv/writer.rs
```diff
@@ -885,10 +885,36 @@ impl Writer {
...
+                        );
+
+                        // Parse the expression
+                        let value_expression = &function.expressions[value];
+                        let (value_id, value_ty) =
+                            self.parse_expression(ir_module, function, value_expression, output);
+
+                        // Check that the actual returned type matches the expected one
+                        let expected_ty = &ir_module.types[expected_ty].inner;
+                        assert_eq!(expected_ty, value_ty, "Type of returned expression does not match declared function return type");
```
Napokue, line 902: I am not sure if we want add this assert_eq here, but I have to ping you @kvark. This would be a case for issue https://github.com/gfx-rs/naga/issues/59, as the back-end has to expect to get the correct data.

src/back/spv/writer.rs
```diff
@@ -885,10 +885,36 @@ impl Writer {
...
+                        // Construct the return value instruction
+                        let mut instruction = Instruction::new(Op::ReturnValue);
+                        instruction.set_result(value_id);
+                        instruction
+                    }
+                    None => {
+                        assert!(
+                            value.is_none(),
+                            "Trying to return value from function returning nothing"
+                        );
```
Napokue, line 913: Same for this one, @kvark. This validation should be done at IR level too I think?

kvark: Correct! The backend should *assume* that IR is valid. We don't want every backend doing its own validation.

GabrielMajeri:
Thanks for the suggestions, I've updated the PR.

PR #81, opened 2020-06-25
Remove spirv-headers use from IR, make it optional

### LINKED ISSUE #70: Make spirv-headers optional?
It's likely a small dependency, but still - Naga's IR doesn't have to include any SPIR-V types directly. We could detach the IR from spir-v and only enabled the latter when either front-end or back-end is requested as a feature.

kvark:
Closes #70 
I think it's much nicer now, since we can make IR tighter with respect to various enumeration modes we used to borrow from SPIR-V.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ bitflags = "1"
 fxhash = "0.2"
 log = "0.4"
 num-traits = "0.2"
-spirv = { package = "spirv_headers", version = "1.4.2" }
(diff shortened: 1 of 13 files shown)


Napokue:
Way better this way, thanks!

kvark:
@Napokue I'm going to assign you to review things next time, if you are interested:)
We are in dire need of reviewers here.

Napokue:
@kvark Sure no problem!

PR #83, opened 2020-06-25
Refactor error handling in GLSL frontend

kristoff3r:
I tried adding some stuff to the GLSL frontend, but I got annoyed with tracking down all the unwraps so I refactored the error handling. Let me know if you want any changes or if this is even a good direction to go in.

It also contains a few additions to the type checker and support for two unary operations, which is what I was adding when I got sidetracked. I can move them to another PR if you want.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ num-traits = "0.2"
 spirv = { package = "spirv_headers", version = "1.4.2", optional = true }
 glsl = { version = "4", optional = true }
 pomelo = { version = "0.1.4", optional = true }
+thiserror = "1.0"
 
 [features]
 glsl_preprocessor = ["glsl"]
--- a/src/front/glsl/mod.rs
+++ b/src/front/glsl/mod.rs
@@ -8,139 +8,68 @@ use glsl::{
     syntax::*,
 };
 use parser::{Token, TokenMetadata};
-use std::fmt;
 
 mod helpers;
 mod parser;
 
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum ErrorKind {
+    #[error("Unexpected token:\nexpected: {}\ngot: {}", expected.iter().map(|t| t.type_to_string()).collect::<Vec<_>>().join(" |"), got.token)]
     UnexpectedToken {
         expected: Vec<Token>,
         got: TokenMetadata,
     },
+    #[error("Unexpected word:\nexpected: {}\ngot: {got}", expected.join("|"))]
     UnexpectedWord {
         expected: Vec<&'static str>,
         got: String,
     },
-    ExpectedEOL {
-        got: TokenMetadata,
-    },
-    UnknownPragma {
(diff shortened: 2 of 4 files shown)


JCapucho:
Sorry for the lack of error handling it was a prototype to integrate very basic glsl support glad that you could extend it beware though that someone is working on a new parser because the glsl crate is very bloated for our needs

kristoff3r:
> Sorry for the lack of error handling it was a prototype to integrate very basic glsl support glad that you could extend it beware though that someone is working on a new parser because the glsl crate is very bloated for our needs

Shouldn't that work be pretty self-contained to the lexer/parser itself? If it isn't then I'll wait a bit before I do more work on the code generator.

JCapucho:
I don't know how many parts it will touch but I guess that some of the enums and structs might change

Cargo.toml
```diff
@@ -16,6 +16,7 @@ log = "0.4"
 num-traits = "0.2"
 spirv = { package = "spirv_headers", version = "1.4.2", optional = true }
 glsl = { version = "4", optional = true }
+thiserror = "1.0"
```
kvark, line 19: I feel uneasy about this. Question is basically, is this the best way of dealing with errors today?
Code seems nice. I'm just concerned about the build time and extra dependencies.

src/front/glsl/mod.rs
```diff
@@ -8,140 +8,69 @@ use glsl::{
...
 use parser::{Token, TokenMetadata};
-use std::fmt;
 
 mod helpers;
 mod parser;
 
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum ErrorKind {
+    #[error("Unexpected token:\nexpected: {}\ngot: {}", expected.iter().map(|t| t.type_to_string()).collect::<Vec<_>>().join(" |"), got.token)]
```
kvark, line 17: is it idiomatic to use newlines in error strings?

Cargo.toml
kvark: We discussed this with @﻿cwfitzgerald on "#wgpu:matrix.org" and concluded that rolling in `thiserror` dependency here, followed by `wgpu-core`, is the right way to go.

kristoff3r: `thiserror` is nice because it doesn't affect the APIs, it just reduces boilerplate. If you ever want to get rid of it due to compile times or some other reason, it is easy to do a mechanical translation back to `Display` and `From` traits.

src/front/glsl/mod.rs
kristoff3r: Not really I think, but I copied it verbatim from the previous Display trait. I haven't managed to trigger that particular error even when I tried, so I don't know how useful it is currently.

kvark:
I believe this is ready to land. Anything you want to do before hitting "Ready to Review" button?

kristoff3r:
I fixed the merge conflict, so it should be good to go.

PR #84, opened 2020-06-26
Spirv backend - Extract instructions to own functions

Napokue:
- Extracted all the instructions to their own functions. 
- Added all the instructions in order of the spec with comments used to categorize the instructions.
- Categorized the whole writer.rs to have the following order:
    - Lookups
    - Instructions
    - Parsing
    - Writing

To follow up, I will start working on unit testing all these instructions. This will be another PR.

Benefit of these changes are that it all these instructions can potentially also be used in the spirv front-end. Maybe it is an idea to create a shared module for front- and back-ends

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1,7 +1,12 @@
 /*! Standard Portable Intermediate Representation (SPIR-V) backend !*/
 use super::{helpers, Instruction, LogicalLayout, PhysicalLayout, WriterFlags};
-use crate::{FastHashMap, FastHashSet};
+use crate::{FastHashMap, FastHashSet, ImageFlags, VectorSize};
 use spirv::{Op, Word};
+
+enum Signedness {
+    Unsigned = 0,
+    Signed = 1,
+}
 
 trait LookupHelper<T> {
     type Target;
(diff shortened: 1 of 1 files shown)


src/back/spv/writer.rs
```diff
@@ -87,12 +92,166 @@ impl Writer {
...
+        &self,
+        arena: &crate::Arena<crate::Type>,
+        kind: crate::ScalarKind,
+        width: u8,
+    ) -> crate::Handle<crate::Type> {
+        let mut scalar_handle = None;
+        for (handle, ty) in arena.iter() {
+            match ty.inner {
+                crate::TypeInner::Scalar {
+                    kind: _kind,
```
kvark, line 198: let's only use `_` prefixes for unused variables. Here, we can add `_` postfix instead

src/back/spv/writer.rs
```diff
@@ -87,12 +92,166 @@ impl Writer {
...
+                    width: _width,
+                } => {
+                    if kind == _kind && width == _width {
+                        scalar_handle = Some(handle);
```
kvark, line 202: why not just `return handle`?

src/back/spv/writer.rs
```diff
@@ -450,147 +838,92 @@ impl Writer {
...
-                instruction.set_result(id);
-
                 let ty = &ir_module.types[constant.ty];
-                match ty.inner {
+                let values = match ty.inner {
                     crate::TypeInner::Scalar { kind: _, width } => match width {
-                        32 => {
-                            instruction.add_operand(val as u32);
-                        }
+                        32 => vec![val as u32],
```
kvark, line 862: nit: in cases like this you can avoid heap allocation by just returning slices, i.e. `&[val as u32][..]`

src/back/spv/writer.rs
```diff
@@ -450,147 +838,92 @@ impl Writer {
...
-                        32 => {
-                            instruction.add_operand((val as f32).to_bits());
-                        }
+                        32 => vec![(val as f32).to_bits()],
                         64 => {
                             let bits = f64::to_bits(val);
                             let (low, high) = ((bits >> 32) as u32, bits as u32);
-                            instruction.add_operand(low);
-                            instruction.add_operand(high);
+                            vec![low, high]
```
kvark, line 884: same here and in other places - let's be cautious about heap allocations

src/back/spv/writer.rs
Napokue: Renamed it to `lookup_kind` and `lookup_width`.

src/back/spv/writer.rs
Napokue: I could, but the goal of this function would be determining if a scalar exists in the IR, and ultimately create the scalar if it is not present in the IR. This would solve #85. To fix this, I have to refactor the whole Lookup architecture I created in this first version.

I experimented with a new lookup architecture in boids. I think best option would be to not solve it now, but to do this _**before**_ I am going to work on the boids example and PR this separately.

src/back/spv/writer.rs
Napokue: Yes thanks for keep telling me that, I still tend to use vectors everywhere, but I am learning it 😄

PR #87, opened 2020-06-29
New GLSL frontend using pomelo

pjoe:
Work in progress, writing a new GLSL parser using [pomelo.](https://docs.rs/pomelo)

Goal is to reduce output size (especially for wasm), as well as making the parser read better against the spec.

Status:
- [x] port lexer
- [x] syntax parser using pomelo
- [ ] error handling
- [ ] generate parse tree
- [ ] tests

For now done i separate `glsl_new`module.
Also added `PREFER_GLSL_NEW` env var that you can set to 1, to to try this in the convert example.


src/front/glsl_new/error.rs
```diff
@@ -0,0 +1,44 @@
+use std::fmt;
+use std::io;
+
+#[derive(Debug)]
+pub enum ErrorKind {
+    InvalidInput,
+    IoError { error: io::Error },
```
kvark, line 7: nit: any reason to have the named field here? seems very straightforward to just have `IoError(io::Error)`

src/front/glsl_new/error.rs
```diff
@@ -0,0 +1,44 @@
+use std::fmt;
```
kvark, line 1: nit: let's use the include style of other modules, i.e. one statement per top-level use. In this case, `use std::{fmt, io}`

src/front/glsl_new/lex.rs
```diff
@@ -0,0 +1,314 @@
...
+            self.input = String::from(rest);
+            let meta = token.extra_mut();
+            let end = meta.chars.end;
+            meta.line = self.line;
+            meta.chars.start += self.offset;
+            meta.chars.end += self.offset;
+            self.offset += end;
+            Some(token)
+        } else {
+            match self.lines.next() {
```
kvark, line 280: nit: can do `let (line, input) = self.lines.next()?;` and proceed without a match

src/front/glsl_new/error.rs
pjoe: Done

src/front/glsl_new/error.rs
pjoe: Done

src/front/glsl_new/lex.rs
pjoe: Done

Cargo.toml
```diff
@@ -16,6 +16,7 @@ log = "0.4"
 num-traits = "0.2"
 spirv = { package = "spirv_headers", version = "1.4.2", optional = true }
 glsl = { version = "4", optional = true }
+pomelo = "0.1.4"
```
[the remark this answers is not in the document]
pjoe: Ok, I put everything into a `glsl-new` feature, so all optional now :)

Made the logic in convert example a bit complex, but that will go away once mature enough to replace 'old' glsl front.

Issue #85, opened 2020-06-26
Scalar type in Vector is not added to the IR


Napokue:
Converting this WGSL:
```
[[location 0]] var<in> a_pos : vec2<f32>;
[[location 0]] var<out> v_uv : vec2<f32>;

fn main_vert() -> void {
  v_uv = a_pos * a_pos;
  return;
}
entry_point vertex as "main" = main_vert;
```

To the following IR module (omitted most things):
```
types: Arena {
        data: [
            Type {
                name: None,
                inner: Vector {
                    size: Bi,
                    kind: Float,
                    width: 32,
                },
            },
        ],
    },
```

As you can see, the Float scalar is not created, the SPIR-V back-end has to check if the scalar type already exists, if not, create it.

The reason the [quad.wgsl](https://github.com/gfx-rs/naga/blob/master/test-data/quad.wgsl) works, is because at the top of the file is the following line declared:
`const c_scale: f32 = 1.2;`

This causes the front-end to create the Float scalar for the Vectors that need this.

Napokue:
Talked with @kvark at Matrix about this, the problems needs to be fixed in the back-ends. The back-end has to check if the scalar type is available, if not, create it.

kvark:
More specifically, it's a problem with SPIR-V backend. I don't think anything else need to adjust for this.

pjoe:
Also ran into this using glsl frontend

pjoe:
Wondering if this would cause issues for other backends too? Maybe we should have things like this handle in middle-out - ehh middle-end 😄

Napokue:
Fixed in #88, closing issue.

PR #88, opened 2020-06-30
Spirv lookup refactor

Napokue:
Fix for #85.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1,45 +1,37 @@
 /*! Standard Portable Intermediate Representation (SPIR-V) backend !*/
 use super::{helpers, Instruction, LogicalLayout, PhysicalLayout, WriterFlags};
-use crate::{FastHashMap, FastHashSet, ImageFlags, VectorSize};
+use crate::{Bytes, FastHashMap, FastHashSet, ImageFlags, VectorSize};
 use spirv::{Op, Word};
+use std::collections::hash_map::Entry;
 
-const BITS_PER_BYTE: u8 = 8;
+const BITS_PER_BYTE: Bytes = 8;
 
 enum Signedness {
     Unsigned = 0,
     Signed = 1,
 }
 
-trait LookupHelper<T> {
-    type Target;
-    fn lookup_id(&self, handle: crate::Handle<T>) -> Option<Word>;
-    fn lookup_handle(&self, word: Word) -> Option<crate::Handle<T>>;
+#[derive(Debug, PartialEq, Hash, Eq, Copy, Clone)]
+enum LocalType {
+    Scalar {
+        kind: crate::ScalarKind,
+        width: Bytes,
+    },
+    Vector {
+        size: crate::VectorSize,
+        kind: crate::ScalarKind,
+        width: Bytes,
+    },
+    Pointer {
+        base: crate::Handle<crate::Type>,
+        class: spirv::StorageClass,
+    },
 }
 
-impl<T> LookupHelper<T> for FastHashMap<Word, crate::Handle<T>> {
-    type Target = T;
-
-    fn lookup_id(&self, handle: crate::Handle<T>) -> Option<Word> {
-        let mut word = None;
-        for (k, v) in self.iter() {
-            if *v == handle {
-                word = Some(*k);
-                break;
-            }
-        }
-        word
-    }
-
-    fn lookup_handle(&self, word: u32) -> Option<crate::Handle<T>> {
-        let mut handle = None;
-        for (k, v) in self.iter() {
-            if *k == word {
-                handle = Some(*v);
-                break;
-            }
-        }
-        handle
-    }
+#[derive(Debug, PartialEq, Hash, Eq, Copy, Clone)]
+enum LookupType {
+    Handle(crate::Handle<crate::Type>),
+    Local(LocalType),
 }
 
 fn map_dim(dim: crate::ImageDimension) -> spirv::Dim {
@@ -51,7 +43,7 @@ fn map_dim(dim: crate::ImageDimension) -> spirv::Dim {
     }
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone, Hash, Eq)]
 struct LookupFunctionType {
     parameter_type_ids: Vec<Word>,
     return_type_id: Word,
@@ -66,11 +58,11 @@ pub struct Writer {
     annotations: Vec<Instruction>,
     writer_flags: WriterFlags,
     void_type: Option<u32>,
-    lookup_type: FastHashMap<Word, crate::Handle<crate::Type>>,
-    lookup_function: FastHashMap<Word, crate::Handle<crate::Function>>,
-    lookup_function_type: FastHashMap<Word, LookupFunctionType>,
-    lookup_constant: FastHashMap<Word, crate::Handle<crate::Constant>>,
-    lookup_global_variable: FastHashMap<Word, crate::Handle<crate::GlobalVariable>>,
+    lookup_type: FastHashMap<LookupType, Word>,
+    lookup_function: FastHashMap<crate::Handle<crate::Function>, Word>,
+    lookup_function_type: FastHashMap<LookupFunctionType, Word>,
+    lookup_constant: FastHashMap<crate::Handle<crate::Constant>, Word>,
+    lookup_global_variable: FastHashMap<crate::Handle<crate::GlobalVariable>, Word>,
 }
 
 impl Writer {
@@ -103,17 +95,13 @@ impl Writer {
         }
     }
 
-    fn get_type_id(
-        &mut self,
-        arena: &crate::Arena<crate::Type>,
-        handle: crate::Handle<crate::Type>,
-    ) -> Word {
-        match self.lookup_type.lookup_id(handle) {
-            Some(word) => word,
-            None => {
-                let (instruction, id) = self.parse_type_declaration(arena, handle);
-                instruction.to_words(&mut self.logical_layout.declarations);
-                id
+    fn get_type_id(&mut self, arena: &crate::Arena<crate::Type>, lookup_ty: LookupType) -> Word {
+        if let Entry::Occupied(e) = self.lookup_type.entry(lookup_ty) {
+            *e.get()
+        } else {
+            match lookup_ty {
+                LookupType::Handle(handle) => self.write_type_declaration_arena(arena, handle),
+                LookupType::Local(local_ty) => self.write_type_declaration_local(arena, local_ty),
             }
         }
     }
@@ -123,10 +111,10 @@ impl Writer {
         handle: crate::Handle<crate::Constant>,
         ir_module: &crate::Module,
     ) -> Word {
-        match self.lookup_constant.lookup_id(handle) {
-            Some(word) => word,
-            None => {
-                let (instruction, id) = self.parse_constant_type(handle, ir_module);
+        match self.lookup_constant.entry(handle) {
+            Entry::Occupied(e) => *e.get(),
+            _ => {
+                let (instruction, id) = self.write_constant_type(handle, ir_module);
                 instruction.to_words(&mut self.logical_layout.declarations);
                 id
             }
@@ -139,11 +127,11 @@ impl Writer {
         global_arena: &crate::Arena<crate::GlobalVariable>,
         handle: crate::Handle<crate::GlobalVariable>,
     ) -> Word {
-        match self.lookup_global_variable.lookup_id(handle) {
-            Some(word) => word,
-            None => {
+        match self.lookup_global_variable.entry(handle) {
+            Entry::Occupied(e) => *e.get(),
+            _ => {
                 let global_variable = &global_arena[handle];
-                let (instruction, id) = self.parse_global_variable(arena, global_variable, handle);
+                let (instruction, id) = self.write_global_variable(arena, global_variable, handle);
                 instruction.to_words(&mut self.logical_layout.declarations);
                 id
             }
@@ -156,7 +144,7 @@ impl Writer {
         arena: &crate::Arena<crate::Type>,
     ) -> Word {
         match ty {
-            Some(handle) => self.get_type_id(arena, handle),
+            Some(handle) => self.get_type_id(arena, LookupType::Handle(handle)),
             None => match self.void_type {
                 Some(id) => id,
                 None => {
@@ -177,47 +165,33 @@ impl Writer {
         class: spirv::StorageClass,
     ) -> Word {
         let ty = &arena[handle];
-        let type_id = self.get_type_id(arena, handle);
+        let ty_id = self.get_type_id(arena, LookupType::Handle(handle));
         match ty.inner {
-            crate::TypeInner::Pointer { .. } => type_id,
+            crate::TypeInner::Pointer { .. } => ty_id,
             _ => {
-                let pointer_id = self.generate_id();
-                let instruction = self.instruction_type_pointer(pointer_id, class, type_id);
-                instruction.to_words(&mut self.logical_layout.declarations);
-
-                /* TODO
-                    Not able to lookup Pointer, because there is no Handle in the IR for it.
-                    Idea would be to not have any handles at all in the lookups, so we aren't bound
-                    to the IR. We can then insert, like here runtime values to the lookups
-                */
-                // self.lookup_type.insert(pointer_id, global_variable.ty);
-                pointer_id
-            }
-        }
-    }
-
-    fn find_scalar_handle(
-        &self,
-        arena: &crate::Arena<crate::Type>,
-        kind: crate::ScalarKind,
-        width: u8,
-    ) -> crate::Handle<crate::Type> {
-        let mut scalar_handle = None;
-        for (handle, ty) in arena.iter() {
-            match ty.inner {
-                crate::TypeInner::Scalar {
-                    kind: lookup_kind,
-                    width: lookup_width,
-                } => {
-                    if kind == lookup_kind && width == lookup_width {
-                        scalar_handle = Some(handle);
-                        break;
+                match self
+                    .lookup_type
+                    .entry(LookupType::Local(LocalType::Pointer {
+                        base: handle,
+                        class,
+                    })) {
+                    Entry::Occupied(e) => *e.get(),
+                    _ => {
+                        let pointer_id = self.generate_id();
+                        let instruction = self.instruction_type_pointer(pointer_id, class, ty_id);
+                        instruction.to_words(&mut self.logical_layout.declarations);
+                        self.lookup_type.insert(
+                            LookupType::Local(LocalType::Pointer {
+                                base: handle,
+                                class,
+                            }),
+                            pointer_id,
+                        );
+                        pointer_id
                     }
                 }
-                _ => continue,
             }
         }
-        scalar_handle.unwrap()
     }
 
     ///
@@ -293,10 +267,8 @@ impl Writer {
         ir_module: &crate::Module,
     ) -> Instruction {
         let mut instruction = Instruction::new(Op::EntryPoint);
-        let function_id = self
-            .lookup_function
-            .lookup_id(entry_point.function)
-            .unwrap();
+
+        let function_id = *self.lookup_function.get(&entry_point.function).unwrap();
 
         let exec_model = match entry_point.stage {
             crate::ShaderStage::Vertex => spirv::ExecutionModel::Vertex,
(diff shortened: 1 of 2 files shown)


src/back/spv/writer.rs
```diff
@@ -743,13 +792,13 @@ impl Writer {
...
-                self.lookup_type.insert(id, handle);
+                self.lookup_type.insert(id, lookup_type);
             }
             crate::TypeInner::Vector { size, kind, width } => {
                 let scalar_handle = self.find_scalar_handle(arena, kind, width);
-                let scalar_id = self.get_type_id(arena, scalar_handle);
+                let scalar_id = self.get_type_id(arena, &LookupType::Handle(scalar_handle));
                 instruction = self.instruction_type_vector(id, scalar_id, size);
-                self.lookup_type.insert(id, handle);
+                self.lookup_type.insert(id, lookup_type);
```
Napokue, line 801: This still gives a compile error, because it expects a value type instead of the reference type. Problem is if I dereference it, I get the error: "Cannot move".

src/back/spv/writer.rs
```diff
@@ -724,9 +769,13 @@ impl Writer {
...
         &mut self,
         arena: &crate::Arena<crate::Type>,
-        handle: crate::Handle<crate::Type>,
+        lookup_type: &LookupType<crate::Type>,
     ) -> (Instruction, Word) {
-        let ty = &arena[handle];
+        let ty = match lookup_type {
+            LookupType::Handle(handle) => &arena[*handle],
+            LookupType::Local(value) => &value,
+        };
```
Napokue, line 777: This feels pretty wrong to do it like this. Would it be better to let both LookupTypes implement a trait, where I can retrieve type T from, instead of matching and trying to get the Type (in this case).

src/back/spv/writer.rs
```diff
@@ -724,9 +769,13 @@ impl Writer {
     fn parse_type_declaration(
         &mut self,
         arena: &crate::Arena<crate::Type>,
-        handle: crate::Handle<crate::Type>,
+        lookup_type: &LookupType<crate::Type>,
```
Napokue, line 772: I can change this to a value type, and deference it at the inserts in this function. Problem is that in `get_type_id` I get the "Cannot move" error.

src/back/spv/writer.rs
```diff
@@ -724,9 +769,13 @@ impl Writer {
...
     ) -> (Instruction, Word) {
-        let ty = &arena[handle];
+        let ty = match lookup_type {
```
kvark, line 774: please don't match on a reference, instead let's match the type `match *lookup_type`

src/back/spv/writer.rs
kvark: it's fine if we don't do this many times :)

src/back/spv/writer.rs
kvark: it looks like you are doing this `self.lookup_type.insert(id, lookup_type);` pretty much uniformly in this function.
Could it be moved outside of `match`?
And then, since this function needs it by value, we likely need to change the parameter to be passed by value

src/back/spv/writer.rs
kvark: please tell more!

src/back/spv/writer.rs
kristoff3r: That doesn't work because you don't have ownership over lookup_type, only a shared reference. Handles implement the Copy trait which is why it worked before, because it automatically copied the value.

While I haven't looked much at the code it looks like you actually do need to copy the lookup type here, but that would require `crate::Type` to be Clone and that you write `LookupType<T: Clone>` everywhere, which is messy. All the problems stem from the `LocalType::Local(T)` part, so I would look at another way to represent that.

kvark: Did you consider `Local(&'a T)` instead?

Napokue: I haven't tried using a lifetime yet, I am still exploring the options. But I indeed stumbled across having to implement Clone for all generics that will be using this.

kvark: Let's try to avoid IR changes for this an see if `&'a T` would be sufficient

src/back/spv/writer.rs
```diff
@@ -8,6 +8,51 @@ enum Signedness {
     Signed = 1,
 }
 
+enum LookupType<T> {
+    Handle(crate::Handle<T>),
+    Local(T),
+}
+
+impl<T> PartialEq for LookupType<T>
```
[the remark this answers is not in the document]
Napokue: Yes, because when I don't implement this, I get the error: "Binary operations '==' cannot be applied to type '&back::spv::LookupType<>'

src/back/spv/writer.rs
```diff
@@ -172,52 +207,73 @@ impl Writer {
...
+                }
+            }
+        } else {
+            let new_ty: &'a crate::Type = &crate::Type {
+                name: None,
+                inner: crate::TypeInner::Scalar {
+                    width,
+                    kind
+                },
+            };
```
Napokue, line 270: I try to create a new type here, but I am not sure why it does not work. Getting this lifetime error:
image

kvark: You are constructing the type on the stack, and then getting a reference to it. This doesn't fulfill "`a".

I think your current `LookupType::Local` variant is dysfunctional: the only thing in "'a" that owns the types is the module's arenat, and if you have the type there, you might as well use `LookupType::Handle`.

src/back/spv/writer.rs
```diff
@@ -10,6 +10,18 @@ enum Signedness {
     Signed = 1,
 }
 
+#[derive(Debug)]
+enum LocalType {
```
kvark, line 14: we probably want vectors as well here, since it's implicitly required by the matrix types in our IR

src/back/spv/writer.rs
```diff
@@ -66,7 +78,7 @@ pub struct Writer {
     annotations: Vec<Instruction>,
     writer_flags: WriterFlags,
     void_type: Option<u32>,
-    lookup_type: FastHashMap<Word, crate::Handle<crate::Type>>,
+    lookup_type_new: FastHashMap<Word, LookupType>,
```
kvark, line 81: is the name change accidental?

Napokue, line 81: Oops have to of course rename this to `lookup_type`.

src/back/spv/writer.rs
```diff
@@ -108,13 +120,28 @@ impl Writer {
...
+
+        if let Some(id) = self
+            .lookup_type_new
+            .iter()
+            .find(|&(_, v)| match v {
+                LookupType::Handle(lookup_handle) => handle == *lookup_handle,
+                LookupType::Local(local_ty) => {
+                    match local_ty {
+                        LocalType::Scalar(ty_inner) => &ty.inner == ty_inner,
+                        // TODO Pointers
```
Napokue, line 133: Had an equivalent check for `LocalType::Pointer`, but it caused a weird behavior that a `OpTypePointer` referencing a result type of `OpTypePointer` would be created. Which is not allowed in SPIR-V.

src/back/spv/writer.rs
```diff
@@ -177,21 +204,35 @@ impl Writer {
...
+                                lookup_inner == &ty.inner && &class == lookup_class
+                            }
+                            _ => false,
+                        },
+                        _ => false,
+                    })
+                    .map(|(k, _)| k)
+                {
+                    *id
+                } else {
```
Napokue, line 226: Not sure I like this code too much, it is quite big for what it has to do.

src/back/spv/writer.rs
```diff
@@ -760,33 +831,61 @@ impl Writer {
...
+                    self.lookup_type_new.insert(
+                        id,
+                        LookupType::Local(LocalType::Scalar(crate::TypeInner::Scalar {
+                            kind,
+                            width,
+                        })),
+                    );
+                    instruction.to_words(&mut self.logical_layout.declarations);
+                    id
+                };
```
Napokue, line 852: Will create a new function of this part, because it is also used for Matrices.

src/back/spv/writer.rs
kvark, line 81: I believe all of the lookup tables here need to be reversed (were they copied from the front-end?). I.e. you should never need to lookup IR stuff *by* a spirv ID. You need the opposite: given IR handles and stuff, be able to produce spirv IDs that you are saving to the file. So this table (and all others) need to look like this:
```rust
lookup_type: FastHashMap<LocalType, Word>,
```

src/back/spv/writer.rs
```diff
@@ -108,13 +120,28 @@ impl Writer {
...
-                let (instruction, id) = self.parse_type_declaration(arena, handle);
-                instruction.to_words(&mut self.logical_layout.declarations);
-                id
-            }
+        let ty = &arena[handle];
+
+        if let Some(id) = self
+            .lookup_type_new
+            .iter()
+            .find(|&(_, v)| match v {
```
kvark, line 128: once we reverse the table, the iteration here will go away, and a single lookup remain

src/back/spv/writer.rs
```diff
@@ -732,6 +773,36 @@ impl Writer {
     /// Primitive Instructions
     ///
 
+    fn parse_type_declaration_local(&mut self, local_ty: LocalType) -> (Instruction, Word) {
```
kvark, line 776: we shouldn't have the "parse" in the names, since the backend isn't supposed to parse anything :)
consider "write" or "convert", or something along the lines

src/back/spv/writer.rs
Napokue: True, I already commented on it, forgot to rename it back to `lookup_type`

src/back/spv/writer.rs
Napokue: I agree with the idea of reversing the Lookups, but how will this iteration go away? Maybe I am not yet understanding what you mean, but as far as I can tell is that we still have to have the exact same code, but instead of matching against the value, we match against the key?

src/back/spv/writer.rs
Napokue: Completely agreed, `write` it is, as we want to have the same terminology as the metal back-end 👍

src/back/spv/writer.rs
Napokue: Good thing you bring this up. Today I have extensively tested the matrix capabilities, but the SPIR-V output is totally not correct yet. I am already working on fixing this, and adding the Vector `LocalType`.

src/back/spv/writer.rs
kvark: The way I think this would work is that you construct the `LookupType::Handle(handle)` as a key, and do `lookup_types.entry(key)`. If it's `Occupied`, just return the value, which is SPIR-V ID. If it's vacant, do the `write_type_declaration` stuff, which produces a new ID, and fill the vacant entry with this.

src/back/spv/writer.rs
Napokue: Done this, as result, I had to modify the lib.rs to include `Hash` and `Eq` on a couple of structs.

src/back/spv/writer.rs
```diff
@@ -10,38 +11,19 @@ enum Signedness {
     Signed = 1,
 }
 
-#[derive(Debug)]
+#[derive(Hash, Eq, PartialEq, Debug)]
 enum LocalType {
     Scalar(crate::TypeInner),
```
kvark, line 16: why are we storing `TypeInner` here? I.e. if we are going to store `TypeInner`, we wouldn't need `LocalType` at all :) The idea was to have these types fully specified, i.e.
```rust
enum LocalType {
  Scalar { kind, width },
  Vector { size, kind, width },
  Pointer { base, class },
}
```

src/lib.rs
```diff
@@ -181,7 +181,7 @@ pub struct Type {
 
 /// Enum with additional information, depending on the kind of type.
 // Clone is used only for error reporting and is not intended for end users
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Hash, Eq)]
```
kvark, line 184: I think with the changes to `LocalType` that I suggested, these derives are no longer going to be needed

src/lib.rs
```diff
@@ -134,15 +134,15 @@ pub enum MemberOrigin {
 
 /// Member of a user-defined structure.
 // Clone is used only for error reporting and is not intended for end users
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Hash, Eq)]
```
kvark, line 137: these shouldn't be needed either

src/lib.rs
```diff
@@ -89,7 +89,7 @@ pub type Bytes = u8;
 
 /// Number of components in a vector.
 #[repr(u8)]
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, PartialEq, Hash, Eq)]
```
Napokue, line 92: Unfortunately still needed for VectorSize and ScalarKind.

src/back/spv/writer.rs
```diff
@@ -742,51 +865,33 @@ impl Writer {
 
         let instruction;
 
-        match ty.inner {
+        match ty.inner.clone() {
```
kvark, line 868: We shouldn't need cloning here. What was the problem that made you clone?

src/back/spv/writer.rs
```diff
@@ -1145,20 +1230,24 @@ impl Writer {
                             crate::TypeInner::Vector { size, kind, width } => {
                                 vector_id = Some(left_id);
                                 for (k, v) in self.lookup_type.iter() {
```
kvark, line 1232: why are we not just issuing a lookup into the map? we shouldn't need to iterate the arenas anywhere more than once, really

src/back/spv/writer.rs
```diff
@@ -1169,20 +1258,24 @@ impl Writer {
                             crate::TypeInner::Vector { size, kind, width } => {
                                 vector_id = Some(right_id);
                                 for (k, v) in self.lookup_type.iter() {
```
kvark, line 1260: same here

src/lib.rs
kvark: yes, sure :)

Napokue:
No worries at all, we are working towards the same goal. I will look into it later today

src/back/spv/writer.rs
```diff
@@ -732,7 +759,103 @@ impl Writer {
...
+        match local_ty {
+            LocalType::Scalar { kind, width } => (self.write_scalar(id, kind, width), id),
+            LocalType::Vector { size, kind, width } => (
+                {
+                    let scalar_id = self.get_or_create_scalar(arena, kind, width);
+                    self.instruction_type_vector(id, scalar_id, size)
+                },
+                id,
+            ),
+            _ => unimplemented!(),
```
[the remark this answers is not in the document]
Napokue: Correct, `Pointer` is the only remaining option. I will specifically specify `LocalType::Pointer`, is indeed better in my opinion.

src/back/spv/writer.rs
Napokue: True, I think I got an intellisense error that `TypeInner` could not be moved, but when I remove it now, the "error" is gone. Probably a caching bug in my IDE.

src/back/spv/writer.rs
Napokue: I completely agree, this is probably the most dirtiest piece of code in this back-end. I will take a look to fix it, so it won't iterate the arena again. This piece of code will be completely refactored in the future I presume, as it only supports very basic cases.

Napokue: Oh and by the way, as I see now, LocalType's arent even supported here. I will fix that too.

src/back/spv/writer.rs
```diff
@@ -762,13 +762,16 @@ impl Writer {
     fn write_scalar(&self, id: Word, kind: crate::ScalarKind, width: u8) -> Instruction {
         match kind {
             crate::ScalarKind::Sint => {
-                self.instruction_type_int(id, (width * BITS_PER_BYTE) as u32, Signedness::Signed)
+                let bits = (width * BITS_PER_BYTE) as u32;
```
kvark, line 765: we should move this outside of the `match`

src/back/spv/writer.rs
```diff
@@ -865,15 +858,21 @@ impl Writer {
 
         let instruction;
 
-        match ty.inner.clone() {
+        match ty.inner {
             crate::TypeInner::Scalar { kind, width } => {
                 instruction = self.write_scalar(id, kind, width);
```
kvark, line 863: I think this should use `get_or_create_scalar`

src/back/spv/writer.rs
```diff
@@ -865,15 +858,21 @@ impl Writer {
...
                 self.lookup_type.insert(LookupType::Handle(handle), id);
+                self.lookup_type
+                    .insert(LookupType::Local(LocalType::Scalar { kind, width }), id);
             }
             crate::TypeInner::Vector { size, kind, width } => {
                 let scalar_id = self.get_or_create_scalar(arena, kind, width);
                 instruction = self.instruction_type_vector(id, scalar_id, size);
```
kvark, line 870: this should use `get_or_create_vector`

src/back/spv/writer.rs
```diff
@@ -1226,57 +1232,33 @@ impl Writer {
...
-                                                {
-                                                    result_type_id = Some(*v);
-                                                    break;
-                                                }
-                                            }
-                                            _ => continue,
-                                        }
-                                    }
+                                if let Entry::Occupied(e) =
+                                    self.lookup_type.entry(LookupType::Local(LocalType::Vector {
```
kvark, line 1239: we don't have to use the Entry API here, we should just do `if let Some(id) = self.lookup_type.get(...)`

src/back/spv/writer.rs
```diff
@@ -1226,57 +1232,33 @@ impl Writer {
...
```
kvark, line 1255: same here

src/back/spv/writer.rs
Napokue: I wouldn't do that. I also believed that to be doable, but `crate:ScalarKind::Bool` doesn't use this. 

Question is: do we want to move it out, so we only have it once in the code, but let one match not use it. We can also choose to not move it, like this, but having 3 times the same variable.

kvark: it's fine if one of the match arms don't use it

src/back/spv/writer.rs
Napokue: No that is not possible, `get_or_create_scalar` and `get_or_create_vector` only returns a Word/Id, not an instruction.

src/back/spv/writer.rs
```diff
@@ -10,36 +11,27 @@ enum Signedness {
...
 
-trait LookupHelper<T> {
-    type Target;
-    fn lookup_id(&self, handle: crate::Handle<T>) -> Option<Word>;
-    fn lookup_handle(&self, word: Word) -> Option<crate::Handle<T>>;
+#[derive(Debug, PartialEq, Hash, Eq, Copy, Clone)]
+enum LocalType {
+    Scalar {
+        kind: crate::ScalarKind,
+        width: u8,
```
kvark, line 18: let's use `Bytes` typedef like IR does

src/back/spv/writer.rs
```diff
@@ -103,17 +95,13 @@ impl Writer {
...
-            Some(word) => word,
-            None => {
-                let (instruction, id) = self.parse_type_declaration(arena, handle);
-                instruction.to_words(&mut self.logical_layout.declarations);
-                id
+    fn get_type_id(&mut self, arena: &crate::Arena<crate::Type>, lookup_ty: LookupType) -> Word {
+        if let Entry::Occupied(e) = self.lookup_type.entry(lookup_ty) {
+            *e.get()
+        } else {
+            match lookup_ty {
```
kvark, line 102: when we add a new type, we have an ID. So we should write down it in the map, i.e.
```rust
match self.lookup_type(lookup_ty) {
  Entry::Occupied(e) => e.get(),
  Entry::Vacant(e) => {
    let id = ...;
    *e.insert(id)
  }
}
```

src/back/spv/writer.rs
```diff
@@ -123,10 +111,10 @@ impl Writer {
...
         ir_module: &crate::Module,
     ) -> Word {
-        match self.lookup_constant.lookup_id(handle) {
-            Some(word) => word,
-            None => {
-                let (instruction, id) = self.parse_constant_type(handle, ir_module);
+        match self.lookup_constant.entry(handle) {
+            Entry::Occupied(e) => *e.get(),
+            _ => {
+                let (instruction, id) = self.write_constant_type(handle, ir_module);
```
kvark, line 117: same here, we should add it to the map with `e.insert(id)`

src/back/spv/writer.rs
```diff
@@ -139,11 +127,11 @@ impl Writer {
         global_arena: &crate::Arena<crate::GlobalVariable>,
         handle: crate::Handle<crate::GlobalVariable>,
     ) -> Word {
-        match self.lookup_global_variable.lookup_id(handle) {
-            Some(word) => word,
-            None => {
+        match self.lookup_global_variable.entry(handle) {
```
kvark, line 130: and here

src/back/spv/writer.rs
```diff
@@ -177,47 +165,33 @@ impl Writer {
...
-                        break;
+                match self
+                    .lookup_type
+                    .entry(LookupType::Local(LocalType::Pointer {
+                        base: handle,
+                        class,
+                    })) {
+                    Entry::Occupied(e) => *e.get(),
+                    _ => {
+                        let pointer_id = self.generate_id();
```
kvark, line 180: and here, we should insert it into the map

src/back/spv/writer.rs
```diff
@@ -732,61 +704,93 @@ impl Writer {
...
+        let instruction = match local_ty {
+            LocalType::Scalar { kind, width } => self.write_scalar(id, kind, width),
+            LocalType::Vector { size, .. } => {
+                let scalar_id = self.get_type_id(arena, LookupType::Local(local_ty));
+                self.instruction_type_vector(id, scalar_id, size)
+            }
+            LocalType::Pointer { .. } => unimplemented!(),
+        };
+
+        self.lookup_type.insert(LookupType::Local(local_ty), id);
```
kvark, line 745: I see what you did here. Let's cut the loop. Instead of doing `match (entry) -> call function -> insert entry` just do `match (entry) -> insert entry` as we have the `Entry::Vacant` there.

src/back/spv/writer.rs
```diff
@@ -103,30 +96,36 @@ impl Writer {
...
+        let id = match self.lookup_type.entry(lookup_ty) {
+            Entry::Occupied(e) => *e.get(),
+            Entry::Vacant(e) => {
+                let id = match lookup_ty {
+                    LookupType::Handle(handle) => self.write_type_declaration_arena(arena, handle, &mut sink),
+                    LookupType::Local(local_ty) => self.write_type_declaration_local(arena, local_ty),
+                };
+                *e.insert(id)
             }
+        };
```
Napokue, line 111: I am getting the "cannot borrow `*self` as mutable more than once at a time" over here. I understand why, but I am not sure yet how to resolve this yet. 

The functions have to be mutable, even if we extract the part `let id = self.generate_id()` from these functions and use it here.

src/back/spv/writer.rs
```diff
@@ -103,30 +96,36 @@ impl Writer {
...
+
+        for (k, v) in sink.iter() {
+          self.lookup_type.insert(*k, *v);
```
Napokue, line 114: I think this works, but it is not clean, as the sink will always be initialized even though a LookupType is local.

src/back/spv/writer.rs
kvark: hmm, right. I suppose we need to refactor `write_type_declaration_arena` to not work with `self`. Or at least, that's one way to address this...
if it's too difficult, we can go back to your original code, which I think was correct, and I'll look at it later to try to improve

Napokue: Yes I would prefer that. This means, the rest of the comments are also resolved?

kvark: sure! Are we all set for landing then?

PR #89, opened 2020-07-01
Add fuzzing support

GabrielMajeri:
Adds some `cargo-fuzz` targets for the SPIR-V and WGSL parsers.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,32 @@
+[package]
+name = "naga-fuzz"
+version = "0.0.0"
+authors = ["Automatically generated"]
+publish = false
+edition = "2018"
+
+[package.metadata]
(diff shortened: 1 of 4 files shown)


GabrielMajeri:
And it found it's first bug :) #90

fuzz/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "naga-fuzz"
```
kvark, line 2: since this is local to the repository, does it make sense to call it the full "naga-fuzz", or could it be just "fuzz"?

GabrielMajeri: By default it will scaffold a project with the `<original crate>-fuzz` name. I presume it's done to avoid conflicts, in case there's some dependency also called `fuzz`.

PR #98, opened 2020-07-03
Add DepthImage IR type

kvark:
See https://github.com/gpuweb/gpuweb/issues/573 and https://github.com/gpuweb/gpuweb/issues/889
Basically, WGSL needs to distinguish between comparison/regular samplers and depth/regular images at the type level. This is a bit complicated in SPIR-V because they are only separated at run-time. So this PR introduces the logic to see how each image and sampler is used, and mutates the type in the arena at the end of parsing accordingly.

- Fix all the usage of Bytes type to really mean bytes.
- Have logic in the SPIR-V front-end to detect the comparison properties of samplers and images.
- Change the fields to hold `MemberPlace` instead of the `Binding`. This is more specific and allows expressing the offsets cleaner, especially with regards to built-ins.


src/arena.rs
```diff
@@ -139,6 +139,11 @@ impl<T> Arena<T> {
             self.append(value)
         }
     }
+
+    /// Get a mutable reference to an element in the arena.
+    pub fn mutate(&mut self, handle: Handle<T>) -> &mut T {
```
Napokue, line 144: The function name is a bit misleading, as I initially thought this would mutate the handle directly, while the intend is to get a mutable reference to it and then mutate a value in the handle.

I am not sure what we would want to call it, maybe `get_mut_handle`?

src/back/spv/writer.rs
```diff
@@ -765,17 +774,25 @@ impl Writer {
...
+                let dim = map_dim(dim);
+                self.try_add_capabilities(dim.required_capabilities());
+
+                let flags = if arrayed {
+                    crate::ImageFlags::ARRAYED
+                } else {
+                    crate::ImageFlags::empty()
+                };
+                instruction = self.instruction_type_image(id, type_id, dim, flags, true);
+                //self.lookup_type.insert(id, base);
```
Napokue, line 794: I am not that familiar with images yet, but how do we retrieve the handle here? Should that be a pointer to an image? I have read that you mentioned that this is determined at runtime for SPIR-V, so we are waiting for the Local value implementation am working on, right?

src/lib.rs
```diff
@@ -122,14 +122,22 @@ pub enum ArraySize {
     Dynamic,
 }
 
+/// Describes where a struct member is placed.
+#[derive(Clone, Debug, PartialEq)]
+pub enum MemberPlace {
```
Napokue, line 127: I am not really sure about this name, would not something like `MemberOrigin` be a little more descriptive? I am just thinking out loud, maybe you had a reason to use the word `Place`

src/front/spv.rs
```diff
@@ -1535,26 +1768,10 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
-                            }
-                        }
-                        None
-                    }
-                    _ => Some(dec.get_binding().ok_or(Error::InvalidBinding(id))?),
-                }
-            }
+            } => match module.types[base].inner {
+                crate::TypeInner::Struct { members: _ } => None,
+                _ => Some(dec.get_binding().ok_or(Error::InvalidBinding(id))?),
```
Napokue, line 1773: This is looking clean! I like it 😄

src/back/spv/writer.rs
kvark: I think it depends on the local type, yeah, since in know that the result is Float here but we don't have an ID to it.

PR #101, opened 2020-07-05
[spv] remove the pointer indirection to variables

### LINKED ISSUE #100: Don't use pointers for accessing global variables
In spir-v any access to variables is done via load/stores in pointers. In WGSL that pointer step is skipped.
I think we should change the SPIR-V front-end to ignore the indirection, and write the IR validation accordingly (#59).

kvark:
Closes #100 
(there may be some leftovers of that semantics, but given the overall state of the SPIR-V frontend we can clean them up later)
Adds support to SPIR-V frontend:
  - vector shuffle
  - use of a few GLSL extension functions
  - a few binary operatos

Removes a few places where SPIR-V frontend was too eager to validate the input. It shouldn't bother: the IR will have to be validated anyway. The *actual* constraint of a front-end is producing a valid IR on a valid input. It can assume things.

Finally, adds the cube SPV shaders from wgpu-rs into the test suite 🎉


kvark:
@Napokue that sounds good to me, as long as it doesn't make it slower. Sharing code with the backend seems sensible today, since they are gated by the same feature.

PR #102, opened 2020-07-05
WGSL front end: Only accept ASCII chars in identifiers

LingMan:
Add some basic lexer tests while we're at it.

### DIFF
--- a/src/front/wgsl.rs
+++ b/src/front/wgsl.rs
@@ -7,7 +7,7 @@ use crate::{
     FastHashMap,
 };
 
-#[derive(Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum Token<'a> {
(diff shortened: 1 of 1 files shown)


src/front/wgsl.rs
```diff
@@ -7,7 +7,7 @@ use crate::{
     FastHashMap,
 };
 
-#[derive(Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
```
kvark, line 10: why do we need Copy+Clone here?

LingMan: We don't quite "need" it. It just made the loop in the test a tiny bit simpler. Without it the line `if result == Token::End` would result in a compile error because the value of `result` was moved into the Vec in the line before.

I can instead check `if results.last() == Some(&Token::End)`,if you want.

Issue #82, opened 2020-06-25
Isolate panics in the code


kvark:
Naga is a pure data processing library. There are two kinds of unexpected here:
  1. input data is weird/bad - all of this should strictly be returning an `Err` without panics!
  2. internal state is inconsistent - this needs to be heavily chased by our testing, but we can panic there.

The problem is - unlike with `unsafe {}` - Rust doesn't have a way to capture/detect the places we panic, easily. For example, just accessing some slice by index has an implicit panic in it.

It would be very useful to have some mechanism where we'd explicitly mark panic places where we are allowed to panic, and warn/error in other places. Perhaps, something like a compiler plugin could be crafted?

cwfitzgerald:
There are clippy warnings in the restrictions category that catch exactly that, warn on the use of implicit panics: https://rust-lang.github.io/rust-clippy/master/index.html search under the restriction category. `indexing_slicing` for the example given, `integer_arithmetic` to catch wrapping panics, etc.

kvark:
@cwfitzgerald thank you, that will be helpful! Even though, it's not going to warn us about any implicit panics in the code outside of Naga, but still a win.
I imagine most of our indexed accesses are done via `Handle`/`Arena`, which should be validated in `proc/validation.rs` separately, at IR level. So we can put a clippy annotation on the `impl Index for Arena` to allow the implicit panic, or turn it into an explicit one (this is better!), and that will cover most of the ergonomics / warning noise, hopefully.

scoopr:
Couple of earlier discussions on reddit 
https://www.reddit.com/r/rust/comments/4lyxtd/is_there_a_tool_to_find_panic_paths/
https://www.reddit.com/r/rust/comments/9mj5xc/nonpanicking_rust/
from those I found, for example
https://github.com/dtolnay/no-panic 

I have no insights on how useful idea those are

paulkernfeld:
Looking at the links posted by @scoopr (thanks @scoopr!), what I'm seeing is that https://github.com/dtolnay/no-panic will fail the build if there is any panic in this crate or dependent code. However, the only way to allow some function to panic requires a pretty crazy technique (https://github.com/dtolnay/no-panic/issues/16), which I think might be necessary in order to use the `no-panic` crate (https://github.com/dtolnay/no-panic/issues/17 is related). So I think the proposed clippy strategy would be more likely to work in the short term.

kvark:
I think this can be closed now with #104. Thanks @paulkernfeld !
[...]

PR #104, opened 2020-07-05
Deny on panics with clippy

paulkernfeld:
This takes a step towards #82

### DIFF
--- a/src/back/spv/layout.rs
+++ b/src/back/spv/layout.rs
@@ -53,12 +53,14 @@ impl Instruction {
         }
     }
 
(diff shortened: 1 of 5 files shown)


paulkernfeld:
Commentary on this change:
- I think there is a bug in clippy that causes it to not detect some panics: https://github.com/rust-lang/rust-clippy/issues/5767
- The `pomelo!` macro generates some panics, which I allowed.
- I allowed panics in `front::glsl` because it looks like the intention is to replace it with pomelo.

PR #108, opened 2020-07-11
Fix SPIR-V back-end, width will now be multiplied by 8

Napokue:
With the change of the bits to Bytes in the IR, the SPIR-V back-end failed to correctly validate OpTypeFloat and OpTypeInt. Their width operand requires the bits, not the bits.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -3,6 +3,8 @@ use super::{helpers, Instruction, LogicalLayout, PhysicalLayout, WriterFlags};
 use crate::{FastHashMap, FastHashSet, ImageFlags, VectorSize};
 use spirv::{Op, Word};
 
+const BITS_PER_BYTE: u8 = 8;
+
 enum Signedness {
     Unsigned = 0,
     Signed = 1,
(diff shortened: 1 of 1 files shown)


src/back/spv/writer.rs
```diff
@@ -744,12 +744,12 @@ impl Writer {
             crate::TypeInner::Scalar { kind, width } => {
                 instruction = match kind {
                     crate::ScalarKind::Sint => {
-                        self.instruction_type_int(id, width, Signedness::Signed)
+                        self.instruction_type_int(id, width << 3, Signedness::Signed)
```
kvark, line 747: It would be cleaner to define a constant like `BITS_PER_BYTE = 8` and use `* BITS_PER_BYTE` here. Note that LLVM will convert it to a bit shift for us, anyway.

kvark, line 747: I think these functions need to be changed. They accept `u8` for width, which is technically neither what SPIR-V needs (`u32`) or what IR has (`Bytes`). We need to either make them receive `Bytes` and do the conversion internally (inside the functions), or switch the arguments to `spirv::Word`

PR #109, opened 2020-07-13
Glsl new handle version

pjoe:
A little bit more work on parser/ast :)

Includes tests 🧪

### DIFF
--- a/src/front/glsl_new/ast.rs
+++ b/src/front/glsl_new/ast.rs
@@ -23,9 +23,7 @@ impl Program {
 
 #[derive(Debug)]
 pub enum Profile {
-    Compatibility,
     Core,
-    Es,
 }
 
 #[derive(Debug)]
--- a/src/front/glsl_new/ast.rs
+++ b/src/front/glsl_new/ast.rs
@@ -23,9 +23,9 @@ impl Program {
 
 #[derive(Debug)]
 pub enum Profile {
-    // Compatibility,
+    Compatibility,
     Core,
-    // Es,
+    Es,
 }
 
 #[derive(Debug)]
--- a/src/front/glsl_new/parser.rs
+++ b/src/front/glsl_new/parser.rs
@@ -41,7 +41,24 @@ pomelo! {
     %type function_definition Function;
 
     root ::= version_pragma translation_unit;
-    version_pragma ::= Version IntConstant Identifier?;
+    version_pragma ::= Version IntConstant(V) Identifier?(P) {
+        match V.1 {
+            450 => (),
+            _ => Err(ErrorKind::InvalidVersion(V.0, V.1))?
+        }
+        extra.version = V.1 as u16;
+        match P {
+            Some((meta, profile)) => {
+                match profile.as_str() {
+                    "compatibility" => extra.profile = Profile::Compatibility,
+                    "core" => extra.profile = Profile::Core,
+                    "es" => extra.profile = Profile::Es,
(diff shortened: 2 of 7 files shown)


pjoe:
Hmmm seems I'm having issues with clippy again, fixing ...

src/front/glsl_new/ast.rs
```diff
@@ -23,9 +23,9 @@ impl Program {
 
 #[derive(Debug)]
 pub enum Profile {
```
kvark, line 25: why do we have that, anyway? We are only interested in GLSL in Vulkan flavor. Does, say, ES profile support Vulkan flavor?

src/front/glsl_new/parser.rs
```diff
@@ -41,7 +41,24 @@ pomelo! {
     %type function_definition Function;
 
     root ::= version_pragma translation_unit;
-    version_pragma ::= Version IntConstant Identifier?;
+    version_pragma ::= Version IntConstant(V) Identifier?(P) {
+        match V.1 {
+            450 => (),
```
kvark, line 46: Can we support at least 440 as well?
When going through glslang, 450 is annoying since it inserts implicit `gl_CullDistance` output, which isn't supported by Naga or WGSL today. This isn't a problem for your code, but if we are to test it on the same shaders versus glslang, it would be nice to support 440

src/front/glsl_new/parser.rs
```diff
@@ -41,7 +41,24 @@ pomelo! {
...
+            _ => return Err(ErrorKind::InvalidVersion(V.0, V.1))
+        }
+        extra.version = V.1 as u16;
+        match P {
```
kvark, line 50: let's do `extra.profile = match P {..};`

src/front/glsl_new/ast.rs
pjoe: You're right, only core profile would actually work. Can't help thinking it would be fantastic if it was somehow possible to support the version used for webgl - to ease developers transitioning from webgl > webgpu, but guessing that is a pipe dream 🤔

src/front/glsl_new/parser.rs
pjoe: I don't really have any strong feelings for 450, just wanna start with a single version, so maybe 440 would be better starting point :)

kvark: can we accept both versions, but for now treat them equally?

pjoe: sure :)

pjoe: maybe include 460 as well then?

kvark: sure

src/front/glsl_new/parser.rs
kvark: still needs to be done?

pjoe: Also done in latest commit

kvark: which isn't pushed?

pjoe: weird, visible to me: https://github.com/gfx-rs/naga/pull/109/commits/914d3d78e22610463042383d253f5fe8f5ca6849#diff-34be000f1f7510bb72876c176236e89eR54

kvark: can we move it out of the outer match?

pjoe: done

PR #110, opened 2020-07-14
Add more glsl-new handling of funcs and types

pjoe:
No tests .. yet 😟


kvark:
@pjoe let's use the commit messages in the format of "[glsl-new] something something" to be able to easily see later what the commits are about

PR #114, opened 2020-07-22
[glsl-new] handle gl_Position builtin

pjoe:
- Also fix width of vec4 type


src/front/glsl_new/ast.rs
```diff
@@ -1,28 +1,30 @@
...
+    pub global_variables: Arena<GlobalVariable>,
 }
 
 impl Program {
     pub fn new() -> Program {
         Program {
             version: 0,
             profile: Profile::Core,
-            ext_decls: vec![],
+            shader_stage: ShaderStage::Vertex,
```
kvark, line 21: why is Vertex the default? Let it be a `new` argument instead

pjoe: Done :)

PR #116, opened 2020-07-28
[glsl-new] Add initial handling of stmts and exprs

pjoe:
Make the parser generate naga statements and expressions.

Still lots of disabled stuff, as well as todos and not implemented errors.

But it now fully parses `simple.vert`, generating working naga IR :)

After this PR probably time for adding some tests, before looking at refactoring. Main issue is that some of the grammar expression rules need to generate both naga expression and statements. I think will need something like a scope stack that can be passed around, so individual rules can easily add statements to current scope.

### DIFF
--- a/src/front/glsl_new/ast.rs
+++ b/src/front/glsl_new/ast.rs
@@ -1,4 +1,7 @@
-use crate::{Arena, Constant, FastHashMap, Function, GlobalVariable, Handle, ShaderStage, Type};
+use crate::{
+    Arena, Constant, Expression, FastHashMap, Function, GlobalVariable, Handle, LocalVariable,
+    ShaderStage, Type,
+};
 
 #[derive(Debug)]
 pub struct Program {
@@ -11,6 +14,7 @@ pub struct Program {
     pub types: Arena<Type>,
     pub constants: Arena<Constant>,
     pub global_variables: Arena<GlobalVariable>,
+    pub context: Option<Context>,
 }
 
 impl Program {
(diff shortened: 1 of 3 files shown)


src/front/glsl_new/parser.rs
```diff
@@ -96,6 +133,9 @@ pomelo! {
                     }),
                 },
             );
+            extra.get_context().expressions.append(Expression::GlobalVariable(h))
+        } else {
+            return Err(ErrorKind::NotImplemented(String::from("var")))
```
pjoe, line 138: I first tried without the `return`, but that gives compiler error. I think this might be a pomelo quirk, like it only sets return type to `Result` if it sees `return Err`

src/front/glsl_new/ast.rs
```diff
@@ -25,11 +29,35 @@ impl Program {
             types: Arena::<Type>::new(),
             constants: Arena::<Constant>::new(),
             global_variables: Arena::<GlobalVariable>::new(),
+            context: None,
         }
     }
+
+    pub fn get_context(&mut self) -> &mut Context {
```
kvark, line 36: why are we lazily creating the context?

src/front/glsl_new/error.rs
```diff
@@ -12,6 +12,7 @@ pub enum ErrorKind {
     IoError(io::Error),
     ParserFail,
     ParserStackOverflow,
+    NotImplemented(String),
```
kvark, line 15: should this be `&'static str`

src/front/glsl_new/ast.rs
pjoe: Hmmm, my thinking was that it needs to be created when ever a grammar rule needs it, as I'm not entirely sure which order grammar rules are called. But lemme try if can just created it up front.

src/front/glsl_new/error.rs
pjoe: Heh, I had no luck with `&'a`, but `&'static` should work of course :)

PR #118, opened 2020-08-04
[glsl-new] Add ExpressionRule for expr/stmts

pjoe:
- Initial work for handling expressions that also generates statements

- Not all expression rules handled yet

### DIFF
--- a/src/front/glsl_new/ast.rs
+++ b/src/front/glsl_new/ast.rs
@@ -1,6 +1,6 @@
 use crate::{
     Arena, Constant, Expression, FastHashMap, Function, GlobalVariable, Handle, LocalVariable,
-    ShaderStage, Type,
+    ShaderStage, Statement, Type,
 };
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


src/front/glsl_new/ast.rs
```diff
@@ -47,3 +47,9 @@ pub struct Context {
     pub expressions: Arena<Expression>,
     pub local_variables: Arena<LocalVariable>,
 }
+
+#[derive(Debug)]
+pub struct ExpressionRule {
```
kvark, line 52: I don't understand this structure. Why do we have a single expression and a vector of statements?

pjoe: I came up with this to be able to have expression rules in the grammar also create statements.

This is e.g. needed for ternary expressions, prefix ops (like ++v, --v), assignment expressions and so on.

As pomelo generates a table based parser, which works bottom up, I thought of this way to have statements needed by expressions be returned and collected up through the parser rules. Not sure if this is the best way, but what I could come up with 😄

PR #120, opened 2020-08-07
[glsl-new] initial declarations support

pjoe:
Still lots of todos, but can now parse global in/out vars 😄

### DIFF
--- a/src/front/glsl_new/ast.rs
+++ b/src/front/glsl_new/ast.rs
@@ -1,6 +1,6 @@
 use crate::{
-    Arena, Constant, Expression, FastHashMap, Function, GlobalVariable, Handle, LocalVariable,
-    ShaderStage, Statement, Type,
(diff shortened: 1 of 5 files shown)


pjoe:
@kvark I guess there is always pros and cons :)

I find it very useful as a parser generator in the same way you would use something like yacc or bison. It does allow to write the parser code in a way that matches up to the spec very closely. Also error handling/recovery 'just works'.

My main difficulties has been not so much from pomelo itself, but from the fact that it generates a table based parser that works bottom up. This doesn't allow parsing extra info down the tree - everything has to be dealt with from the bottom and then parsed up.

One solution could be to have the parser output more of a 'dumb' AST/parse tree, and then do the semantic analysis in a second pass on that output. I wanted to avoid multiple passes if possible to keep the code size down, but not so sure anymore if this is the best approach.

Anyway, I'll try to also take a closer look at the WGSL parser code, for some more inspiration :)

PR #123, opened 2020-08-10
WIP: glsl 450 backend and common glsl module

JCapucho:
Adds a glsl450 backend and a module that implements functionality that can be shared between glsl backends


Cargo.toml
```diff
@@ -23,6 +23,7 @@ thiserror = "1.0"
 default = []
 glsl_preprocessor = ["glsl"]
 glsl-new = ["pomelo"]
+glsl450-backend = []
```
kvark, line 26: Looks like we need to separate frontend and backend features. There are cases where we'd need GLSL backend but not the frontend, for example, and their code is completely different.
How do you feel about calling your feature just "glsl-out"?

examples/convert.rs
```diff
@@ -157,6 +157,19 @@ fn main() {
 
             fs::write(&args[2], bytes.as_slice()).unwrap();
         }
+        #[cfg(feature = "glsl450-backend")]
+        "vert" | "frag" => {
+            use naga::back::glsl450;
```
kvark, line 162: let's not call the module "glsl450", since, as we discussed, we need a more generic glsl backend than just 450

src/arena.rs
```diff
@@ -77,7 +77,7 @@ impl<T> Handle<T> {
 /// Adding new items to the arena produces a strongly-typed [`Handle`].
 /// The arena can be indexed using the given handle to obtain
 /// a reference to the stored item.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
kvark, line 80: why do we need to clone the arena?

src/back/glsl450.rs
```diff
@@ -0,0 +1,201 @@
...
+                structs.insert(handle, (name, idxs));
+            }
+            _ => continue,
+        }
+    }
+
+    // Do a second pass to build the structs
+    // TODO: glsl is order dependent so we need to build structs in order
+    for (handle, ty) in module.types.iter() {
+        match &ty.inner {
```
kvark, line 51: please don't match on a reference. Rust makes smarts to let you do this in 2018 edition, but I'd really like our code base to not rely on these smarts. It's easier to reason about borrowing vs copying without that.

src/back/glsl450.rs
```diff
@@ -0,0 +1,201 @@
...
+            _ => continue,
+        }
+    }
+
+    let mut globals_lookup = FastHashMap::default();
+
+    for (handle, global) in module.global_variables.iter() {
+        if let Some(Binding::BuiltIn(built_in)) = global.binding {
+            match built_in {
+                BuiltIn::Position => globals_lookup.insert(handle, String::from("gl_position")),
```
kvark, line 75: should this be simplified as
```rust
let semantic = match built_in {...};
global_lookup.insert(handle, String::from(semantic));
```

src/back/glsl_common.rs
```diff
@@ -0,0 +1,704 @@
...
+                builder.locals_lookup.get(&handle).unwrap().clone()
+            }
+            Expression::Load { pointer } => todo!(),
+            Expression::ImageSample {
+                image,
+                sampler,
+                coordinate,
+                depth_ref,
+            } => {
+                let image_ty = crate::proc::Typifier::new().resolve(
```
kvark, line 293: we should retain the typifier instead of creating a new one: it caches types as it traverses the module

src/back/glsl_common.rs
```diff
@@ -0,0 +1,704 @@
...
+impl Constant {
+    pub(crate) fn write_glsl(
+        &self,
+        module: &Module,
+        builder: &StatementBuilder<'_>,
+    ) -> Result<String, Error> {
+        Ok(match &self.inner {
+            ConstantInner::Sint(int) => int.to_string(),
+            ConstantInner::Uint(int) => int.to_string(),
+            ConstantInner::Float(float) => float.to_string(),
```
kvark, line 479: probably want to ensure here that there is at least one fraction digit, to not confuse GLSL

src/back/glsl_common.rs
```diff
@@ -0,0 +1,704 @@
...
+    }
+}
+
+impl Handle<Type> {
+    pub(crate) fn write_glsl<'a>(
+        &self,
+        types: &'a Arena<Type>,
+        structs: &'a FastHashMap<Handle<Type>, (String, Vec<String>)>,
+    ) -> Result<String, Error> {
+        Ok(match &types[*self].inner {
```
kvark, line 512: same here, let's not match on a reference

src/back/glsl_common.rs
```diff
@@ -0,0 +1,704 @@
...
+                    .collect::<Result<Vec<_>, _>>()?
+                    .join(","),
+            ),
+        })
+    }
+}
+
+impl Handle<Type> {
+    pub(crate) fn write_glsl<'a>(
+        &self,
```
kvark, line 508: I find it weird that this function is implemented for a handle. Could we have it as a free function instead, i.e. `write_type`?

src/back/glsl_common.rs
```diff
@@ -0,0 +1,704 @@
...
+                "sampler"
+            } else {
+                "samplerShadow"
+            }),
+        })
+    }
+}
+
+impl StorageClass {
+    pub(crate) fn write_glsl<'a>(&'a self) -> impl fmt::Display + 'a {
```
kvark, line 610: same here. I understand this is a matter of preference. My confusion comes from the fact that `self` is just a piece of data here. It feels like it's abusing the "method" semantic on types, which makes more sense for types that have actual contents in them, and meant to have associated logic.

src/back/glsl_common.rs
```diff
@@ -0,0 +1,704 @@
...
+        struct StorageClassWriter<'a> {
```
kvark, line 611: what's going on here? why does it need to borrow the storage class?

src/back/glsl_common.rs
```diff
@@ -0,0 +1,704 @@
...
+                )
+            }
+        }
+
+        StorageClassWriter { inner: self }
+    }
+}
+
+impl ArraySize {
+    pub(crate) fn write_glsl<'a>(&'a self) -> impl fmt::Display + 'a {
```
kvark, line 639: I see the pattern now. I suggest we do the following instead:
```rust
struct ArraySize(crate::ArraySize);
impl fmt::Disaplay for ArraySize {..}
```
I believe it's also how WGSL backend does this.

src/lib.rs
```diff
@@ -171,7 +171,7 @@ bitflags::bitflags! {
 }
 
 /// A data type declared in the module.
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
```
kvark, line 174: why do we need a clone here?

examples/convert.rs
JCapucho: I named it glsl450 because the code that it has can only generate glsl 450 and 460 and since most of the code is already in glsl-common we can add other versions and disable them trough features to reduce code size and compile time

src/arena.rs
JCapucho: I added clone because the glsl backend copies the type out of the module to later mutably borrow it and use it in the Typifier

src/lib.rs
JCapucho: Same reason why i added it to the Arena

examples/convert.rs
kvark: We need features to turn on/off big chunks of code, which is what mostly cared by browsers and WASM targets. Do you think it makes sense to turn GLSL450 output separately ON from GLSLES300 ? My understanding is that the latter is much more involved, but also is what browsers would need, so at least the users I mentioned don't need GLSL450 feature to be separate. Would your app need it?

src/arena.rs
kvark: Where are they borrowed mutably? The backends aren't supposed to mutate the source module.

examples/convert.rs
JCapucho: No i dont need it for any particular reason if you think it is okay to bundle them all together than for me it's okay

src/arena.rs
JCapucho: the backend doesnt mutate but the Typifier requires a mutable reference in resolve

examples/convert.rs
kvark: yeah, I think it's fine to have both behind `glsl-out` feature

PR #125, opened 2020-08-13
Serialize and Deserialize Module

### LINKED ISSUE #32: Serialization of Module

lachlansneff:
This pr adds `serialize` and `deserialize` features to naga, with `Serialize` and `Deserialize` implemented on all types necessary for `Module` when the respective feature is enabled. ~~Examples of using serialize and deserialize  are also included.~~ The `convert` example is expanded to support input and output of `.ron`.

During testing of this patch, I noticed an bug when unserializing a module from ron into spirv. It appears that the initialization of a local variable is not included in the output. It's possible this is a bug in serialization, but I'm not sure how that could be.

```glsl
#version 450 core

void main() {
    float w = 1.0;
    gl_Position = vec4(position, 0.0, w);
}
```

turns into

```glsl
#version 450 core
layout(location=0) in vec2 position;
void main() {
float w;
gl_position = vec4(position,0,w);
}
```

<details>
<summary>Here is the <code>.ron</code> serialization of the shader</summary>

```rust
(
    header: (
        version: (1, 0, 0),
        generator: 0,
    ),
    types: [
        (
            name: None,
            inner: Vector(
                size: Bi,
                kind: Float,
                width: 4,
            ),
        ),
        (
            name: None,
            inner: Scalar(
                kind: Float,
                width: 4,
            ),
        ),
        (
            name: None,
            inner: Scalar(
                kind: Sint,
                width: 4,
            ),
        ),
        (
            name: None,
            inner: Vector(
                size: Quad,
                kind: Float,
                width: 4,
            ),
        ),
    ],
    constants: [
        (
            name: None,
            specialization: None,
            inner: Sint(1),
            ty: 3,
        ),
        (
            name: None,
            specialization: None,
            inner: Sint(0),
            ty: 3,
        ),
    ],
    global_variables: [
        (
            name: Some("position"),
            class: Input,
            binding: Some(Location(0)),
            ty: 1,
        ),
        (
            name: Some("gl_Position"),
            class: Output,
            binding: Some(BuiltIn(Position)),
            ty: 4,
        ),
    ],
    functions: [
        (
            name: Some("main"),
            parameter_types: [],
            return_type: None,
            global_usage: [],
            local_variables: [
                (
                    name: Some("w"),
                    ty: 2,
                    init: Some(1),
                ),
            ],
            expressions: [
                Constant(1),
                GlobalVariable(2),
                GlobalVariable(1),
                Constant(2),
                LocalVariable(1),
                Compose(
                    ty: 4,
                    components: [
                        3,
                        4,
                        5,
                    ],
                ),
            ],
            body: [
                Store(
                    pointer: 2,
                    value: 6,
                ),
            ],
        ),
    ],
    entry_points: [
        (
            stage: Vertex,
            name: "main",
            function: 1,
        ),
    ],
)
```

</details>

Resolves #32 and replaces #48


pjoe:
For the examples, instead of separate serialize/deserialize, could it make sense to simply support `.ron` as input/output format in `convert.rs`?

kvark:
Right, this is all good stuff!
Please make the existing `convert` example just support "ron" extension, instead of having 3 different examples.

lachlansneff:
@pjoe @kvark Done!

examples/convert.rs
```diff
@@ -99,7 +104,7 @@ fn main() {
         return;
     }
 
-    let param_path = std::path::PathBuf::from(&args[1]).with_extension("ron");
+    let param_path = std::path::PathBuf::from(&args[1]).with_extension("param.ron");
```
kvark, line 107: we have a "boids.ron" file in test-data that now needs to be renamed

PR #126, opened 2020-08-13
[glsl-new] Handle local vars scopes

pjoe:
I think this was the biggest remaining issue I needed to figure out - the rest should more or less be just work 😊

Probably also a good time to start figure out some proper tests.

### DIFF
--- a/src/front/glsl_new/ast.rs
+++ b/src/front/glsl_new/ast.rs
@@ -34,7 +34,7 @@ impl Program {
             context: Context {
                 expressions: Arena::<Expression>::new(),
                 local_variables: Arena::<LocalVariable>::new(),
-                lookup_local_variables: FastHashMap::default(),
+                scopes: vec![FastHashMap::default()],
             },
         }
     }
@@ -49,7 +49,47 @@ pub enum Profile {
 pub struct Context {
     pub expressions: Arena<Expression>,
     pub local_variables: Arena<LocalVariable>,
-    pub lookup_local_variables: FastHashMap<String, Handle<LocalVariable>>,
+    pub scopes: Vec<FastHashMap<String, Handle<LocalVariable>>>,
+}
+
+impl Context {
+    pub fn lookup_local_var(&self, name: &str) -> Option<&Handle<LocalVariable>> {
+        for scope in self.scopes.iter().rev() {
+            if let Some(var) = scope.get(name) {
+                return Some(var);
+            }
+        }
+        None
+    }
+
+    pub fn lookup_local_var_current_scope(&self, name: &str) -> Option<&Handle<LocalVariable>> {
+        if let Some(current) = self.scopes.last() {
+            current.get(name)
+        } else {
+            None
+        }
+    }
+
+    pub fn clear_scopes(&mut self) {
+        self.scopes.clear();
+        self.scopes.push(FastHashMap::default());
+    }
+
+    /// Add variable to current scope
+    pub fn add_local_var(&mut self, name: String, handle: Handle<LocalVariable>) {
+        if let Some(current) = self.scopes.last_mut() {
+            (*current).insert(name, handle);
+        }
+    }
+
+    /// Add new empty scope
+    pub fn add_scope(&mut self) {
+        self.scopes.push(FastHashMap::default());
+    }
+
+    pub fn remove_current_scope(&mut self) {
+        self.scopes.pop();
+    }
 }
 
 #[derive(Debug)]
--- a/src/front/glsl_new/ast.rs
+++ b/src/front/glsl_new/ast.rs
@@ -49,22 +49,23 @@ pub enum Profile {
 pub struct Context {
     pub expressions: Arena<Expression>,
     pub local_variables: Arena<LocalVariable>,
+    //TODO: Find less allocation heavy representation
     pub scopes: Vec<FastHashMap<String, Handle<LocalVariable>>>,
 }
 
 impl Context {
-    pub fn lookup_local_var(&self, name: &str) -> Option<&Handle<LocalVariable>> {
+    pub fn lookup_local_var(&self, name: &str) -> Option<Handle<LocalVariable>> {
         for scope in self.scopes.iter().rev() {
             if let Some(var) = scope.get(name) {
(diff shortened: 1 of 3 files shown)


src/front/glsl_new/ast.rs
```diff
@@ -49,7 +49,47 @@ pub enum Profile {
 pub struct Context {
     pub expressions: Arena<Expression>,
     pub local_variables: Arena<LocalVariable>,
-    pub lookup_local_variables: FastHashMap<String, Handle<LocalVariable>>,
+    pub scopes: Vec<FastHashMap<String, Handle<LocalVariable>>>,
+}
+
+impl Context {
+    pub fn lookup_local_var(&self, name: &str) -> Option<&Handle<LocalVariable>> {
```
kvark, line 56: no need to return a reference to a handle, since the are copyable - just return the handle

src/front/glsl_new/ast.rs
```diff
@@ -49,7 +49,47 @@ pub enum Profile {
...
+        for scope in self.scopes.iter().rev() {
+            if let Some(var) = scope.get(name) {
+                return Some(var);
+            }
+        }
+        None
+    }
+
+    pub fn lookup_local_var_current_scope(&self, name: &str) -> Option<&Handle<LocalVariable>> {
```
kvark, line 65: same here

src/front/glsl_new/ast.rs
```diff
@@ -49,7 +49,47 @@ pub enum Profile {
 pub struct Context {
     pub expressions: Arena<Expression>,
     pub local_variables: Arena<LocalVariable>,
-    pub lookup_local_variables: FastHashMap<String, Handle<LocalVariable>>,
+    pub scopes: Vec<FastHashMap<String, Handle<LocalVariable>>>,
```
kvark, line 52: This data structure is rather allocation-heavy. Maybe leave a TODO to make it more memory friendly?

src/front/glsl_new/ast.rs
```diff
@@ -49,7 +49,47 @@ pub enum Profile {
...
+
+    /// Add variable to current scope
+    pub fn add_local_var(&mut self, name: String, handle: Handle<LocalVariable>) {
+        if let Some(current) = self.scopes.last_mut() {
+            (*current).insert(name, handle);
+        }
+    }
+
+    /// Add new empty scope
+    pub fn add_scope(&mut self) {
```
kvark, line 86: let's rename to `push_scope`

src/front/glsl_new/error.rs
```diff
@@ -37,6 +38,9 @@ impl fmt::Display for ErrorKind {
             ErrorKind::UnknownVariable(meta, val) => {
                 write!(f, "Unknown variable {} at {:?}", val, meta)
             }
+            ErrorKind::VariableAlreadyDeclared(val) => {
```
kvark, line 41: who forbids this? I think, if you encounter a variable with the same name, it can just replace the current one in the scope map

src/front/glsl_new/parser.rs
```diff
@@ -156,18 +156,17 @@ pomelo! {
...
-                }
-            } else if let Some(local_var) = extra.context.lookup_local_variables.get(&v.1) {
-                ExpressionRule{
-                    expression: extra.context.expressions.append(Expression::LocalVariable(*local_var)),
-                    statements: vec![],
-                }
-            } else {
-                return Err(ErrorKind::UnknownVariable(v.0, v.1))
+            let expr =
+                if let Some(global_var) = extra.lookup_global_variables.get(&v.1) {
```
kvark, line 160: Should we first look the local scopes and then the global vars?

src/front/glsl_new/error.rs
pjoe: At least glslang barfs on this: http://shader-playground.timjones.io/53f99a8f0acb42b00c1b20a6888878d2

src/front/glsl_new/parser.rs
pjoe: of course :D

src/front/glsl_new/ast.rs
pjoe: done

src/front/glsl_new/ast.rs
pjoe: done

src/front/glsl_new/ast.rs
pjoe: done

src/front/glsl_new/ast.rs
pjoe: done

src/front/glsl_new/error.rs
pjoe: Will try checking if the spec has anything to say on this ...

kvark: Naga IR doesn't really care about names, other than for entry points (since they connect to the API). So IR has no limitations on what the names are, might as well treat them as debug labels.
The limitation you are talking about comes from GLSL, and the front-end should assume it has valid GLSL on the input, you don't need to go into the forest of validating incoming GLSL sources :) So you don't need to worry about overlaps.

pjoe: Where would the GLSL source be validated? Is naga always supposed to be used with a validator in front of it?

kvark: Naga has one and only place in its pipeline where the data is guaranteed to be valid: that's the IR:
  1. front-end inputs are *assumed* to be valid. We don't care if they actually are, as long as the front-end doesn't crash or hangs (which is a universal requirement to the code base).
  2. front-end output IR has to be valid, and we are testing this, and any clients will be testing this, using the validator in of #59 . Any breaking of this contract is a bug in the front-end.
  3. similarly, backends can *assume* the IR is valid
  4. similarly, backend's produced low-level shaders have to be valid, and we are either testing, or will be testing this

So Naga doesn't care if your GLSL is valid or not. Just like a Vulkan driver doesn't care if the use of the API is valid or not. Unlike the Vulkan driver, Naga can't afford to behave strangely though.

PR #128, opened 2020-08-14
Add support for glsl es in the glsl backend

JCapucho:
The vertex stage of the boids doesnt yet compile and the compute stage cant compile because it uses the buffer storage class


Makefile
```diff
@@ -14,3 +14,15 @@ clean:
 
 %.metallib: %.air
 	xcrun -sdk macosx metallib $< -o $@
+
+%.vert: test-data/%.wgsl $(wildcard src/*.rs src/**/*.rs examples/*.rs)
```
kvark, line 18: I wonder if that wildcard can be put into a separate target, since otherwise we are repeating it too many times
(not required for landing this)

src/back/glsl.rs
```diff
@@ -38,16 +39,93 @@ impl fmt::Display for Error {
...
+            }
+
+            (v, true)
+        }
+    };
+
+    writeln!(
+        out,
+        "#version {} {}",
+        version,
```
kvark, line 106: we are implementing `Display` but not using it here?

src/back/glsl.rs
```diff
@@ -38,16 +39,93 @@ impl fmt::Display for Error {
...
+    }
 
     let mut counter = 0;
     let mut names = FastHashMap::default();
 
     let mut namer = |name: Option<&String>| {
         if let Some(name) = name {
-            names.insert(name.clone(), ());
-            name.clone()
+            if name.starts_with("gl_") || name == "main" || names.get(name).is_some() {
```
kvark, line 119: would it be simpler to have a `Map<&str, usize>`, which stores the last index used for each name? This way, we could use the borrowed strings for keys (instead of `String`) and avoid iteration altogether

src/back/glsl.rs
```diff
@@ -119,17 +255,33 @@ pub fn write(module: &Module, out: &mut impl Write) -> Result<(), Error> {
...
+        } else {
+            global.name.clone().ok_or_else(|| {
+                Error::Custom(String::from("Global names must be specified in es"))
+            })?
+        };
+
         if let Some(ref binding) = global.binding {
-            write!(out, "layout({}) ", Binding(binding.clone()))?;
+            if !es {
+                write!(out, "layout({}) ", Binding(binding.clone()))?;
```
kvark, line 268: I'm wondering now... if the GLES doesn't have the layout qualifier, then matching has to happen with names. But names depend on the order they've been encountered. So technically, would we hit a case where the same varying, when exported from VS and FS shaders, will end up with different names? I.e. "myvar4" vs "myvar5"?

We may need to find a more elaborate scheme. For example, see how MSL backend does this.

src/back/glsl.rs
```diff
@@ -138,14 +290,46 @@ pub fn write(module: &Module, out: &mut impl Write) -> Result<(), Error> {
...
     let mut functions = FastHashMap::default();
 
-    // Do a first pass to collect names
     for (handle, func) in module.functions.iter() {
-        functions.insert(handle, namer(func.name.as_ref()));
+        // Discard all entry points
+        if module
+            .entry_points
+            .iter()
+            .any(|entry| entry.function == handle && entry_point.function != handle)
```
kvark, line 298: should `entry_point.function != handle` moved out of the loop completely?

src/back/glsl.rs
```diff
@@ -759,34 +960,42 @@ fn write_constant(
...
-fn write_type<'a>(
+fn write_type(
     ty: Handle<Type>,
-    types: &'a Arena<Type>,
-    structs: &'a FastHashMap<Handle<Type>, String>,
+    types: &Arena<Type>,
+    structs: &FastHashMap<Handle<Type>, String>,
+    block: Option<String>,
+    features: SupportedFeatures,
 ) -> Result<String, Error> {
```
kvark, line 991: Let's return `Cow<'static, str>` instead of `String` here, it would allow us to avoid heap allocating constant strings

src/back/glsl.rs
```diff
@@ -822,50 +1031,89 @@ fn write_type<'a>(
...
+            }
+
+            format!(
+                "{}texture{}{}",
+                match types[base].inner {
+                    TypeInner::Scalar { kind, .. } => match kind {
+                        ScalarKind::Sint => "i",
+                        ScalarKind::Uint => "u",
+                        ScalarKind::Float => "",
+                        _ =>
```
kvark, line 1102: let's list the remaining variants (just `Bool`?) explicitly here

src/back/glsl.rs
```diff
@@ -879,16 +1127,32 @@ fn write_type<'a>(
     })
 }
 
-fn write_storage_class(class: StorageClass) -> Result<String, Error> {
+fn write_storage_class(class: StorageClass, features: SupportedFeatures) -> Result<String, Error> {
```
kvark, line 1130: this should return `&'static str` instead of `String`

src/back/glsl.rs
```diff
@@ -899,17 +1163,31 @@ fn write_array_size(size: ArraySize) -> Result<String, Error> {
     })
 }
 
-fn write_image_flags(flags: ImageFlags) -> Result<String, Error> {
+fn write_image_flags(flags: ImageFlags, features: SupportedFeatures) -> Result<String, Error> {
     let mut out = String::new();
```
kvark, line 1167: another way to write this (and that's how MSL backend does it) would be:
```
let ms = if flags.contains(ImageFlags::MULTISAMPLED) {.. } else { "" };
let array = if flags.contains(ImageFlags::ARRAYED) {"Array"} else { "" };
Ok(format!("{}{}", ms, array))
```
This has the benefit of not using the `write!` on a string, since it may be a bit confusing

src/back/glsl.rs
JCapucho: I'm not sure what the problem is but its already required that the name is defined in es

PR #130, opened 2020-08-16
[glsl-new] Fix expression ordering for compose

pjoe:
Fix issue with `Compose` expression depending on later expressions.

Now `Compose` isn't generated until all args have been collected.

### DIFF
--- a/src/front/glsl_new/ast.rs
+++ b/src/front/glsl_new/ast.rs
@@ -112,3 +112,16 @@ pub struct VarDeclaration {
     pub ids_initializers: Vec<(String, Option<ExpressionRule>)>,
     pub ty: Handle<Type>,
 }
+
+#[derive(Debug)]
+pub enum FunctionCallKind {
+    TypeConstructor(Handle<Type>),
+    Function(Handle<Expression>),
+}
+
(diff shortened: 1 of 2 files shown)


src/front/glsl_new/ast.rs
```diff
@@ -112,3 +112,16 @@ pub struct VarDeclaration {
     pub ids_initializers: Vec<(String, Option<ExpressionRule>)>,
     pub ty: Handle<Type>,
 }
+
+#[derive(Debug)]
+pub enum FunctionCallKind {
+    TypeConstructor(Handle<Type>),
+    Function(Handle<Expression>),
```
pjoe, line 119: NOTE: GLSL spec defines the function identifier in a call as a postfix expression, not sure how this will work with naga, looks like the `Call` expression needs a `FunctionOrigin` directly. Will need to figure out when implementing function calls (for now it only handles type ctors)

pjoe: After some more digging, I think this is done in order to support GLSL subroutines (didn't even know there was such a thing). They work by dispatching the function call based on a uniform input.

I suggest we don't support this for now.

PR #133, opened 2020-08-17
Add more scalar and vector types to glsl-new

lachlansneff:
This might be messier than necessary. I wasn't sure if pomelo could provide more generic definitions (like VecN, etc).


lachlansneff:
Oops, I have some more changes here, will make a further pr.

PR #135, opened 2020-08-18
[spirv-out] - Refactor Function Base

Napokue:
Refactor of the base logic for functions. Strong base is needed to support complex shaders in the future (boids for example). As the quad sample only works now, which is very basic, there was no need to have a "great" parsing for functions.

Recently, there has been more and more interest in more complex features of spirv like: if/else statements, loops, access chains etc.

This PR includes:
- [x] Implement `Statement::Empty`.
- [x] Implement `OpFunctionParameter`.
- [x] Implement `OpFunctionCall`.
- [x] Implement new function structs to make it more maintainable.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -647,6 +647,13 @@ impl Writer {
         instruction
     }
 
+    fn instruction_function_parameter(&self, result_type_id: Word, id: Word) -> Instruction {
+        let mut instruction = Instruction::new(Op::FunctionParameter);
+        instruction.set_type(result_type_id);
+        instruction.set_result(id);
+        instruction
+    }
+
     fn instruction_function_end(&self) -> Instruction {
         Instruction::new(Op::FunctionEnd)
     }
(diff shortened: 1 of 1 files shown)


src/back/spv/writer.rs
```diff
@@ -1075,199 +1160,269 @@ impl Writer {
...
+                            block.body.push(self.instruction_load(
+                                value_ty_id,
+                                load_id,
+                                id,
+                                None,
+                            ));
+                            load_id
+                        }
+                        _ => id
+                    };
```
Napokue, line 1389: The only thing I am not yet happy about is this piece of code. This is also used (for now) at `Statement::Store`. The reason for manually checking is because if the following conclusion I made:

There is a difference between what id to pass, a local variable is in an unique case I think that it can be both at the left side and the right side of an operator, e.g:

```wgsl
var test: f32 = 1.0; // Left side of the operator
var test_2: f32 = test; // Right side of the operator
```

So the reason for an OpLoad here is because if a variable is used at the right side of an operator (an expression), then we need to load it. If it is referenced/used as an actual variable reference, then we need to have its direct id.

Above example outputs the following spirv:
```spirv
%4 = OpVariable %_ptr_Function_float Function %float_1
%6 = OpVariable %_ptr_Function_float Function
%11 = OpLoad %float %4 None
OpStore %6 %11
```

I think this all sounds pretty logical, and I believe this case only exists for variables, but I wanted to give attention to this piece of code.

kvark: that makes sense! I'm not sure if we'll need to special case any global variable types as well, but what you have sounds reasonable

PR #140, opened 2020-08-19
Interpolation support for wgsl, glsl, glsl-new, spirv (front and back), and glsl-out

lachlansneff:
This pr adds interpolation qualifier support to naga itself, as well as all frontends and all backends except `msl` (I don't know `msl` well enough). You may notice that includes `wgsl`. While there has been no official declaration of how interpolation qualifiers will make their way into wgsl, [this](https://github.com/gpuweb/gpuweb/issues/802) issue gave some good options, and I went with the one (the second) that seemed better to me. Easy enough to change down the line.

While I was able to test most of these, the spirv frontend was having issues that were unrelated to my changes and so I couldn't test it.

### DIFF
--- a/src/front/glsl/mod.rs
+++ b/src/front/glsl/mod.rs
@@ -1243,12 +1243,10 @@ impl<'a> Parser<'a> {
                     }
                 }
                 TypeQualifierSpec::Interpolation(interpolation_qualifier) => {
-                    match interpolation_qualifier {
-                        InterpolationQualifier::NoPerspective => {
-                            interpolation = Interpolation::Linear
-                        }
-                        InterpolationQualifier::Flat => interpolation = Interpolation::Flat,
-                        InterpolationQualifier::Smooth => {}
+                    interpolation = match interpolation_qualifier {
+                        InterpolationQualifier::NoPerspective => Interpolation::Linear,
+                        InterpolationQualifier::Flat => Interpolation::Flat,
+                        InterpolationQualifier::Smooth => Interpolation::Perspective,
                     }
                 }
                 _ => unimplemented!(),
--- a/src/front/glsl/mod.rs
+++ b/src/front/glsl/mod.rs
@@ -196,7 +196,6 @@ impl<'a> Parser<'a> {
                                     } else {
                                         ty
                                     },
-                                    interpolation: None,
                                 });
 
                                 if name.is_none() {
@@ -1175,7 +1174,7 @@ impl<'a> Parser<'a> {
     ) -> (StorageClass, Option<Binding>, Option<Interpolation>) {
         let mut storage = None;
         let mut binding = None;
-        let mut interpolation = None;
+        let mut interpolation = Interpolation::Perspective;
 
         for qualifier in qualifier.qualifiers {
             match qualifier {
@@ -1244,12 +1243,11 @@ impl<'a> Parser<'a> {
                     }
                 }
                 TypeQualifierSpec::Interpolation(interpolation_qualifier) => {
-                    assert!(interpolation.is_none());
                     match interpolation_qualifier {
                         InterpolationQualifier::NoPerspective => {
-                            interpolation = Some(Interpolation::NoPerspective)
+                            interpolation = Interpolation::Linear
                         }
-                        InterpolationQualifier::Flat => interpolation = Some(Interpolation::Flat),
+                        InterpolationQualifier::Flat => interpolation = Interpolation::Flat,
                         InterpolationQualifier::Smooth => {}
                     }
                 }
@@ -1257,10 +1255,15 @@ impl<'a> Parser<'a> {
             }
         }
 
+        let class = storage.unwrap_or(StorageClass::Private);
+
         (
-            storage.unwrap_or(StorageClass::Private),
+            class,
             binding,
-            interpolation,
+            match class {
+                StorageClass::Input | StorageClass::Output => Some(interpolation),
+                _ => None,
+            },
         )
     }
 }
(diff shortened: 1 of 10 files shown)


lachlansneff:
Did some more research and it seems like msl supports structs fields with interpolation qualifiers, but glsl clearly doesn't, so will modify accordingly.

src/lib.rs
```diff
@@ -316,6 +337,8 @@ pub struct GlobalVariable {
     pub binding: Option<Binding>,
     /// The type of this variable.
     pub ty: Handle<Type>,
+    /// The interpolation qualifier, if any.
```
kvark, line 340: on IR level, we should be more strict and say that this is required for global inputs in FS and global outputs in VS

src/lib.rs
```diff
@@ -153,6 +153,26 @@ pub enum MemberOrigin {
     Offset(u32),
 }
 
+/// The interpolation qualifier of a binding or struct field.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+pub enum Interpolation {
+    /// Indicates that linear, non-perspective, correct
```
kvark, line 161: we should have `Perspective` here as well, explicitly

src/lib.rs
```diff
@@ -153,6 +153,26 @@ pub enum MemberOrigin {
...
+    //// interpolation must be used.
+    NoPerspective,
```
kvark, line 163: perhaps we can call it `Linear` instead?

src/front/wgsl.rs
```diff
@@ -970,6 +981,18 @@ impl Parser {
                             offset = lexer.next_uint_literal()?;
                             ready = false;
                         }
+                        Token::Word(word) if ready => {
```
kvark, line 984: where are you getting the syntax? WGSL doesn't have any interpolation qualifiers defined yet
If you want to invent it, I'd much like to see the built-in decorator called "interpole", i.e. `[[interpolate linear]]` instead of just `[[linear]]`

src/lib.rs
lachlansneff: Should I try to encode that invariant in the type system somehow? Or just add it as a comment?

src/lib.rs
lachlansneff: Perspective would be default interpolation (aka, None)? I added all the kinds of interpolation supported by spirv.

src/lib.rs
lachlansneff: It's called `NoPerspective` in most of the literature (spirv, glsl, etc). Happy to rename it though,.

src/front/wgsl.rs
lachlansneff: Yeah, I'm getting this syntax from [this issue](https://github.com/gpuweb/gpuweb/issues/802). Obviously, it's not official, but it seemed reasonable to me. This syntax here is easy enough to change when an official decision is reached, but I'd like to add it (or at least error out explicitly) here.

`[[interpolate <qualifier>]]` seems reasonable to me. I can change it to that if you'd like.

pjoe:
Looks very nice indeed, glad too see you managed to do this in glsl-new.

👍  for the glsl-new part

src/lib.rs
kvark: The IR doesn't need defaults, nobody writes it by hand. I want it to be explicit and clear.

src/lib.rs
kvark: I believe it's only called that because the human-written languages have a default, which is perspective. Once you don't have a default, the name `Linear` starts to make much more sense. And in IR we don't need to have defaults.

src/lib.rs
kvark: Just as a comment. I tried to think about how to encode this more, and didn't find a way.

Napokue:
Thanks for adding this! Looked over all the changes and specifically the spirv parts, love it!

src/front/glsl/mod.rs
```diff
@@ -1228,11 +1242,29 @@ impl<'a> Parser<'a> {
                         panic!()
                     }
                 }
+                TypeQualifierSpec::Interpolation(interpolation_qualifier) => {
+                    match interpolation_qualifier {
```
kvark, line 1246: can we do `interpolation = match interpolation_qualifier {.. }`?

src/front/glsl_new/lex.rs
```diff
@@ -113,6 +113,23 @@ pub fn consume_token(mut input: &str) -> (Option<Token>, &str) {
...
                 "out" => (Some(Token::Out(meta)), rest),
+                "flat" => (
+                    Some(Token::Interpolation((meta, crate::Interpolation::Flat))),
+                    rest,
+                ),
+                "noperspective" => (
+                    Some(Token::Interpolation((meta, crate::Interpolation::Linear))),
+                    rest,
+                ),
+                "smooth" => (None, rest),
```
kvark, line 124: why is this `None`?

src/front/spv.rs
```diff
@@ -496,6 +497,21 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                 inst.expect(base_words + 2)?;
                 dec.array_stride = NonZeroU32::new(self.next()?);
             }
+            spirv::Decoration::NoPerspective
+            | spirv::Decoration::Flat
+            | spirv::Decoration::Patch
+            | spirv::Decoration::Centroid
+            | spirv::Decoration::Sample => {
+                use spirv::Decoration::*;
+                dec.interpolation = Some(match dec_typed {
```
kvark, line 506: hmm, perhaps it would be simpler to avoid matching the same set of values twice

src/front/glsl_new/lex.rs
lachlansneff: Oops, forgot to change this after adding `Interpolation::Perspective`. Thanks for catching it!

PR #141, opened 2020-08-19
[wgsl] Support for composite constant expressions.

lachlansneff:
Adds support for composite constant expressions to the wgsl frontend, as well as fixes a bug with broken array template endings.

### DIFF
--- a/src/front/wgsl.rs
+++ b/src/front/wgsl.rs
@@ -474,24 +474,17 @@ impl Parser {
         index: usize,
     ) -> Result<Handle<crate::Type>, Error<'static>> {
         let ty = match type_arena[ty].inner {
-            crate::TypeInner::Vector {
-                ref kind,
-                ref width,
-                ..
+            crate::TypeInner::Vector { kind, width, .. }
+            | crate::TypeInner::Matrix { kind, width, .. } => {
+                type_arena.fetch_or_append(crate::Type {
+                    name: None,
+                    inner: crate::TypeInner::Scalar { kind, width },
+                })
             }
-            | crate::TypeInner::Matrix {
-                ref kind,
-                ref width,
-                ..
-            } => Err(crate::TypeInner::Scalar {
-                kind: *kind,
-                width: *width,
-            }),
-            crate::TypeInner::Array { ref base, .. } => Ok(*base),
-            crate::TypeInner::Struct { ref members } => Ok(members[index].ty),
+            crate::TypeInner::Array { base, .. } => base,
+            crate::TypeInner::Struct { ref members } => members[index].ty,
             ref inner => return Err(Error::NotCompositeType(inner.clone())),
-        }
-        .unwrap_or_else(|inner| type_arena.fetch_or_append(crate::Type { name: None, inner }));
+        };
 
         Ok(ty)
     }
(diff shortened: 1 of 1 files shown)


src/front/wgsl.rs
```diff
@@ -456,6 +458,27 @@ impl Parser {
         }
     }
 
+    fn deconstruct_composite_type(
+        type_arena: &mut Arena<crate::Type>,
+        ty: Handle<crate::Type>,
+        index: usize,
+    ) -> Result<Handle<crate::Type>, Error<'static>> {
+        let ty = match &type_arena[ty].inner {
```
kvark, line 466: let's not match on a reference, please

lachlansneff: How should I do this then? If I match on the value, it'll try to move it and it can't.

kvark: Use `ref` in pattern matching instead, where the data is not `Copy`:
```rust
let ty = match type_arena[ty].inner {
 ...
  crate::TypeInner::Struct { ref members } => {..}
};
```

lachlansneff: There's some ownership issues with that, but I'll work around them. For future reference, why is matching on a reference discouraged?

kvark: `ref` is precisely what you get behind the sugar of matching on a reference, and that's why I don't like it.
Sorry, it could be one of my pet peeves. I'm used to Rust that doesn't even allow one to match on a reference, and you had to write out `ref` where applicable. That has the benefit of you knowing precisely what's going on there upon reading some code, and it allows you to use `ref` even on `Copy` types in case they are big, and you don't really want them to be copied anywhere.

TL;DR: there shouldn't be any more ownership issues with `ref` than you'd get with matching on a reference

lachlansneff: Weird, maybe I was doing it wrong, but it was borrowing the `type_arena` through the whole match expression when I need to use it mutably. Whereas, matching on references didn't do that.

kvark: if you need mutability, you do `ref mut` instead of `ref`

src/front/wgsl.rs
```diff
@@ -466,6 +468,34 @@ impl Parser {
...
     }
 
+    fn deconstruct_composite_type(
+        type_arena: &mut Arena<crate::Type>,
+        ty: Handle<crate::Type>,
+        index: usize,
+    ) -> Result<Handle<crate::Type>, Error<'static>> {
+        let ty = match type_arena[ty].inner {
+            crate::TypeInner::Vector {
+                ref kind,
```
kvark, line 478: you don't need to take `kind` and `width` by reference, they are small and copyable

src/front/wgsl.rs
```diff
@@ -466,6 +468,34 @@ impl Parser {
...
+            }
+            | crate::TypeInner::Matrix {
+                ref kind,
+                ref width,
+                ..
+            } => Err(crate::TypeInner::Scalar {
+                kind: *kind,
+                width: *width,
+            }),
+            crate::TypeInner::Array { ref base, .. } => Ok(*base),
```
kvark, line 490: same for `base`

src/front/wgsl.rs
lachlansneff: Yeah, I attempted writing that match again, and it now works out normally with `ref`s and all that.

PR #142, opened 2020-08-20
Better floating point support for wgsl

lachlansneff:
I think this follows all the rules.

https://gpuweb.github.io/gpuweb/wgsl.html#literals

### DIFF
--- a/src/front/wgsl.rs
+++ b/src/front/wgsl.rs
@@ -44,6 +44,30 @@ mod lex {
         input.split_at(pos)
     }
 
+    fn consume_number(input: &str) -> (&str, &str) {
+        let mut is_first_char = true;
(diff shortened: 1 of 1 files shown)


src/front/wgsl.rs
```diff
@@ -44,6 +44,28 @@ mod lex {
         input.split_at(pos)
     }
 
+    fn consume_number(input: &str) -> (&str, &str) {
+        let mut is_first_char = true;
```
kvark, line 48: sounds like it should be a state machine, i.e. `enum State` and such...

lachlansneff: Yeah, I was trying to use an enum state machine, and it just ended up more complex.

PR #144, opened 2020-08-20
[wgsl] Add more complete function calling support

### LINKED ISSUE #136: WGSL - Support for function calling
Function calling is not yet supported in the WGSL front-end. The following WGSL throws an error:

```
fn main_vert() -> void {
  test_function(1.0);
  return;
}

fn test_function(test: f32) -> void {
	return;
}
entry_point vertex as "main" = main_vert;
```

Error: `ParseError { error: UnknownIdent("test_function"), scopes: [FunctionDecl, Block, Statement], pos: (6, 15) }`

lachlansneff:
This pr adds support statement function calls, as well as fixes an issue with importing a namespace with multiple namespaces, aka: `import "GLSL.std.450" as std::glsl;`.

Resolves #136

### DIFF
--- a/src/front/wgsl.rs
+++ b/src/front/wgsl.rs
@@ -278,6 +278,21 @@ impl<'a> Lexer<'a> {
             input: &original_input[..used_len],
         })
     }
+
+    fn skip_same_tokens(&mut self, what: &str) -> bool {
+        let mut this = self.clone();
+        let mut what_lexer = Lexer { input: what };
+
+        loop {
+            if what_lexer.peek() == Token::End {
+                *self = this;
+                return true;
+            }
+            if what_lexer.next() != this.next() {
+                return false;
+            }
+        }
+    }
 }
 
 trait StringValueLookup<'a> {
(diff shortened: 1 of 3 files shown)


Napokue:
I forgot how it is formally called, but does WGSL require functions to be declared before being used or can we declare them later in the code, but still able to resolve it? I believe the current WGSL front-end is built to parse from top to bottom, I am just wondering if this is something we want or not.

kvark:
Currently, it requires functions to be declared prior to being used.

src/front/wgsl.rs
```diff
@@ -276,6 +278,21 @@ impl<'a> Lexer<'a> {
             input: &original_input[..used_len],
         })
     }
+
+    fn skip_same_tokens(&mut self, what: &str) -> bool {
```
kvark, line 282: I wonder why we aren't simply skipping the string here, i.e. `self.input.starts_with(what)`

src/front/wgsl.rs
```diff
@@ -503,6 +522,45 @@ impl Parser {
         }
     }
 
+    fn parse_function_call<'a>(
+        &mut self,
+        lexer: &mut Lexer<'a>,
+        mut ctx: ExpressionContext<'a, '_, '_>,
+    ) -> Result<crate::Expression, Error<'a>> {
```
kvark, line 529: the error model of this function is weird: it tries to return parsing errors, but they never used by the callers
Perhaps, it should instead just return an `Option<>`?

src/front/wgsl.rs
```diff
@@ -614,22 +672,9 @@ impl Parser {
...
-                            lexer.expect(Token::Separator(','))?;
-                        }
-                        let arg = self.parse_general_expression(lexer, ctx.reborrow())?;
-                        arguments.push(arg);
-                    }
-                    crate::Expression::Call {
-                        origin: crate::FunctionOrigin::External(name.to_owned()),
-                        arguments,
-                    }
+                if let Ok(expr) = self.parse_function_call(&mut backup, ctx.reborrow()) {
```
kvark, line 675: why are we passing the backup here mutably? That seems suspiciously weird

src/front/wgsl.rs
lachlansneff: I tried just skipping if the same thing, but I think that tokenizing is better because we want to be a little loose with whether spaces between namespace and double colons are allowing, I think. The standard is a little vague, but I think this is what it's saying (https://gpuweb.github.io/gpuweb/wgsl.html#imports).

src/front/wgsl.rs
lachlansneff: You're right, this should return `Option`.

src/front/wgsl.rs
lachlansneff: Also good catch. I tried to make this simpler, but the reason why it is so weird is that it needs access to a lexer from before the current token was taken. I could pass a `&backup` and a `&mut lexer` at the same time, if that would seem better.

src/front/wgsl.rs
kvark: isn't it called on a single namespace anyway? what kind of tokens are we expecting to be skipped here?

src/front/wgsl.rs
lachlansneff: Okay, I think I fixed this by returning the new lexer from `parse_function_call`.

src/front/wgsl.rs
lachlansneff: It is just called on a single namespace, but I wanted `std :: glsl:: distance`  to  parse as well as `std::glsl::distance`.

lachlansneff: Confirmed by https://matrix.to/#/!MFogdGJfnZLrDmgkBN:matrix.org/$cmtoMc-VHCJTK0BTRxv6zPwMrjnlpz-M9rZAP5x7q8o?via=matrix.org&via=mozilla.org&via=kyju.org.

kvark: just to be clear on the motivation here, I want the lexer to be as simple as possible. Every new method on the lexer has to be carefully considered. In this case, what you really are looking for is this: given a list of `namespaces: &[&str]`, you want to check if the current stream of tokens corresponds to
```
[
  `Token::ident(namespace[0])`,
  `Token::Separator('::')`,
  `Token::ident(namespace[1])`,
 ... etc
]
```
So it seems to me that this could be implemented in terms of `expect_token` instead of providing a string that is re-parsed again and again.

lachlansneff: That's a good point, I think you're right. I want to make sure the code for this is easily adaptable to the future when there may be many places something may be imported from, so I'll try doing it this way? Would this involve essentially making a `Vec` of tokens that represent the namespace?

kvark: please try, I'm sure we can get it to work with tokens

PR #146, opened 2020-08-20
Make "None" interpolation mean default or nothing, depending on context

lachlansneff:
Essentially, removes the `Perspective` variant from `Interpolation` and makes `None: Option<Interpolation>` mean either `perspective`/`smooth` or no interpolation qualifier, depending on context.

### DIFF
--- a/src/back/glsl.rs
+++ b/src/back/glsl.rs
@@ -899,6 +899,7 @@ fn write_storage_class(class: StorageClass) -> Result<String, Error> {
 
 fn write_interpolation(interpolation: Interpolation) -> Result<&'static str, Error> {
     Ok(match interpolation {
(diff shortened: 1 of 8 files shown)


src/back/glsl.rs
```diff
@@ -899,7 +899,6 @@ fn write_storage_class(class: StorageClass) -> Result<String, Error> {
 
 fn write_interpolation(interpolation: Interpolation) -> Result<&'static str, Error> {
     Ok(match interpolation {
-        Interpolation::Perspective => "smooth",
```
kvark, line 902: hold on, let's not rush into removing this one, ok? there is no harm to keep an explicit variant, in case we want to make it required later

Issue #105, opened 2020-07-05
Rosetta testing method


kvark:
We have many front-ends, many back-ends, and we need to test them all using the a limited number of shaders in the tree (in `test-data`), and we need to do it in a way that is straightforward and extendable. Here is an idea on how to do it, inspired by [Rosetta stone](https://en.wikipedia.org/wiki/Rosetta_Stone).

Steps:
  1. We select a number of shaders (or shader modules, to be precise), e.g. deferred shaders, forward shaders, flat shaders, compute boids, etc.
  2. For each we compose the "canonical" IR (e.g. by parsing the shaders from one of the languages).
  3. We then write the same shader (each of them) in *all* the front-ends that we can test, and the tests ensure that parsing those sources produces the same IR (possibly with #41).
  4. We also test conversion from IR of this shader (for each of them) to all the backends. If the backend format also has a front-end, we just compare it with the shader we wrote in (3). If it doesn't, we write a shader (e.g. MSL one) and compare against that.

The main idea is to bring structure to the tested material, so that we know how to write the tests and configure them. Basically, the test suite is a collection of those base IR shaders. For each we can define a list of front-ends to test (with the source files) and a list of back-ends to test (with the source files), plus possibly some configuration tables.

One way to structure this is `test-data/boids/` folder having all of `shader.wgsl`, `shader.spv`, `shader.glsl`, `config.ron` (testing configuration), and `shader.ron` (the IR).

Caveats:
  - Classy glsl -> SPIR-V conversion doesn't support multiple entry points, so that could limit the variety of modules we use for testing. However, I think *our* GLSL parser can be configured to support multiple entry points with low effort, so we should do it, resolving the issue.

pjoe:
I like this idea ♥

I'm not sure if there are some issues in practice, but guess we just have to try it out:

- Is there always a "canonical" IR, or can things like ordering or different handle IDs differ, while output is still valid? Maybe some of this can be fixed by post-processing IR, e.g. abstracting out handle IDs.

- Will round-trip e.g. WGSL > IR > WGSL always produce the same? Or do we have cases where the IR doesn't capture all details of the input, or again ordering might differ?

kvark:
Good concerns, thank you for evaluating this!
For ordering in canonical IR: we don't have many unoredered sets/maps, and ones we have are likely to be populated with the same elements in the same order. So it could be that we don't need to do anything specific. We'll see ;)

[...]

PR #148, opened 2020-08-21
Add initial Rosetta testing

pjoe:
Initial attempt at getting started with Rosetta testing #105 

To run:
```sh
$ ./test-data/simple/run.sh
``` 

For now it:
- Compiles naga in debug with all features
- Runs convert on `glsl` (using glsl-new) and `wgsl` input, outputting `ron`
- Diffs against expected output

Exit code is 0 for no diffs and 1 if a diff is found, so should fit right in with CI


kvark:
This gets me super excited, thank you for nailing this!

PR #151, opened 2020-08-25
glsl-out: Fixed names beginning with digits and arguments without names

JCapucho:
Just some small fixes to fix names that begin with digits and actually put names on functions arguments

### DIFF
--- a/src/back/glsl.rs
+++ b/src/back/glsl.rs
@@ -102,7 +102,11 @@ pub fn write<'a>(module: &'a Module, out: &mut impl Write, options: Options) -> 
 
     let mut namer = |name: Option<&'a String>| {
         if let Some(name) = name {
-            if name.starts_with("gl_") || name == "main" || names.get(name.as_str()).is_some() {
+            if name.chars().next().map(char::is_numeric).unwrap_or(true)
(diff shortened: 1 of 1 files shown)


src/back/glsl.rs
```diff
@@ -102,7 +102,11 @@ pub fn write<'a>(module: &'a Module, out: &mut impl Write, options: Options) ->
 
     let mut namer = |name: Option<&'a String>| {
         if let Some(name) = name {
-            if name.starts_with("gl_") || name == "main" || names.get(name.as_str()).is_some() {
+            if name.chars().next().map(char::is_numeric).unwrap_or(true)
```
kvark, line 105: nit: there is `map_or` for this

src/back/glsl.rs
```diff
@@ -102,7 +102,11 @@ pub fn write<'a>(module: &'a Module, out: &mut impl Write, options: Options) ->
...
+                || name.starts_with("gl_")
+                || name == "main"
+                || names.get(name.as_str()).is_some()
```
kvark, line 108: nit: maybe `names.contains(...)`?

kvark:
The nits I noted here are non-blocking. Merging.

PR #153, opened 2020-08-26
Image IR refactor

kvark:
Improves our handling of images in the IR, adds the "imageLoad" expression.


src/front/spv/mod.rs
```diff
@@ -817,18 +817,16 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                     match type_arena[image_type_handle].inner {
                         //TODO: compare the result type
                         crate::TypeInner::Image {
-                            base: _,
+                            kind: _,
```
Napokue, line 820: Nit: can't this be changed to using the syntax with dots? e.g. :

```rust
crate::TypeInner::Image {
    dim,
    arrayed,
    class: crate::ImageClass::Sampled,
    ..
} => {
```

src/front/spv/mod.rs
```diff
@@ -877,7 +876,12 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                     };
                     match type_arena[image_type_handle].inner {
                         //TODO: compare the result type
-                        crate::TypeInner::DepthImage { dim, arrayed } => {
+                        crate::TypeInner::Image {
+                            kind: _,
```
Napokue, line 880: Same here

Napokue (changes requested):
Thank you! Overall looking good, only the front/glsl_new/parser.rs has quite a lot problems with clippy it seems. Got one nitpick comment, rest LGTM.

src/front/spv/mod.rs
kvark: I believe explicit syntax `kind: _` is superior. If new fields are added to the struct, the author would have to revisit these places and decided if these fields need to be considered.

kvark:
All should be fixed now. Added a clippy setting to allow the explicit pattern.

Issue #155, opened 2020-08-26
[meta] requirements for wgpu-rs examples validation


kvark:
This is what we are missing to load SPIR-Vs of wgpu-rs examples for introspection

- [x] (boids)
  - [x] `WorkgroupSize` builtin
  - [x] `UnsupportedInstruction(Function, Bitcast)`
  - [x] `UnsupportedExtInst(25)`
- [x] (mipmap, skybox) `InvalidWordCount` after `Switch` instruction
- [x] (shadow):
  - [x] `UnsupportedInstruction(Function, MatrixTimesMatrix)`
  - [x] `UnsupportedInstruction(Function, ConvertSToF)`
  - [x] `MissingDecoration(Offset)`
  - [x] `UnsupportedInstruction(Function, Bitcast)`
- [ ] (texture-array)
  - [ ] `InvalidImageExpression(Handle(10))`
- [x] (water)
  - [x] `UnsupportedExtInst(69)`
  - [x] `UnsupportedExtInst(49)`
  - [x] `UnsupportedInstruction(Function, Kill)`
  - [x] `UnsupportedInstruction(Function, ConvertSToF)`

kvark:
Filed https://github.com/gfx-rs/naga/issues/161 for the bitcast operator

kvark:
I believe all the examples (except for `texture-array`) are now validated for their interface. Issue can be closed 🎉

PR #157, opened 2020-08-27
Improve SPIR-V frontend with image operands, dot and transpose

kvark:
Addresses a large portion of  #155
See commit names. I checked every single one to pass full checks locally.


Napokue (approved):
Thank you, looks nice!

PR #159, opened 2020-08-27
Make CI test on multiple rust versions

pjoe:
Use 1.41.0 and nightly

Big surprise: nightly seems to run a bit faster :D


pjoe:
Huh, what is that 4th `Test` check GitHub is waiting for?

Even the `Checks` tab at the top of the PR only says 3 Checks.

Maybe some GitHub problem 🤔

PR #164, opened 2020-08-27
glsl-out: fixes

JCapucho:
Fixes `UnaryOp::Not` generating the wrong op (`~` instead of `!`) on booleans and correctly handles storage images

### DIFF
--- a/src/back/glsl.rs
+++ b/src/back/glsl.rs
@@ -2,7 +2,7 @@ use crate::{
     Arena, ArraySize, BinaryOperator, BuiltIn, Constant, ConstantInner, DerivativeAxis, Expression,
     FastHashMap, Function, FunctionOrigin, GlobalVariable, Handle, ImageClass, Interpolation,
     IntrinsicFunction, LocalVariable, MemberOrigin, Module, ScalarKind, ShaderStage, Statement,
-    StorageClass, StructMember, Type, TypeInner, UnaryOperator,
+    StorageAccess, StorageClass, StorageFormat, StructMember, Type, TypeInner, UnaryOperator,
 };
 use std::{
     borrow::Cow,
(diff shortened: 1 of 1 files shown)


src/back/glsl.rs
```diff
@@ -251,9 +251,43 @@ pub fn write<'a>(module: &'a Module, out: &mut impl Write, options: Options) ->
             })?
         };
 
-        if let Some(ref binding) = global.binding {
-            if !es {
-                write!(out, "layout({}) ", Binding(binding))?;
+        let format = match module.types[global.ty].inner {
```
kvark, line 254: nit: call it `storage_format`

src/back/glsl.rs
```diff
@@ -251,9 +251,43 @@ pub fn write<'a>(module: &'a Module, out: &mut impl Write, options: Options) ->
...
+            TypeInner::Image {
+                class: ImageClass::Storage(format, _),
+                ..
+            } => Some(format),
+            _ => None,
+        };
+
+        if format.is_some() || global.binding.is_some() {
```
kvark, line 262: if it's a storage texture type, it has to have a binding. I think we can omit the `format.is_some()` check

src/back/glsl.rs
```diff
@@ -499,7 +533,7 @@ fn write_statement<'a, 'b>(
                     )?;
                 }
 
-                if fallthrough.is_some() {
+                if fallthrough.is_none() {
```
kvark, line 536: heh

src/back/glsl.rs
```diff
@@ -874,25 +911,47 @@ fn write_expression<'a, 'b>(
...
-            } else {
-                todo!()
+            let expr = match class {
+                ImageClass::Sampled | ImageClass::Multisampled => {
+                    let ms = match class {
+                        crate::ImageClass::Multisampled => true,
+                        _ => false,
+                    };
+
+                    //TODO: fix this
```
kvark, line 921: is this `TODO` still needed?

src/back/glsl.rs
```diff
@@ -874,25 +911,47 @@ fn write_expression<'a, 'b>(
...
+                    let sampler_constructor = format!(
+                        "{}sampler{}{}{}({})",
+                        match kind {
+                            ScalarKind::Sint => "i",
```
kvark, line 925: this little piece is repeated at least twice, asks to be moved out :)

src/back/glsl.rs
JCapucho: This part will be removed when we implement combined image samplers

src/back/glsl.rs
JCapucho: This part will be removed when we implement combined image samplers

src/back/glsl.rs
```diff
@@ -251,9 +251,43 @@ pub fn write<'a>(module: &'a Module, out: &mut impl Write, options: Options) ->
...
+        if global.binding.is_some() {
+            write!(out, "layout(")?;
+
+            if let Some(ref binding) = global.binding {
+                if !es {
+                    write!(out, "{}", Binding(binding))?;
+                }
+            }
+
+            if storage_format.is_some() && global.binding.is_some() {
```
LingMan, line 271: This is already inside an `if global.binding.is_some()` block. No need to check twice.

PR #170, opened 2020-08-29
[wgsl-in] Fix constant initialization

Napokue:
Constant initalizations, e.g. would get a `Statement::Empty` in the IR:
```
fn main_vert() -> void {
	var test: f32 = 1.0;	
	return;
}
entry_point vertex as "main" = main_vert;
```

PR fixed this, and it will now get a `Statement::Store` in the IR.

Back-ends tested:
- [x] Spirv
- [x] Metal
- [x] GLSL

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -1142,10 +1142,12 @@ impl Parser {
                             .append(crate::Expression::LocalVariable(var_id));
                         context.lookup_ident.insert(name, expr_id);
                         match init {
-                            Init::Variable(value) => crate::Statement::Store {
-                                pointer: expr_id,
-                                value,
-                            },
+                            Init::Variable(value) | Init::Uniform(value) => {
+                                crate::Statement::Store {
+                                    pointer: expr_id,
+                                    value,
+                                }
+                            }
                             _ => crate::Statement::Empty,
                         }
                     }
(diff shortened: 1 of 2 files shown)


src/front/wgsl/mod.rs
```diff
@@ -1142,10 +1142,12 @@ impl Parser {
                             .append(crate::Expression::LocalVariable(var_id));
                         context.lookup_ident.insert(name, expr_id);
                         match init {
-                            Init::Variable(value) => crate::Statement::Store {
-                                pointer: expr_id,
-                                value,
-                            },
+                            Init::Variable(value) | Init::Uniform(value) => {
```
kvark, line 1145: I believe the old code was correct. If it's `Init::Uniform`, then the initial value is already stored in `crete::LocalVariable::init`, so no `Statement::Store` is expecged

Napokue: Lets go in detail at Matrix about this subject, as this is a tricky one to answer.

pjoe:
FWIW: this is the place where this is handled in glsl-new: https://github.com/gfx-rs/naga/blob/master/src/front/glsl_new/parser.rs#L575

Napokue:
PR will be closed, discussed this with @kvark, conclusion is to iterate over the local variables in the spirv back-end and not introduce a change in the front-ends/IR for this.

PR #171, opened 2020-08-31
[glsl-new] Add bool consts, update logical exprs

pjoe:
Handle "true" and "false" constants.

Refactor logical expression rules in parser to use `ExpressionRule`.

Still need to change remaining expressions to `ExpressionRule`.


pjoe:
Rebased to latest master

PR #173, opened 2020-09-01
[glsl-new] Update last exprs to ExpressionRule

pjoe:
Refactor out binary_expr to keep parser.rs 'small' and readable


pjoe:
Hmmm monocodus playing tricks again. I wonder if it is related to those 2 warnings the CI Test on 1.41 also reports 🤔

Anyway I have no clippy issues locally:
```
code@code-server:~/git/naga$ git branch
* glsl-new-expression-rule
  master
  wasm-build
code@code-server:~/git/naga$ cargo clippy --all-features -- -D warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
```

src/front/glsl_new/expressions.rs
```diff
@@ -0,0 +1,15 @@
+use super::ast::*;
+use crate::{BinaryOperator, Expression};
+
+pub fn binary_expr(
+    program: &mut Program,
```
kvark, line 5: would it makes sense to have as a mutable method on the `Program`?

pjoe: Yeah, that could make sense. I like the idea of having functionality split into different files like: expressions, types, variables, but right now that may be a bit overkill for a single function.

pjoe: Tried it out - much simpler :D

src/front/glsl_new/ast.rs
```diff
@@ -39,6 +39,19 @@ impl Program {
             },
         }
     }
+
+    pub fn binary_expr(
+        self: &mut Program,
```
kvark, line 44: why is this not just `&mut self`?

pjoe: Because I'm still a rust noob :P

Simply copy pasted the external function and made it compile 🤦‍♂️

PR #174, opened 2020-09-03
[Spirv-out] Move instructions to own module

Napokue:
The [writer.rs](https://github.com/gfx-rs/naga/blob/master/src/back/spv/writer.rs#L255) in the spirv back-end was becoming a little bit too cluttered with instructions, especially after I wanted to add even more of them.  So I decided to move them to their own module. This is done for maintainability, as the instructions module will reach a point where it will always be correct (as the spirv spec doesn't really change).

- [x] Move all instructions to the instructions module.
- [x] Remove the instructions_ prefix of the instructions.
- [x] Move all current instruction unit tests to the instructions module.
- [x] Consistent parameter naming and order for each instruction (sometimes its `result_id`, sometimes `id`).
- [x] Check if all the correct Instruction methods are called (there are some cases where an instruction doesn't have an `result_id`, but it is still called).
- [x] Check compliance with spirv v1 spec (as when I implemented the spirv back-end, I was implementing it via the unified spec, instead of v1).
- [x] Update all current instruction unit tests.
- [x] Add instruction unit tests for all current instructions that don't have a test yet.

After this, the spirv back-end will be ready for its growth towards compiling more complex shaders (boids).

### DIFF
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -619,7 +619,6 @@ mod tests {
             result_id: true,
             operands: true,
         };
-
         validate_spec_requirements(requirements, &instruction);
 
         instruction.to_words(&mut output);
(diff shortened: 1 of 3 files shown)


kvark (approved):
Looks reasonable!

PR #177, opened 2020-09-05
[glsl-out]: combined image sampler

JCapucho:
I have also included some changes that are related to the fact that we were using some vulkan glsl only parts

### DIFF
--- a/src/back/glsl.rs
+++ b/src/back/glsl.rs
@@ -802,14 +802,7 @@ fn write_expression<'a, 'b>(
             module.borrow_type(builder.locals[handle].ty),
         ),
         Expression::Load { pointer } => {
-            let (expr, ty) = write_expression(&builder.expressions[pointer], module, builder)?;
-
-            let ty = match *ty.borrow() {
-                TypeInner::Pointer { base, .. } => base,
-                _ => panic!("{:#?}", ty.borrow()),
-            };
-
-            (expr, MaybeOwned::Borrowed(&module.types[ty].inner))
+            write_expression(&builder.expressions[pointer], module, builder)?
         }
         Expression::ImageSample {
             image,
@@ -1406,15 +1399,7 @@ fn write_image_type(
 
     Ok(format!(
         "{}{}{}{}{}",
-        match kind {
-            ScalarKind::Sint => "i",
-            ScalarKind::Uint => "u",
-            ScalarKind::Float => "",
-            ScalarKind::Bool =>
-                return Err(Error::Custom(String::from(
-                    "Cannot build image of booleans",
-                ))),
-        },
+        map_scalar(kind, 4, features)?.prefix,
         match class {
             ImageClass::Storage(_) => "image",
             _ => "sampler",
--- a/src/back/glsl.rs
+++ b/src/back/glsl.rs
@@ -265,6 +265,7 @@ pub fn write<'a>(
 
     let texture_mappings = collect_texture_mapping(module, &functions)?;
     let mut mappings_map = FastHashMap::default();
+    let mut globals_lookup = FastHashMap::default();
 
     for ((handle, global), _) in module
         .global_variables
@@ -272,41 +273,30 @@ pub fn write<'a>(
         .zip(func.global_usage.iter())
         .filter(|(_, usage)| !usage.is_empty())
     {
-        if let TypeInner::Image {
-            kind,
-            dim,
-            arrayed,
-            class,
-        } = module.types[global.ty].inner
-        {
-            let mapping =
-                if let Some(map) = texture_mappings.iter().find(|map| map.texture == handle) {
-                    map
-                } else {
-                    warn!(
-                        "Couldn't find a mapping for {:?}, handle {:?}",
-                        global, handle
-                    );
-                    continue;
-                };
-
-            if let Some(ref binding) = global.binding {
-                write!(out, "layout(")?;
-
-                if !es {
-                    write!(out, "{}", Binding(binding))?;
-
-                    write!(out, ",")?;
-                }
+        match module.types[global.ty].inner {
+            TypeInner::Image {
+                kind,
+                dim,
+                arrayed,
+                class,
+            } => {
+                let mapping =
+                    if let Some(map) = texture_mappings.iter().find(|map| map.texture == handle) {
+                        map
+                    } else {
+                        warn!(
+                            "Couldn't find a mapping for {:?}, handle {:?}",
+                            global, handle
+                        );
+                        continue;
+                    };
 
                 if let TypeInner::Image {
                     class: ImageClass::Storage(storage_format),
                     ..
                 } = module.types[global.ty].inner
                 {
-                    write!(out, "{}) ", write_format_glsl(storage_format),)?;
-                } else {
-                    write!(out, ") ")?;
+                    write!(out, "layout({}) ", write_format_glsl(storage_format))?;
                 }
 
                 if global.storage_access == StorageAccess::LOAD {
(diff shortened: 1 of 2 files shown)


src/back/glsl.rs
```diff
@@ -61,23 +62,32 @@ pub struct Options {
     pub entry_point: (String, ShaderStage),
 }
 
+#[derive(Debug, Clone)]
+pub struct TextureMapping {
+    pub texture: Handle<GlobalVariable>,
+    pub sampler: Handle<GlobalVariable>,
```
kvark, line 68: this should be `Option<>`, since not all textures are used with samplers

src/back/glsl.rs
```diff
@@ -1632,3 +1663,191 @@ fn write_format_glsl(format: StorageFormat) -> &'static str {
...
+    functions: &FastHashMap<Handle<Function>, String>,
+) -> Result<Vec<TextureMapping>, Error> {
+    fn collect_texture_mapping_expr(
+        func: &Function,
+        expr: Handle<Expression>,
+        mappings: &mut Vec<TextureMapping>,
+    ) -> Result<(), Error> {
+        match func.expressions[expr] {
+            Expression::Access { base, index } => {
+                collect_texture_mapping_expr(func, base, mappings)?;
```
kvark, line 1678: We have the logic of walking through expressions in `proc::Interface`. It would be nice to share this code wider by separating the visitor logic from the traversal.

src/back/glsl.rs
```diff
@@ -1632,3 +1663,191 @@ fn write_format_glsl(format: StorageFormat) -> &'static str {
...
+                    crate::SampleLevel::Exact(expr) | crate::SampleLevel::Bias(expr) => {
+                        collect_texture_mapping_expr(func, expr, mappings)?
+                    }
+                    crate::SampleLevel::Auto => {}
+                }
+                if let Some(expr) = depth_ref {
+                    collect_texture_mapping_expr(func, expr, mappings)?;
+                }
+
+                let mapping = mappings.iter().find(|map| map.texture == tex_handle);
```
kvark, line 1728: If I understand correctly, this is an internal map we are collecting. It looks like we are often searching it by the texture handle. Would it make things easier if this internal map was `FastHashMap<Handle<GlobalVar>, Handle<GlobalVar>>`? (addressable by texture handles)

src/back/glsl.rs
```diff
@@ -1596,6 +1620,13 @@ fn builtin_to_glsl(builtin: BuiltIn) -> &'static str {
     }
 }
 
+fn get_base_ty<'a>(types: &'a Arena<Type>, ty: &'a TypeInner) -> &'a TypeInner {
+    match ty {
+        TypeInner::Pointer { base, .. } => get_base_ty(types, &types[*base].inner),
```
kvark, line 1625: this function specifically works with pointers? why did you need it?

src/back/glsl.rs
```diff
@@ -214,13 +223,145 @@ pub fn write<'a>(module: &'a Module, out: &mut impl Write, options: Options) ->
 
     writeln!(out)?;
 
-    let mut globals_lookup = FastHashMap::default();
+    let mut functions = FastHashMap::default();
 
-    for ((handle, global), usage) in module.global_variables.iter().zip(func.global_usage.iter()) {
-        if usage.is_empty() {
+    for (handle, func) in module.functions.iter() {
```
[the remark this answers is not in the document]
JCapucho: Right now this exposes all the functions except for the ones that are entry points

src/back/glsl.rs
JCapucho: I had some issues where i got pointers where other types where expected this is a temporary solution

src/back/glsl.rs
```diff
@@ -1,10 +1,12 @@
 use super::{BorrowType, MaybeOwned};
 use crate::{
+    proc::{Interface, Visitor},
     Arena, ArraySize, BinaryOperator, BuiltIn, Constant, ConstantInner, DerivativeAxis, Expression,
     FastHashMap, Function, FunctionOrigin, GlobalVariable, Handle, ImageClass, Interpolation,
     IntrinsicFunction, LocalVariable, MemberOrigin, Module, ScalarKind, ShaderStage, Statement,
     StorageAccess, StorageClass, StorageFormat, StructMember, Type, TypeInner, UnaryOperator,
 };
+use log::warn;
```
kvark, line 9: note that in other places we just use the full path for logging, i.e. `log::warn!("...")`

src/back/glsl.rs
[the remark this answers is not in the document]
kvark: Filed #186 to follow-up

src/back/glsl.rs
kvark: Please tell more. I thought we fixed this in #100. If not, this should be fixed in the front-ends, not worked around here.

src/back/glsl.rs
```diff
@@ -1624,3 +1659,100 @@ fn write_format_glsl(format: StorageFormat) -> &'static str {
...
+                    Expression::GlobalVariable(global) => global,
+                    // Temp
+                    Expression::Load { pointer } => match self.expressions[pointer] {
+                        Expression::GlobalVariable(global) => global,
+                        _ => unreachable!(),
+                    },
+                    _ => unreachable!(),
+                };
+
+                let mapping = self.map.get(&tex_handle);
```
kvark, line 1693: let's use the entry API. This would be `match self.map.entry(tex_handle) {...}`

src/back/glsl.rs
```diff
@@ -1624,3 +1659,100 @@ fn write_format_glsl(format: StorageFormat) -> &'static str {
...
+
+                if mapping.map_or(false, |sampler| *sampler != None) {
+                    self.error = Some(Error::Custom(String::from(
+                        "Cannot use texture with two different samplers",
+                    )));
+                }
+
+                if mapping.is_none() {
```
kvark, line 1724: same here

kvark (changes requested):
Looks great! We just need one more iteration on it.

src/back/glsl.rs
JCapucho: I just checked right now `test-data/spv/cube.frag.spv` and the frontend is producing multiple loads to globals

src/back/glsl.rs
```diff
@@ -743,7 +802,14 @@ fn write_expression<'a, 'b>(
             module.borrow_type(builder.locals[handle].ty),
         ),
         Expression::Load { pointer } => {
-            write_expression(&builder.expressions[pointer], module, builder)?
+            let (expr, ty) = write_expression(&builder.expressions[pointer], module, builder)?;
+
+            let ty = match *ty.borrow() {
+                TypeInner::Pointer { base, .. } => base,
```
kvark, line 808: I think what we want is to not have pointers where they are implied. I.e. when accessing any global or local variables, doing pointer expression indirection is pointless

src/back/glsl.rs
```diff
@@ -1344,46 +1383,54 @@ fn write_type<'a>(
...
+    {
+        return Err(Error::Custom(String::from(
+            "Arrayed non 2d images aren't supported",
+        )));
+    }
+
+    Ok(format!(
+        "{}{}{}{}{}",
+        match kind {
+            ScalarKind::Sint => "i",
```
kvark, line 1410: let's use the `map_scalar` function here

kvark (approved):
Looks great! I have a few minor-ish notes left

kvark (approved):

PR #182, opened 2020-09-08
Move storage class from image type into the global declaration

kvark:
It appears that the access flags are not really a part of the image type in SPIR-V. The `ReadOnly` and `WriteOnly` flags are only for the "Kernel" shader model, which we don't use.
Therefore, it's totally controlled by the decorations `NonReadable` and `NonWritable`, which are put on the global variables. Therefore, it makes sense to have the `storage_access` on the global vars instead of the image types, and now it can apply to buffers as well!

Unresolved: perhaps, the flags should be a part of the `StorageClass::Uniform` variant? I'm hesitating because SPIR-V confusingly considers storage buffers and images in this class. Also, it has an extension for the `StorageBuffer` class, which confuses me even more...

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -145,6 +145,14 @@ struct Block {
     buffer: bool,
 }
 
+bitflags::bitflags! {
+    #[derive(Default)]
+    struct DecorationFlags: u32 {
+        const NON_READABLE = 0x1;
+        const NON_WRITABLE = 0x2;
+    }
+}
+
 #[derive(Debug, Default)]
 struct Decoration {
     name: Option<String>,
@@ -156,6 +164,7 @@ struct Decoration {
     offset: Option<spirv::Word>,
     array_stride: Option<NonZeroU32>,
     interpolation: Option<crate::Interpolation>,
+    flags: DecorationFlags,
 }
 
 impl Decoration {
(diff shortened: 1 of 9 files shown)


src/front/spv/mod.rs
```diff
@@ -2277,13 +2282,35 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+            crate::TypeInner::Struct { .. } => class == crate::StorageClass::StorageBuffer,
+            crate::TypeInner::Image {
+                class: crate::ImageClass::Storage(_),
+                ..
+            } => true,
+            _ => false,
+        };
+
+        let storage_access = if is_storage {
+            let mut access = crate::StorageAccess::all();
```
Napokue, line 2295: So if I understand correctly, the `StorageAccess` for Structs and Images are by default both Readable and Writable? I am more in favor of using it defensively and start with `StorageAccess::empty()`, or isn't this an expected behavior?

Napokue (commented):
So to recap your investigation: Kernel capability is not used/will not be used, ever (I did not know this)? Which means the optional `Access Qualifier` operand for `OpTypeImage` becomes irrelevant, and if we want to have to same behavior we just add a decoration flag with `NonWritable` or `NonReadable`? Which then can also be used for an `Uniform Storage Class` with the `BufferBlock` decoration, correct?

I'm not sure about including it into the `StorageClass::Uniform` as to my understanding images and uniform storage classes are not the same, but I don't have experience with this, so take that with a grain of salt.

Outside of this and the one question I have in the PR, it LGTM.

src/front/spv/mod.rs
kvark: this is SPIRV frontend: stuff is RW by default unless one of  those decorations is present
In the IR we are explicit about this with this PR :)

kvark:
> So to recap your investigation

Oll Korrect!

> I'm not sure about including it into the StorageClass::Uniform as to my understanding images and uniform storage classes are not the same, but I don't have experience with this, so take that with a grain of salt.

That's what I assumed too, but I tried it in the shader playground, and it shows a storage writable image being `UniformConstant` class... yikes
I think we should reconsider the storage classes in IR, but let's do it in the working group first.

PR #183, opened 2020-09-09
Extended Interface to be used in more places

JCapucho:


src/proc/interface.rs
```diff
@@ -1,11 +1,19 @@
...
-struct Interface<'a> {
+pub struct Interface<'a, T> {
     expressions: &'a Arena<crate::Expression>,
-    uses: Vec<crate::GlobalUse>,
+    visitor: &'a mut T,
 }
 
-impl<'a> Interface<'a> {
+pub trait Visitor {
+    fn expr_visitor(&mut self, _: &crate::Expression) {}
```
kvark, line 9: maybe `visit_expr` and `visit_stmt` instead?

src/proc/interface.rs
```diff
@@ -92,9 +96,10 @@ impl<'a> Interface<'a> {
                 }
             }
         }
+        self.visitor.expr_visitor(&self.expressions[handle])
```
kvark, line 99: can we move it up so that the current element is visited before its children?

src/proc/interface.rs
```diff
@@ -1,11 +1,19 @@
...
+pub trait Visitor {
+    fn expr_visitor(&mut self, _: &crate::Expression) {}
+    fn sta_visitor(&mut self, _: &crate::Statement, _: &Arena<crate::Expression>) {}
+}
+
+impl<'a, T> Interface<'a, T>
+where
+    T: Visitor,
+{
     fn add_inputs(&mut self, handle: Handle<crate::Expression>) {
```
kvark, line 17: this should now be renamed to `traverse_expr` or something like that

src/proc/interface.rs
```diff
@@ -92,9 +96,10 @@ impl<'a> Interface<'a> {
                 }
             }
         }
+        self.visitor.expr_visitor(&self.expressions[handle])
     }
 
-    fn collect(&mut self, block: &[crate::Statement]) {
+    pub fn collect(&mut self, block: &[crate::Statement]) {
```
kvark, line 102: and this - to just `traverse`

src/proc/interface.rs
```diff
@@ -145,16 +150,44 @@ impl<'a> Interface<'a> {
...
-                                self.uses[handle.index()] |= crate::GlobalUse::STORE;
-                                break;
-                            }
                             _ => break,
                         }
                     }
                     self.add_inputs(value);
                 }
             }
+            self.visitor.sta_visitor(statement, &self.expressions)
```
kvark, line 159: similarly, I think it would make sense to visit the statements first, and then expressions

src/proc/interface.rs
```diff
@@ -145,16 +150,44 @@ impl<'a> Interface<'a> {
                             crate::Expression::AccessIndex { base, .. } => {
                                 left = base;
                             }
-                            crate::Expression::GlobalVariable(handle) => {
```
kvark, line 148: as I'm thinking about this now, we don't want the visitor to repeat this descend themselves in order to reach that variable. How about this: instead of `visit_stmt` we have `visit_lhs_expr`, called on the final destination of this loop? It would mean "left-hand side" expression, and for the visitor it would mean that the IR wants to store/write into it.

src/proc/interface.rs
```diff
@@ -145,16 +150,44 @@ impl<'a> Interface<'a> {
                             crate::Expression::AccessIndex { base, .. } => {
```
kvark, line 150: also, I see a bug here: all the expressions starting from `pointer` down to the last element in this sub-tree should be added to `self.add_inputs`

kvark (changes requested):
That's a great start! Have a few ideas on how we can shape it better ;)

src/proc/interface.rs
```diff
@@ -1,34 +1,41 @@
...
+}
+
+impl<'a, T> Interface<'a, T>
+where
+    T: Visitor,
+{
+    fn traverse_expr(&mut self, handle: Handle<crate::Expression>) {
         use crate::Expression as E;
+
+        self.visitor.visitor_expr(&self.expressions[handle]);
```
kvark, line 20: can we assign this reference to a temporary to avoid indexing by `handle` twice?

src/proc/interface.rs
```diff
@@ -1,34 +1,41 @@
...
+pub struct Interface<'a, T> {
     expressions: &'a Arena<crate::Expression>,
-    uses: Vec<crate::GlobalUse>,
+    visitor: &'a mut T,
 }
 
-impl<'a> Interface<'a> {
-    fn add_inputs(&mut self, handle: Handle<crate::Expression>) {
+pub trait Visitor {
+    fn visitor_expr(&mut self, _: &crate::Expression) {}
```
kvark, line 9: this start with a verb, which is "visit", not "visitor". So this would be `visit_expr` and `visit_lhs_expr`

src/proc/interface.rs
```diff
@@ -37,139 +44,209 @@ impl<'a> Interface<'a> {
...
+                    self.traverse_expr(value);
                 }
             }
         }
     }
 }
 
+struct GlobalUseVisitor(Vec<crate::GlobalUse>);
+
+impl Visitor for GlobalUseVisitor {
```
kvark, line 168: this is just beautiful now!

kvark (changes requested):
A few last touches, we are almost there!

kvark (approved):

PR #184, opened 2020-09-10
[glsl] Promote glsl-new to glsl

pjoe:
- Delete `glsl` front-end
- Rename `glsl-new` to `glsl`
- Adjust `Cargo.toml`
  - Remove `glsl` crate from deps
  - Remove `glsl_preprocessor` for now
- Fixup `examples/convert.rs`
- Disable `constant_expressions` in `tests/convert.rs`


Cargo.toml
```diff
@@ -15,16 +15,14 @@ fxhash = "0.2"
...
 pomelo = { version = "0.1.4", optional = true }
 thiserror = "1.0"
 serde = { version = "1.0", features = ["derive"], optional = true }
 petgraph = { version ="0.5", optional = true }
 
 [features]
 default = []
-glsl_preprocessor = ["glsl"]
-glsl-new = ["pomelo"]
+glsl = ["pomelo"]
```
kvark, line 25: let's call this `glsl-in` to be consistent

kvark (changes requested):
Thank you! We just need to rename the feature, everything else looks good

Cargo.toml
pjoe: done

kvark (approved):
thank you!

PR #185, opened 2020-09-10
ir: bitcast expression

### LINKED ISSUE #161: Implement the bitcast operator
Incoming SPIR-Vs have `Bitcast`.  That's the last major blocker in validating wgpu-rs examples.
We need to do a research about how this looks like in WGSL, what the previous investigations are, and find a way to process this instruction.

kvark:
Fixes #161

### DIFF
--- a/src/back/glsl.rs
+++ b/src/back/glsl.rs
@@ -684,20 +684,7 @@ fn write_expression<'a, 'b>(
             let constructor = match module.types[ty].inner {
                 TypeInner::Vector { size, kind, width } => format!(
                     "{}vec{}",
-                    match kind {
-                        ScalarKind::Sint => "i",
-                        ScalarKind::Uint => "u",
-                        ScalarKind::Float => match width {
-                            4 => "",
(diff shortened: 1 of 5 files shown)


MatusT:
This PR doesn't cover instructions Op*Convert and OpQuantiOpQuantizeToF16. However, from what I can see the IR itself doesn't have any support for types outside of 32-bit width therefore relegating this issue to another PR.

Therefore, I am giving this PR the green light.

MatusT (approved):

src/back/glsl.rs
```diff
@@ -1255,25 +1237,45 @@ fn write_constant(
     })
 }
 
+struct ScalarString<'a> {
+    prefix: &'a str,
```
JCapucho, line 1241: Maybe we should use an `Option` here so that we can support 2 bytes float and others in the future

JCapucho (changes requested):
Just one small thing but it isn't blocking merge

PR #187, opened 2020-09-10
[spirv-in] make pointers totally transparent

### LINKED ISSUE #100: Don't use pointers for accessing global variables

kvark:
Fixes #100 for good, as a follow-up to #101 

In SPIR-V all load/stores/access chains have to go through pointers.
I don't think it makes sense for the IR - it only clutters it, and doesn't allow us to do more than we can, already.
All the backends have an easier time working with load/stores/accesses directly into expressions and variables.

### DIFF
--- a/src/back/msl.rs
+++ b/src/back/msl.rs
@@ -546,6 +546,19 @@ impl<W: Write> Writer<W> {
                             &crate::TypeInner::Vector { size, kind, width },
                             &crate::TypeInner::Vector { .. },
                         ) => MaybeOwned::Owned(crate::TypeInner::Vector { size, kind, width }),
+                        (
+                            &crate::TypeInner::Matrix {
+                                rows,
+                                columns: _,
+                                kind,
+                                width,
+                            },
+                            &crate::TypeInner::Vector { .. },
+                        ) => MaybeOwned::Owned(crate::TypeInner::Vector {
+                            size: rows,
+                            kind,
+                            width,
+                        }),
                         (_, _) => {
                             return Err(Error::UnableToInferBinaryOpOutput(ty_left, op, ty_right))
                         }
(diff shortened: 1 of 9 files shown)


src/back/msl.rs
```diff
@@ -546,6 +546,19 @@ impl<W: Write> Writer<W> {
                             &crate::TypeInner::Vector { size, kind, width },
                             &crate::TypeInner::Vector { .. },
                         ) => MaybeOwned::Owned(crate::TypeInner::Vector { size, kind, width }),
+                        (
```
Napokue, line 549: I don't understand this matching arm. Is it possible to have the following expression in Metal?

Napokue: Ah yes I think I already answered my own question. I have search through the Metal documentation and it should be possible: "A right vector operand is treated as a column vector and a left vector operand as a row vector.".

The left operand as vector is not yet implemented, are there any plans for this?

src/front/spv/rosetta.rs
```diff
@@ -0,0 +1,24 @@
+use std::{fs, path::Path};
+
+const TEST_PATH: &str = "test-data";
+
+fn rosetta_test(file_name: &str) {
+    if true {
+        return; //TODO: fix this test
```
Napokue, line 7: Does this still need to be fixed in this PR or is this for later?

Napokue (commented):
I really like this idea, thanks! I only have two questions commented, but those are minor, LGTM!

src/back/msl.rs
kvark: I was implemented as much as needed to get the SPV examples running (more concretely, the cube shaders here, and the wgpu-rs shaders). We can consider expanding this in other PRs. Honestly, I don't know if IR needs to care about "vector * matrix" syntax... since it's trivial to polyfill as `translate(matrix * vector)`. Let's see where the working group goes with this and follow-up?

src/front/spv/rosetta.rs
kvark: this needs more work, and it's not critical to the PR, so we'll fix it in the follow-ups

kvark:
Thank you for the review!

src/back/msl.rs
Napokue: Sure no problem, I was just wondering, never made a Metal shader :)

PR #188, opened 2020-09-11
Fix lexer operator issues (with tests)

pjoe:


src/front/glsl/lex_tests.rs
```diff
@@ -158,3 +158,89 @@ fn glsl_lex_version() {
...
+fn glsl_lex_operators() {
+    let source = "+ - * | & % / += -= *= |= &= %= /= ++ -- || && ^^";
+    let lex = Lexer::new(source);
+    let tokens: Vec<Token> = lex.collect();
+    assert_eq!(tokens.len(), 19);
+
+    let mut iter = tokens.iter();
+    assert_eq!(
+        format!("{:?}", iter.next().unwrap()),
+        "Plus(TokenMetadata { line: 0, chars: 0..1 })"
```
kvark, line 172: why are we converting the data to string, just to then compare with the other piece of data? Would it be simpler to compare the data directly, i.e.:
```rust
    assert_eq!(
        iter.next(),
        Some(Plus(TokenMetadata { line: 0, chars: 0..1 })),
  );
```

kvark (changes requested):
Just one note here, looks good overall!

src/front/glsl/lex_tests.rs
pjoe: The parser's `Token` doesn't implement `PartialEq`, guessing that could be added, but don't think that could be done only for tests.

Also not sure if output for failing tests would be as nice. Here is how it looks with string:

```
---- front::glsl::lex_tests::glsl_lex_version stdout ----                                                 
thread 'main' panicked at 'assertion failed: `(left == right)`                                            
  left: `"Identifier((TokenMetadata { line: 0, chars: 13..17 }, \"core\"))"`,                             
 right: `"Identifier((TokenMetadata { line: 0, chars: 12..17 }, \"core\"))"`', src/front/glsl/lex_tests.rs:156:5
```

kvark: Let's try that! To implement it for tests only you could do
```rust
#[cfg_attr(test, derive(PartialEq))]`
```

pjoe: Ok, will give that a try and see how it looks for failures

pjoe: Seems to work well, update pushed.

src/front/glsl/lex_tests.rs
```diff
@@ -1,5 +1,7 @@
 use super::lex::Lexer;
 use super::parser::Token;
+use super::parser::Token::*;
```
kvark, line 3: nit: generally the style we are trying to keep here is one crate per use statement, i.e. this would be something like:
```rust
use super::{lex::Lexer, parser::Token::{self, *}, token::TokenMetadata};
```
this is not a blocker for merging though:)

src/front/glsl/lex_tests.rs
```diff
@@ -10,28 +12,49 @@ fn glsl_lex_simple() {
 
     let mut iter = tokens.iter();
```
kvark, line 13: why do we need to collect them? can we just call `next` on the iterator, and then at the end assert that the next() is None (instead of checking the count here)?

kvark (commented):
Looks great, thank you!
Just a few minorish concerns left

src/front/glsl/lex_tests.rs
pjoe: done

src/front/glsl/lex_tests.rs
pjoe: good idea, makes writing the tests a bit simpler. Fixing ...

kvark (approved):

PR #189, opened 2020-09-11
[glsl-in] Add gl_VertexIndex

pjoe:
Refactor variable handling out of parser.rs

### DIFF
--- a/src/front/glsl/variables.rs
+++ b/src/front/glsl/variables.rs
@@ -1,0 +1,81 @@
+use crate::{
+    Binding, BuiltIn, Expression, GlobalVariable, Handle, ScalarKind, ShaderStage, StorageAccess,
+    StorageClass, Type, TypeInner, VectorSize,
+};
+
+use super::ast::*;
+
+pub fn lookup_variable(program: &mut Program, name: &str) -> Option<Handle<Expression>> {
+    let mut expression: Option<Handle<Expression>> = None;
+    if name == "gl_Position" {
+        if let ShaderStage::Vertex | ShaderStage::Fragment { .. } = program.shader_stage {
+            let h = program.global_variables.fetch_or_append(GlobalVariable {
+                name: Some(name.into()),
+                class: match program.shader_stage {
+                    ShaderStage::Vertex => StorageClass::Output,
+                    ShaderStage::Fragment { .. } => StorageClass::Input,
+                    _ => StorageClass::Input,
+                },
+                binding: Some(Binding::BuiltIn(BuiltIn::Position)),
+                ty: program.types.fetch_or_append(Type {
+                    name: None,
+                    inner: TypeInner::Vector {
+                        size: VectorSize::Quad,
+                        kind: ScalarKind::Float,
+                        width: 4,
+                    },
+                }),
+                interpolation: None,
+                storage_access: StorageAccess::empty(),
+            });
+            program.lookup_global_variables.insert(name.into(), h);
+            let exp = program
+                .context
+                .expressions
+                .append(Expression::GlobalVariable(h));
+            program
+                .context
+                .lookup_global_var_exps
+                .insert(name.into(), exp);
+
+            expression = Some(exp);
+        }
+    }
+    if program.shader_stage == ShaderStage::Vertex && name == "gl_VertexIndex" {
+        let h = program.global_variables.fetch_or_append(GlobalVariable {
+            name: Some(name.into()),
+            class: StorageClass::Input,
+            binding: Some(Binding::BuiltIn(BuiltIn::VertexIndex)),
+            ty: program.types.fetch_or_append(Type {
+                name: None,
+                inner: TypeInner::Scalar {
+                    kind: ScalarKind::Uint,
+                    width: 4,
+                },
+            }),
+            interpolation: None,
+            storage_access: StorageAccess::empty(),
+        });
+        program.lookup_global_variables.insert(name.into(), h);
+        let exp = program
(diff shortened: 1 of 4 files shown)


src/front/glsl/variables.rs
```diff
@@ -0,0 +1,81 @@
+use crate::{
+    Binding, BuiltIn, Expression, GlobalVariable, Handle, ScalarKind, ShaderStage, StorageAccess,
+    StorageClass, Type, TypeInner, VectorSize,
+};
+
+use super::ast::*;
+
+pub fn lookup_variable(program: &mut Program, name: &str) -> Option<Handle<Expression>> {
```
kvark, line 8: would it make sense to have it as a method on the `Program`? It can still be in a different module (like `variables` module here)

src/front/glsl/variables.rs
```diff
@@ -0,0 +1,81 @@
...
+    let mut expression: Option<Handle<Expression>> = None;
+    if name == "gl_Position" {
+        if let ShaderStage::Vertex | ShaderStage::Fragment { .. } = program.shader_stage {
+            let h = program.global_variables.fetch_or_append(GlobalVariable {
+                name: Some(name.into()),
+                class: match program.shader_stage {
+                    ShaderStage::Vertex => StorageClass::Output,
+                    ShaderStage::Fragment { .. } => StorageClass::Input,
+                    _ => StorageClass::Input,
```
kvark, line 17: what is this? compute? we should generate an error here

src/front/glsl/variables.rs
```diff
@@ -0,0 +1,81 @@
...
+                .append(Expression::GlobalVariable(h));
+            program
+                .context
+                .lookup_global_var_exps
+                .insert(name.into(), exp);
+
+            expression = Some(exp);
+        }
+    }
+    if program.shader_stage == ShaderStage::Vertex && name == "gl_VertexIndex" {
```
kvark, line 44: you probably want to shape it as a big match:
```rust
match name {
  "gl_Position" => {
    // check stage, etc
  }
  "gl_VertexIndex" => {
    // check stage, etc
  }
}
```

src/front/glsl/variables.rs
pjoe: Sure, is expecting this to grow large (lots of builtins), but if can still be in it's own file, that would be fine. I'll figure out how.

src/front/glsl/variables.rs
pjoe: I don't think we should error, in compute you could still make your own var named 'gl_Position' (maybe not very sensible, but you coudl). Anyway we are already checking on the stage in outer if, so I guess this is only to satisfy that match must cover all variants, but maybe there is a better way?

pjoe:
Added commit attempting to fix mentioned issue, waiting for the verdict :)

src/front/glsl/variables.rs
kvark: if it's already checked, then it's not expected here, and thus it should be `unreachable!()`

src/front/glsl/variables.rs
```diff
@@ -0,0 +1,82 @@
...
+                    let exp = self
+                        .context
+                        .expressions
+                        .append(Expression::GlobalVariable(h));
+                    self.context.lookup_global_var_exps.insert(name.into(), exp);
+
+                    expression = Some(exp);
+                }
+            }
+            "gl_VertexIndex" => {
```
kvark, line 43: These can't be legally used for user variables: `identifiers starting with "gl_" are reserved`
Therefore, instead of filtering the stage out, we should just assume the stage is right (or validate the stages, if you want validation)

kvark (changes requested):

src/front/glsl/variables.rs
pjoe: Ok, I think I'll try to have it error when referencing a `gl_` var not valid for the stage - at least that's what would make most sense to me when thinking as a user passing in a shader :)

pjoe: For ref: what glslang does is error `undeclared identifier`, when trying to reference a `gl_` var in wrong stage. And if trying to declare a user var with `gl_` it errors `identifiers starting with "gl_" are reserved`

pjoe: Looking over it once more, this is how current code behaves (for variable references). Could add the `gl_` prefix check for declarations, but maybe best in separate PR?

pjoe: thinking some more, as a user I'd prefer a better error than just `undeclared identifier`, something like `gl_Position is not available in this stage`. I'll see if I can do something like that, feature flagged with `glsl-validate`

src/front/glsl/variables.rs
```diff
@@ -0,0 +1,93 @@
...
+        let mut expression: Option<Handle<Expression>> = None;
+        match name {
+            "gl_Position" => {
+                #[cfg(feature = "glsl-validate")]
+                match self.shader_stage {
+                    ShaderStage::Vertex | ShaderStage::Fragment { .. } => {}
+                    _ => {
+                        return Err(ErrorKind::VariableNotAvailable(name.into()));
+                    }
+                };
```
pjoe, line 20: Maybe there is a better way than `match`, I tried `if ! let`, but that doesn't work.

pjoe, line 20: Maybe there is a better way than `match`, I tried `if ! let`, but that doesn't work.

kvark (approved):

PR #190, opened 2020-09-12
[glsl-in] Fix unused braces in parser.rs

pjoe:
Added newlines to `parser.rs` in order to avoid `unused_braces` issue.

Maybe not the prettiest, but pomelo needs those braces - and guess clippy can't really understand why :D


kvark (approved):
Thank you!

PR #191, opened 2020-09-12
glsl-out: More glsl versions

JCapucho:


src/back/glsl.rs
```diff
@@ -68,18 +69,202 @@ pub struct TextureMapping {
     pub sampler: Option<Handle<GlobalVariable>>,
 }
 
-const SUPPORTED_CORE_VERSIONS: &[u16] = &[450, 460];
+const SUPPORTED_CORE_VERSIONS: &[u16] = &[330, 400, 410, 420, 430, 440, 450];
```
kvark, line 72: Could you explain more (in the PR description, or in the chat) why you think this is needed (even if it's just "just because we can")?

src/back/glsl.rs
```diff
@@ -75,18 +76,193 @@ pub struct TextureMapping {
...
+
+    #[allow(clippy::collapsible_if)]
+    pub fn write(&self, version: Version, out: &mut impl Write) -> Result<(), Error> {
+        let (v, es) = match version {
+            Version::Desktop(v) => (v, false),
+            Version::Embedded(v) => (v, true),
+        };
+
+        if self.0.contains(Features::COMPUTE_SHADER) {
+            if (es && v < 310) || (!es && v < 420) {
```
kvark, line 117: This may be a bit error-prone. It would better to encode these restrictions as `Version` values, basically, and then define a `PartialOrd` for them in a way that `Desktop` vs `Embedded` is considered to have no order. See https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=276f348602ba91fb57265ce50d524bc4 for inspiration

src/front/wgsl/mod.rs
```diff
@@ -563,7 +563,7 @@ impl Parser {
                 Token::Separator('.') => {
                     let _ = lexer.next();
                     let name = lexer.next_ident()?;
-                    println!("Resolving '{}' for {:?}", name, ctx.resolve_type(handle)); //TEMP
```
kvark, line 566: oops!

src/front/wgsl/mod.rs
```diff
@@ -563,7 +563,7 @@ impl Parser {
...
+                    log::debug!("Resolving '{}' for {:?}", name, ctx.resolve_type(handle)); //TEMP
```
kvark, line 566: please remove the `//TEMP` as this is changed to be properly logged

src/front/wgsl/mod.rs
JCapucho: Had to change it because the tests were being spammed I'll remove

src/back/glsl.rs
```diff
@@ -108,34 +128,29 @@ impl FeaturesManager {
...
-            if (es && v < 310) || (!es && v < 400) {
+            if version < Version::Embedded(310) || version < Version::Desktop(400) {
                 return Err(Error::Custom(format!(
                     "Version {} doesn't support buffer storage class",
                     version
                 )));
             }
 
-            if !es {
+            if let Version::Desktop(_) = version {
```
kvark, line 153: was this meant to be `!version.is_es()`?

kvark (approved):
Please squash commits in a way that makes every commit testable

kvark (approved):

Issue #193, opened 2020-09-13
spirv/rosetta test: constants come first


kvark:
In spir-v front-end, constants are always declared up-front, and then come first in the expression list.
In other front-ends constants can be inlined, and can come later...

Issue #192, opened 2020-09-13
spirv/rosetta test: type declaration order


kvark:
Since SPIR-V requires `float` to be declared before `vec4<float>`, for example, it can't possibly express a case where `vec4` comes first in the source (e.g. WGSL).
What can we do?
  1. Always write Rosetta tests in a way that primitive types/scalars come first. In convenient, but a relatively light-handed solution/workaround.
  2. Make SPIR-V front-end aware of implicit type declarations and somehow produce a module without them. This seems like it would require quite a bit of engineering (based on a test 10 minute dive)
  3. Rethink how tests are made in general?

PR #194, opened 2020-09-13
Minor fixes in feature checks

kvark:
Also adds the WIP `simple.spv` shader that I've been trying to massage into Rosetta shape (using asm/disasm) before hitting #192 and #193


Napokue (approved):
Thank you, LGTM!

PR #195, opened 2020-09-14
Use the typifier in both front-ends and backends

### LINKED ISSUE #162: Use typifier in the backends
Currently it's used in the frontends only, and it mutates the type arena.
GLSL and MSL backends would benefit from having this automated too, but we need to do something about the mutable arena.

kvark:
Fixes #162


src/back/glsl.rs
```diff
@@ -481,19 +502,20 @@ pub fn write<'a>(
...
-        let mut builder = StatementBuilder {
+        let builder = StatementBuilder {
             functions: &functions,
             globals: &globals_lookup,
             locals_lookup: &locals,
             structs: &structs,
             args: &args,
             expressions: &func.expressions,
-            locals: &func.local_variables,
+            //locals: &func.local_variables,
```
JCapucho, line 512: Any reason not to remove this?

src/back/glsl.rs
```diff
@@ -503,20 +525,21 @@ pub fn write<'a>(
...
-    pub expressions: &'a Arena<Expression>,
-    pub locals: &'a Arena<LocalVariable>,
-    pub features: SupportedFeatures,
+    functions: &'a FastHashMap<Handle<Function>, String>,
+    globals: &'a FastHashMap<Handle<GlobalVariable>, String>,
+    locals_lookup: &'a FastHashMap<Handle<LocalVariable>, String>,
+    structs: &'a FastHashMap<Handle<Type>, String>,
+    args: &'a FastHashMap<u32, String>,
+    expressions: &'a Arena<Expression>,
+    //locals: &'a Arena<LocalVariable>,
```
JCapucho, line 534: same thing

JCapucho (approved):
Only reviewed the glsl backend, everything seems fine left only a comment about removing some commented parts

PR #197, opened 2020-09-15
[glsl-in] Add uniforms and structs

pjoe:
Also reorder things slightly in parser, in order to match the GLSL spec

### DIFF
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -6,8 +6,8 @@ pomelo! {
     %include {
         use super::super::{error::ErrorKind, token::*, ast::*};
         use crate::{Arena, BinaryOperator, Binding, Block, Constant, ConstantInner, Expression,
-            Function, GlobalVariable, Handle, LocalVariable, ScalarKind,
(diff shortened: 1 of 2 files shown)


src/front/glsl/parser.rs
```diff
@@ -503,22 +539,37 @@ pomelo! {
...
 
+    interpolation_qualifier ::= Interpolation((_, i)) {
+        i
+    }
+
     layout_qualifier ::= Layout LeftParen layout_qualifier_id_list(l) RightParen {
-        l
+        if let Some((_, loc)) = l.iter().find(|&q| q.0.as_str() == "location") {
+            Binding::Location(*loc)
+        } else if let Some((_, binding)) = l.iter().find(|&q| q.0.as_str() == "binding") {
```
kvark, line 549: this is implicitly pattern matches by reference. The type on the right is `Option<&Something>`. So if you do this as `Some(&(_, binding))` you'd make it more explicit, clear, and avoid the need to do `*binding*` later

kvark (approved):

PR #198, opened 2020-09-16
Add ArrayLength expression

kvark:
See https://github.com/gpuweb/gpuweb/issues/989
This PR cares about IR, the backend and validator support will need to come later.

### DIFF
--- a/src/back/glsl.rs
+++ b/src/back/glsl.rs
@@ -1042,6 +1042,10 @@ fn write_expression<'a, 'b>(
                     .join(","),
             ))
         }
+        Expression::ArrayLength(expr) => {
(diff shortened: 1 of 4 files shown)


Napokue (approved):
Thank you, LGTM!

Why is this only applicable to dynamic-sized arrays? In C# for instance, a List construct (dynamic-sized array) has a function Count(), that calculates the length just like we do here. Static-sized arrays have a property Length that is immediately calculated upon array creation and will be static after that.

Maybe that would also be nice to have? I don't know, I'm just throwing some thoughts.

kvark:
I agree. I think we can make it work for non-dynamic arrays too, since we know the size.

PR #200, opened 2020-09-17
Refactor entry point IR

kvark:
Another innocent change that turned out to be rather big. I actually was just trying to update WGSL, and the changes started being contagious...

On the IR side, this changes entry points to:
  - be a map, so no collisions by construction
  - store the function directly instead of a handle, so that nothing can call this function

It also renames the "set" to "group" in the decorations.

And it updates WGSL attribute syntax to the one that just got accepted.

### DIFF
--- a/src/back/msl.rs
+++ b/src/back/msl.rs
@@ -31,7 +31,7 @@ pub struct BindTarget {
 
 #[derive(Clone, Debug, Hash, Eq, Ord, PartialEq, PartialOrd)]
 pub struct BindSource {
-    pub set: u32,
+    pub group: u32,
     pub binding: u32,
 }
 
@@ -64,7 +64,6 @@ pub enum Error {
     Format(FmtError),
     Type(ResolveError),
     UnexpectedLocation,
-    MixedExecutionModels(Handle<crate::Function>),
     MissingBinding(Handle<crate::GlobalVariable>),
     MissingBindTarget(BindSource),
     InvalidImageAccess(crate::StorageAccess),
@@ -125,8 +124,8 @@ impl Options<'_> {
                 }),
                 LocationMode::Uniform => Err(Error::UnexpectedLocation),
             },
-            crate::Binding::Descriptor { set, binding } => {
-                let source = BindSource { set, binding };
+            crate::Binding::Resource { group, binding } => {
+                let source = BindSource { group, binding };
                 self.binding_map
                     .get(&source)
                     .cloned()
@@ -180,22 +179,6 @@ impl Indexed for ParameterIndex {
     const CLASS: &'static str = "param";
     fn id(&self) -> usize {
         self.0
-    }
-}
-struct InputStructIndex(Handle<crate::Function>);
-impl Indexed for InputStructIndex {
-    const CLASS: &'static str = "Input";
-    const PREFIX: bool = true;
-    fn id(&self) -> usize {
-        self.0.index()
-    }
-}
-struct OutputStructIndex(Handle<crate::Function>);
-impl Indexed for OutputStructIndex {
-    const CLASS: &'static str = "Output";
-    const PREFIX: bool = true;
-    fn id(&self) -> usize {
-        self.0.index()
     }
 }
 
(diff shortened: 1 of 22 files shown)


src/lib.rs
```diff
@@ -718,12 +714,12 @@ pub struct Function {
...
-    pub stage: ShaderStage,
-    /// Name identifying this entry point.
-    pub name: String,
-    /// The function to be used.
-    pub function: Handle<Function>,
+    /// Early depth test for fragment stages.
+    pub early_depth_test: Option<EarlyDepthTest>,
+    /// Workgroup size for compute stages
+    pub workgroup_size: [u32; 3],
+    /// The function body.
```
Napokue, line 721: Minor, but maybe crucial for people new to the codebase: this isn't actually only the function body, but also the signature and all other details associated with a function. I think this comment is more confusing than having no comment.

src/lib.rs
```diff
@@ -416,8 +412,8 @@ pub enum Binding {
     BuiltIn(BuiltIn),
     /// Indexed location.
     Location(u32),
-    /// Binding within a descriptor set.
-    Descriptor { set: u32, binding: u32 },
+    /// Binding within a resource group.
+    Resource { group: u32, binding: u32 },
```
Napokue, line 416: I'm just curious, but why did you rename it to `group` instead of `set`?

src/front/wgsl/mod.rs
```diff
@@ -1373,30 +1361,56 @@ impl Parser {
         let mut binding = None;
         // Perspective is the default qualifier.
         let mut interpolation = None;
+        let mut stage = None;
```
Napokue, line 1364: Maybe our code base isn't there yet with syntax validation, but what happens if an user forgets to add the `stage` attribute? I think it will completely go through the front-end -> IR, and crashes at the back-end?

I see at line [1509](https://github.com/gfx-rs/naga/pull/200/files#diff-37ba88792a802ec39755d9da39194f86R1509) that it then gets interpreted as a normal function.

src/front/glsl/parser.rs
```diff
@@ -828,9 +828,19 @@ pomelo! {
...
-        let name = f.name.clone();
-        let handle = extra.functions.append(f);
-        if let Some(name) = name {
+        if f.name == extra.entry {
+            let name = extra.entry.take().unwrap();
+            extra.module.entry_points.insert(
+                (extra.shader_stage, name),
+                EntryPoint {
+                    early_depth_test: None,
+                    workgroup_size: [1; 3], //TODO
```
Napokue, line 837: Do you still need to do something here, or is it for a later?

src/back/msl.rs
```diff
@@ -924,8 +907,56 @@ impl<W: Write> Writer<W> {
...
+
+            for (index, &ty) in fun.parameter_types.iter().enumerate() {
+                let name = Name::from(ParameterIndex(index));
+                let member_type_name = module.types[ty].name.or_index(ty);
+                let separator = separate(index + 1 == fun.parameter_types.len());
+                writeln!(self.out, "\t{} {}{}", member_type_name, name, separator)?;
+            }
+            writeln!(self.out, ") {{")?;
+
+            // write down function body
```
Napokue, line 924: Nit: I think this is a leftover comment from down below, as I have seen it at line [1137](https://github.com/gfx-rs/naga/pull/200/files#diff-aaa1f30b3e995f65d9bd7557d98981fdR1137).

src/back/msl.rs
```diff
@@ -941,206 +972,175 @@ impl<W: Write> Writer<W> {
...
-                }
-            }
-            let output_name = fun.name.or_index(OutputStructIndex(fun_handle));
+
+            let output_name = Name {
+                class: "Output",
+                source: NameSource::Custom { name, prefix: true },
+            };
 
             // make dedicated input/output structs
```
Napokue, line 981: Nit: Leftover comment?

Napokue (changes requested):
Thank you, I love the changes, especially those new Stage attributes!

test-data/simple/simple.wgsl
```diff
@@ -1,10 +1,10 @@
 # vertex
-[[location 0]] var<in> a_pos : vec2<f32>;
-[[location 0]] var<out> o_pos : vec4<f32>;
+[[location(0)]] var<in> a_pos : vec2<f32>;
+[[location(0)]] var<out> o_pos : vec4<f32>;
 
+[[stage(vertex)]]
```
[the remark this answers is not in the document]
kvark: I proposed it :)

test-data/simple/simple.expected.ron
```diff
@@ -64,58 +64,56 @@
...
+                    ),
+                    Return(
+                        value: None,
+                    ),
+                ],
+            ),
         ),
-    ],
+    },
 )
```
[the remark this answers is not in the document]
kvark: seems like it's passing, so no?

src/front/wgsl/mod.rs
kvark: If there is no `stage`, then it's not an entry point. IR will be built just fine. If we are talking about wgpu-core as the client, it will try to get the entry point information from IR, and it will not find the entry point, so this is where an error will be produced.

src/front/glsl/parser.rs
kvark: that matches the current behavior, so there is no need to fix this in this PR. I wanted to leave a note here to address later by somebody.

src/back/msl.rs
kvark: moved it to a better place

test-data/simple/simple.expected.ron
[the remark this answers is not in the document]
kvark: I'll add a newline

PR #201, opened 2020-09-17
Consolidate rosetta tests

kvark:
Based on  #200
It seems a bit awkward to have exactly the same testing module in each frontend/backend, all of which load the same expectations module.
r? @pjoe


tests/rosetta.rs
```diff
@@ -0,0 +1,59 @@
+use std::{fs, path::Path};
+
+fn test_rosetta(dir_name: &str) {
+    let dir_path = Path::new("test-data").join(dir_name);
+    let expected = fs::read_to_string(dir_path.join("module.ron")).unwrap();
```
pjoe, line 5: I think this may need to be wrapped in `#[cfg(feature = "serialize")]` in order to get rid of unused var warnings

pjoe (changes requested):
Looks pretty good to me (only looked over the glsl-in parts.

Great to just use `module` directly in `ast::Program`, was actually thinking of doing this, just never got around to it 😄 

Small fix needed for `expected` unused var warning.

PR #203, opened 2020-09-17
Validation of global variables and entry points

kvark:
Another step towards  #59
This PR establishes the structure of the validator, and covers global variables and entry points.

There is an annoying bit that it's hard to say if a global is a storage buffer/texture or not, I wasn't able to figure that out yet, but it's not a blocker.

### DIFF
--- a/src/proc/validator.rs
+++ b/src/proc/validator.rs
@@ -1,5 +1,10 @@
 use super::typifier::{ResolveContext, ResolveError, Typifier};
-use crate::arena::Handle;
+use crate::arena::{Arena, Handle};
+
+const MAX_BIND_GROUPS: u32 = 8;
+const MAX_LOCATIONS: u32 = 64; // using u64 mask
+const MAX_BIND_INDICES: u32 = 64; // using u64 mask
+const MAX_WORKGROUP_SIZE: u32 = 0x4000;
 
 #[derive(Debug)]
 pub struct Validator {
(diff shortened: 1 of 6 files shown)


src/proc/validator.rs
```diff
@@ -28,8 +81,242 @@ impl Validator {
...
+        if ep.early_depth_test.is_some() && stage != crate::ShaderStage::Fragment {
+            return Err(EntryPointError::UnexpectedEarlyDepthTest);
+        }
+        if stage == crate::ShaderStage::Compute {
+            if ep
+                .workgroup_size
+                .iter()
+                .any(|&s| s == 0 || s > MAX_WORKGROUP_SIZE)
+            {
+                return Err(EntryPointError::InvalidWorkgroupSize);
```
Napokue, line 211: Shouldn't the error be something like: `OutOfRangeWorkgroupSize`? The word `Invalid` seems a little bit too generic here in my opinion.

src/proc/validator.rs
```diff
@@ -28,8 +81,242 @@ impl Validator {
...
+        for ((var_handle, var), &usage) in module
+            .global_variables
+            .iter()
+            .zip(&ep.function.global_usage)
+        {
+            if usage.is_empty() {
+                continue;
+            }
+
+            println!("var {:?}", var); //TEMP
```
Napokue, line 229: This can be removed? :)

src/proc/validator.rs
```diff
@@ -28,8 +81,242 @@ impl Validator {
...
+                    //TODO: built-in checks?
+                    let mut storage_usage = crate::GlobalUse::empty();
+                    if var.storage_access.contains(crate::StorageAccess::LOAD) {
+                        storage_usage |= crate::GlobalUse::LOAD;
+                    }
+                    if var.storage_access.contains(crate::StorageAccess::STORE) {
+                        storage_usage |= crate::GlobalUse::STORE;
+                    }
+                    //TODO: make this more explicit somehow
+                    if storage_usage.is_empty() {
```
Napokue, line 290: I don't really understand this statement. What needs to happen when the user uses an Uniform or StorageBuffer storage class? Don't we just need to explicitly state that this is a syntax error?

Or is it possible to not have a storage usage specified beforehand?

Napokue (changes requested):
Thank you! I've got a couple of things.

src/proc/validator.rs
kvark: wow, this is why we need reviews :)
thank you for spotting!

src/proc/validator.rs
kvark: I'll remove the todo for now. Technically, the storage access flags being non-zero means that it's a storage, so IR is actually fine.

PR #205, opened 2020-09-18
[ir, wgsl] Texture support

### LINKED ISSUE #204: Hide image scalar kind for non-sampleable images
We have the image type variant define as follows:
```rust
    Image {
        kind: ScalarKind,
        dim: ImageDimension,
        arrayed: bool,
        class: ImageClass,
    },
```
However, the `kind` member is redundant for `ImageClass::Depth` (always `Float`) and `ImageClass:Storage(format)` (always computed as `ScalarKind::from(format)`).

Therefore, we may consider to make this less redundant:
```rust
    Image {
        dim: ImageDimension,
        arrayed: bool,
        class: ImageClass,
    },

pub enum ImageClass {
    Sampled { kind: ScalarKind, multi: bool },
    Depth,
    Storage(StorageFormat),
}
```

I think getting less redundancy in IR is always welcome!
The possible downside is that all the front-ends already know this `kind`, and all the backends need it. But that is covered by the typifier anyway, which the backends use.

kvark:
See https://github.com/gpuweb/gpuweb/pull/909 and multiple follow-ups.
also, Fixes #204

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ clean:
 	rm *.metal *.air *.metallib *.vert *.frag *.comp
 
 %.metal: test-data/%.wgsl $(wildcard src/*.rs src/**/*.rs examples/*.rs)
-	cargo run --example convert -- $< $@
+	cargo run --example convert --features wgsl-in,msl-out -- $< $@
 
 %.air: %.metal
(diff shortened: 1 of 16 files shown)


pjoe (approved):
looks pretty good to my lazy eyes 👍

PR #206, opened 2020-09-18
[glsl-in] consolidate tests

pjoe:
- remove simple lex test as was covered by others
- combine related tests to single function
- remove redundant prefix in test names

### DIFF
--- a/src/front/glsl/lex_tests.rs
+++ b/src/front/glsl/lex_tests.rs
@@ -1,60 +1,8 @@
 use super::{lex::Lexer, parser::Token::*, token::TokenMetadata};
 
 #[test]
-fn glsl_lex_simple() {
-    let source = "void main() {\n}";
-    let mut lex = Lexer::new(source);
-
-    assert_eq!(
-        lex.next().unwrap(),
-        Void(TokenMetadata {
-            line: 0,
-            chars: 0..4
-        })
-    );
-    assert_eq!(
-        lex.next().unwrap(),
-        Identifier((
-            TokenMetadata {
-                line: 0,
-                chars: 5..9
-            },
-            "main".into()
-        ))
-    );
-    assert_eq!(
-        lex.next().unwrap(),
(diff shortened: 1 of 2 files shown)


pjoe:
@kvark though that does make the test code a bit cleaner, I don't like the way failures are then reported at the line number in the `sub_test` function, instead of at a specific 'instance' where a comparison fails. What do you think?

pjoe:
Not really sure I understand, how this would change things?

kvark:
> I don't like the way failures are then reported at the line number in the sub_test function

If we put `[track_caller]` I was thinking that the assert will report the line number of the caller, which is the `sub_test`. Anyhow, this can be addressed by just returning a `Result` in `sub_test`, so we don't need the fancy stuff.

pjoe:
Ok, so the remaining question is more about asserts in loops vs. pointing to individual line.

Keeping them individual:
- pro: will point failures to specific instance
- con: will require a bit more code, for each token to verify

pjoe:
As en example here is how failure looks with individual assert:
image

kvark:
Seems fine to me. Are you concerned about this look?

pjoe:
No, this is the way it looks with individual asserts (like currently). In terms of how failures are reported, this is how I'd prefer.

I'll try adding similar screenshot tomorrow for the other solution with a helper function, looping over expected tokens, and returning a `Result`

kvark:
Just to clarify, reducing the number of `#[test]` functions does not mean having less asserts.

pjoe:
Right, as it stands the PR reduces number of lexer test functions  to 4, would you prefer it all the way down to 1?

The matter of number of asserts relates to using a common helper function to check against an array of expected tokens, instead of asserting them individually.

pjoe:
As agreed on matrix, consolidated lex tests to a single function

kvark (approved):

PR #208, opened 2020-09-18
[spv-out] Combined all instruction spec tests to one test

Napokue:
This PR combines all spec requirement verify unit tests for the spv back-end into one unit test. Reason for this is because it cluttered the test log with all the separate unit tests, and it can be compiled down to one unit test just fine.


kvark (approved):
Thank you!

PR #211, opened 2020-09-21
[glsl-in] Support anonymous global uniforms

### LINKED ISSUE #210: Unable to parse Bevy's ".vert" files
I faced with the following issue: `ParseError { kind: UnknownVariable(TokenMetadata { line: 19, chars: 16..21 }, "Model") }` when tried to parse `.vert` file.

I noticed this [commented test](https://github.com/gfx-rs/naga/blob/081b90f11765c8afaf9f26cbe7030781bb4155c9/tests/convert.rs#L194) and now curious - what happened? It seems like it worked before but now - it's not.

How to fix it?

pjoe:
Partly handles issue in  #210

Example shader:
```glsl
#version 450

layout(set = 0, binding = 0) uniform Matrices {
    mat4 modelViewProjection;
};

void main() {
    gl_Position = modelViewProjection * vec4(position, 0.0, 1.0);
}
```

### DIFF
--- a/src/front/glsl/variables.rs
+++ b/src/front/glsl/variables.rs
@@ -7,6 +7,41 @@ use super::ast::*;
 use super::error::ErrorKind;
 
 impl Program {
+    fn lookup_global_struct_member(&mut self, name: &str) -> Option<Handle<Expression>> {
+        let global_struct_member = self
+            .module
+            .global_variables
+            .iter()
+            // find globals without name
+            .filter(|(_, var)| var.name.is_none())
+            // find those of struct type and map to their members
+            .filter_map(|(h, var)| {
+                let ty = &self.module.types[var.ty];
+                if let TypeInner::Struct { members } = &ty.inner {
+                    Some((h, members))
+                } else {
+                    None
+                }
+            })
+            // find member matching name
+            .find_map(|(h, members)| {
+                members
+                    .iter()
+                    .position(|m| m.name == Some(name.into()))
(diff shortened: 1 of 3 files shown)


src/front/glsl/variables.rs
```diff
@@ -92,6 +127,8 @@ impl Program {
             Ok(Some(local_var))
         } else if let Some(global_var) = self.context.lookup_global_var_exps.get(name) {
             Ok(Some(*global_var))
+        } else if let Some(global_member) = self.lookup_global_struct_member(name) {
```
kvark, line 130: Would it be possible, instead of calling `lookup_global_struct_member` here, make sure these fields are present in the `lookup_global_var_exps` table. I.e. once we encounter an anonymous global struct, just make sure we add the members to that map, and avoid any wacky lookups later on.

kvark (commented):
Could you post an example code of what you are trying to support in this PR?

src/front/glsl/variables.rs
pjoe: Will take a look 👀
First step was to make it work :)

pjoe: Done. Simpler code - though a bit sad to see all my fancy FP code gone 😁

pjoe: Done. Simpler code - a bit sad to see all my fancy FP fun gone though 😁

kvark (approved):

PR #213, opened 2020-09-24
[glsl-in] Implement swizzle for r-values

pjoe:
Related to #210

### DIFF
--- a/src/front/glsl/variables.rs
+++ b/src/front/glsl/variables.rs
@@ -104,8 +104,7 @@ impl Program {
         name: &str,
         meta: TokenMetadata,
     ) -> Result<Handle<Expression>, ErrorKind> {
-        let type_inner = self.resolve_type(expression)?;
-        match type_inner {
+        match self.resolve_type(expression)? {
             TypeInner::Struct { members } => {
                 let index = members
                     .iter()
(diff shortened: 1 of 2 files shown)


src/front/glsl/variables.rs
```diff
@@ -96,4 +97,88 @@ impl Program {
...
     }
+
+    pub fn field_selection(
+        &mut self,
+        expression: Handle<Expression>,
+        name: &str,
+        meta: TokenMetadata,
+    ) -> Result<Handle<Expression>, ErrorKind> {
+        let type_inner = self.resolve_type(expression)?;
+        match type_inner {
```
kvark, line 108: let's match on the value always, i.e. `match *self.resolve_type(expression)? {..}`

pjoe: Done, except borrow checker wouldn't let me deref with `*` (something about unable to move out of shared ref), so had to settle for `match self.resolve_type(expression)? {..}`

kvark: The whole point of borrowing by value is that it lets you see which things *are* borrowed, and which not.
So, you see that the borrow checker complains about `members` field being moved out? Mark it as `ref members`.

pjoe: Ahhh, maybe one day I will learn to actually listen to what the borrow checker tells me :)

kvark (approved):
Thank you!

PR #214, opened 2020-09-28
hello-compute validation fixes

kvark:


Napokue (approved):
Thank you, LGTM!

kvark:
There is a bit of a hiccup on https://github.com/gpuweb/gpuweb/issues/1113 coming from the fact that things like `color.x = yyy` may involve reading internally. For now, I'm going to fix the example shaders to not do that.

Napokue:
Sure, we'll see what the group thinks about this.

PR #216, opened 2020-09-30
[spv-out] Fix write_function returning block id

Napokue:
The variable [`function_id`](https://github.com/gfx-rs/naga/blob/master/src/back/spv/writer.rs#L341) in `write_entry_point` was using a wrong id. Namely, the `write_function` method returned the block id instead of the function id. spirv-val tool told me it failed due to the entrypoint having a wrong id. 

At the same time, the `lookup_function` probably stopped working as block ids were inserted, instead of function ids.

This was introduced during the entry point refactor, due to the fact that it used to override variables, see [commit](https://github.com/gfx-rs/naga/commit/2ebaadaf0ca62e558ac027606f6f18c807308ee7#diff-4516391251ac170e6db90c0fa6f07022L1074).

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -311,23 +311,23 @@ impl Writer {
             parameter_type_ids,
         };
 
-        let id = self.generate_id();
+        let function_id = self.generate_id();
         let function_type =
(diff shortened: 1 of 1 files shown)


kvark (approved):
nice!

PR #217, opened 2020-10-01
[glsl-in] Handle spacing for preproces directives

pjoe:
Handle spacing between `#` and directive.
Also require directive to be first token on a line, to be recognized.

Examples:
```glsl
  #  version 450 // now valid
...
void main() {} #version 450 // now invalid
```

This is in preparation for supporting other pre-processor directives

### DIFF
--- a/src/front/glsl/lex.rs
+++ b/src/front/glsl/lex.rs
@@ -16,7 +16,7 @@ fn consume_any(input: &str, what: impl Fn(char) -> bool) -> (&str, &str, usize) 
     (o, i, pos)
 }
 
-pub fn consume_token(mut input: &str) -> (Option<Token>, &str) {
+pub fn consume_token(mut input: &str, start_of_line: bool) -> (Option<Token>, &str) {
     let start = input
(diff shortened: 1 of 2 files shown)


src/front/glsl/lex.rs
```diff
@@ -302,7 +313,7 @@ impl<'a> Lexer<'a> {
 
     #[must_use]
     pub fn next(&mut self) -> Option<Token> {
-        let (token, rest) = consume_token(&self.input);
+        let (token, rest) = consume_token(&self.input, self.offset == 0);
```
kvark, line 316: is `self.offset` the offset from the line, or the offset from the whole file?

pjoe: It's char offset from start of line, get's reset further down (line 356). Would `char_offset` be a better name?

It's used for calculating position of the token (in TokenMetadata).

kvark (approved):

src/front/glsl/lex.rs
kvark: I think it's typically called a "column" fwiw

PR #218, opened 2020-10-01
[spv-out] Add matrix construct support

Napokue:
Add support for a matrix construct. 

This WGSL input:
```
[[stage(fragment)]]
fn main() -> void {
  var matrix: mat3x3<f32> = mat3x3<f32>(1.0, 2.0, 3.0);
  return;
}
```

Will result in this SPIR-V output:
```
%main = OpFunction %void None %13
 %14 = OpLabel
  %9 = OpVariable %_ptr_Function_mat3v3float Function
 %15 = OpCompositeConstruct %v3float %float_0_1 %float_0_2 %float_0_3
 %16 = OpCompositeConstruct %v3float %float_0_1 %float_0_2 %float_0_3
 %17 = OpCompositeConstruct %v3float %float_0_1 %float_0_2 %float_0_3
 %18 = OpCompositeConstruct %mat3v3float %15 %16 %17
       OpStore %9 %18 None
       OpReturn
       OpFunctionEnd
```

There isn't (yet) support for diagonal matrix initialization like in GLSL as far as I know, so I didn't take that into account with the changes.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -776,6 +776,23 @@ impl Writer {
         }
     }
 
+    fn write_composite_construct(
+        &mut self,
+        base_type_id: Word,
+        constituent_ids: &[Word],
+        block: &mut Block,
+    ) -> Word {
+        let id = self.generate_id();
+        block
+            .body
+            .push(super::instructions::instruction_composite_construct(
+                base_type_id,
+                id,
+                constituent_ids,
+            ));
+        id
+    }
+
     fn write_expression<'a>(
         &mut self,
         ir_module: &'a crate::Module,
@@ -800,8 +817,7 @@ impl Writer {
                 Some((id, Some(var.ty)))
             }
             crate::Expression::Compose { ty, components } => {
-                let id = self.generate_id();
-                let type_id = self.get_type_id(&ir_module.types, LookupType::Handle(*ty));
+                let base_type_id = self.get_type_id(&ir_module.types, LookupType::Handle(*ty));
 
                 let mut constituent_ids = Vec::with_capacity(components.len());
                 for component in components {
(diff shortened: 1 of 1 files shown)


src/back/spv/writer.rs
```diff
@@ -811,13 +827,49 @@ impl Writer {
                         .unwrap();
                     constituent_ids.push(component_id);
                 }
+                let constituent_ids_slice = constituent_ids.as_slice();
+
+                let id = match &ir_module.types[*ty].inner {
```
kvark, line 832: let's not match on a reference please!
Instead, add `ref` near the non-movable fields. It's clear to the reader what's going on therefore

kvark (changes requested):
Nice, just one small change is needed!

src/back/spv/writer.rs
Napokue: Always forgetting that, I think you already told me 10 times! Really need to think about that, change incoming in 1 minute.

src/back/spv/writer.rs
```diff
@@ -811,13 +827,49 @@ impl Writer {
...
+                                vector_type_id,
+                                constituent_ids_slice,
+                                block,
+                            );
+                            vector_ids.push(vector_id);
+                        }
+
+                        self.write_composite_construct(base_type_id, vector_ids.as_slice(), block)
+                    }
+                    _ => unreachable!(),
```
kvark, line 870: Cases like this should actually `panic` instead of being `unreachable`. The condition to fail here is basically working on an IR that wasn't validated.
Unreachable should be used where some of the other code *has* to ensure that this is good, so it's truly no ever expected to come there.

kvark (approved):

PR #219, opened 2020-10-01
[spv-out] Add multiply instructions

Napokue:
Add support for the long-awaited multiply instructions:
- OpIMul
- OpFMul
- OpVectorTimesScalar
- OpVectorTimesMatrix
- OpMatrixTimesVector
- OpMatrixTimesScalar
- OpMatrixTimesMatrix

Tested with the following WGSL:
```
[[stage(fragment)]]
fn main() -> void {
    # OpIMul
    var int_1: i32 = 1;
    var int_2: i32 = int_1 * int_1;

    # OpFMul
    var float_1: f32 = 1.0;
    var float_2: f32 = float_1 * float_1;

    # OpVectorTimesScalar
    var vec_1: vec2<f32> = vec2<f32>(1.0, 1.0);
    var vec_2: vec2<f32> = vec_1 * float_1;

    # OpVectorTimesMatrix
    var mat_1: mat2x2<f32> = mat2x2<f32>(1.0, 1.0);
    var vec_3: vec2<f32> = vec_2 * mat_1;

    # OpMatrixTimesVector
    var vec_4: vec2<f32> = mat_1 * vec_3;

    # OpMatrixTimesScalar
    var mat_2: mat2x2<f32> = mat_1 * float_2;

    # OpMatrixTimesMatrix
    var mat_3: mat2x2<f32> = mat_2 * mat_2;

    return;
}
```

### DIFF
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -520,6 +520,34 @@ pub(super) fn instruction_composite_construct(
 // Arithmetic Instructions
 //
 
+pub(super) fn instruction_i_mul(
+    result_type_id: Word,
+    id: Word,
+    operand_1: Word,
+    operand_2: Word,
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::IMul);
+    instruction.set_type(result_type_id);
+    instruction.set_result(id);
+    instruction.add_operand(operand_1);
+    instruction.add_operand(operand_2);
+    instruction
+}
+
+pub(super) fn instruction_f_mul(
+    result_type_id: Word,
+    id: Word,
+    operand_1: Word,
+    operand_2: Word,
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::FMul);
+    instruction.set_type(result_type_id);
+    instruction.set_result(id);
+    instruction.add_operand(operand_1);
+    instruction.add_operand(operand_2);
+    instruction
+}
+
 pub(super) fn instruction_vector_times_scalar(
     result_type_id: Word,
     id: Word,
@@ -531,6 +559,62 @@ pub(super) fn instruction_vector_times_scalar(
     instruction.set_result(id);
     instruction.add_operand(vector_type_id);
     instruction.add_operand(scalar_type_id);
+    instruction
+}
+
+pub(super) fn instruction_matrix_times_scalar(
+    result_type_id: Word,
+    id: Word,
+    matrix_id: Word,
+    scalar_id: Word,
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::MatrixTimesScalar);
+    instruction.set_type(result_type_id);
+    instruction.set_result(id);
+    instruction.add_operand(matrix_id);
+    instruction.add_operand(scalar_id);
+    instruction
+}
+
+pub(super) fn instruction_vector_times_matrix(
+    result_type_id: Word,
+    id: Word,
+    vector_id: Word,
+    matrix_id: Word,
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::VectorTimesMatrix);
+    instruction.set_type(result_type_id);
+    instruction.set_result(id);
+    instruction.add_operand(vector_id);
+    instruction.add_operand(matrix_id);
+    instruction
+}
+
+pub(super) fn instruction_matrix_times_vector(
+    result_type_id: Word,
+    id: Word,
+    matrix_id: Word,
+    vector_id: Word,
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::MatrixTimesVector);
+    instruction.set_type(result_type_id);
+    instruction.set_result(id);
+    instruction.add_operand(matrix_id);
+    instruction.add_operand(vector_id);
+    instruction
+}
+
+pub(super) fn instruction_matrix_times_matrix(
+    result_type_id: Word,
+    id: Word,
+    left_matrix: Word,
+    right_matrix: Word,
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::MatrixTimesMatrix);
+    instruction.set_type(result_type_id);
+    instruction.set_result(id);
+    instruction.add_operand(left_matrix);
+    instruction.add_operand(right_matrix);
     instruction
 }
 
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -875,7 +875,6 @@ impl Writer {
             crate::Expression::Binary { op, left, right } => {
                 match op {
                     crate::BinaryOperator::Multiply => {
-                        // TODO OpVectorTimesScalar is only supported
                         let id = self.generate_id();
                         let left_expression = &ir_function.expressions[*left];
                         let right_expression = &ir_function.expressions[*right];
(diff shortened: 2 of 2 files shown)


src/back/spv/instructions.rs
```diff
@@ -478,6 +478,34 @@ pub(super) fn instruction_composite_construct(
 // Arithmetic Instructions
 //
 
+pub(super) fn instruction_i_mul(
```
kvark, line 481: would you consider generalizing it as `instruction_binary_op()` that receives and `Op` and can be used for all of the binary operators as well as multiplications, to save code?
If you see the clear names like `instruction_matrix_times_scalar` bringing value, you can still have them as thin layers on top of `instruction_binary_op`

src/back/spv/instructions.rs
```diff
@@ -1161,6 +1251,40 @@ mod tests {
...
+                result_id: true,
+                operands: true,
+            };
+            validate_spec_requirements(requirements, &instruction);
+
+            instruction.to_words(&mut output);
+            validate_instruction(output.as_slice(), &instruction);
+        }
+
+        fn test_instruction_f_mul(&self) {
```
kvark, line 1271: similarly, having a generalized `instruction_binary_op` would allow to test all of them in a simpler way.
in general, not totally sure that all these tests are valuable, but this is up to you :)

src/back/spv/writer.rs
```diff
@@ -901,49 +901,111 @@ impl Writer {
...
-                            load_id,
-                            vector_id,
+                        let right_result_type_id = self
+                            .get_type_id(&ir_module.types, LookupType::Handle(right_ty.unwrap()));
+
+                        // Load left- and right-hand side via OpLoad
+                        let left_load_id = self.generate_id();
+                        let right_load_id = self.generate_id();
+
+                        block.body.push(super::instructions::instruction_load(
```
kvark, line 914: why are there `load` instructions here?

Napokue: That is, because the values are from an `OpVariable` for instance, and to use an `OpVariable`, you have to load them through `OpLoad`.

src/back/spv/instructions.rs
Napokue: Yes I understand what you are suggesting, but I rather would not have the developer of the SPIR-V back-end use Op codes directly, that is the reason for this instructions module.

What we can do is still creating the `instruction_binary_op`, but make it private, and exposing the instructions like `instruction_matrix_times_scalar`, but in their body, they will call `instruction_binary_op` with the op as argument.

Napokue: Update: I see not all binary instructions follow the layout, and will diverge. Can I suggest to keep it like this, and when we have all binary instructions implemented, we can clearly see where "duplicate" is?

kvark: sure thing

kvark (approved):

src/back/spv/writer.rs
kvark: hmm, but there isn't any check here about whether the target is OpVariable or not. So how do we know that unconditionally "load"-ing is the right thing to do?

kvark (changes requested):
Just one concern

src/back/spv/writer.rs
Napokue: That is true, but as far as I know there are only two ways to use binary instructions at high level: via pointers, and via constants.

I have done some research at http://shader-playground.timjones.io/, and when doing it via pointers, there is always an `OpLoad` needed.

GLSL:
```glsl
void main()
{
    vec2 vec_1 = vec2(1.0, 1.0);
    float float_1 = vec_1.x * vec_1.y;
    float float_3 = float_1 * float_1;
}
```
This will always translate back to `OpLoad`, whether it is via an `OpAccessChain` or `OpVariable`. There probably are more ways of accessing value via pointers, but all pointers always need to be loaded in via OpLoad to be useful (as far as I know).

What is good to see, is that when I am using constants, instead of variables at the shader playground, it pre-calculates the value, and uses it directly (which is allowed).

GLSL:
```glsl
void main()
{
    float float_1 = 1.0 * 2.0;
}
```

However, because of the pre-calculation, the resulting SPIR-V output does not even contain a binary instruction, as it is already calculated.

I think it is safe to say for now to always use `OpLoad` here, but maybe you think about a case where this might not be, please let me know, so we can reconsider this piece of code.

kvark: We can't *always* use `OpLoad` there. It should only happen if the target expression is a variable. Once it's a different expression, there should be no load. For example, if you do `float x = 2 * (y + 1);`

src/back/spv/instructions.rs
Napokue: "Fixed" in #223

src/back/spv/writer.rs
Napokue: Ah yes, I understand what you mean. I have a rough idea, and will try to implement it tomorrow: what if we never check for the specific `Expression::LocalVariable`, but let the match arm of `Expression::LocalVariable` always create and return an `OpLoad`? 

That should fix this issue, and the rest of the code will be cleaner, as [`Expression::Return`](https://github.com/gfx-rs/naga/blob/master/src/back/spv/writer.rs#L1147), and [`Expression::Store`](https://github.com/gfx-rs/naga/blob/master/src/back/spv/writer.rs#L1191) have a specific match for if the id of the expression is of the expression type `Expression::LocalVariable`. This is pretty dirty to be honest, but I didn't see that back then.

kvark: yes, both `LocalVariable` and `GlobalVariable` variants need to generate `OpLoad`.

Napokue: Fixed and now it supports even "unlimited" expression nesting:
```
var float_1: f32 = 2.0 * (2.0 * (4.0 * 2.0));
```

Giving the predicted result without an `OpLoad`, like you said (and correctly validates):
```
%5 = OpVariable %_ptr_Function_float Function
%13 = OpFMul %float %float_4 %float_2
%12 = OpFMul %float %float_2 %13
%11 = OpFMul %float %float_2 %12
      OpStore %5 %11 None
```

src/back/spv/writer.rs
```diff
@@ -898,53 +897,129 @@ impl Writer {
...
+
+                        let left_result_type_id =
+                            self.get_type_id(&ir_module.types, LookupType::Handle(left_ty));
+
+                        let right_result_type_id =
+                            self.get_type_id(&ir_module.types, LookupType::Handle(right_ty));
+
+                        let left_id = match *left_expression {
+                            crate::Expression::LocalVariable(_)
+                            | crate::Expression::GlobalVariable(_) => {
```
Napokue, line 914: Little bit dirty this way, but these matches were already there in other expressions, but it will suffice for now.

kvark: This is something we'll need to do for *any* variable used in any expression. So a helper method is needed, and we need to call this from more places. Do you want to leave this for a follow-up?

kvark (approved):

src/back/spv/writer.rs
Napokue: Yes please, I will take a note and store it as task in my todo.

PR #220, opened 2020-10-03
[spv-out] Add scalar conversion support

Napokue:
Add support for scalar conversions via `Expression::As`.

Following conversions are implemented:
- `OpConvertFToU` (float to unsigned int)
- `OpConvertFToS` (float to signed int)
- `OpConvertSToF` (signed int to float)
- `OpConvertUToF` (unsigned int to float)

Tested with the following WGSL:
```
[[stage(fragment)]]
fn main() -> void {
  var uint_1: u32 = u32(1.0);
  var int_1: i32 = i32(1.0);

  var float_1: f32 = f32(1);
  var temp_uint_2: u32 = u32(1);
  var float_2: f32 = f32(temp_uint_2);
  # var float_2: f32 = f32(u32(1));
  return;
}
```

I've commented out two particular cases where double casting didn't result in a double `Expression::As` expression, so I had to use an temp variable to hold the first conversion in, see #222.

### DIFF
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -453,7 +453,49 @@ pub(super) fn instruction_function_call(
 //
 // Conversion Instructions
 //
+fn instruction_unary(op: Op, result_type_id: Word, id: Word, value: Word) -> Instruction {
+    let mut instruction = Instruction::new(op);
+    instruction.set_type(result_type_id);
+    instruction.set_result(id);
+    instruction.add_operand(value);
+    instruction
+}
 
+pub(super) fn instruction_convert_f_to_u(
+    result_type_id: Word,
+    id: Word,
+    float_value: Word,
+) -> Instruction {
+    instruction_unary(Op::ConvertFToU, result_type_id, id, float_value)
+}
+
+pub(super) fn instruction_convert_f_to_s(
+    result_type_id: Word,
+    id: Word,
+    float_value: Word,
+) -> Instruction {
+    instruction_unary(Op::ConvertFToS, result_type_id, id, float_value)
+}
+
+pub(super) fn instruction_convert_s_to_f(
+    result_type_id: Word,
+    id: Word,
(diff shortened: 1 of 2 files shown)


Napokue:
After having a conversation with @kvark at Matrix, we came to the conclusion to remove `OpSatConvertSToU` and `OpSatConvertUToS`, as `Kernel` capability is not something we want to support in Naga.

This effectively means that for now, the SPIR-V back-end is unable to convert signed integers to unsigned integers and vice versa.

src/back/spv/instructions.rs
```diff
@@ -454,6 +454,54 @@ pub(super) fn instruction_function_call(
 // Conversion Instructions
 //
 
+pub(super) fn instruction_convert_f_to_u(
```
kvark, line 457: similarly to feedback of the other PR, maybe it would help to have `instruction_unary_op` to contain all the code that's repeated between these methods?

src/back/spv/writer.rs
```diff
@@ -1033,6 +1033,76 @@ impl Writer {
                 }
                 _ => unimplemented!("{:?}", origin),
             },
+            crate::Expression::As {
+                expr,
+                kind,
+                convert,
+            } => {
+                // TODO Convert = false, a bitcast, is not yet implemented in the front-ends
```
kvark, line 1041: it would actually be relatively simple to add, there is a single `OpBitcast`

src/back/spv/instructions.rs
Napokue: Yes I looked at the `instruction_unary_op`, and that is possible, so I added it!

src/back/spv/writer.rs
Napokue: Added it, but I didn't test it, as we don't pass `convert = true` from the front-ends yet.

kvark (approved):

PR #223, opened 2020-10-03
[spv-out] Remove instruction spec tests

Napokue:
Little did they contribute to for testing. Discussed with @kvark that due to its LOC increase versus the benefit it gives, and the benefit it gives in general, that it is not worth to have these tests.

Conclusion: when instructions are written, they are pretty much set in stone, and won't ever be touched again.


kvark (approved):
Thank you!

PR #224, opened 2020-10-03
[spv-out] Add access and access index support

Napokue:
# Description
Add support for `Expression::Access` and `Expression::AccessIndex`.

`Expression::Access`:
- [x] Vector
- ~~Matrix~~ - Will be done at a later moment.
- ~~Array~~ - Will be done at a later moment.

`Expression::AccessIndex`:
- [x] Vector
- ~~Matrix~~ - Will be done at a later moment.
- ~~Array~~ - Will be done at a later moment.
- [x] Struct

# WGSL test file
```
type Data = struct{
    [[offset(0)]] e: f32;
};

[[group(0), binding(0)]] var<uniform> u_data: Data;

[[stage(vertex)]]
fn main() -> void {
  var a: vec2<f32> = vec2<f32>(1.0, 1.0);
  
  # Expression::Access - Vector
  var b: f32 = a[0];
  
  # Expression::AccessIndex - Vector
  var c: f32 = a.x;
  
  # Expression::AccessIndex - Struct
  #var d: f32 = u_data.k;
  
  return;
}
```

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1,7 +1,7 @@
 /*! Standard Portable Intermediate Representation (SPIR-V) backend !*/
 use super::{Instruction, LogicalLayout, PhysicalLayout, WriterFlags};
 use spirv::Word;
-use std::collections::hash_map::Entry;
+use std::{collections::hash_map::Entry, ops};
 use thiserror::Error;
 
 const BITS_PER_BYTE: crate::Bytes = 8;
@@ -16,6 +16,8 @@ pub enum Error {
     NotImage,
     #[error("empty value")]
     EmptyValue,
+    #[error("feature is not yet implemented")]
+    FeatureNotImplemented(),
 }
 
 struct Block {
@@ -90,7 +92,7 @@ enum LocalType {
     },
     Pointer {
         base: crate::Handle<crate::Type>,
-        class: spirv::StorageClass,
+        class: crate::StorageClass,
     },
     SampledImage {
         image_type: crate::Handle<crate::Type>,
@@ -118,6 +120,21 @@ struct LookupFunctionType {
     return_type_id: Word,
 }
 
+enum MaybeOwned<'a, T> {
+    Owned(T),
+    Borrowed(&'a T),
+}
+
+impl<'a, T> ops::Deref for MaybeOwned<'a, T> {
+    type Target = T;
+    fn deref(&self) -> &T {
+        match *self {
+            MaybeOwned::Owned(ref value) => value,
+            MaybeOwned::Borrowed(reference) => reference,
+        }
+    }
+}
+
 pub struct Writer {
     physical_layout: PhysicalLayout,
     logical_layout: LogicalLayout,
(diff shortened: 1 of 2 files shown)


src/back/spv/writer.rs
```diff
@@ -807,6 +823,91 @@ impl Writer {
         id
     }
 
+    fn get_type_inner(
```
Napokue, line 826: I am totally not satisfied with this, but I couldn't think of any other way to make a function that will get a type inner based on `LookupType`, without modifying the IR. Any comments to refine this is welcome.

src/back/spv/writer.rs
```diff
@@ -240,24 +242,35 @@ impl Writer {
...
+        instruction.to_words(&mut self.logical_layout.declarations);
+        id
+    }
+
+    fn create_constant(&mut self, type_id: Word, value: &[Word]) -> Word {
+        let id = self.generate_id();
+        let instruction = super::instructions::instruction_constant(type_id, id, value);
+        instruction.to_words(&mut self.logical_layout.declarations);
+        id
+    }
```
Napokue, line 272: Both cannot easily be stored in a lookup now without refactoring some more pieces. For now whenever the same pointer or same constant is found, it just "duplicates" it, resulting in duplicate values in SPIR-V.

I preferably don't want to block on this, but I can create an issue for it.

src/back/spv/writer.rs
kvark: Why didn't you go with something like `crate::proc::MaybeOwned`? This function signature would be:
```rust
fn get_type_inner<'a>(
  arena: &'a crate::Arena<crate::Type>,
  lookup_ty: LookupType,
) -> MaybeOwned<'a, crate::TypeInner> {..}
```

kvark (approved):

src/back/spv/writer.rs
```diff
@@ -809,6 +840,43 @@ impl Writer {
...
+            LookupType::Local(local_ty) => match local_ty {
+                LocalType::Scalar { kind, width } => {
+                    MaybeOwned::Owned(crate::TypeInner::Scalar { kind, width })
+                }
+                LocalType::Vector { size, kind, width } => {
+                    MaybeOwned::Owned(crate::TypeInner::Vector { size, kind, width })
+                }
+                LocalType::Pointer { base, class } => {
+                    let storage_class = match class {
+                        spirv::StorageClass::UniformConstant => crate::StorageClass::Handle,
```
kvark, line 859: It seems backwards that we map from SPIR-V into IR here. Can we make it so `LocalType::Pointer` stores IR's StorageClass instead? That would make the conversion only one-way, and we could have a single helper conversion function.

Napokue: Yes agreed! We already have a mapper from `crate::StorageClass` to `spirv::StorageClass`, so that will be easier, and more maintainable. 

Only the instructions would require the `spirv::StorageClass`, rest can be the `crate::StorageClass`.

src/back/spv/writer.rs
Napokue: Thank you very much for your commit, it helped me a lot!

kvark (approved):
Thank you!

PR #225, opened 2020-10-04
[spv-out] Add check for global variables if name is set

Napokue:
Add check if name for global variables is set. When the debug flag is set for spv-out, `OpName` would be added for the global variable. It can happen, for instance when having a uniform variable, that the name is not set. 

As the code always called `unwrap` when the debug flag was enabled, the code crashed.

This bug was found in issue: #210.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -670,10 +670,10 @@ impl Writer {
         let instruction = super::instructions::instruction_variable(pointer_id, id, class, None);
 
         if self.writer_flags.contains(WriterFlags::DEBUG) {
-            self.debugs.push(super::instructions::instruction_name(
-                id,
-                global_variable.name.as_ref().unwrap().as_str(),
-            ));
+            if let Some(ref name) = global_variable.name {
+                self.debugs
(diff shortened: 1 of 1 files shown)


src/back/spv/writer.rs
```diff
@@ -669,7 +669,7 @@ impl Writer {
         let pointer_id = self.get_pointer_id(arena, global_variable.ty, class);
         let instruction = super::instructions::instruction_variable(pointer_id, id, class, None);
 
-        if self.writer_flags.contains(WriterFlags::DEBUG) {
+        if self.writer_flags.contains(WriterFlags::DEBUG) && global_variable.name.is_some() {
```
kvark, line 672: It would be better to avoid checking for this twice (first `is_some()`, then in `unwrap()`.
instead, we should be doing something like:
```rust
if self.writer_flags.contains(WriterFlags::DEBUG) {
  if let Some(ref name) = global_variable.name {
    ..
  }
}
```

Napokue: Sure, that sounds good! Update incoming.

kvark (approved):
Thank you!

PR #226, opened 2020-10-05
Added support for matrix access on the typifier

JCapucho:
I talked about with kvark about this in the matrix room, we didn't reach any conclusion but i think it's supposed to be supported, here are some links
- https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpAccessChain
- https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#CompositeType


kvark (approved):
Thank you!

PR #227, opened 2020-10-06
[glsl-in] make consume_token a lexer method

pjoe:
This gives `consume_token`access to lexer state. This simplifies the code a bit, and is also in preparation for dealing with lexer states for e.g. pre-processor.

### DIFF
--- a/src/front/glsl/lex.rs
+++ b/src/front/glsl/lex.rs
@@ -16,268 +16,6 @@ fn consume_any(input: &str, what: impl Fn(char) -> bool) -> (&str, &str, usize) 
     (o, i, pos)
 }
 
-pub fn consume_token(mut input: &str, start_of_line: bool) -> (Option<Token>, &str) {
-    let start = input
-        .find(|c: char| !c.is_whitespace())
-        .unwrap_or_else(|| input.chars().count());
-    input = &input[start..];
-
-    let mut chars = input.chars();
-    let cur = match chars.next() {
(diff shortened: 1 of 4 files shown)


src/front/glsl/lex.rs
```diff
@@ -16,278 +16,272 @@ fn consume_any(input: &str, what: impl Fn(char) -> bool) -> (&str, &str, usize)
...
+                    "continue" => Some(Token::Continue(meta)),
+                    "break" => Some(Token::Break(meta)),
+                    "return" => Some(Token::Return(meta)),
+                    "discard" => Some(Token::Discard(meta)),
+                    // types
+                    "void" => Some(Token::Void(meta)),
+                    word => {
+                        let token = match parse_type(word) {
+                            Some(t) => Token::TypeName((meta, t)),
+                            None => Token::Identifier((meta, String::from(word))),
```
kvark, line 147: This is concerning for performance. We don't want to be heap-allocating `String::from` on every token, that's really bad.
Notably, it's been there before the PR, so technically unrelated.
The token itself should just borrow strings. See how WGSL frontend does it.

Do you want to try making this happen in this PR? Or follow up?

kvark (approved):

src/front/glsl/lex.rs
pjoe: i think this relates a bit to how pomelo handles tokens, so better in separate PR

PR #230, opened 2020-10-06
[ir] Remove Matrix Kind from the IR

### LINKED ISSUE #229: Remove `kind` from the matrix type variants
Apparently, SPIR-V only works with floating-point matrices!

Napokue:
Should fix #229. I haven't included any additional validation, as I don't know how to do that (yet). I don't know if that is necessary for this PR or if that can be added later.

### DIFF
--- a/src/back/glsl.rs
+++ b/src/back/glsl.rs
@@ -940,11 +940,10 @@ fn write_expression<'a, 'b>(
                 TypeInner::Matrix {
                     columns,
                     rows,
-                    kind,
                     width,
                 } => format!(
                     "{}mat{}x{}",
-                    map_scalar(kind, width, builder.manager)?.prefix,
+                    map_scalar(crate::ScalarKind::Float, width, builder.manager)?.prefix,
                     columns as u8,
                     rows as u8,
                 ),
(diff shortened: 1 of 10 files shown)


src/back/glsl.rs
```diff
@@ -1405,20 +1404,15 @@ fn write_type<'a>(
         TypeInner::Matrix {
             columns,
             rows,
-            kind,
             width,
         } => {
-            if kind != ScalarKind::Float {
```
kvark, line 1411: oh, nice!

src/proc/validator.rs
```diff
@@ -228,14 +228,18 @@ impl Validator {
...
-                            | crate::TypeInner::Matrix { kind, .. } => {
+                            | crate::TypeInner::Vector { kind, .. } => {
                                 if kind != crate::ScalarKind::Float
                                     && var.interpolation != Some(crate::Interpolation::Flat)
                                 {
                                     return Err(EntryPointError::InvalidIntegerInterpolation);
                                 }
                             }
+                            crate::TypeInner::Matrix { .. } => {
+                                if var.interpolation != Some(crate::Interpolation::Flat) {
```
kvark, line 239: this condition should be removed from here, the code above checked "kind != crate::ScalarKind::Float &&  this", so in this branch it's guaranteed false

kvark (changes requested):
Thank you!

src/proc/validator.rs
Napokue: I'm sorry, but I have no idea what interpolation means. So I can remove this entire match arm? As this is the only check.

kvark: This would essentially be `crate::TypeInner::Matrix { .. } => {}`

kvark (approved):

PR #237, opened 2020-10-12
[clippy] Fix clippy issues after upgrading to 1.47

pjoe:
- Disable match_like_matches_macro for now

Still need to deal with `panic`s found in:
- [x] spv-out
  - [/src/back/spv/writer.rs](../tree/master/src/back/spv/writer.rs#L1035)
- [x] spv-in
  - [/src/front/spv/mod.rs](../tree/master/src/front/spv/mod.rs#L1480)
- [x] wgsl-in
  - [/src/front/wgsl/mod.rs](../tree/master/src/front/wgsl/mod.rs#L1790)
- [x] typifier
  - numerous

### DIFF
--- a/src/proc/typifier.rs
+++ b/src/proc/typifier.rs
@@ -53,9 +53,9 @@ pub enum ResolveError {
     TypeNotFound,
     #[error("Incompatible operand: {op} {operand}")]
     IncompatibleOperand { op: String, operand: String },
-    #[error("Incompatible operands: {left} {op:?} {right}")]
+    #[error("Incompatible operands: {left} {op} {right}")]
     IncompatibleOperands {
-        op: crate::Expression,
+        op: String,
         left: String,
         right: String,
     },
@@ -213,7 +213,7 @@ impl Typifier {
                             }),
                             _ => {
                                 return Err(ResolveError::IncompatibleOperands {
-                                    op: expr.clone(),
+                                    op: "x".to_string(),
                                     left: format!("{:?}", ty_left),
                                     right: format!("{:?}", ty_right),
                                 })
--- a/src/proc/typifier.rs
+++ b/src/proc/typifier.rs
@@ -29,6 +29,7 @@ impl Clone for Resolution {
                     columns,
                     width,
                 },
+                #[allow(clippy::panic)]
                 _ => panic!("Unepxected clone type: {:?}", v),
             }),
         }
@@ -50,6 +51,8 @@ pub enum ResolveError {
     FunctionReturnsVoid,
     #[error("Type is not found in the given immutable arena")]
     TypeNotFound,
+    #[error("{msg}")]
+    Other { msg: String },
 }
 
 pub struct ResolveContext<'a> {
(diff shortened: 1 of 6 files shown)


src/proc/typifier.rs
```diff
@@ -50,6 +51,8 @@ pub enum ResolveError {
     FunctionReturnsVoid,
     #[error("Type is not found in the given immutable arena")]
     TypeNotFound,
+    #[error("{msg}")]
+    Other { msg: String },
```
kvark, line 55: can we get away without the `Other` loophole? I.e. we could have `IncompatibleOperands` error covering a few cases.

kvark (commented):
Thank you for fixing this!
just one concern about the typifier

src/proc/typifier.rs
pjoe: Ok, I've made an attempt, maybe you can change exact wording/params if needed :)

src/proc/typifier.rs
```diff
@@ -51,8 +51,14 @@ pub enum ResolveError {
...
     #[error("Type is not found in the given immutable arena")]
     TypeNotFound,
-    #[error("{msg}")]
-    Other { msg: String },
+    #[error("Incompatible operand: {op} {operand}")]
+    IncompatibleOperand { op: String, operand: String },
+    #[error("Incompatible operands: {left} {op} {right}")]
+    IncompatibleOperands {
+        op: String,
+        left: String,
```
kvark, line 59: can the operands be `Expression` instead of `String`?

src/proc/typifier.rs
```diff
@@ -50,6 +51,14 @@ pub enum ResolveError {
     FunctionReturnsVoid,
     #[error("Type is not found in the given immutable arena")]
     TypeNotFound,
+    #[error("Incompatible operand: {op} {operand}")]
+    IncompatibleOperand { op: String, operand: String },
+    #[error("Incompatible operands: {left} {op:?} {right}")]
+    IncompatibleOperands {
+        op: crate::Expression,
```
kvark, line 58: I actually meant the left/right/operand to be `Expression`, not the `op`. I guess `op` works too

pjoe: Yeah sorry, my bad ... already reverted

pjoe: Hmm ... at this point in the code it only seems to have `Handle<Expression>` and no way to look it up in the `Arena<Expression>` - or am I missing something?

pjoe: @kvark if you have the time, maybe easier for you to do this edit - than me making a mess 😄

kvark: ok, will do. Sorry about all the change requests :sweat_smile:

pjoe: np, just not too familier with this code, so could take quite some iteration before I got it right 😄

kvark: I took a few attempts and wasn't able to find a nice way to expose this. Let's proceed!

PR #238, opened 2020-10-14
[glsl-in] Add initial pre-processor

pjoe:
Supported:
- `#define KEY VALUE`
- `#undef`
- `#ifdef`
- `#ifndef`
- `#else`
- `#endif`

Not supported (yet):
- `#define MACRO(x, y, z)`
- `#if` (need to handle const expressions)
- `#include`

Pre-configured defines can now also be passed in to the `parse_str` function as `Vec<String, String>` (key, value).

The preprocessor works on a line by line basis, and should be easy to apply to other front-ends if needed.

As the implementation is currently of reasonable size (<150 lines), this has not (yet) been feature gated.

### DIFF
--- a/src/front/glsl/mod.rs
+++ b/src/front/glsl/mod.rs
@@ -1,4 +1,4 @@
-use crate::{Module, ShaderStage};
+use crate::{FastHashMap, Module, ShaderStage};
 
 mod lex;
 #[cfg(test)]
@@ -25,13 +25,13 @@ pub fn parse_str(
     source: &str,
     entry: &str,
     stage: ShaderStage,
-    defines: Vec<(String, String)>,
+    defines: Option<FastHashMap<String, String>>,
 ) -> Result<Module, ParseError> {
     let mut program = Program::new(stage, entry);
 
     let mut lex = Lexer::new(source);
-    for (k, v) in defines {
-        lex.pp.defines.insert(k, v);
+    if let Some(defines) = defines {
+        lex.pp.defines = defines;
     }
 
     let mut parser = parser::Parser::new(&mut program);
--- a/src/front/glsl/mod.rs
+++ b/src/front/glsl/mod.rs
@@ -25,14 +25,12 @@ pub fn parse_str(
     source: &str,
     entry: &str,
     stage: ShaderStage,
-    defines: Option<FastHashMap<String, String>>,
+    defines: FastHashMap<String, String>,
 ) -> Result<Module, ParseError> {
     let mut program = Program::new(stage, entry);
 
     let mut lex = Lexer::new(source);
-    if let Some(defines) = defines {
-        lex.pp.defines = defines;
-    }
+    lex.pp.defines = defines;
 
     let mut parser = parser::Parser::new(&mut program);
 
--- a/src/front/glsl/mod.rs
+++ b/src/front/glsl/mod.rs
@@ -4,6 +4,10 @@ mod lex;
 #[cfg(test)]
 mod lex_tests;
 
+mod preprocess;
+#[cfg(test)]
+mod preprocess_tests;
+
 mod ast;
 use ast::Program;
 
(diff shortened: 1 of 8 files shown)


src/front/glsl/mod.rs
```diff
@@ -17,11 +21,19 @@ mod token;
...
 mod variables;
 
-pub fn parse_str(source: &str, entry: &str, stage: ShaderStage) -> Result<Module, ParseError> {
-    log::debug!("------ GLSL-pomelo ------");
-
+pub fn parse_str(
+    source: &str,
+    entry: &str,
+    stage: ShaderStage,
+    defines: Vec<(String, String)>,
```
kvark, line 28: Why not get it in the format we expect, right away (i.e. `FastHashMap<String, String>`)?

src/front/glsl/preprocess.rs
```diff
@@ -0,0 +1,145 @@
...
+    #[error("unmatched else")]
+    UnmatchedElse,
+    #[error("unmatched endif")]
+    UnmatchedEndif,
+}
+
+#[derive(Clone, Debug)]
+pub struct LinePreProcessor {
+    pub defines: FastHashMap<String, String>,
+    if_stack: Vec<(bool, bool)>,
```
kvark, line 16: this looks quite weird, it's totally unclear what this `(bool, bool)` is

src/front/glsl/preprocess.rs
```diff
@@ -0,0 +1,145 @@
...
+                        }
+                    }
+                    "ifndef" => {
+                        if let Some(rest) = iter.next() {
+                            let key = rest.trim();
+                            self.if_stack.push((!self.defines.contains_key(key), false));
+                        }
+                    }
+                    "else" => {
+                        if let Some(if_state) = self.if_stack.last_mut() {
```
kvark, line 113: Another way to right this would be `let if_state = self.if_stack.last_mut().ok_or(Error::UnmatchedElse)?;`
Similar technique can be used in other places too

kvark (commented):
Nice work here!
A few non-critical notes

src/front/glsl/mod.rs
pjoe: How about `Option<FastHashMap<String, String>>`? Somewhat easier on your fingers in the cases where you don't really car about initial defines (and can just write `None`)

src/front/glsl/preprocess.rs
pjoe: Indeed, will try turning into small struct instead.

src/front/glsl/mod.rs
pjoe: Hmmm after thinking some more about this, wouldn't this be imposing our `FastHashMap` on the user, as this is exposed in the external interface? Or is `FastHashMap` with `fxhash` so widely used anyway that this won't be a problem?

src/front/glsl/preprocess.rs
pjoe: think I'm starting to love `.ok_or()` 😄

src/front/glsl/mod.rs
```diff
@@ -17,11 +21,19 @@ mod token;
...
 mod variables;
 
-pub fn parse_str(source: &str, entry: &str, stage: ShaderStage) -> Result<Module, ParseError> {
-    log::debug!("------ GLSL-pomelo ------");
-
+pub fn parse_str(
+    source: &str,
+    entry: &str,
+    stage: ShaderStage,
+    defines: Option<FastHashMap<String, String>>,
```
kvark, line 28: I think the `Option` here is not really needed. If the users don't care, they can always pass `Default::default()`.

pjoe: ok, will fix

pjoe: done

kvark (approved):

Issue #243, opened 2020-10-25
[spv-in] Embark SPIRV shaders fail to parse with multiple errors


repi:
Did a quick test to take 20 of our GLSL -> SPIRV (compiled by shaderc) test shaders we have at Embark and just read them into a naga module with the SPIRV front-end with the latest `master` version of naga. 

Was a collection of vertex, fragment and (a few) compute shaders. 

Unfortunately all of them failed to parse the valid SPIRV, they failed on these set of errors:

- InvalidId(14)
- InvalidId(17)
- UnsupportedCapability
- UnsupportedCapability(MultiView)
- UnsupportedInstruction
- UnsupportedInstruction(Function, BitwiseAnd)
- UnsupportedStorageClass(9)

This was more of a first experiment to see if `naga` could handle our current fairly simple shaders, we are not reliant on this at this point but thought it would be good to report.

If there are any later additional improvements or fixes in `naga` with the SPIRV input parser, ping me here and I can re-run the tests. May be able to later also open source these shaders as we'll use them in the open [Rust GPU](https://shader.rs) project as well (convert them from GLSL to Rust, still output SPIRV).

kvark:
Thank you for filing! It would be ideal if we had the shaders to test on: I don't want to distract you asking to test stuff if we can do it on our end.

repi:
Thanks, we'll think about how/if we can extract them, but likely not in the next few weeks. But would be useful to repro the GLSL frontend issues we also ran into

repi:
Filed PR #341 which adds a set of our current SPIRV shader binaries and a SPIRV parsing test case here to repro this. These SPIRV shaders are generated with our [Rust GPU compiler](https://shaders.rs) now.

PR #244, opened 2020-10-27
Simplify bitshift IR

kvark:
Aligns our IR to WGSL according to https://github.com/gpuweb/gpuweb/pull/914
TL;DR: right shifts are always arithmetic for signed integer and logical for unsigned.
Fixes one of the points in https://github.com/gfx-rs/naga/issues/243


Napokue (approved):
Thank you, LGTM!

PR #245, opened 2020-10-27
Clean up the storage classes

kvark:
Mostly aligns us to https://github.com/gpuweb/gpuweb/pull/1093

I think the storage classes were one of the most confusing parts of the IR, and I've long wanted to make them understandable. I believe, they are now.

The push constant class is not really supported, but we'll need it for future expansion (outside of WebGPU).

See https://github.com/KhronosGroup/SPIRV-Reflect/pull/96 for the logic about where SPIR-V chooses which classes.

### DIFF
--- a/src/back/spv/layout.rs
+++ b/src/back/spv/layout.rs
@@ -24,6 +24,10 @@ impl PhysicalLayout {
         sink.extend(iter::once(self.bound));
         sink.extend(iter::once(self.instruction_schema));
     }
+
+    pub(super) fn supports_storage_buffers(&self) -> bool {
+        self.version >= 0x10300
+    }
 }
 
 impl LogicalLayout {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,14 +108,24 @@ pub enum ShaderStage {
 #[cfg_attr(feature = "deserialize", derive(Deserialize))]
 #[allow(missing_docs)] // The names are self evident
 pub enum StorageClass {
-    Constant,
+    /// Function locals.
     Function,
+    /// Pipeline input, per invocation.
     Input,
+    /// Pipeline output, per invocation, mutable.
     Output,
+    /// Private data, per invocation, mutable.
     Private,
-    StorageBuffer,
-    Uniform,
+    /// Workgroup shared data, mutable.
     WorkGroup,
+    /// Uniform buffer data.
+    Uniform,
+    /// Storage buffer data, potentially mutable.
+    Storage,
+    /// Opaque handles, such as samplers and images.
+    Handle,
+    /// Push constants.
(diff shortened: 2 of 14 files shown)


src/back/spv/layout.rs
```diff
@@ -24,6 +24,10 @@ impl PhysicalLayout {
         sink.extend(iter::once(self.bound));
         sink.extend(iter::once(self.instruction_schema));
     }
+
+    pub(super) fn supports_storage_buffers(&self) -> bool {
+        self.version >= 0x10300
+    }
```
Napokue, line 30: Its nice to already have this, but for now the back-end only supports SPIR-V 1.0. I don't mind including it already, but just to avoid confusing, there is a chance that SPIR-V version > 1.0 won't work.

src/lib.rs
```diff
@@ -108,14 +108,24 @@ pub enum ShaderStage {
 #[cfg_attr(feature = "deserialize", derive(Deserialize))]
 #[allow(missing_docs)] // The names are self evident
 pub enum StorageClass {
-    Constant,
+    /// Function locals.
     Function,
+    /// Pipeline output, per invocation.
```
Napokue, line 113: This comment, and the next one: "/// Pipeline input, per invocation, mutable.", are swapped. But, ultimately they are very handy to have, as I now instantly know where to use them for.

test-data/quad.wgsl
```diff
@@ -14,8 +14,8 @@ fn main() -> void {
 
 # fragment
 [[location(0)]] var<in> v_uv : vec2<f32>;
-[[group(0), binding(0)]] var<uniform> u_texture : texture_sampled_2d<f32>;
-[[group(0), binding(1)]] var<uniform> u_sampler : sampler;
+[[group(0), binding(0)]] var u_texture : texture_sampled_2d<f32>;
+[[group(0), binding(1)]] var u_sampler : sampler;
```
Napokue, line 18: So by default `var` are now `uniform`, sweet!

Napokue (changes requested):
Thank you, I've a couple small things

src/back/spv/layout.rs
kvark: yeah, I understand. If we have a separate version check and error on anything other than 1.0, that's not going to make this code less needed. In the end, we'll want to support higher SPIR-V versions

test-data/quad.wgsl
kvark: Actually, our old names (and SPIR-V storage class names) are rather confusing. In new terms, there isn't a "default" class. However, samplers and images belong to `Handle` class unconditionally.

Napokue (approved):

PR #248, opened 2020-10-30
A bunch of fixes in MSL to get the compute example going

kvark:


src/back/msl.rs
```diff
@@ -247,6 +247,15 @@ impl<I: Indexed> From<I> for Name<'_> {
     }
 }
 
+fn is_valid_identifier_name(s: &str) -> bool {
+    let mut chars = s.chars();
+    match chars.next() {
+        Some(c) if c.is_ascii_alphabetic() => (),
+        _ => return false,
+    };
+    chars.all(|c| c.is_alphanumeric() || c == '_')
```
grovesNL, line 256: We'd probably want `is_ascii_alphanumeric` to be consistent with the first check, or use the non-ASCII versions for both.

I guess even if non-ASCII is allowed we might still want to restrict it.

src/back/msl.rs
```diff
@@ -247,6 +247,15 @@ impl<I: Indexed> From<I> for Name<'_> {
     }
 }
 
+fn is_valid_identifier_name(s: &str) -> bool {
+    let mut chars = s.chars();
+    match chars.next() {
+        Some(c) if c.is_ascii_alphabetic() => (),
```
grovesNL, line 253: `_` should probably be allowed here

grovesNL (approved):

PR #249, opened 2020-10-30
Option to dump spirv flow graphs

kvark:


MatusT (commented):
I like it. Especially the separation of each function into its own file.

PR #250, opened 2020-10-30
msl-out: Use io::Write instead of fmt::Write

JCapucho:
Makes the msl out use `io::Write` instead of `fmt::Write`.

`fmt::Write` is only implemented for `String` and even the rust documentation says that
> This is similar to the standard library's io::Write trait, but it is only intended for use in libcore.

### DIFF
--- a/src/back/msl.rs
+++ b/src/back/msl.rs
@@ -19,9 +19,11 @@ use crate::{
     proc::{ResolveContext, ResolveError, Typifier},
     FastHashMap,
 };
-use std::fmt::{Display, Error as FmtError, Formatter};
-use std::io::{Error as IoError, Write};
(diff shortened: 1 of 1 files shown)


src/back/msl.rs
```diff
@@ -19,7 +19,9 @@ use crate::{
     proc::{ResolveContext, ResolveError, Typifier},
     FastHashMap,
 };
-use std::fmt::{Display, Error as FmtError, Formatter, Write};
+use std::fmt::{Display, Error as FmtError, Formatter};
```
kvark, line 22: this is super nit, can we please use a single `use` per crate?

kvark (approved):
Very nice! Just one nit

PR #258, opened 2020-11-02
GLSL local variables, interface blocks, and XOR

kristoff3r:
I did some more GLSL work. With this all the shaders from #210 gets through glsl-in -> glsl-out with what looks like a sensible result - except for a small bug in glsl-out, which I'll open another pull request for.

### DIFF
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -8,7 +8,7 @@ pomelo! {
         use crate::{proc::Typifier, Arena, BinaryOperator, Binding, Block, Constant,
             ConstantInner, EntryPoint, Expression, FallThrough, FastHashMap, Function, GlobalVariable, Handle, Interpolation,
             LocalVariable, MemberOrigin, SampleLevel, ScalarKind, Statement, StorageAccess,
-            StorageClass, StructMember, Type, TypeInner};
+            StorageClass, StructMember, Type, TypeInner, UnaryOperator};
     }
     %token #[derive(Debug)] #[cfg_attr(test, derive(PartialEq))] pub enum Token {};
     %parser pub struct Parser<'a> {};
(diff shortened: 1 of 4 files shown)


src/front/glsl/parser.rs
```diff
@@ -1071,6 +1072,27 @@ pomelo! {
             );
             if let Some(id) = id {
                 extra.lookup_global_variables.insert(id, h);
+            } else {
+                // variables in interface blocks without an instance name are in the global namespace
+                // https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL)
+                if let TypeInner::Struct { members } = &extra.module.types[d.ty].inner {
```
kvark, line 1078: Let's not match on a reference, instead we should just do `ref members`. This is more explicit and less sugary.

kvark (approved):

src/front/glsl/parser.rs
```diff
@@ -1071,6 +1072,27 @@ pomelo! {
...
+                                    name: Some(name.into()),
+                                    class,
+                                    binding: binding.clone(),
+                                    ty: m.ty,
+                                    interpolation,
+                                    storage_access: StorageAccess::empty(), // TODO
+                                });
+                            extra.lookup_global_variables.insert(name.into(), h);
+                        }
+                    }
```
pjoe, line 1094: I don't really see how this works - how does it relate to the code here: https://github.com/kristoff3r/naga/blob/c436606d09188efab0057db469802ea01df3230f/src/front/glsl/parser.rs#L971.

Also won't this register the members as 'plain' global vars? What happens when these are referenced - then I don't think it will translate into AccessIndex. (OpAccessChain in spv).

I made a small sample where you can see the spv output for named vs. anon interface block: http://shader-playground.timjones.io/12dc04ae724cbb6aa1ddcf812e638302

kristoff3r: Ah, I didn't notice that there was some handling of this already. I'll take a look at it and see if it needs changing.

pjoe: No worries, we'll sort it out :)

kvark: I shall wait for @pjoe before merging the GLSL-in fixes

PR #259, opened 2020-11-02
[glsl-out] Handle initializers for locals

kristoff3r:
This pretty trivial change unfortunately required a loop to mutably borrow both
`builder` and `manager` at the same time, so I had to separate them. Perhaps a
cleaner solution would be to use an object with internal state instead of free
functions.

### DIFF
--- a/src/back/glsl.rs
+++ b/src/back/glsl.rs
@@ -684,23 +684,6 @@ pub fn write<'a>(
             })
             .collect();
 
-        for (handle, name) in locals.iter() {
-            writeln!(
-                &mut buf,
-                "\t{} {};",
-                write_type(
-                    func.local_variables[*handle].ty,
-                    &module.types,
-                    &structs,
-                    None,
-                    &mut manager
-                )?,
-                name
-            )?;
-        }
-
-        writeln!(&mut buf)?;
-
         let mut builder = StatementBuilder {
             functions: &functions,
             globals: &globals_lookup,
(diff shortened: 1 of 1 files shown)


JCapucho (commented):
Thanks for the PR but there might be an issue here, the initialiser might refer to a local which hasn't been declared yet

kristoff3r:
> Thanks for the PR but there might be an issue here, the initialiser might refer to a local which hasn't been declared yet

Ah, I didn't realize that was allowed. In that case I guess we need to do a topological sort first?

JCapucho:
> > Thanks for the PR but there might be an issue here, the initialiser might refer to a local which hasn't been declared yet
> 
> Ah, I didn't realize that was allowed. In that case I guess we need to do a topological sort first?

The way I was thinking doing it was to write the local declaration

```
Type name;
```

And then write the initialiser

```
name = initialiser;
```

But now I'm not sure if a sort would be faster

kvark:
It could be a design choice for the IR to require that the local variable initializers can only depend on previous local vars.
I guess we should check what SPIR-V requires in that sense.

kvark:
Actually, it may be possible only in SPIRV frontend, and it makes no benefit to the backends.
So let's just enforce this at the IR level, so GLSL backend can expect these to be ordered already.

kvark:
Actually actually, this isn't possible in SPIR-V either:
> Initializer must be an <id> from a constant instruction or a global (module scope) OpVariable instruction.

So a local can't be initialized with the value of another local. So we are all good :)

kvark (approved):

PR #267, opened 2020-11-11
Fix and refactor the constant array sizes

kvark:
Our old semantics was entirely broken wrt SPIR-V :sweat_smile:


Napokue (approved):
Thank you! Looks good that we are actually validating now if the constant exists.

PR #268, opened 2020-11-11
Naming processor, used for [msl]

### LINKED ISSUE #264: Do something about reserved keywords in metal
I have the following in a shader that works fine on vulkan:
```glsl
void main() {
    out_colour = colour;


    // Remove the y coordinate because we don't want to scale on that axis.
    vec3 position_no_y = vec3(position.x, 0.0, position.z);
    // Calculate what the new distance from the center should be.
    // When the radius is 1 there should be no change.
    float new_length = length(position_no_y) + radius - 1;
    // Rescale the position by the new length and add the y coordinate back.
    vec3 new = normalize(position_no_y) * new_length;
    new.y = position.y;
    
    mat4 transform = mat4_from_translation(center);
    mat4 modelview = view * transform;
    gl_Position = perspective * modelview * vec4(new, 1.0);
}
```
but on osx I get a runtime error because `new` is a reserved keyword or something.
```
[2020-11-07T23:36:25Z ERROR gfx_backend_metal::device] Error compiling the shader CompilationFailed("
Compilation failed: 
program_source:44:12: error: expected unqualified-id
float3 new = normalize(position_no_y) * new_length;
program_source:45:8: error: \'new\' is not supported in Metal
new.y = in.position.y;
program_source:49:66: error: \'new\' is not supported in Metal
out.gl_Position = (_104.perspective * modelview) * float4(new, 1.0);
")
```

It's not best practice to name a variable `new` in general, but I still feel like it'd be nice to have this code work on all platforms without having to change it. The best way might be to 'mangle' variables like this when translating to metal shaders? I'm not sure.

kvark:
Fixes #264
The idea is having a global registry of names, to avoid collisions. We fill it up as a separate step before writing down `Module` contents, so that during writing we only need to look up those names in a created map.

### DIFF
--- a/src/proc/namer.rs
+++ b/src/proc/namer.rs
@@ -1,0 +1,122 @@
+use crate::{arena::Handle, FastHashMap};
+use std::collections::hash_map::Entry;
+
+pub type EntryPointIndex = u16;
+
+#[derive(Debug, Eq, Hash, PartialEq)]
+pub enum NameKey {
+    GlobalVariable(Handle<crate::GlobalVariable>),
+    Type(Handle<crate::Type>),
+    StructMember(Handle<crate::Type>, u32),
+    Function(Handle<crate::Function>),
+    FunctionParameter(Handle<crate::Function>, u32),
+    FunctionLocal(Handle<crate::Function>, Handle<crate::LocalVariable>),
+    EntryPoint(EntryPointIndex),
+    EntryPointLocal(EntryPointIndex, Handle<crate::LocalVariable>),
+}
+
+/// This processor assigns names to all the things in a module
+/// that may need identifiers in a textual backend.
+pub struct Namer {
+    unique: FastHashMap<String, u32>,
+}
+
+const MAX_PARAM_COUNT: u32 = 100;
+const PARAM_NAME: &str = "param";
+
+impl Namer {
+    fn sanitize(string: &str) -> String {
+        let mut base = string
+            .chars()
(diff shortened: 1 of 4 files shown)


src/proc/namer.rs
```diff
@@ -0,0 +1,117 @@
...
+        reserved: &[&str],
+        output: &mut FastHashMap<NameKey, String>,
+    ) {
+        let mut this = Namer {
+            unique: reserved
+                .iter()
+                .map(|string| (string.to_string(), 0))
+                .collect(),
+        };
+        let max_parameters = 100;
```
Napokue, line 75: Can't this be a `const`, just like `PARAM_NAME`?

src/proc/namer.rs
```diff
@@ -0,0 +1,117 @@
...
+        this.unique.insert(PARAM_NAME.to_string(), max_parameters);
```
Napokue, line 76: I am only confused to why we need to insert this here, is this an always reserved keyword?

Napokue (changes requested):
Looks good, thank you! I only have two minor comments.

src/proc/namer.rs
kvark: This needs to be reworked...

kvark: Added a comment to explain

kvark:
Thanks for the review! I believe I addressed the comments, also tested this a bit and found one more thing I fixed :)

PR #270, opened 2020-11-11
[ir] support function argument names

kvark:


src/proc/namer.rs
```diff
@@ -97,12 +91,9 @@ impl Namer {
...
             let fun_name = this.call_or(&fun.name, "function");
             output.insert(NameKey::Function(fun_handle), fun_name);
-            if fun.parameter_types.len() >= MAX_PARAM_COUNT as usize {
-                unreachable!()
-            }
-            for (index, _) in fun.parameter_types.iter().enumerate() {
-                let name = format!("{}{}", PARAM_NAME, index + 1);
-                output.insert(NameKey::FunctionParameter(fun_handle, index as u32), name);
+            for (index, arg) in fun.arguments.iter().enumerate() {
+                let name = this.call_or(&arg.name, "param");
```
Napokue, line 95: Maybe suffix the fallback value in the following format: "param_[index]"? No real benefit, other than having unique values for the name.

src/back/spv/writer.rs
```diff
@@ -324,21 +324,18 @@ impl Writer {
 
         let return_type_id =
             self.get_function_return_type(ir_function.return_type, &ir_module.types);
-        let mut parameter_type_ids = Vec::with_capacity(ir_function.parameter_types.len());
+        let mut parameter_type_ids = Vec::with_capacity(ir_function.arguments.len());
 
         let mut function_parameter_pointer_ids = vec![];
 
-        for parameter_type in ir_function.parameter_types.iter() {
+        for argument in ir_function.arguments.iter() {
```
Napokue, line 331: Briefly mentioned in Matrix, and I quote what you said:
> I'm going to add names to function parameters next. SPIR-V doesn't appear to have that, but it's fine. Everything else has it.
I looked through the SPIR-V specification and I tested in the [Shader Playground](http://shader-playground.timjones.io/015c96e25bbccc898704cb9ab42d8295), but there is a way to "name" your arguments. Namely: via `OpName`.

`OpName` will only be written when `WriterFlags::DEBUG` is enabled. So what you can do is something like this:

```rust
if self.writer_flags.contains(WriterFlags::DEBUG) {
    if let Some(ref name) = argument.name {
        self.debugs
            .push(super::instructions::instruction_name(id, name.as_str()));
    }
}
```

Maybe this is also a case for `spv-in` to check if there is an `OpName` associated with the function argument?

Napokue (changes requested):

src/back/spv/writer.rs
Napokue: Discussed with @kvark via Matrix, and we concluded to follow this up in another PR. I've created #272 for this.

Napokue (approved):

PR #273, opened 2020-11-12
[ir] select expression

kvark:
I figured this to be require to support `OpSelect` from SPIR-V, which is used in WebGPU samples.
An alternative would involve splitting the expressions into statements, blocks, and merging with `OpPhi` - the whole eternal pain :)


Napokue (approved):
Love it, thanks!

PR #274, opened 2020-11-14
[spv-out] Add and implement i_sub and f_sub

Napokue:
- Add and implement i_sub and f_sub.
- Reduce DRY code from binary instructions.
- Move out binary operator logic out of `crate::BinaryOperator::Multiply` match to easily implement other binary instructions.

There is however an edge case, and I am not sure yet how we want to solve this. Both instructions are also compatible with vectors, but as the SPIR-V spec says: "The types of Operand 1 and Operand 2 both must be the same as Result Type.".

What this means, and I have checked this in the [shader playground](http://shader-playground.timjones.io/ce2a44d27e0012a232a04aa94fdda636), is that a float vector can be subtracted by an integer. This contradicts the SPIR-V spec, but it does work in the playground as follows:

1)  Convert integer to float
2) Create an CompositeConstruct (a vector) that matches the vector operand result type (in this case a vec2)
3) Syntax has now basically become `vec2(1.0)` which translates in GLSL to a vector with 2 elements set to the value of 1.0
4) As the result types are now equal, subtract them from each other

It gets weirder as it is not possible to do this same "trick" with an `ivec2`. So `ivec2 vector = ivec2(1, 1) - 1` is not valid, because apparently the integer is still converted to a float.

Do we also want to compile in this way? The problem is that I think that it is a fairly common case to substract a float vector with an integer, as the Hello-triangle example already uses this in its shader.

### DIFF
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -571,6 +571,38 @@ pub(super) fn instruction_composite_construct(
 //
 // Arithmetic Instructions
 //
+fn instruction_binary(
+    op: Op,
+    result_type_id: Word,
+    id: Word,
+    operand_1: Word,
+    operand_2: Word,
+) -> Instruction {
+    let mut instruction = Instruction::new(op);
+    instruction.set_type(result_type_id);
+    instruction.set_result(id);
+    instruction.add_operand(operand_1);
+    instruction.add_operand(operand_2);
+    instruction
+}
+
+pub(super) fn instruction_i_sub(
+    result_type_id: Word,
+    id: Word,
+    operand_1: Word,
+    operand_2: Word,
+) -> Instruction {
+    instruction_binary(Op::ISub, result_type_id, id, operand_1, operand_2)
+}
+
+pub(super) fn instruction_f_sub(
+    result_type_id: Word,
+    id: Word,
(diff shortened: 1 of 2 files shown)


kvark (approved):
Nice, great stuff!

> "The types of Operand 1 and Operand 2 both must be the same as Result Type.".

That's equally the same with our IR. The types of expressions on both sides of add/remove have to be the same.

kvark:
> What this means, and I have checked this in the shader playground, is that a float vector can be subtracted by an integer. This contradicts the SPIR-V spec

That example doesn't contradict SPIR-V spec, because it's GLSL example :)
In desktop (not ES!) GLSL there are (nasty!) implicit conversions:
  - from int to float
  - from scalar to vector

Generated SPIR-V obviously don't have them implicit.

> So ivec2 vector = ivec2(1, 1) - 1 is not valid, because apparently the integer is still converted to a float.

Are you sure about that? Just added that line to your playground, and it compiles just fine. Integers don't *unconditionally* convert to float, but only as needed in GLSL.

> Do we also want to compile in this way?

It's up to GLSL frontend (@JCapucho and @pjoe ) to decide if they want to support these implicit conversions. This doesn't concern IR or spv-out.

PR #275, opened 2020-11-14
[spv-out] Add and implement bitwise_and

Napokue:
Support the following cases:

GLSL
```glsl
#version 450

void main() {
  ivec2 test_1 = ivec2(1, 1) & ivec2(1, 1);    
  int test_2 = 1 & 1;
}
```

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1214,6 +1214,15 @@ impl Writer {
                         },
                         _ => unreachable!(),
                     },
+                    crate::BinaryOperator::And => (
+                        super::instructions::instruction_bitwise_and(
+                            left_result_type_id,
+                            id,
+                            left_id,
+                            right_id,
+                        ),
+                        left_lookup_ty,
+                    ),
                     _ => unimplemented!("{:?}", op),
                 };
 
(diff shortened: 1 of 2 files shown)


src/back/spv/writer.rs
```diff
@@ -1214,6 +1214,41 @@ impl Writer {
                         },
                         _ => unreachable!(),
                     },
+                    crate::BinaryOperator::And => match *left_ty_inner {
+                        crate::TypeInner::Vector { .. } => match *right_ty_inner {
```
kvark, line 1218: Here and in other binary operators we don't actually need to match the left and right types, at all. The valid IR guarantees that they are the same, and they are applicable. You just need to generate the proper SPIR-V instruction with that in mind, no need to re-validate the types.
The only places where types need to be checked are places where instructions are different, e.g. multiply instructions f_add/i_add, etc

kvark (changes requested):

src/back/spv/writer.rs
Napokue: Yes that is true! This is the first binary operator that uses one instruction for all its result types.

kvark (approved):
May be first, but not the last :)

PR #276, opened 2020-11-14
[spv-out] Reduce BinaryOperator::Multiply indentation

Napokue:
Combined matches together to reduce indentation. Later when the `BinaryOperator` match grows, we can consider to split each match to its own function.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -135,6 +135,21 @@ impl<'a, T> ops::Deref for MaybeOwned<'a, T> {
     }
 }
 
+enum Dimension {
+    Scalar,
+    Vector,
+    Matrix,
+}
+
+fn get_dimension(ty_inner: &crate::TypeInner) -> Dimension {
+    match *ty_inner {
+        crate::TypeInner::Scalar { .. } => Dimension::Scalar,
+        crate::TypeInner::Vector { .. } => Dimension::Vector,
+        crate::TypeInner::Matrix { .. } => Dimension::Matrix,
+        _ => unreachable!(),
+    }
+}
+
 pub struct Writer {
     physical_layout: PhysicalLayout,
     logical_layout: LogicalLayout,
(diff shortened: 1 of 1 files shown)


src/back/spv/writer.rs
```diff
@@ -1115,59 +1115,56 @@ impl Writer {
...
-                        crate::TypeInner::Vector { .. } => match *right_ty_inner {
-                            crate::TypeInner::Scalar { .. } => (
-                                super::instructions::instruction_vector_times_scalar(
-                                    left_result_type_id,
-                                    id,
-                                    left_id,
-                                    right_id,
-                                ),
-                                left_lookup_ty,
+                    crate::BinaryOperator::Multiply => match (*left_ty_inner, *right_ty_inner) {
```
Napokue, line 1118: Why can I match on reference with a single value, but not with two? Has it something to do that ( ) is a tuple like structure where Rust stores the values in?

Do you have any idea how to fix this, @kvark ? Maybe your proposed solution like discussed at Matrix, e.g.:
```rust
fn dimension(ty: &crate::TypeInner) -> u8 {
  match *ty {
    Ty::Scalar { .. } => 0,
    Ty::Vector { .. } => 1,
    Ty::Matrix { .. } => 2,
    _ => unreachable!(),
  }
}
```

kvark: the main concept to understand is that pattern matching on something doesn't take it by value, i.e. `match x` doesn't require moving of `x` anywhere. So `match *x` works fine if you just have `x: &Something`. However, as soon as you write `(*x, *y)`, Rust can't see through this expression, the smartness is limited to the immediate argument of `match`. So `(*x, *y)` has to move things out, and we have to resort to `(x, y)` (where these are references) instead.

kvark: I think we should do with a helper `fn dimension(ty: &crate::TypeInner) -> Dimension` to work around that matching limitation

src/back/spv/writer.rs
```diff
@@ -1114,60 +1129,57 @@ impl Writer {
...
+                                id,
+                                left_id,
+                                right_id,
                             ),
-                            _ => unreachable!(),
-                        },
-                        crate::TypeInner::Scalar { kind, .. } => match kind {
+                            left_lookup_ty,
+                        ),
+                        (Dimension::Scalar { kind, .. }, Dimension::Scalar { .. }) => match kind {
```
kvark, line 1182: can we match more aggressively here? i.e.
```rust
(Dimension::Scalar { kind: crate::ScalarKind::Float }, Dimension::Scalar { .. }) => {...} // floating-point path
(Dimension::Scalar { .. }, Dimension::Scalar { .. }) => {...} // integer path
```

kvark (commented):
Note that we have `TypeInner::scalar_kind` in the IR, which we could use here.
So you technically could remove the `kind` from your `Dimension::Scalar` variant and instead check separately, e.g.
```rust
(Dimension::Scalar, Dimension::Scalar) if something.scalar_kind() == Some(crate::ScalarKind::Float) => {...} // floating point path
(Dimension::Scalar, Dimension::Scalar)  => {..} // integer path
```

src/back/spv/writer.rs
Napokue: I wasn't even aware that matching that way works 😮 I removed the kind, so I went with the `scalar_kind()` function.

src/back/spv/writer.rs
```diff
@@ -1114,80 +1129,83 @@ impl Writer {
...
+                                right_id,
                             ),
-                            _ => unreachable!(),
-                        },
-                        crate::TypeInner::Scalar { kind, .. } => match kind {
-                            crate::ScalarKind::Sint | crate::ScalarKind::Uint => (
+                            left_lookup_ty,
+                        ),
+                        (Dimension::Scalar, Dimension::Scalar)
+                            if left_ty_inner.scalar_kind() == Some(crate::ScalarKind::Sint)
```
kvark, line 1183: let's first put the one with `Float` condition, and this arm would go after it. It doesn't need the `if` at all, since naturally it's either integer or unreachable. We can safely assume it's integer.

kvark (commented):
sorry one more little step!

src/back/spv/writer.rs
Napokue: True yes, I used this order as I followed the ScalarKind enum order. Makes sense to put Float first.

kvark (approved):
Great, thank you!

Issue #277, opened 2020-11-15
[glsl-in] Support for casting


Napokue:
It is not yet possible to cast. Using the following GLSL code:

```glsl
#version 450

void main() {
    float a = float(1);
}
```

This does not crash the compilation, but rather results for some reason to `Expression::Compose`, instead of the expected `Expression::As`.

pjoe:
Currently glsl-in sees this as a type constructor, that always result in compose. Thinking it should be easy enough to have it use `Expression::As` when there is only one component.

Napokue:
Fixed in #278

PR #278, opened 2020-11-16
[glsl-in] Handle simple casts

pjoe:
Related to #277 

Example:
```glsl
#version 450

void main() {
    float w = float(1);
} 
```

This should now result in `Expression::As` instead of `Expression::Compose`

### DIFF
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -244,10 +244,25 @@ pomelo! {
     function_call ::= function_call_or_method(fc) {
         match fc.kind {
             FunctionCallKind::TypeConstructor(ty) => {
-                let h = extra.context.expressions.append(Expression::Compose {
-                    ty,
-                    components: fc.args.iter().map(|a| a.expression).collect(),
-                });
+                let h = if fc.args.len() == 1 {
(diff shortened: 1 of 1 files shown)


Napokue (approved):

src/front/glsl/parser.rs
```diff
@@ -244,10 +244,25 @@ pomelo! {
     function_call ::= function_call_or_method(fc) {
         match fc.kind {
             FunctionCallKind::TypeConstructor(ty) => {
-                let h = extra.context.expressions.append(Expression::Compose {
-                    ty,
-                    components: fc.args.iter().map(|a| a.expression).collect(),
-                });
+                let h = if fc.args.len() == 1 {
+                    let kind = match extra.module.types[ty].inner {
```
kvark, line 248: could we use the existing `scalar_kind()` method here? See https://github.com/gfx-rs/naga/blob/9041cd4b10ec6105d9041d103fe0a78022c3472c/src/proc/mod.rs#L54

pjoe: Ahh, of course - hadn't spotted that one. Guess need to do a new PR :)

PR #279, opened 2020-11-16
[glsl-in] use TypeInner::scalar_kind for casts

pjoe:
Small followup fix for #278

### DIFF
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -245,12 +245,10 @@ pomelo! {
         match fc.kind {
             FunctionCallKind::TypeConstructor(ty) => {
                 let h = if fc.args.len() == 1 {
-                    let kind = match extra.module.types[ty].inner {
-                        TypeInner::Scalar{kind, ..} => kind,
-                        TypeInner::Vector{kind, ..} => kind,
-                        _ => {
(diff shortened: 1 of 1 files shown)


src/front/glsl/parser.rs
```diff
@@ -245,12 +245,10 @@ pomelo! {
         match fc.kind {
             FunctionCallKind::TypeConstructor(ty) => {
                 let h = if fc.args.len() == 1 {
-                    let kind = match extra.module.types[ty].inner {
-                        TypeInner::Scalar{kind, ..} => kind,
-                        TypeInner::Vector{kind, ..} => kind,
-                        _ => {
-                            return Err(ErrorKind::SemanticError("Can only cast to scalar or vector"));
-                        }
+                    let kind = if let Some(kind) = extra.module.types[ty].inner.scalar_kind() {
```
kvark, line 248: this could be written simpler as
```rust
let kind = extra.module.types[ty].inner
    .scalar_kind()
    .ok_or(ErrorKind::SemanticError("Can only cast to scalar or vector"))?;
```

pjoe: Damn, when do I learn to remember those nice FP helper functions Rust has 😃

kvark (approved):

PR #284, opened 2020-11-16
[spv-out] Load variables when using CompositeConstruct

### LINKED ISSUE #283: [spv-out] Using variables in CompositeConstruct causes validation error
When using a variable in a CompositeConstruct, spv-out does not load the variables via an `OpLoad`, but uses them directly, which is not possible and throws the following error when validating the output:
```
"Expected Constituents to be scalars or vectors of the same type as Result Type components"
```

Simply checking if the constituent is a variable and loading it via `OpLoad` should be enough to fix this.

Napokue:
Should fix a problem 3 wgpu-rs examples are facing currently. This change is a duplicate, and used by the matching arm of `Expression::Binary`. I want to refactor this, so we can easily use this for other Expressions (without DRY code), but I will do that a little bit later when spv-out is more feature complete.

Fixes #283

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1017,13 +1017,28 @@ impl Writer {
                 let mut constituent_ids = Vec::with_capacity(components.len());
                 for component in components {
                     let expression = &ir_function.expressions[*component];
-                    let (component_id, _) = self.write_expression(
+                    let (component_id, component_local_ty) = self.write_expression(
                         ir_module,
                         &ir_function,
                         expression,
                         block,
                         function,
                     )?;
+
+                    let component_id = match expression {
+                        crate::Expression::LocalVariable(_)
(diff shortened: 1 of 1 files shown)


src/back/spv/writer.rs
```diff
@@ -1017,14 +1017,29 @@ impl Writer {
...
-                    let (component_id, _) = self.write_expression(
+                    let (component_id, component_local_ty) = self.write_expression(
                         ir_module,
                         &ir_function,
                         expression,
                         block,
                         function,
                     )?;
 
+                    let component_id = match expression {
```
kvark, line 1028: we should start abstracting this block away, given how often we have to repeat it

kvark (approved):

src/back/spv/writer.rs
Napokue: Agreed, will add it to my todo

PR #287, opened 2020-11-18
[spv-out] Add support for vector with vector multiplication

Napokue:
I went through the Bevy shader thread (#210) and a requirement for a shader to work was vector with vector multiplication. I have used the same strategy as we came up with the normal multiplication: if it is not a float vector times float vector, we will default to an `ivec`/`uvec multiplication.

There are however two other vector types in GLSL at least: `bvec` and `dvec`, but I don't think they are that commonly used, but that is just an assumption?

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1194,7 +1194,8 @@ impl Writer {
                             ),
                             left_lookup_ty,
                         ),
-                        (Dimension::Scalar, Dimension::Scalar)
+                        (Dimension::Vector, Dimension::Vector)
+                        | (Dimension::Scalar, Dimension::Scalar)
                             if left_ty_inner.scalar_kind() == Some(crate::ScalarKind::Float) =>
                         {
                             (
@@ -1207,7 +1208,8 @@ impl Writer {
                                 left_lookup_ty,
                             )
                         }
-                        (Dimension::Scalar, Dimension::Scalar) => (
+                        (Dimension::Vector, Dimension::Vector)
+                        | (Dimension::Scalar, Dimension::Scalar) => (
                             super::instructions::instruction_i_mul(
                                 left_result_type_id,
                                 id,


kvark:
We should support dvec eventually, but multiplying `bvec` seems like an invalid op.

src/back/spv/writer.rs
```diff
@@ -1158,6 +1158,28 @@ impl Writer {
                             ),
                             left_lookup_ty,
                         ),
+                        (Dimension::Vector, Dimension::Vector)
+                            if left_ty_inner.scalar_kind() == Some(crate::ScalarKind::Float) =>
+                        {
+                            (
```
kvark, line 1164: this is formatted inconsistently with the follow-up arm, can we remove  the `{` and `}` around the tuple?

src/back/spv/writer.rs
```diff
@@ -1158,6 +1158,28 @@ impl Writer {
                             ),
                             left_lookup_ty,
                         ),
+                        (Dimension::Vector, Dimension::Vector)
```
Napokue, line 1161: We can reduce code and have no DRY code, that is with using this match together with their scalar matching. So we would get:

```rust
(Dimension::Vector, Dimension::Vector) |
(Dimension::Scalar, Dimension::Scalar)
    if left_ty_inner.scalar_kind() == Some(crate::ScalarKind::Float) =>
{
    (
        super::instructions::instruction_f_mul(
            left_result_type_id,
            id,
            left_id,
            right_id,
        ),
        left_lookup_ty,
    )
}
(Dimension::Vector, Dimension::Vector) |
(Dimension::Scalar, Dimension::Scalar) => (
    super::instructions::instruction_i_mul(
        left_result_type_id,
        id,
        left_id,
        right_id,
    ),
    left_lookup_ty,
),
```

src/back/spv/writer.rs
Napokue: Yes I can, `cargo fmt` put those `{` and `}` around it. Let me check if I remove it what happens.

kvark (approved):
Thank you!

PR #289, opened 2020-11-19
[mtl] Restructure the backend a bit

### LINKED ISSUE #239: [MSL] Don't include metal namespace
Shaders are read perfectly fine when the metal namespace functions are called with the full path.
We don't need this:
```
using namespace metal;
```

kvark:
Extends the reserved words. Makes the backend modular.
Fixes #239


kvark:
Clippy errors are unrelated, they belong to GLSL and WGSL frontends.

src/back/msl/keywords.rs
```diff
@@ -0,0 +1,100 @@
...
+    "or_eq",
+    "xor",
+    "xor_eq",
+    "compl",
+    // Metal-specific
+    "constant",
+    "device",
+    "threadgroup",
+    "threadgroup_imageblock",
+    "compute",
```
grovesNL, line 94: I wonder how we could easily find keywords like `kernel` to include here. We might also want to consider adding most of the C++ keywords which includes words like `using`

grovesNL (approved):
Looks great, thanks!

kvark:
Agreed! Let's follow-up with more words (like `kernel` and `using`). This PR ended up doing lots of re-org on top of the reserved words stuff.

src/back/msl/keywords.rs
kvark: Added in https://github.com/gfx-rs/naga/commit/6e42c3468306d410a7473397525a45da0f65e420

Issue #290, opened 2020-11-19
Flatten the special vertex output structures


kvark:
Things like this needs to be flattened and not visible in the IR:
```cpp
out gl_PerVertex {
    vec4 gl_Position;
};
```
See also - #288

kvark:
Closing in favor of #359

PR #291, opened 2020-11-19
[glsl-in] Skip gl_PerVertex

pjoe:
For now simple ignore explicit `gl_PerVertex` struct. Will treat builtins like `gl_Position` like when implicit.

### DIFF
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -98,7 +98,7 @@ pomelo! {
     %type initializer ExpressionRule;
 
     // declarations
-    %type declaration VarDeclaration;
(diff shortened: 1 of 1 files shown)


pjoe:
Related to #290

kvark (approved):

kvark:
legitimate errors?

pjoe:
Their is a conflict between clippy and the 1.41 build.

Clippy says: you should use [strip_prefix](https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix), but that only got into stable in 1.45 :(

I guess we could either disbale that clippy rule, or update 'the old' build to 1.45.

PR #293, opened 2020-11-20
[spv-out] fix generation of OpConvert

kvark:
Follow-up to #220 to help with https://github.com/gfx-rs/wgpu-rs/issues/620


Napokue (approved):
Thank you, much cleaner by reducing the instruction functions!

PR #297, opened 2020-11-23
[glsl-out] Refractoring

JCapucho:
Splitted the backend into two more modules (features, errors)
Started documenting the backend
Some minor changes


src/back/glsl/error.rs
```diff
@@ -0,0 +1,44 @@
+use crate::proc::ResolveError;
+use std::{
+    fmt::{self, Error as FmtError},
+    io::Error as IoError,
+};
+
+pub type BackendResult = std::result::Result<(), Error>;
+
+#[derive(Debug)]
+pub enum Error {
```
kvark, line 10: let's just use `thiserror` here, see https://github.com/gfx-rs/naga/blob/96c80738650822de35f77ab6a589f309460c8f39/src/proc/validator.rs#L16-L33 for inspiration. For example , doing `IoError(#[from] IoError)` would implement `From<IoError>` for you.

kvark: And with this ergonomic improvement, we can just put the declaration into `glsl/mod.rs` and not have this module

kvark (approved):
Nice comments, and amazing work!
I just have one suggestion :)

kvark (approved):

PR #299, opened 2020-12-03
Add spv::write_vec helper

kvark:
~~All my prior resistance to making it cloneable was futile. Yes, we get away from clone internally. But this is IR, it's a part of API, and for the users it's just convenient to be able to clone it. For example, `gfx` would need a copy for its purposes, and `wgpu` needs a copy for validation.~~

### DIFF
--- a/src/back/msl/mod.rs
+++ b/src/back/msl/mod.rs
@@ -204,8 +204,20 @@ impl ResolvedBinding {
     }
 }
 
-pub fn write_string(module: &crate::Module, options: &Options) -> Result<String, Error> {
+/// Information about a translated module that is required
+/// for the use of the result.
+pub struct TranslationInfo {
+    /// Mapping of the entry point names. Each item in the array
+    /// corresponds to an entry point in `module.entry_points.iter()`.
+    pub entry_point_names: Vec<String>,
+}
+
+pub fn write_string(
+    module: &crate::Module,
+    options: &Options,
+) -> Result<(String, TranslationInfo), Error> {
     let mut w = writer::Writer::new(Vec::new());
-    w.write(module, options)?;
-    Ok(String::from_utf8(w.finish())?)
+    let info = w.write(module, options)?;
+    let string = String::from_utf8(w.finish())?;
+    Ok((string, info))
 }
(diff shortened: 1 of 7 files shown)


src/back/msl/mod.rs
```diff
@@ -204,8 +204,20 @@ impl ResolvedBinding {
...
+pub fn write_string(
+    module: &crate::Module,
+    options: &Options,
+) -> Result<(String, TranslationInfo), Error> {
     let mut w = writer::Writer::new(Vec::new());
-    w.write(module, options)?;
-    Ok(String::from_utf8(w.finish())?)
+    let info = w.write(module, options)?;
+    let string = String::from_utf8(w.finish())?;
+    Ok((string, info))
```
grovesNL, line 222: nit: most of the time both of the tuple items will be used, so it might be slightly more convenient in the long run to keep them on one `Output` struct containing the MSL + any translation info

src/back/msl/mod.rs
```diff
@@ -204,8 +204,20 @@ impl ResolvedBinding {
     }
 }
 
-pub fn write_string(module: &crate::Module, options: &Options) -> Result<String, Error> {
+/// Information about a translated that is required
+/// for the use of the result.
+pub struct TranslationInfo {
+    /// Mapping of the entry point names. Each item in the array
+    /// corresponds to an entry point in `module.entry_points.iter()`.
+    pub entry_point_names: Vec<String>,
```
grovesNL, line 212: nit: `HashMap<(String, ExecutionModel), String>)` might be more convenient instead of iterating to find the index of our entry point

grovesNL (approved):

src/back/msl/mod.rs
kvark: I'm not sure about this. With the way this PR sets it up, gfx-backend-metal doesn't have to do any work at all, i.e. this vector is most convenient. Switching it to a hashmap would be more user-oriented (versus gfx backend oriented), and would require more code on the gfx side, while the benefit isn't clear.

src/back/msl/mod.rs
kvark: I wanted to do this, but the `write` function needs to return the same thing but without the string. So the string is kinda special here, maybe it's fine to return a tuple with it.

kvark:
Thank you for the review!
I rolled back the `Clone` changes, and now this PR is much less controversial. Going forward with this.

PR #302, opened 2020-12-04
Terminator processor, used in WGSL and GLSL frontends

### LINKED ISSUE #301: Required `return` statement in the IR
SPIR-V requires a terminator instruction at the end of each block. For the tails of a graph of blocks in a function, this terminator has to be `return`. However, when writing GLSL, for example, one doesn't have to use explicit `return` statement if there is no value.

I believe we should require the IR to have those returns. That means all the textual frontends need to make sure to insert that statement when necessary. This can be tricky.... Let me explain why I think we should do this:
  1. there is always a need to be able to do a `return` without values. So we have to handle that anyway, we might as well make it non-optional.
  2. it's a balance between doing the work in the frontends and the backends. I believe our endgame is to generate low-level backend forms, such as AIR/DXIL, for which the explicit return is needed (I presume).
  3. our IR validation has to do control flow analysis to check that all the code paths return something for a function that has a return value. Using the same logic to validate void functions is natural, no extra work is needed.

kvark:
Fixes #301

### DIFF
--- a/src/proc/terminator.rs
+++ b/src/proc/terminator.rs
@@ -1,0 +1,40 @@
+/// Ensure that the given block has return statements
+/// at the end of its control flow.
+///
+/// Note: we don't want to blindly append a return statement
+/// to the end, because it may be either redundant or invalid,
+/// e.g. when the user already has returns in if/else branches.
+pub fn ensure_block_returns(block: &mut crate::Block) {
+    match block.last_mut() {
+        Some(&mut crate::Statement::Block(ref mut b)) => {
+            ensure_block_returns(b);
+        }
+        Some(&mut crate::Statement::If {
+            condition: _,
+            ref mut accept,
+            ref mut reject,
+        }) => {
+            ensure_block_returns(accept);
+            ensure_block_returns(reject);
+        }
+        Some(&mut crate::Statement::Switch {
+            selector: _,
+            ref mut cases,
+            ref mut default,
+        }) => {
+            for case in cases.values_mut() {
+                if let (ref mut b, None) = *case {
+                    ensure_block_returns(b);
+                }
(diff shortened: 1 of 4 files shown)


src/proc/terminator.rs
```diff
@@ -0,0 +1,40 @@
...
+///
+/// Note: we don't want to blindly append a return statement
+/// to the end, because it may be either redundant or invalid,
+/// e.g. when the user already has returns in if/else branches.
+pub fn ensure_block_returns(block: &mut crate::Block) {
+    match block.last_mut() {
+        Some(&mut crate::Statement::Block(ref mut b)) => {
+            ensure_block_returns(b);
+        }
+        Some(&mut crate::Statement::If {
```
Kangz, line 12: Why do we need to add it on both sides of the if? It seems it could be sufficient to add a single return at the end of `block` (and maybe it is unreachable). It would make this function much simpler because it would just check if the last statement is a `Return` and if not, then add one.

kvark: I'm not totally sure if that way is better, but the reasoning is expressed at the top of the function:
>/// Note: we don't want to blindly append a return statement
/// to the end, because it may be either redundant or invalid,
/// e.g. when the user already has returns in if/else branches.

kvark: I.e. we don't want to generate IR that looks like this: 
```
if xxx { return; } else { return; }
return
```
Because then we'd hit our own validation error that there is stuff after `return`.

Kangz: Ack, LGTM then!

Kangz (approved):

kvark:
Weird, GHA got stuck?

Issue #312, opened 2020-12-07
Required SPIR-V capabilities


kvark:
The way it's defined by SPIR-V spec is - for each instruction there can be a set of capabilities, such that if any of the set is enabled, the instruction is valid to use. If we are considering all the instructions, then their requirement would be a mix of different sets. This is not the same as just a union of all sets that we currently use.

teoxoy:
This was iterated on and fixed in:

- https://github.com/gfx-rs/naga/pull/315
- https://github.com/gfx-rs/naga/pull/777
- https://github.com/gfx-rs/naga/pull/881
- https://github.com/gfx-rs/naga/pull/1256

PR #313, opened 2020-12-07
[spv] write down workgroup size

kvark:
I commented out the capabilities, since these are currently broken - #312
There is a test CI run - https://github.com/gfx-rs/wgpu/runs/1507885603


kvark:
I think that should be OK to land.

PR #315, opened 2020-12-08
Reworked Spirv capabilities, more tests

### LINKED ISSUE #312: Required SPIR-V capabilities

kvark:
Fixes #312

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ bitflags = "1"
 fxhash = "0.2"
 log = "0.4"
 num-traits = "0.2"
-spirv = { package = "spirv_headers", version = "1.4.2", optional = true }
+spirv = { package = "spirv_headers", version = "1.5", optional = true }
 pomelo = { version = "0.1.4", optional = true }
 thiserror = "1.0.21"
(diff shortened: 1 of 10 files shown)


Napokue (approved):
Thank you! I really like the changes, also the Binary Operator changes! The only thing I'm uncertain about is the name of the function `check`, as it is not really descriptive, but maybe you have had some thought behind it.

kvark:
@Napokue right, I understand the concern! It's an unusual situation, since in all cases where `check` is called, the parameter has `.required_capabilities()`, so it's never ambiguous. If `spirv-headers` makes it a trait, which we generically accept, then we'll need to make the name more descriptive.
Anyway, I don't have hard feelings, just figured it's non-ambiguous for this reason, and it's used a lot, so short may be fine. If you feel different, please rename, and I'll be happy to accept.

Issue #295, opened 2020-11-20
Convert builtin types from GLSL


kvark:
In GLSL, `gl_VertexIndex` is integer, while in Naga IR it needs to be unsigned integer.
I'll follow-up with some validation code for the IR, but the frontend needs to ultimately fix it.

pjoe:
here is what currently happens:
```glsl
#version 450

out gl_PerVertex {
    vec4 gl_Position;
};

void main() {
    float x = float(gl_VertexIndex - 1);
    float y = float(((gl_VertexIndex & 1) * 2) - 1);
    gl_Position = vec4(x, y, 0.0, 1.0);
}
```

which results in:
```spirv
...
         %16 = OpLoad %uint %gl_VertexIndex
         %10 = OpISub %uint %16 %int_1
         %17 = OpConvertUToF %float %10
               OpStore %2 %17
         %21 = OpLoad %uint %gl_VertexIndex
         %20 = OpBitwiseAnd %uint %21 %int_1
         %19 = OpIMul %uint %20 %int_2
         %18 = OpISub %uint %19 %int_1
         %23 = OpConvertUToF %float %18
               OpStore %5 %23
         %27 = OpLoad %float %2
         %28 = OpLoad %float %5
         %31 = OpCompositeConstruct %v4float %27 %28 %float_0 %float_1
               OpStore %gl_Position %31
               OpReturn
               OpFunctionEnd
```

pjoe:
One solution could be to cast `gl_VertexIndex` to int when we are generating expressions for globals during the function prelude.

PR #319, opened 2020-12-08
[spv] force treating the built-in scalar types as unsigned

kvark:
After #306 landed, all GLSL-generated SPIR-V code began failing validation, because GLSL built-in semantics is signed int.
cc @pjoe  - may need to do something like this for GLSL-in?


grovesNL (approved):

kvark:
Found #295 for GLSL

PR #320, opened 2020-12-09
Add all standard library functions to the IR

### LINKED ISSUE #158: Remove `Expression::Cross` variant
It's not an opcode in SPIR-V, it's an extension method.
Arguably, we can even move the `Dot` variant as well, but that should be another issue.

kvark:
Fixes #158
Hopefully, the last big disruptive IR change here.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -516,19 +516,6 @@ pub enum BinaryOperator {
     ShiftRight,
 }
 
-/// Built-in shader function.
-#[derive(Clone, Copy, Debug, Hash, Eq, Ord, PartialEq, PartialOrd)]
-#[cfg_attr(feature = "serialize", derive(Serialize))]
-#[cfg_attr(feature = "deserialize", derive(Deserialize))]
-pub enum IntrinsicFunction {
-    Any,
-    All,
-    IsNan,
-    IsInf,
-    IsFinite,
-    IsNormal,
-}
-
 /// Axis on which to compute a derivative.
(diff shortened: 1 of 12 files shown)


src/lib.rs
```diff
@@ -628,17 +674,26 @@ pub enum Expression {
         accept: Handle<Expression>,
         reject: Handle<Expression>,
     },
-    /// Call an intrinsic function.
-    Intrinsic {
-        fun: IntrinsicFunction,
+    /// Compute the derivative on an axis.
```
grovesNL, line 677: I wonder if these need special handling in the IR itself? If we were able to reuse `Call` with something like `FunctionOrigin::Intrinsic` then it seems like we'd be able to simplify the writers a little

grovesNL (approved):

src/lib.rs
kvark: I thought derivatives are "special" because they depend on the hidden state of the pipeline - the quad-based rasterization. Using them also requires uniform control flow. So that makes them different enough to differentiate from other functions that are mostly math.

kvark:
Thank you for the quick review! I'm super excited to land this 🚀

src/lib.rs
grovesNL: Ah yeah I guess derivatives are probably a bit different. I was mostly wondering about whether we could collapse `Math` and `Relational` into `Call` somehow

kvark: Yeah, we could totally do that. I'm not feeling strongly, just got inspired by the fact the relational stuff is separate in both current WGSL spec draft and the MSL "specification".

PR #324, opened 2020-12-10
Update simple.wgsl to latest syntax

Napokue:
Example works again, as it still used somewhat outdated WGSL syntax.


kvark:
Hmm, why are we not testing it on CI?

kvark (approved):

Napokue:
> Hmm, why are we not testing it on CI?

No idea, we aren't testing any test-data on the CI as far as I know. Maybe something to follow up on in the future?

PR #326, opened 2020-12-11
WGSL updated structs, storage access, and textures

kvark:
Also updates IR for `TypeInner::Struct` and enhances our snapshot tests.
The texture definition changes follow https://github.com/gpuweb/gpuweb/pull/1182, which we recently agreed upon (but needs rebasing to land).


Napokue (approved):
I'm not familiar yet with all theory on why the changes were made, but code-wise it looks good to me.

As far as I understand you introduced in this code:
- Being able to have implicit void return type.
- Distinguish between a block and non-block struct, by introducing the block concept for structs.
- Reduce tokens, as we can now have read/write/readwrite attributes tied to storage textures.

PR #329, opened 2020-12-12
[spv-out] Refactor the load/store logic

kvark:
This PR brings a number of fixes and improvements to the SPIR-V backend.

The goal was to get both the `cube` and `skybox` wgpu-rs examples converted to WGSL with validated SPIR-V output.
The main improvement here is the consistent logic of inserting the Load/Store instructions. There is no code duplication any 
more, and all the meat is in one place (`write_expression_raw`).
Based on #328 . Please review stuff starting with "[spv]".

### DIFF
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -375,13 +375,13 @@ pub(super) fn instruction_variable(
 pub(super) fn instruction_load(
     result_type_id: Word,
     id: Word,
-    pointer_type_id: Word,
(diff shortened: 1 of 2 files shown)


Napokue (approved):
Thanks! This looks good and improves the overall maintainability of the spv-out. So essentially we can control when to write a "pointer" of a "value" expression, which is just a wrapper function around the raw expressions, but the `pointer` expressions add an `OpLoad`. Without having to worry about that loading manually.

PR #331, opened 2020-12-12
[msl] namespace and matrix fixes to get Skybox running

kvark:
Based on #329, this PR gets us to a point where we can generate the MSL shader for the `skybox` example directly from IR 🎉 
(there is only one new commit here)
The things missing were:
  - inserting the namespace qualifier `metal::` before the types (we can discuss this on Matrix if there are concerns)
  - fixing matrix composition and indexing

Update: also got GLSL output working for this! And added it to the snapshot tests.

### DIFF
--- a/src/back/glsl/features.rs
+++ b/src/back/glsl/features.rs
@@ -193,7 +193,7 @@ impl<'a, W> Writer<'a, W> {
             match ty.inner {
                 TypeInner::Scalar { kind, width } => self.scalar_required_features(kind, width),
                 TypeInner::Vector { kind, width, .. } => self.scalar_required_features(kind, width),
-                TypeInner::Matrix { .. } => self.scalar_required_features(ScalarKind::Float, 8),
+                TypeInner::Matrix { width, .. } => self.scalar_required_features(ScalarKind::Float, width),
                 TypeInner::Array { base, .. } => {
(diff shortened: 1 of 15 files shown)


JCapucho (changes requested):
Reviewing the glsl backend, there's just one small thing that should be fixed (was wrong before this PR), everything else looks fine to me

src/back/glsl/features.rs
```diff
@@ -193,7 +193,7 @@ impl<'a, W> Writer<'a, W> {
             match ty.inner {
                 TypeInner::Scalar { kind, width } => self.scalar_required_features(kind, width),
                 TypeInner::Vector { kind, width, .. } => self.scalar_required_features(kind, width),
-                TypeInner::Matrix { .. } => self.scalar_required_features(ScalarKind::Float, 8),
+                TypeInner::Matrix { .. } => self.scalar_required_features(ScalarKind::Float, 4),
```
JCapucho, line 196: This should be
```rust
TypeInner::Matrix {width, .. } => self.scalar_required_features(ScalarKind::Float, width),
```

Napokue (approved):
Reviewed everything outside the GLSL back-end. LGTM!

src/back/glsl/features.rs
kvark: uuh, how did I miss that :) thanks for spotting!

PR #332, opened 2020-12-14
[glsl-in] Cast gl_VertexIndex to SInt

pjoe:
Adds a cast to SInt (`Expression::As`), when `gl_VertexIndex` is used.

Now getting this kind of output:
```
...
         %14 = OpLoad %uint %gl_VertexIndex
         %10 = OpISub %int %14 %int_1
         %17 = OpConvertSToF %float %10            
               OpStore %2 %17 
         %21 = OpLoad %uint %gl_VertexIndex
         %20 = OpBitwiseAnd %int %21 %int_1
         %19 = OpIMul %int %20 %int_2                                                                             
         %18 = OpISub %int %19 %int_1
         %23 = OpConvertSToF %float %18                                                                           
               OpStore %5 %23
         %27 = OpLoad %float %2                                                                                   
         %28 = OpLoad %float %5
         %31 = OpCompositeConstruct %v4float %27 %28 %float_0 %float_1
               OpStore %gl_Position %31
               OpReturn
               OpFunctionEnd 
```

### DIFF
--- a/src/front/glsl/variables.rs
+++ b/src/front/glsl/variables.rs
@@ -24,29 +24,26 @@ impl Program {
                         return Err(ErrorKind::VariableNotAvailable(name.into()));
                     }
                 };
-                let h = self
-                    .module
-                    .global_variables
-                    .append(GlobalVariable {
-                        name: Some(name.into()),
-                        class: if self.shader_stage == ShaderStage::Vertex {
-                            StorageClass::Output
-                        } else {
-                            StorageClass::Input
+                let h = self.module.global_variables.append(GlobalVariable {
+                    name: Some(name.into()),
+                    class: if self.shader_stage == ShaderStage::Vertex {
+                        StorageClass::Output
+                    } else {
+                        StorageClass::Input
+                    },
+                    binding: Some(Binding::BuiltIn(BuiltIn::Position)),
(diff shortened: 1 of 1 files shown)


pjoe:
Relates to #295

src/front/glsl/variables.rs
```diff
@@ -19,6 +19,9 @@ impl Program {
                         return Err(ErrorKind::VariableNotAvailable(name.into()));
                     }
                 };
+                if let Some(global_var) = self.context.lookup_global_var_exps.get(name) {
```
kvark, line 22: there is a bit of a problem here - we have this check both inside the match arms and outside of it (near the end of the function). How about the following instead: we move this check to the *beginning* of the function. Then within match arms we can safely assume that the global is not known. So we can use `append` instead of `fetch_or_append`, and we can add the expressions like we do today.

kvark (changes requested):

src/front/glsl/variables.rs
pjoe: Right, was kind of contemplating this, but wasn't quite sure how it would affect precedence. I'll try moving both this and the local_vars check up front - I think that should work well.

kvark (approved):
Looking good, thank you!

PR #333, opened 2020-12-15
Fix SPIR-V global binding check, add GLES 320 support

kvark:
For SPIR-V - the global variable can be in `Private` storage class, for which there isn't necessarily a binding.
For GLSL - the ES 320 target exists, although we may need to re-think the target version checks in the future anyway.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -67,9 +67,9 @@ mod features;
 mod keywords;
 
 /// List of supported core glsl versions
-const SUPPORTED_CORE_VERSIONS: &[u16] = &[330, 400, 410, 420, 430, 440, 450];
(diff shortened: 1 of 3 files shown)


src/back/glsl/mod.rs
```diff
@@ -287,6 +287,7 @@ impl<'a, W: Write> Writer<'a, W> {
     pub fn new(out: W, module: &'a Module, options: &'a Options) -> Result<Self, Error> {
         // Check if the requested version is supported
         if !options.version.is_supported() {
+            log::error!("Version {:?}", options.version);
```
JCapucho, line 290: You can use `Display` here

JCapucho (approved):
Just touched on glsl part, everything looks fine to me, just one small comment that in no way blocks merging

Issue #336, opened 2020-12-17
Generated SPIR-V structures are missing the offset decorations


kvark:
> VALIDATION [UNASSIGNED-CoreValidation-Shader-InconsistentSpirv (7060244)] : Validation Error: [ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object 0: handle = 0x555557536d20, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x6bbb14 | SPIR-V module not valid: Structure id 19 decorated as Block must be explicitly laid out with Offset decorations.
  %_struct_19 = OpTypeStruct %mat4v4float

This is a blocker for enabling experimental shader support in wgpu-rs examples.

parasyte:
The WGSL parser is also missing offset decorations.

kvark:
The parser now expects spans (only when necessary), see https://github.com/gpuweb/gpuweb/issues/1303

PR #337, opened 2020-12-18
Implement structure member span

kvark:
Gets https://github.com/gpuweb/gpuweb/issues/1303 into our IR.

We used to have support for struct members as built-ins, locations, or offsets. The builtins and locations were added because of the GLSL->SPIR-V weirdness, but I believe we decided to move them out of structs and have the IR not bothered by this. So with these variants removed, the new span is the only thing left.

Most complex parts of this change would have been mapping from SPIR-V and into SPIR-V. The former part is something that will work even if we just skip it. The latter part wasn't implemented at all, i.e. today we aren't properly outputting the offsets - #336. So there was nothing to do with this IR change for me.

I added a simple processor that helps counting sizes of types, so that SPIR-V backend is easier to write.


Napokue (approved):
Looks good to me, thank you! I haven't found anything strange in the code.

src/front/spv/mod.rs
```diff
@@ -2039,15 +2014,13 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                 .future_member_decor
                 .remove(&(id, i))
                 .unwrap_or_default();
-            let origin = decor.get_origin()?;
             members.push(crate::StructMember {
                 name: decor.name,
-                origin,
+                span: None, //TODO
```
kw-mut, line 2019: potential to note here if this might be related to https://github.com/gfx-rs/naga/issues/336 or some other leading instructions, also there are other `span: None` data types not flagged `//TODO` i.e. in changes to `src/front/glsl/parser.rs`

PR #338, opened 2020-12-18
Glsl in builtin functions

pjoe:


kvark (approved):

src/back/spv/writer.rs
```diff
@@ -1277,6 +1277,27 @@ impl Writer {
...
                     }
+                    Mf::Ceil | Mf::Round | Mf::Floor | Mf::Fract | Mf::Trunc => {
+                        let result_lookup_ty = arg0_lookup_ty;
+                        let result_type_id =
+                            self.get_type_id(&ir_module.types, result_lookup_ty)?;
+
+                        block.body.push(super::instructions::instruction_ext_inst(
+                            result_type_id,
+                            id,
+                            1, //TODO get the actual import set id
```
Napokue, line 1288: Can you file a ticket that we can retrieve set ids by import name or something similar, thanks.

Napokue (approved):
LGTM

PR #342, opened 2020-12-20
Tidy up constants IR

kvark:
Another chapter in the series of IR refinements. Previously, we carried a type handle in every constant.
However, this is redundant for the scalar constants mostly. New API is more dense, more correct by constructions. It allows a clear 1:1 between `ScalarValue` and `ScalarKind`.

Note: in some cases it seems that moving the `Bool` variant out of `ScalarValue` into `ConstantInner` would be more convenient, but we'd lose this 1:1 relationship, so I didn't do it.


Napokue (approved):
Very clean change. Reduces a lot of complexity in my opinion.

PR #344, opened 2021-01-04
[spv-out] Handle import id for GLSL.std.450

pjoe:
Small fix for math functions, to properly lookup id for the GLSL.std.450 import


kvark (approved):
Nice, thank you!

PR #345, opened 2021-01-06
WGSL // comments

parasyte:
- Update parser to match the latest spec: https://github.com/gpuweb/gpuweb/pull/1326


kvark (approved):
Very nice, thank you!

Issue #346, opened 2021-01-06
SPIR-V writer panics with local array indexing


parasyte:
I created a minimal test case.

```
$ cat test.wgsl
const foo: array<f32, 3> = array<f32, 3>(1.0, -1.0, 0.0);
fn main() {
    var bar: f32 = foo[1];
}

$ cat test.param.ron
(
    spv_capabilities: [ Shader ],
    mtl_bindings: {},
)

$ RUST_BACKTRACE=1 cargo run --example convert --features=wgsl-in,spv-out -- test.wgsl test.spv
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/examples/convert test.wgsl test.spv`
thread 'main' panicked at 'no entry found for key', src/back/spv/writer.rs:643:41
stack backtrace:
   0: rust_begin_unwind
             at /rustc/0b644e419681835bd0f5871c3bfbd648aa04f157/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/0b644e419681835bd0f5871c3bfbd648aa04f157/library/core/src/panicking.rs:92:14
   2: core::option::expect_failed
             at /rustc/0b644e419681835bd0f5871c3bfbd648aa04f157/library/core/src/option.rs:1260:5
   3: core::option::Option<T>::expect
             at /Users/parasyte/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:349:21
   4: <std::collections::hash::map::HashMap<K,V,S> as core::ops::index::Index<&Q>>::index
             at /Users/parasyte/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/library/std/src/collections/hash/map.rs:1096:9
   5: naga::back::spv::writer::Writer::write_type_declaration_arena
             at ./src/back/spv/writer.rs:643:41
   6: naga::back::spv::writer::Writer::get_type_id
             at ./src/back/spv/writer.rs:228:26
   7: naga::back::spv::writer::Writer::write_constant_type
             at ./src/back/spv/writer.rs:768:31
   8: naga::back::spv::writer::Writer::get_constant_id
             at ./src/back/spv/writer.rs:243:37
   9: naga::back::spv::writer::Writer::write_expression_raw
             at ./src/back/spv/writer.rs:1118:26
  10: naga::back::spv::writer::Writer::write_expression_pointer
[...]

PR #348, opened 2021-01-08
[spv] indexing arrays

kvark:
Opens up the workaround for #346
r? @parasyte


parasyte:
@kvark r+

Thank you! I think this is a good step toward fixing #346.

PR #349, opened 2021-01-08
Unlock arbitrary binding indices in the validator

kvark:


Napokue (approved):
I'm not exactly sure what the changes do, but code LGTM, thank you!

kvark:
Basically, it allows shaders with something like `group(123)` and `binding(43531)`. This makes little sense, but Naga is not the right place to enforce these limits anyway.

PR #354, opened 2021-01-12
[glsl-in] Implement function call

pjoe:
Handle calls to user defined functions

### DIFF
--- a/src/front/glsl/functions.rs
+++ b/src/front/glsl/functions.rs
@@ -108,7 +108,25 @@ impl Program {
                                 .collect(),
                         })
                     }
-                    _ => Err(ErrorKind::NotImplemented("Function call")),
+                    func_name => {
+                        let function = *self
+                            .lookup_function
+                            .get(func_name)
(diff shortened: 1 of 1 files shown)


src/front/glsl/functions.rs
```diff
@@ -108,7 +108,25 @@ impl Program {
...
+                        Ok(ExpressionRule {
+                            expression: self.context.expressions.append(Expression::Call {
+                                function,
+                                arguments: fc.args.iter().map(|a| a.expression).collect(),
+                            }),
+                            sampler: None,
+                            statements: fc
+                                .args
+                                .into_iter()
+                                .map(|a| a.statements)
```
kvark, line 125: probably want `flat_map`?

kvark (approved):

src/front/glsl/functions.rs
pjoe: Ahh right, nice catch. Will fix :)

PR #356, opened 2021-01-15
Change ErrorKind::SemanticError to use Cow<'static, str> instead of &'static str

sotrh:
I was messing around with migrating my build scripts for https://github.com/sotrh/learn-wgpu to use `naga` instead of `shaderc`, and I ran into issues with errors not having enough information. While this PR doesn't fix that issue entirely, it makes `SemanticError`s more flexible as they can now still borrow static strings, and they can be formatted to include additional details such as line numbers and names of variables.

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -72,7 +72,7 @@ impl Program {
             &resolve_ctx,
         ) {
             //TODO: better error report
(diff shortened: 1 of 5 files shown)


kvark (approved):
Thank you, this seems useful!

kvark:
Uuh, something went wrong. The committed suggestion ruined the code? Clippy is angry

PR #360, opened 2021-01-21
[spv] emit proper member layout decorations

### LINKED ISSUE #336: Generated SPIR-V structures are missing the offset decorations

### LINKED ISSUE #350: MatrixStride is not implemented
This looks like a validation error. According to VK spec, all array and matrix types in composite objects must provide explicit layout; https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#ShaderValidation

```
$ cat test.wgsl
[[block]] struct Locals {
    transform: [[stride(64)]] mat4x4<f32>;
};

$ cargo run --example convert --features=wgsl-in -- test.wgsl
// ... snip
            Type {
                name: None,
                inner: Matrix {
                    columns: Quad,
                    rows: Quad,
                    width: 4,
                },
            },
// ... snip
```

The `Array` type does include a stride member... I won't copy-paste it here, but you can see that in the code. https://github.com/gfx-rs/naga/blob/bd998f6c32f94dd66b98f23866ede7ed2878e87b/src/lib.rs#L371-L375

kvark:
Fixes #350 
Fixes #336


examples/convert.rs
```diff
@@ -53,7 +53,12 @@ fn main() {
     let param_path = std::path::PathBuf::from(&args[1]).with_extension("param.ron");
     let params = match fs::read_to_string(param_path) {
         Ok(string) => ron::de::from_str(&string).unwrap(),
-        Err(_) => Parameters::default(),
+        Err(_) => {
+            let mut param = Parameters::default();
+            // very useful to have this by default
+            param.spv_capabilities.insert(spirv::Capability::Shader);
```
Napokue, line 59: Very useful indeed! Maybe something for later to add for all examples by default when a `param.ron` is not found?

Napokue (approved):
LGTM, thanks! Maybe I am not seeing it, but what exactly is the difference in functionality between `proc/sizer.rs` and the new `proc/layouter.rs`?

examples/convert.rs
kvark: yes, I was thinking about this as well, just having a default one in a file. Good for a follow-up!

kvark:
> but what exactly is the difference in functionality between proc/sizer.rs and the new proc/layouter.rs?

The difference is that the old sizer was only computing the sizes, and the new layouter computes both the sizes and the alignments, analogous to https://doc.rust-lang.org/stable/std/alloc/struct.Layout.html

The alignment is necessary, because it affects the layout by introducing pads. For example, a `f32` followed by `vec4<f32>` would need to have a gap, but the old routine wasn't computing this gap (and therefore also mistakenly computing the size of structures).

PR #361, opened 2021-01-22
Rename GlobalUse variants, validate more things, add Collatz

kvark:
Includes #360


tests/snapshots/snapshots__collatz.msl.snap
```diff
@@ -0,0 +1,39 @@
...
+		if (n % static_cast<uint>(2) == static_cast<uint>(0)) {
+			n = n / static_cast<uint>(2);
+		} else {
+			n = static_cast<uint>(3) * n + static_cast<int>(1);
+		}
+		i = i + static_cast<uint>(1);
+	}
+	return i;
+}
+kernel void main1(
```
Napokue, line 33: Just making sure as I have no knowledge about Metal, but is Metal kernel not the same as SPIR-V kernel? If true, my follow up question would be: is this allowed to be used?

src/proc/validator.rs
```diff
@@ -430,14 +473,31 @@ impl Validator {
...
                 .map_err(|error| FunctionError::LocalVariable {
                     handle: var_handle,
                     name: var.name.clone().unwrap_or_default(),
                     error,
                 })?;
         }
+
+        for (index, argument) in fun.arguments.iter().enumerate() {
+            match module.types[argument.ty].inner {
+                crate::TypeInner::Image { .. } | crate::TypeInner::Sampler { .. } => {
```
Napokue, line 491: If I may ask, why are Images and Samplers an invalid argument type?

Napokue (approved):
LGTM, thank you! Only got a couple of non-blocking questions.

tests/snapshots/snapshots__collatz.msl.snap
kvark: Correct, in metal the compute entry points are called "kernel".

src/proc/validator.rs
kvark: Great question! WGSL doesn't support this right now, and in both SPIR-V and GLSL the samplers and images are just globals, so they aren't passed around either. We can come back to this at some point.

PR #363, opened 2021-01-22
[spv-out] Add comparison operators

Napokue:
First PR of the series of PRs to implement flow control. This PR will add comparison operators to enable `==`, `!=`, `<`, `<=`, `>`, `>=` in the near future.

As far as I know, Sint, Uint, and bool are the only comparison operators.

There is maybe only one concern: Floats have two variants for all comparison operators. An ordered and unordered ones. I don't know what is the difference, but I implemented the ordered ones for now. Do we also want to implement the unordered ones?

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1302,6 +1302,44 @@ impl Writer {
                         Some(crate::ScalarKind::Float) => spirv::Op::FMod,
                         _ => unreachable!(),
                     },
+                    crate::BinaryOperator::Equal => match left_ty_inner.scalar_kind() {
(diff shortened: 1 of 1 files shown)


src/back/spv/writer.rs
```diff
@@ -1302,6 +1302,44 @@ impl Writer {
                         Some(crate::ScalarKind::Float) => spirv::Op::FMod,
                         _ => unreachable!(),
                     },
+                    crate::BinaryOperator::Equal => match left_ty_inner.scalar_kind() {
```
kvark, line 1305: since `None` is unreachable, we should just `unwrap()` here - this will also make the match exhaustive

kvark (commented):
Nice!

kvark (approved):
actually, let's get this in!

PR #364, opened 2021-01-23
[wgsl] support unsigned literals

### LINKED ISSUE #196: [wgsl-in] Generating wrong `ConstantInner` variant
Given the following piece of wgsl code:
```js
var index : u32 = gl_GlobalInvocationID.x;
if (index >= 5) {
  return;
}
```
The wgsl frontend will generate `ConstantInner::Sint(5)` instead of `ConstantInner::Uint(5)` this causes issues in backends that don't have implicit conversions, for example the glsl es backend will generate
```c
uint index = gl_GlobalInvocationID[0];
if((index >= 5)) {
	return;
}
```
which is wrong and `glslangValidator` complains about the correct code would be
```c
uint index = gl_GlobalInvocationID[0];
if((index >= 5u)) {
	return;
}
```
hence why generating the right variant is important

kvark:
Fixes #196


src/front/wgsl/lexer.rs
```diff
@@ -13,7 +13,8 @@ fn consume_any(input: &str, what: impl Fn(char) -> bool) -> (&str, &str) {
     input.split_at(pos)
 }
 
-fn consume_number(input: &str) -> (&str, &str) {
+fn consume_number(input: &str) -> (Token, &str) {
+    //Note: I wish this function was simpler and faster...
```
Napokue, line 17: Could you explain why it isn't fast? Simpler I get, but as far as I see it is fast?

Napokue (approved):
LGTM, thank you! Really like that we have support for uint now!

src/front/wgsl/lexer.rs
kvark: We are checking for all these special `-`, `.`, `e`, symbols as we go, and now we are also scanning the string we just processed for the `.` after the fact. And after all that, we'll be doing this again to extract the actual number from the sub-string.

PR #366, opened 2021-01-24
[spv-out] Conditional flow control

Napokue:
Introducing conditional flow control for spv-out, also known as if/else statements. Some notable changes:
- Refactored binary operator parsing, as it always used the result type of either the left or right sided type. However for the comparison operators, this is not true. It always has to return `bool` as result type.
- Statements of the dominator block are getting split when it encounters a branching and resumes its statements after the branching is merged back to the merge block.
- Some optimizations done for the conditional branching. This is probably not something for the back-end and if we follow compiler architecture it should be in the optimizer layer. Optimizations done: When both accept and reject block are empty, ignore the whole conditional branching, when the accept block is empty then branch it to the merge block and when the reject block is empty then remove it from the output.

The following SPIR-V instructions have been implemented:
- OpMemberName
- OpSelectionMerge
- OpLoopMerge
- OpBranch
- OpBranchConditional
- OpFNegate
- OpSNegate
- OpLogicalNot
- OpNot
- OpExclusiveOr
- OpInclusiveOr
- OpLogicalAnd
- OpLogicalOr
- OpShiftLeft
- OpShiftRightArithmetic
- OpShiftRightLogical
- FAbs
- Sabs
- OpCopyObject
- FMin
- SMin
- UMin
- FMax
- SMax
- UMax
- FClamp
- SClamp
- UClamp
- Sin
- Asin
- Cos
- Acos
- Tan
- Atan
- Atan2
- Ceil
- Round
- Floor
- Fract
- Trunc
- Dot
- Cross
- Distance
- Length
- Normalize

### DIFF
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -27,6 +27,14 @@ pub(super) fn instruction_name(target_id: Word, name: &str) -> Instruction {
     instruction
 }
 
+pub(super) fn instruction_member_name(target_id: Word, member: Word, name: &str) -> Instruction {
+    let mut instruction = Instruction::new(Op::MemberName);
+    instruction.add_operand(target_id);
+    instruction.add_operand(member);
+    instruction.add_operands(helpers::string_to_words(name));
+    instruction
+}
+
 //
 // Annotation Instructions
 //
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -656,9 +656,38 @@ pub(super) fn instruction_binary(
 // Control-Flow Instructions
 //
 
+pub(super) fn instruction_selection_merge(
+    id: Word,
+    selection_control: spirv::SelectionControl,
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::SelectionMerge);
+    instruction.add_operand(id);
+    instruction.add_operand(selection_control.bits());
+    instruction
+}
+
 pub(super) fn instruction_label(id: Word) -> Instruction {
     let mut instruction = Instruction::new(Op::Label);
     instruction.set_result(id);
+    instruction
+}
+
+pub(super) fn instruction_branch(id: Word) -> Instruction {
+    let mut instruction = Instruction::new(Op::Branch);
+    instruction.add_operand(id);
+    instruction
+}
+
+// TODO Branch Weights not implemented.
+pub(super) fn instruction_branch_conditional(
+    condition_id: Word,
+    true_label: Word,
+    false_label: Word,
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::BranchConditional);
+    instruction.add_operand(condition_id);
+    instruction.add_operand(true_label);
+    instruction.add_operand(false_label);
     instruction
 }
 
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -73,10 +73,10 @@ pub(super) fn instruction_ext_inst_import(id: Word, name: &str) -> Instruction {
 }
 
 pub(super) fn instruction_ext_inst(
+    set_id: Word,
     op: spirv::GLOp,
     result_type_id: Word,
     id: Word,
-    set_id: Word,
     operands: &[Word],
 ) -> Instruction {
     let mut instruction = Instruction::new(Op::ExtInst);
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -657,11 +657,23 @@ pub(super) fn instruction_binary(
 //
 
 pub(super) fn instruction_selection_merge(
-    id: Word,
+    merge_id: Word,
     selection_control: spirv::SelectionControl,
 ) -> Instruction {
     let mut instruction = Instruction::new(Op::SelectionMerge);
-    instruction.add_operand(id);
+    instruction.add_operand(merge_id);
+    instruction.add_operand(selection_control.bits());
+    instruction
+}
+
+pub(super) fn instruction_loop_merge(
+    merge_id: Word,
+    continuing_id: Word,
+    selection_control: spirv::SelectionControl,
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::LoopMerge);
+    instruction.add_operand(merge_id);
+    instruction.add_operand(continuing_id);
     instruction.add_operand(selection_control.bits());
     instruction
 }
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -73,17 +73,17 @@ pub(super) fn instruction_ext_inst_import(id: Word, name: &str) -> Instruction {
 }
 
 pub(super) fn instruction_ext_inst(
+    op: spirv::GLOp,
     result_type_id: Word,
     id: Word,
     set_id: Word,
-    ext_instruction: u32,
     operands: &[Word],
 ) -> Instruction {
     let mut instruction = Instruction::new(Op::ExtInst);
     instruction.set_type(result_type_id);
     instruction.set_result(id);
     instruction.add_operand(set_id);
-    instruction.add_operand(ext_instruction);
+    instruction.add_operand(op as u32);
     for operand in operands {
         instruction.add_operand(*operand)
     }
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -16,9 +16,19 @@ pub enum Error {
 }
 
 struct Block {
-    label_id: spirv::Word,
+    label_id: Word,
     body: Vec<Instruction>,
     termination: Option<Instruction>,
+}
+
+impl Block {
+    fn new(label_id: Word) -> Self {
+        Block {
+            label_id,
+            body: Vec::new(),
+            termination: None,
+        }
+    }
 }
 
 struct LocalVariable {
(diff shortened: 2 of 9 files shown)


src/back/spv/writer.rs
```diff
@@ -1349,11 +1358,31 @@ impl Writer {
                     }
                 };
 
-                let (result_type_id, result_lookup_ty) = if result_side_left {
-                    (left_result_type_id, left_lookup_ty)
+                let is_comparison = match op {
```
kvark, line 1361: This can be made for the follow up, but ideally we'd not have 2 matches of the the same thing. We should combine the matches. i.e. this would look like:
```rust
enum BinaryResult {
  Left,
  Right,
  Bool,
}

let (spirv_op, bin_result) = match op {
  ...
};

let (result_type_id, result_lookup_ty) = match bin_result {
...
};
```

src/back/spv/writer.rs
```diff
@@ -1349,11 +1358,31 @@ impl Writer {
...
+                    | crate::BinaryOperator::NotEqual
+                    | crate::BinaryOperator::Less
+                    | crate::BinaryOperator::LessEqual
+                    | crate::BinaryOperator::Greater
+                    | crate::BinaryOperator::GreaterEqual => true,
+                    _ => false,
+                };
+
+                let (result_type_id, result_lookup_ty) = if is_comparison {
+                    let local_ty = LookupType::Local(LocalType::Scalar {
```
kvark, line 1372: This would only work for scalars. I think we want IR to also support vector comparisons, i.e.
```
const xx: vec2<bool> = vec2<f32>(1.0, 1.0) == vec2<f32>(0.0, 1.0);
```

src/back/spv/writer.rs
```diff
@@ -407,7 +407,16 @@ impl Writer {
             function_type,
         ));
 
-        self.write_block(&ir_function.body, ir_module, ir_function, &mut function)?;
+        let id = self.generate_id();
+        let mut block = self.write_block(
+            &mut ir_function.body.to_vec(),
```
kvark, line 412: That seems wrong. We are passing down a mutable reference to a temporary vector that we aren't using anyway?

src/back/spv/writer.rs
```diff
@@ -1612,23 +1641,82 @@ impl Writer {
...
-    ) -> Result<spirv::Word, Error> {
+        branch_id: Option<spirv::Word>,
+    ) -> Result<Block, Error> {
         let mut block = Block::default();
-        let id = self.generate_id();
-        block.label = Some(super::instructions::instruction_label(id));
 
-        for statement in statements {
-            match *statement {
+        while !statements.is_empty() {
```
kvark, line 1652: currently, I'd like to understand why this needs to be a vector we are mutating, before getting to the rest

src/back/spv/instructions.rs
```diff
@@ -604,12 +604,41 @@ pub(super) fn instruction_binary(
 // Control-Flow Instructions
 //
 
+pub(super) fn instruction_selection_merge(
```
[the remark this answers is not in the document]
Napokue: Yes I did actually consider this at first. However there was and is at least one case I can remember that doesn't work, without prefixing it. That is `OpReturn`. If we want to change it, it will be called `return`, which is a reserved keyword.

Other than that, at the moment there are no keywords (at the moment) that conflict with Rust.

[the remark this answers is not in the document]
kvark: ok, but that's just naming. We can call it `ret()` fwiw, it's fine. It's all internal API anyway.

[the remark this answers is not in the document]
Napokue: Yes we can indeed, I can do this in a follow up issue. Let me add it to my todo list, so I can create the issue at a later time.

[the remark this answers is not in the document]
kvark: Awesome, thank you! As long as it's on our radar, all is well :)

src/back/spv/writer.rs
Napokue: Agreed, will note it in my todo list and create an issue at a later time.

src/back/spv/writer.rs
Napokue: True, I think at least, never used an expression like that! Will add it to my todo to create an issue for it at a later time.

src/back/spv/writer.rs
Napokue: Yes, I understand and I am not entirely sure if this is the right way code wise and memory wise. What I want to do is iterating through the statements while it is not empty. 

When does it become empty? By either organically iterating over it and removing the first item (like a dequeue, see line 1655), or via a branching.

When we reach branching, we want to essentially "move" the remaining statements to the post dominator/merge block (block after the branching), and continue to process the remaining statements.

As this is a recursive function, the first function call will be the first block and when the branching and merging is finished, it doesn't have any statements left (as it is drained via the `create_post_dominate_block()` function.

Like I said, I am not sure if this is the right way, hence I think this is not a good way, but this was the first solution that came into my mind, without considering memory footprints etc.

kvark: I'm not sure I understand why anything needs to be moved. Let's say you are processing a statement somewhere. In the context for this, you need to know the label ID you are in, and the label ID where you'd need to merge to. If you encounter an if/else statement, you'd create those new block labels, call yourself recursively for true/false statements (in which case you know exactly what the merge point is, and what the parent ID is). Then you'd just continue as usual through the rest of the statements, just with the "my block ID" pointing to the new label you created for the stuff after if/else. So it's supposed to be just a routine that recursively processes the statements and spawns the necessary blocks. The only little unspecified bit here is who is responsible for creation of which block labels, but the answer to that will come naturally as you are implementing the code.

Am I terribly missing something here?

src/back/spv/writer.rs
```diff
@@ -1610,24 +1635,66 @@ impl Writer {
...
+                        condition_id,
+                        accept_block.label_id.unwrap(),
+                        reject_block.label_id.unwrap(),
+                    ));
+                    function.blocks.push(block);
+                    function.blocks.push(accept_block);
+                    function.blocks.push(reject_block);
+
+                    let pdom_block = self.create_block();
+                    block = pdom_block;
```
Napokue, line 1697: Remaining issues are: 
- How do we add the `pdom_block` to the `function.blocks` vector?
- How do we add the main block to the `function.block` vector when not having branching statements?

src/back/spv/writer.rs
```diff
@@ -1610,24 +1635,66 @@ impl Writer {
...
-        statements: &[crate::Statement],
+        statements: &crate::Block,
         ir_module: &crate::Module,
         ir_function: &crate::Function,
         function: &mut Function,
-    ) -> Result<spirv::Word, Error> {
-        let mut block = Block::default();
-        let id = self.generate_id();
-        block.label = Some(super::instructions::instruction_label(id));
+        branch_id: Option<spirv::Word>,
```
kvark, line 1653: should this be called `merge_id`?

src/back/spv/writer.rs
```diff
@@ -1666,8 +1733,11 @@ impl Writer {
             }
         }
 
-        function.blocks.push(block);
-        Ok(id)
+        if branch_id.is_some() {
+            block.termination = Some(super::instructions::instruction_branch(branch_id.unwrap()));
```
kvark, line 1737: note: we don't want to do this if there already a `termination` assigned, since it could be `return`

src/back/spv/writer.rs
kvark: so if it's None, means we are in the root block, correct? I wonder if we should make it non-optional then. We'd just make `block.termination = Return` for the root block (before calling `write_block`), essentially blocking the effect of the branch ID.

kvark (commented):
Looks awesome!

src/back/spv/writer.rs
Napokue: Yes, `merge_id` would be a better name.

To follow up on your second comment: No I don't think so. As Blocks always require a termination (following the spec), essentially it always has to have either a `merge_id` or a termination instruction from other terminations (only case is `Return` that I can think of right now, that is a statement).

src/back/spv/writer.rs
Napokue: Yes true, I think we don't even need to check here if it `is_some`. As it is required to have a termination. So it either has to have a termination already from the statements, otherwise via the `merge_id` parameter.

```rust
  if block.termination.is_none() {
    block.termination = Some(super::instructions::instruction_branch(branch_id.unwrap()));
  }
```

src/back/spv/writer.rs
```diff
@@ -1666,8 +1745,11 @@ impl Writer {
             }
         }
 
-        function.blocks.push(block);
-        Ok(id)
+        if block.termination.is_none() {
+            block.termination = Some(super::instructions::instruction_branch(merge_id.unwrap()));
```
kvark, line 1749: does it make sense to keep `merge_id` optional if we are unwrapping it anyway?

Napokue: Yes, because the "last" root block won't have a `merge_id`, as it should have a return. We start this tail recursion with `None` passed in from the root block. I don't think we can make it non optional, as I don't know what to pass in to start the recursion.

kvark: You can pass 0, technically. Since you are writing the backend, you should assume that the module is valid.
In a valid module, there is a `return` at the end of a function, including the entry point functions. And with `return` comes `block.termination`, so the `merge_id == 0` would never be used.

kvark: it's totally up to you though. I'm just pointing out that we are never actually making sense (or use) of `None` here, so we might as well not have it

kvark (approved):

PR #373, opened 2021-01-26
Slight improvements to whitespace formatting

grovesNL:


src/back/msl/writer.rs
```diff
@@ -1034,14 +1038,14 @@ impl<W: Write> Writer<W> {
 
             match stage {
                 crate::ShaderStage::Vertex | crate::ShaderStage::Fragment => {
-                    writeln!(self.out, "\t{} {};", output_name, OUTPUT_STRUCT_NAME)?;
```
kvark, line 1037: what's the motivation for changing these? I thought '\t` would be better since it consumes less symbols and achieves mostly the same result.

kvark (commented):
Great stuff! Just one concern here.

src/back/msl/writer.rs
grovesNL: I don’t have a strong preference here, but it seems like most codebases have standardized on 4 spaces, so it would be nice to be symmetric with the front ends and not break any default linter rules/auto-formatting.

Another minor issue with tab characters is that generated shaders/diffs will be displayed differently depending on the environment (some treat tabs as 4 spaces, some as 6, 8, etc.). So standardizing on 4 spaces could slightly improve readability in some environments if the generated shaders are being viewed (especially diffs).

Either seems ok to me though so we can change it back to tabs if that’s preferred for backends.

kvark: Trying to make linters happy would be a fairly involving task, and it's probably not worth it. I.e. cbindgen just runs rustfmt as the final step, and we should expect the same from anybody who wants to vendor the generated sources.

The point about tabs being displayed differently plays well for the multi-platform code that's developed in collaboration, and this is where spaces got dominant. In our case, the produced sources are neither multi-platform, or collaborated on, so these points hardly apply.

At the same time, trying to save a few cycles on passing around spaces instead of tabs may also seem very silly :)
I took a simple measurement on the last `shadow.wgsl` I wrote, and converting spaces to tabs saved about 5% of the total byte size. Seems pretty measurable?

No matter which one we choose, could we at least have it so it's easy to switch? I.e. the backend would have some `const INDENT: &str = "  "` or something like this.

grovesNL: Yeah I think either is alright :) I do see SPIRV-Cross output being committed to repositories occasionally (without an additional formatting pass), so it still might be nice to have good defaults where possible.

if we're primarily interested in readability/debuggability, I'd probably prefer to keep the output closer to what people would tend to write normally (except when it would cause problems for us, like Metal namespaces).  On the other hand, if we're trying to reduce byte size then we might want to consider avoiding all unnecessary whitespace, identifier mangling, etc.

`const INDENT` sounds good to me too :+1: I'll make that change

kvark (approved):
Great!

PR #375, opened 2021-01-26
[wgsl] fix and update storage load

kvark:
The old implementation was the *opposite* of what we need: it only read the index for storages, and expected an identifier.


Napokue (approved):
LGTM!

PR #379, opened 2021-01-27
[spv] implement full image sampling

kvark:
Adds support for all the image sampling modifiers, including the array index selection (most complex part here), bias, depth reference, etc.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -243,8 +243,10 @@ impl Writer {
         if let Entry::Occupied(e) = self.lookup_constant.entry(handle) {
             Ok(*e.get())
         } else {
-            let (instruction, id) = self.write_constant_type(handle, ir_module)?;
-            instruction.to_words(&mut self.logical_layout.declarations);
+            let id = self.generate_id();
+            self.lookup_constant.insert(handle, id);
+            let inner = &ir_module.constants[handle].inner;
+            self.write_constant_type(id, inner, ir_module)?;
             Ok(id)
         }
     }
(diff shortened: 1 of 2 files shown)


src/back/spv/writer.rs
```diff
@@ -1617,18 +1690,118 @@ impl Writer {
...
+                            coordinate_id,
+                            depth_id,
+                        );
+
+                        //TODO: cache this!
+                        let zero_id = self.generate_id();
+                        let zero_inner = crate::ConstantInner::Scalar {
+                            width: 4,
+                            value: crate::ScalarValue::Float(0.0),
+                        };
```
Napokue, line 1718: Can't we use [`get_type_id()`](https://github.com/gfx-rs/naga/blob/master/src/back/spv/writer.rs#L215) for this? You should be able to request the scalar, even though it isn't in the IR. If it doesn't exist, it will add it to [`lookup_type`](https://github.com/gfx-rs/naga/blob/master/src/back/spv/writer.rs#L215).

Napokue: I know, I posted this already, and understand the difference. This mechanic is not implemented indeed for consts. I believe I have looked at it a couple of weeks/months ago, but there was a reason why I couldn't get it to work.

Napokue (approved):
LGTM, thank you!

src/back/spv/writer.rs
kvark: Yeah, this is not a type ID, this is a constant. I think we can easily follow-up with the logic of caching these, but for the purposes of this PR, the correctness is more important than optimization.

kvark (commented):
Thank you for the review!

PR #381, opened 2021-01-27
Improve WGSL error reporting

kvark:
Here is how it looks with the PR:
```
[2021-01-27T15:57:52Z ERROR naga::front::wgsl] |	fn vs_bake() {
[2021-01-27T15:57:52Z ERROR naga::front::wgsl] >	    out_position = u_globals.view_proj 1* u_entity.world * vec4<f32>(in_position);
[2021-01-27T15:57:52Z ERROR naga::front::wgsl] |	                                        ^
thread 'main' panicked at 'error while parsing WGSL in scopes [FunctionDecl, Block, Statement] at line 26 pos 40: unexpected token Number { value: "1", ty: 'i', width: "" }, expected separator', examples/convert.rs:54:31
```
Here is how it looks on master right now:
```
[2021-01-27T15:57:34Z ERROR naga::front::wgsl] Unmet expectation for Separator(';')
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseError { error: Unexpected(Number { value: "1", ty: 'i', width: "" }), scopes: [Function```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,9 +33,12 @@ spv-in = ["petgraph", "spirv"]
 spv-out = ["spirv"]
 wgsl-in = []
 
+#Note: it would be nice to do `cargo run` for conversion.
+# Blocked on https://github.com/rust-lang/cargo/issues/4663
+
 [dev-dependencies]
 difference = "2.0"
-env_logger = "0.6"
+env_logger = "0.8"
 ron = "0.6"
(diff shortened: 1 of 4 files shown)


grovesNL (approved):
Nice, this is a great baseline! I think the codespan errors would mostly shine when we start adding some contextual errors into the parser

PR #382, opened 2021-01-27
[wgsl] support simple block statements

kvark:
The code is mostly a copy-paste from `parse_block`, but sharing it would be more trouble than it's worth.


Napokue (approved):
Thank you!

Issue #390, opened 2021-01-30
Aggregate interfaces of called functions


kvark:
When an entry point function A calls helper function B, we need to ensure the the interface of A contains the interface of B.

In addition, Metal backend would require the resources to be passed around in parameters to helper functions.

PR #391, opened 2021-01-30
Assorted MSL fixes

kvark:
MSL:
  - actually implements `continuing` loop statement. Refactors the Namer API to aid that, and reuses the allocation within the writer.
  - fixes scoping of binary expressions
  - fixes unsigned constants
  - fixes ">" operation
  - fixes `sample_compare` calls to avoid `level(0)` option
  - fixes `return` without a value in the middle of functions

In the end, it allows wgpu-rs to run the `boids` example with pure Naga translation, not involving SPIR-V on Metal 🎉 
Unfortunately, the `shadow` example got blocked on #390. It seems good otherwise.

For testing, the `boids.wgsl` is updated, now only contains the compute part. The `shadow.wgsl` is added, it demonstrates quite a few difficulties:
  - depth sampling
  - using a resource only within a helper function (not supported by our `Interface` processor yet!)
  - "fat" constant, which we really need to process differently. Currently the value is repeated all over the place.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -305,7 +305,7 @@ impl<'a, W: Write> Writer<'a, W> {
 
         // Generate a map with names required to write the module
         let mut names = FastHashMap::default();
-        Namer::process(module, keywords::RESERVED_KEYWORDS, &mut names);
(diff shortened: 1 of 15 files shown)


Napokue (approved):
LGTM, thank you!

PR #394, opened 2021-01-30
[glsl-in] Add support for function declarations with arguments

JCapucho:
Also made ParseError public and changed global declarations to default to private instead of throwing error


kvark (approved):
Nicely done!

PR #395, opened 2021-01-30
[glsl-in] Fix double constant lexing and lex uint constants

JCapucho:
Fixes double constants (ending with "lf") being treated as floats and added support for uint constants (ending with "u")

### DIFF
--- a/src/front/glsl/lex.rs
+++ b/src/front/glsl/lex.rs
@@ -93,8 +93,20 @@ impl<'a> Lexer<'a> {
                 }
             }
             '0'..='9' => {
-                let (number, remainder, pos) =
-                    consume_any(input, |c| (('0'..='9').contains(&c) || c == '.'));
+                let next = chars.next().and_then(|c| c.to_lowercase().next());
+
+                let hexadecimal = cur == '0' && next.map_or(false, |c| c == 'x');
+
+                let (number, remainder, pos) = if hexadecimal {
+                    consume_any(&input[2..], |c| {
+                        ('0'..='9').contains(&c)
+                            || ('a'..='f').contains(&c)
+                            || ('A'..='F').contains(&c)
+                    })
+                } else {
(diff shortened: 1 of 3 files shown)


src/front/glsl/lex.rs
```diff
@@ -93,13 +93,12 @@ impl<'a> Lexer<'a> {
...
+                let (number, remainder, pos) =
                     consume_any(input, |c| (('0'..='9').contains(&c) || c == '.'));
                 if number.find('.').is_some() {
-                    if (
-                        chars.next().map(|c| c.to_lowercase().next().unwrap()),
-                        chars.next().map(|c| c.to_lowercase().next().unwrap()),
-                    ) == (Some('l'), Some('f'))
+                    if remainder
+                        .get(..2)
+                        .map_or(false, |c| &c.to_lowercase() == "lf")
```
kvark, line 101: nit: seems unfortunate that we are calling `to_lowercase` multiple times, since it involves a string allocation on the heap. Ideally, we should not call it even once, since we are just comparing slices. I.e. could this be something like
```rust
let mut remainder_chars = remainder.chars();
let first = remainder_chars.next();
let second = remainder_chars.next();
match (first, second) {
  (Some(a), Some(b)) if a.to_lowercase() == 'l' && b.to_lowercase == 'f' => {..}
  ...
}
```

src/front/glsl/lex.rs
```diff
@@ -93,21 +93,50 @@ impl<'a> Lexer<'a> {
                 }
             }
             '0'..='9' => {
-                let (number, _, pos) =
-                    consume_any(input, |c| (('0'..='9').contains(&c) || c == '.'));
+                let next = chars.next().and_then(|c| c.to_lowercase().next());
+
+                let hexadecimal = cur == '0' && next.map_or(false, |c| c == 'x');
```
kvark, line 98: nit: could also do `next == Some('x')`

kvark (approved):
Great!

PR #397, opened 2021-01-30
Update emoji in README

parasyte:
The checkmarks used were a bit confusing. This improves the visual prominence between primary and secondary support (IMHO)


kvark (approved):
Thank you!

PR #398, opened 2021-01-31
Release 0.3.1 with a bunch of SPIRV and WGSL fixes

kvark:


Napokue (approved):
LGTM, thank you! As far as I see this also fixes the duplicate constants in spv-out?

kvark:
I don't think it fixes the duplicate constants, no. We are still generating more constants in cases where there are no associated handles in the IR.
I think I can rework the last commit a bit nicer.

kvark:
Confirmed this fixes our `boids` example! :tada:

PR #399, opened 2021-01-31
[glsl-in] Populate ResolveContext with functions and add more math functions

JCapucho:
Populates the `ResolveContext` with the module functions, otherwise the `Typifier` would panic.

Adds support for `sin`, `abs` and `mix`


kvark (approved):
Nice!

PR #400, opened 2021-01-31
[glsl-in] Added support for initalizers and constants

JCapucho:
Adds another processor to translate expressions to constants if possible

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -1,7 +1,7 @@
 use super::error::ErrorKind;
 use crate::{
-    proc::{ResolveContext, Typifier},
-    Arena, BinaryOperator, Binding, Expression, FastHashMap, Function, FunctionArgument,
+    proc::{ConstantSolver, ResolveContext, Typifier},
+    Arena, BinaryOperator, Binding, Constant, Expression, FastHashMap, Function, FunctionArgument,
     GlobalVariable, Handle, Interpolation, LocalVariable, Module, ShaderStage, Statement,
     StorageClass, Type,
 };
@@ -15,6 +15,7 @@ pub struct Program {
     pub lookup_function: FastHashMap<String, Handle<Function>>,
     pub lookup_type: FastHashMap<String, Handle<Type>>,
     pub lookup_global_variables: FastHashMap<String, Handle<GlobalVariable>>,
+    pub lookup_constants: FastHashMap<String, Handle<Constant>>,
     pub context: Context,
     pub module: Module,
 }
@@ -29,12 +30,14 @@ impl Program {
             lookup_function: FastHashMap::default(),
             lookup_type: FastHashMap::default(),
             lookup_global_variables: FastHashMap::default(),
+            lookup_constants: FastHashMap::default(),
             context: Context {
                 expressions: Arena::<Expression>::new(),
                 local_variables: Arena::<LocalVariable>::new(),
                 arguments: Vec::new(),
                 scopes: vec![FastHashMap::default()],
                 lookup_global_var_exps: FastHashMap::default(),
+                lookup_constant_exps: FastHashMap::default(),
                 typifier: Typifier::new(),
             },
             module: Module::generate_empty(),
@@ -77,6 +80,21 @@ impl Program {
             Ok(()) => Ok(self.context.typifier.get(handle, &self.module.types)),
         }
     }
+
+    pub fn solve_constant(
+        &mut self,
+        root: Handle<Expression>,
+    ) -> Result<Handle<Constant>, ErrorKind> {
+        let mut solver = ConstantSolver {
+            types: &self.module.types,
+            expressions: &self.context.expressions,
+            constants: &mut self.module.constants,
+        };
+
+        solver
+            .solve(root)
+            .map_err(|_| ErrorKind::SemanticError("Can't solve constant".into()))
+    }
 }
 
 #[derive(Debug)]
@@ -92,6 +110,7 @@ pub struct Context {
     //TODO: Find less allocation heavy representation
     pub scopes: Vec<FastHashMap<String, Handle<Expression>>>,
     pub lookup_global_var_exps: FastHashMap<String, Handle<Expression>>,
+    pub lookup_constant_exps: FastHashMap<String, Handle<Expression>>,
     pub typifier: Typifier,
 }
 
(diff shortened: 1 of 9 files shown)


src/front/glsl/ast.rs
```diff
@@ -178,3 +197,9 @@ pub struct FunctionCall {
     pub kind: FunctionCallKind,
     pub args: Vec<ExpressionRule>,
 }
+
+#[derive(Debug, Clone, Copy)]
+pub enum StorageQualifier {
+    StorageClass(StorageClass),
+    Const,
```
kvark, line 204: You reminded me to file something up about this - https://github.com/gpuweb/gpuweb/issues/1394

kvark (approved):
Looks solid!

src/proc/constants.rs
```diff
@@ -1,11 +1,12 @@
...
     arena::{Arena, Handle},
-    ArraySize, Constant, ConstantInner, Expression, ScalarKind, ScalarValue, Type, UnaryOperator,
+    ArraySize, Constant, ConstantInner, Expression, ScalarKind, ScalarValue, Type, TypeInner,
+    UnaryOperator,
 };
 
 #[derive(Debug)]
 pub struct ConstantSolver<'a> {
-    pub types: &'a Arena<Type>,
+    pub types: &'a mut Arena<Type>,
```
kvark, line 9: This is concerning. If the goal of this processor is to solve constants, it shouldn't mutate the types.

src/proc/constants.rs
```diff
@@ -142,7 +143,33 @@ impl<'a> ConstantSolver<'a> {
                         Ok(components[index])
                     }
                 }
-                crate::TypeInner::Matrix { .. } => todo!(),
+                crate::TypeInner::Matrix {
```
kvark, line 146: why is that `match self.types[ty].inner` needed at all? If the constant is `Composite`, and you need to access the `index`, you should be able to just get `components[index]` and be done with it, regardless of the type.

src/proc/constants.rs
JCapucho: I changed this in the last commit because matrix access produces vectors which is a composite and requires a type handle that might need to be created

src/proc/constants.rs
JCapucho: This match serves 3 purposes:
1. A sanity check that the type can be accessed
2. Allows us to extract information to bound check so that we don't panic (the `ConstantSolver` doesn't rely on validation having ran, I can change this if you want)
3. Most importantly, matrix access indexes an entire slice and not just one element

src/proc/constants.rs
```diff
@@ -0,0 +1,564 @@
...
+                    rows,
+                    columns,
+                    width,
+                } => {
+                    if columns as usize <= index {
+                        return Err(ConstantSolvingError::InvalidAccessIndex);
+                    }
+
+                    let components =
+                        components[(index * rows as usize)..((index + 1) * rows as usize)].to_vec();
```
kvark, line 165: This might be a problem. The composite components for a matrix in our IR should be vectors and only vectors.
But this code works with components as if they were scalars. I assume that's because when implementing matrix construction in GLSL it was easier to just fill out the components vector right away? The IR will fail validation on that once the validation matures a bit.

src/proc/constants.rs
```diff
@@ -0,0 +1,564 @@
...
+                            Err(ConstantSolvingError::InvalidAccessIndex)
+                        } else {
+                            Ok(components[index])
+                        }
+                    }
+                    ArraySize::Dynamic => Err(ConstantSolvingError::ArrayLengthDynamic),
+                },
+                crate::TypeInner::Struct { ref members, .. } => {
+                    if members.len() <= index {
+                        Err(ConstantSolvingError::InvalidAccessIndex)
```
kvark, line 196: nit: you can do a fancy `components.get(index).map_or(ConstantSolvingError::InvalidAccessIndex).cloned()` here to avoid bounds-checking twice

src/lib.rs
```diff
@@ -412,7 +412,7 @@ pub enum ScalarValue {
 }
 
 /// Additional information, dependendent on the kind of constant.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
```
kvark, line 415: why does this need a `Clone`?

src/proc/constants.rs
```diff
@@ -0,0 +1,573 @@
...
+            .unwrap()
+        }
+
+        let mut inner = self.constants[constant].inner.clone();
+
+        match inner {
+            ConstantInner::Scalar { ref mut value, .. } => {
+                let intial = value.clone();
+
+                match kind {
```
kvark, line 237: let's move the assignment out of the match:
```rust
*value = match kind {
};
```

kvark (commented):
Looks great! Just one correction needed, plus a few minor things.

src/proc/constants.rs
JCapucho: I did this way because I thought they were supposed to be flattened should have asked

src/lib.rs
JCapucho: I made it `Clone` for convenience so that for example in `cast` we don't need to redefine a `ConstantInner` in each match arm

kvark: where is that code that uses it?

JCapucho: Lines 179 and 219.
For example `unary_op` has
```rust
ConstantInner::Composite {
    ty,
    ref mut components,
} => {
    match self.types[ty].inner {
        crate::TypeInner::Vector { .. } | crate::TypeInner::Matrix { .. } => (),
        _ => return Err(ConstantSolvingError::InvalidCastArg),
    }

    for component in components {
        *component = self.unary_op(op, *component)?
    }
}
```

Instead we would need
```rust
ConstantInner::Composite {
    ty,
    ref components,
} => {
    match self.types[ty].inner {
        crate::TypeInner::Vector { .. } | crate::TypeInner::Matrix { .. } => (),
        _ => return Err(ConstantSolvingError::InvalidCastArg),
    }

   let mut new_components = Vec::new();

    for component in components {
        let new_component = self.unary_op(op, *component)?
        new_components.push(new_component)
    }
    
    ConstantInner::Composite {
       ty,
       components: new_components
    }
}
```

kvark: Yeah, I think it would be worth it. The code clearly becomes more functional, less imperative, less mutating, etc.
It's an improvement, and it's not that much more code to write.

PR #401, opened 2021-02-01
[spv-out] Add missing math functions

lain-dono:
Implementation of `sinh`, `cosh`, `tanh`, `mod`, `frexp`, `ldexp`, `faceForward`, `exp`, `exp2`, `log`, `log2`, `sign`, `fma`, `mix`, `step`, `smoothstep`, `sqrt`, `determintant`, `reflect`.

At least `mod`, `frexp`, `ldexp`, `determintant` require some checking.

Perhaps the [specification](https://gpuweb.github.io/gpuweb/wgsl.html) missed the `refract` function.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1596,10 +1596,13 @@ impl Writer {
                     }
                     // trigonometry
                     Mf::Sin => MathOp::Single(spirv::GLOp::Sin),
+                    Mf::Sinh => MathOp::Single(spirv::GLOp::Sinh),
                     Mf::Asin => MathOp::Single(spirv::GLOp::Asin),
                     Mf::Cos => MathOp::Single(spirv::GLOp::Cos),
+                    Mf::Cosh => MathOp::Single(spirv::GLOp::Cosh),
                     Mf::Acos => MathOp::Single(spirv::GLOp::Acos),
                     Mf::Tan => MathOp::Single(spirv::GLOp::Tan),
+                    Mf::Tanh => MathOp::Single(spirv::GLOp::Tanh),
                     Mf::Atan => MathOp::Single(spirv::GLOp::Atan),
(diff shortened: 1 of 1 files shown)


src/back/spv/writer.rs
```diff
@@ -1680,8 +1686,43 @@ impl Writer {
...
+                        MathOp::Single(op)
+                    }
+                    Mf::Fma => MathOp::Triple(spirv::GLOp::Fma),
+                    Mf::Mix => {
+                        let op = match self
+                            .get_type_inner(&ir_module.types, arg0_lookup_ty)
+                            .scalar_kind()
+                        {
+                            Some(crate::ScalarKind::Float) => spirv::GLOp::FMix,
+                            Some(crate::ScalarKind::Sint) => spirv::GLOp::IMix,
```
kvark, line 1716: There is a problem here. Looks like `IMix` is not in the extended instruction set - https://www.khronos.org/registry/spir-v/specs/1.0/GLSL.std.450.html
So for integers we'll have to actually encode the arithmetic expressions in here

src/back/spv/writer.rs
```diff
@@ -1707,7 +1748,31 @@ impl Writer {
...
+                            self.gl450_ext_inst_id,
+                            spirv::GLOp::Determinant,
+                            result_type_id,
+                            id,
+                            &[arg0_id],
+                        );
+                        MathOp::Other(inst, result_lookup_ty)
+                    }
+
+                    Mf::Outer | Mf::ReverseBits | Mf::CountOneBits => {
```
kvark, line 1775: amazing, love the exhaustive enums!

kvark (commented):
Looks great, just one small concern here.

src/back/spv/writer.rs
lain-dono: Done

kvark (approved):
Thank you!

PR #408, opened 2021-02-02
Start hlsl-out

Napokue:
Start implementation of hlsl-out. This will cover only the empty case. When this is landed a simple empty entrypoint can be generated via the hlsl-out path.

### DIFF
--- a/src/back/hlsl/mod.rs
+++ b/src/back/hlsl/mod.rs
@@ -1,0 +1,23 @@
+mod keywords;
+mod writer;
+
+use std::io::Error as IoError;
+use std::string::FromUtf8Error;
+use thiserror::Error;
+
+pub use writer::Writer;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    IoError(#[from] IoError),
+    #[error(transparent)]
+    Utf8(#[from] FromUtf8Error),
+}
+
+pub fn write_string(module: &crate::Module) -> Result<String, Error> {
+    let mut w = Writer::new(Vec::new());
+    w.write(module)?;
+    let output = String::from_utf8(w.finish())?;
+    Ok(output)
+}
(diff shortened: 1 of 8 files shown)


src/back/hlsl/mod.rs
```diff
@@ -0,0 +1,22 @@
+mod writer;
+
+use std::string::FromUtf8Error;
+pub use writer::Writer;
+
+#[derive(Debug)]
+pub enum Error {
+    Utf8(FromUtf8Error),
```
kvark, line 8: you may use `thiserror` from the start as the better practice

src/back/hlsl/writer.rs
```diff
@@ -0,0 +1,20 @@
+use crate::back::hlsl::Error;
```
kvark, line 1: I don't think any backend should include anything from either `crate::back` or `crate::front`. Instead, let's do `use super::Error`.

Napokue: Yes agreed, probably an auto complete function in my IDE that auto imports.

Napokue:
The scope of just naming this PR: "Implementation hlsl-out" was a bit too ambitious for me to undertake. I have down scaled the scope of this PR, and this PR will add the bare minimum to at least enable the hlsl-out path.

Napokue:
The PR should be ready, the only thing I don't know how to do yet is plugging this into our test framework with the snapshots.

kvark:
Please make CI happy :)
I assume it's fine to squash it on merge, right?
> only thing I don't know how to do yet is plugging this into our test framework with the snapshots.

You can just follow MSL and GLSL in the `snapshots.rs`.

Napokue:
> I assume it's fine to squash it on merge, right?
Sure.

I think I have added hlsl-out to the snapshots. I only can't seem to find how the output snapshots are generated, or are they created by hand?

kvark:
> I only can't seem to find how the output snapshots are generated, or are they created by hand?

they are created by `cargo test` or `make`.

Napokue:
Found it. This `cargo insta` is a pretty handy tool. I created them with `cargo insta review`

kvark (approved):

Issue #410, opened 2021-02-03
[meta] validate Veloren shaders


kvark:
SPIR-V parsing fails with the following:
```
[0.589056 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedCapability(ImageQuery)
[0.589100 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[0.607530 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedCapability(ImageQuery)
[0.607565 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[0.614626 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedCapability(ImageQuery)
[0.614664 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[0.623153 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedCapability(ImageQuery)
[0.623194 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[0.634291 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedInstruction(Function, SMod)
[0.634336 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[0.636301 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedCapability(ImageQuery)
[0.636336 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[0.676557 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedCapability(ImageQuery)
[0.676605 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[0.682822 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedCapability(ImageQuery)
[0.682864 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[0.686978 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedCapability(ImageQuery)
[0.687014 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[0.690405 WARN]()(wgpu_core::device): Failed to parse shader SPIR-V code: UnsupportedCapability(ImageQuery)
[0.690436 WARN]()(wgpu_core::device): Shader module will not be validated or reflected
[...]

PR #411, opened 2021-02-03
[spv-in] image fetching, constants, logical and relational ops

### LINKED ISSUE #406: `MissingBlockDecoration` in a glsl shader. What does that mean?
I have the following shader:

```
#version 450

struct Vertex
{
    vec4 pos;
};

layout(std430, set = 0, binding = 0) readonly buffer VertexBuffer
{
    Vertex data[];
} vertexBuffer;

void main(){
    Vertex v = vertexBuffer.data[gl_VertexIndex];
    gl_Position = v.pos;
}
```
I compiled it with shaderc as well as with glslangValidator. It compiles without errors but at runtime, I receive a validation error in `device.create_shader_module(&wgpu::include_spirv!("../shaders/shader.vert.spv"));`:

```
wgpu error: Validation Error

Caused by:
      note: label = `../shaders/shader.vert.spv`
    Type Handle(4) 'Vertex' is invalid: MissingBlockDecoration
```
I do not understand what does that means, where this validation error comes from (this is definitely not from a Vulkan Validation Layer) and which decoration should I apply, and where.

kvark:
Fixes #406 
Advances towards #410


src/front/spv/mod.rs
```diff
@@ -518,6 +519,12 @@ impl<I: Iterator<Item = u32>> Parser<I> {
             log::debug!("\t\t{:?} [{}]", inst.op, inst.wc);
 
             match inst.op {
+                Op::Undef => {
+                    inst.expect(3)?;
+                    let _result_type_id = self.next()?;
+                    let _result_id = self.next()?;
+                    //TODO?
```
Napokue, line 526: Nit: as we are not doing anything with this (yet), maybe better to not initialize the variable? I do not know if the rust compiler removes the variable declarations when prefixed with a `_`, otherwise I suggest:
```suggestion
                    //TODO When an intermediate object is undefined, set it to undefined?
                    self.next()?;
                    self.next()?;                    
```
If I understand your TODO correctly, we want to mark the said intermediate object as undefined?

src/front/spv/mod.rs
```diff
@@ -1014,7 +1070,7 @@ impl<I: Iterator<Item = u32>> Parser<I> {
 
                     let si_lexp = self.lookup_sampled_image.lookup(sampled_image_id)?;
                     let coord_lexp = self.lookup_expression.lookup(coordinate_id)?;
-                    let coord_type_handle = self.lookup_type.lookup(coord_lexp.type_id)?.handle;
+                    //let coord_type_handle = self.lookup_type.lookup(coord_lexp.type_id)?.handle;
```
Napokue, line 1073: Nit: Can be removed?

src/front/spv/mod.rs
```diff
@@ -1307,10 +1345,38 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+                }
+                Op::LogicalAnd => {
+                    inst.expect(5)?;
+                    self.parse_expr_binary_op(expressions, crate::BinaryOperator::LogicalAnd)?;
+                }
                 op if inst.op >= Op::IEqual && inst.op <= Op::FUnordGreaterThanEqual => {
                     inst.expect(5)?;
                     self.parse_expr_binary_op(expressions, map_binary_operator(op)?)?;
                 }
+                op if inst.op >= Op::Any && inst.op <= Op::IsNormal => {
```
Napokue, line 1360: I am not sure if we want to construct our matchings like this? It is okay for now, but it requires the one that is working with this code to have knowledge about what OP codes are defined between `Op::Any` and `Op::IsNormal`.

Napokue (approved):
Thank you! I have a couple of small things that are non blocking from merging.

src/front/spv/mod.rs
kvark: We'll likely needed when we implement the proper array layer support here. Spirv coordinates contain the array layer, and we need to extract it (from the last component), and use the rest for the coordinates, so we'll need to know their type.

src/front/spv/mod.rs
kvark: It's not super neat, I agree. It's just there to save the code space, and it was already used right above, so nothing new this PR brings on the table. Do you think we should spell out each of the cases individually? That would be about a page long, but I think we could live with that.

src/front/spv/mod.rs
kvark: We know what these are (i.e. result ID), so erasing this information is also not helpful if we are meant to come back to this.

kvark:
Thank you for the review!
Let's resolve the nits in followup discussions and code, I don't think there is anything that needs fixing right away.

PR #414, opened 2021-02-04
FragDepth is a Write Property

tfgast:
FragDepth is written to in a fragment shader, not read.


kvark (approved):
oops, thank you for the fix!

PR #416, opened 2021-02-05
Don't break upon a void function in the typifier

kvark:
Helps with https://github.com/gfx-rs/wgpu/issues/1197

### DIFF
--- a/src/proc/typifier.rs
+++ b/src/proc/typifier.rs
@@ -48,8 +48,6 @@ pub enum ResolveError {
     InvalidAccessIndex,
     #[error("Function {name} not defined")]
     FunctionNotDefined { name: String },
-    #[error("Function without return type")]
-    FunctionReturnsVoid,
     #[error("Type is not found in the given immutable arena")]
     TypeNotFound,
     #[error("Incompatible operand: {op} {operand}")]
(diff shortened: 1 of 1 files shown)


src/proc/typifier.rs
```diff
@@ -439,10 +437,13 @@ impl Typifier {
...
-                let ty = ctx.functions[function]
-                    .return_type
-                    .ok_or(ResolveError::FunctionReturnsVoid)?;
-                Resolution::Handle(ty)
+                match ctx.functions[function].return_type {
+                    Some(ty) => Resolution::Handle(ty),
+                    // return a type that can't ever be returned...
+                    // We don't have `void` in the IR, but we have to return something here.
+                    // This is also not a place to validate anything.
+                    None => Resolution::Value(crate::TypeInner::Sampler { comparison: false }),
```
Napokue, line 445: I am a little concerned about possible side effect that this line causes. Whenever a void function is encountered now, it is recognized as a Sampler?

Theoretically a void is also a return type I believe, as it returns a void pointer, but I am not sure if that applies to all languages. Maybe we can, at a later point in time, choose to change `return_type` to a `Handle<Type>` and not an `Option<Handle<Type>>`?

Napokue (approved):
One little concern, but I think there is a reason why you choose for this solution, for now. Besides that LGTM, thank you!

src/proc/typifier.rs
kvark: I tried to clarify this in the longish comment in there, and I'll try to use different words here.
Void is, as it stands, only used for function return types. Therefore, having it in the general type system is a waste, and just `Option<Type>` seems much nicer. The only problem is the combination of:
  - the fact that function calls are expressions today
  - the typifier needs to know what each expression type is

So the typifier is in a weird position. On one hand, it has to deal with the IR that has a flaw. So re-architecting the typifier to support this flaw isn't the best. On the other hand, typifier doesn't need to validate anything, so returning a "weird" type, like a sampler, is fine, assuming there is an actual validator that will match this type against what's on the other side, for now.

I filed #420 to address this properly, but I don't know how difficult this would be. Will check today!

kvark:
Let's merge as an interim patch.

src/proc/typifier.rs
Napokue: Right, that is clear! Thanks for the extended answer, looking forward to #420!

PR #421, opened 2021-02-05
Statement for void function call

### LINKED ISSUE #420: Turn function calls into statements
This would fix #416 naturally.

kvark:
Fixes #420

TODO:
- [x] IR statement
- [x] backends
- [x] frontends
- [x] typifier


Napokue (approved):
LGTM, thank you! Like what you did with spv-out, reserving those ids to reduce code complexity in the writer.

kvark:
@Napokue please have another look, at spv-in at least. I'm doing a tricky thing, which is somewhat elegant, but I haven't figure out a cleaner way to do so. The problem is that function calls can use SPIR-V IDs that we haven't seen yet. So previously we'd store some meta-data about an expression used to call a function, and patch it after the module is done (and we know about all the functions).

But now with the function call in a statement, we can't patch it easily. There is no such term as "statement index", so we don't know how to address it and find later...
The solution I came up with is creating fake `Handle<crate::Function>` handles that we find later (recursively scanning the statements) and patch into real ones.

Napokue (approved):
LGTM, thank you for the explanation about the fake function creation. I think it is the best way to do it at the moment. Maybe for later we can discuss it more, as I believe I have read something about compilers doing reverse depth traversal to figure out these kind of things (especially for SPIR-V).

To be sure and clear on the wording. You are using the words "deferred" in the code. Its okay to call it that, but that is the same as how SPIR-V is using the words "forward references", correct? I.e. using references (IDs) that are not yet defined.

kvark:
> To be sure and clear on the wording. You are using the words "deferred" in the code. Its okay to call it that, but that is the same as how SPIR-V is using the words "forward references", correct? I.e. using references (IDs) that are not yet defined.

yes, precisely. We can rename it to anything, I don't have hard feelings for the "deferred" name.

Thank you for reviewing!

PR #422, opened 2021-02-05
Makefile commands for native shader validation

kvark:
This is something I wanted to do for a long time, and I think snapshots make it fairly easy to roll in.

The problem we had is that we generate the snapshot shaders, but it's tedious to double-check with the actual native SDK if those updated snapshots are valid at all. I.e. we could have a snapshot in-tree that is simply invalid, which compromises the whole idea of keeping it around. In fact, that's the case today for a few shaders.

So technically we have the native shaders produced, and we have the tools, but remembering all the tool syntax and editing the snapshot shaders to strip the header, makes it very hard to force this into a habit. Just takes too much effort!
Lo and behold - the automation:
```bash
make validate-spv # for Vulkan shaders, requires SPIRV-Tools installed
make validate-msl # for Metal shaders, requires XCode command-line tools installed
make validate-glsl # for OpenGL shaders, requires GLSLang installed
```

It's very straightforward to just call these. They are separate because they require different SDK/platforms. The developer is expected to call the ones applicable to their platforms.

I think the follow-up here should be the relevant Github Action tasks that react on the changes in those files, that install the SDKs and run these makefile commands on CI, automatically.

Interestingly, I instantly discovered 2 issues with our snapshotted shaders:
  1. the GLSL output for global images was broken. I fixed this in the PR.
  2. the SPIR-V output for calling nested functions was not gathering the I/O properly. It's #390, and I haven't fixed this yet. I don't think it should be blocking the PR.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -455,11 +455,7 @@ impl<'a, W: Write> Writer<'a, W> {
 
                     // Finally write the name and end the global with a `;`
                     // The leading space is important
-                    writeln!(
-                        self.out,
-                        " {};",
-                        self.names[&NameKey::GlobalVariable(handle)]
-                    )?;
+                    writeln!(self.out, " {};", self.get_global_name(handle, global))?;
 
                     writeln!(self.out)?;
                 }
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,13 @@
-.PHONY: all clean
+.PHONY: all clean validate-spv validate-msl
 .SECONDARY: boids.metal quad.metal
+SNAPSHOTS=tests/snapshots
 
 all:
 
 clean:
-	rm *.metal *.air *.metallib *.vert *.frag *.comp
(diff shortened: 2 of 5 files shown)


src/back/glsl/mod.rs
```diff
@@ -455,11 +455,7 @@ impl<'a, W: Write> Writer<'a, W> {
 
                     // Finally write the name and end the global with a `;`
                     // The leading space is important
-                    writeln!(
-                        self.out,
-                        " {};",
-                        self.names[&NameKey::GlobalVariable(handle)]
-                    )?;
+                    writeln!(self.out, " {};", self.get_global_name(handle, global),)?;
```
JCapucho, line 458: I don't like this comma in the end

JCapucho (approved):
Everything looks okay, only a very small style nit

src/back/glsl/mod.rs
kvark: hah, nicely spotted!

PR #423, opened 2021-02-05
Add CI action for validation of native shaders

kvark:
Only SPV and GLSL so far.
Follow-up to #422

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -1,0 +1,16 @@
+name: validation
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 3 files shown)


kvark:
So, funny thing. The checks are all in place, but it doesn't stop CI. There needs to be a way for spirv-val to return the error code...

kvark:
I think it's still useful to get at least an ability to look up on how these checks are going. I.e. I may be writing a PR from linux, and I want to make sure it doesn't accidentally break any HLSL or MSL.

PR #424, opened 2021-02-05
Add binary ops support for the ConstantSolver

JCapucho:
Only implements between scalars since composites require more work


src/proc/constants.rs
```diff
@@ -253,6 +261,102 @@ impl<'a> ConstantSolver<'a> {
...
+                ConstantInner::Scalar {
+                    value: right_value, ..
+                },
+            ) if discriminant(left_value) == discriminant(right_value) => {
+                let value = match op {
+                    BinaryOperator::Equal => ScalarValue::Bool(left_value == right_value),
+                    BinaryOperator::NotEqual => ScalarValue::Bool(left_value != right_value),
+
+                    _ => match (left_value, right_value) {
+                        (ScalarValue::Sint(a), ScalarValue::Sint(b)) => match op {
```
kvark, line 291: let's just wrap it up, i.e. `=> ScalarValue::Sint(match op {...})`
can be applied to other places too

kvark (approved):

src/proc/constants.rs
JCapucho: This isn't possible because in the comparison operations we return a boolean

kvark: If we are talking about scalars only, the booleans could be moved outside and processed together with `Equal`/`NotEqual`.

kvark (approved):
Sweet!

PR #425, opened 2021-02-05
[glsl-*] component wise comparisons and more math functions

JCapucho:
Also fixes a mistake when building the `mix` function in glsl-in where the third argument wasn't set


kvark (approved):
Looks reasonable, thank you!

PR #426, opened 2021-02-06
[glsl-in] Fix the way equality is handled

JCapucho:
Vector equality in glsl is now handled as `all(equal(vec1, vec2))` and inequality as `any(notEqual(vec1, vec2))`

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -2,8 +2,8 @@ use super::{constants::ConstantSolver, error::ErrorKind};
 use crate::{
     proc::{ResolveContext, Typifier},
     Arena, BinaryOperator, Binding, Constant, Expression, FastHashMap, Function, FunctionArgument,
-    GlobalVariable, Handle, Interpolation, LocalVariable, Module, ShaderStage, Statement,
-    StorageClass, Type,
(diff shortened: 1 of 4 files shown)


src/front/glsl/ast.rs
```diff
@@ -57,6 +57,50 @@ impl Program {
...
+        equals: bool,
+        left: &ExpressionRule,
+        right: &ExpressionRule,
+    ) -> Result<ExpressionRule, ErrorKind> {
+        let left_is_vector = match self.resolve_type(left.expression)? {
+            crate::TypeInner::Vector { .. } => true,
+            _ => false,
+        };
+
+        let rigth_is_vector = match self.resolve_type(left.expression)? {
```
kvark, line 74: this should be `right.expression`

src/front/glsl/ast.rs
```diff
@@ -57,6 +57,50 @@ impl Program {
...
+                },
+                left: left.expression,
+                right: right.expression,
+            }));
+
+        Ok(if left_is_vector && rigth_is_vector {
+            ExpressionRule::from_expression(self.context.expressions.append(
+                Expression::Relational {
+                    fun: match equals {
+                        true => RelationalFunction::All,
```
kvark, line 93: nit: it would be tidier to match this only once, i.e.
```rust
let (op, fun) = match equals {
};
```

kvark (approved):

PR #427, opened 2021-02-06
[glsl-in] more math functions and relational functions

JCapucho:


src/front/glsl/functions.rs
```diff
@@ -164,6 +167,25 @@ impl Program {
                             statements: fc.args.into_iter().flat_map(|a| a.statements).collect(),
                         })
                     }
+                    "isinf" | "isnan" | "all" | "any" => {
+                        if fc.args.len() != 1 {
```
kvark, line 171: would it be simpler to organize this as `self.parse_relational_fun()` helper, so that you just invoke it with the proper argument for each of `isinf`/`isnan`/etc, instead of matching them twice?

kvark (approved):

PR #429, opened 2021-02-07
Push constant support

### LINKED ISSUE #362: Push constants support
These are not a part of WebGPU upstream, but useful for webgpu-native as well as regular SPIR-V related use cases.
See also - https://github.com/gfx-rs/wgpu/issues/1162

kvark:
Closes #362

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -245,3 +245,9 @@ pub enum StorageQualifier {
     StorageClass(StorageClass),
     Const,
 }
+
+#[derive(Debug, Clone)]
+pub enum StructLayout {
+    Binding(Binding),
+    PushConstant,
+}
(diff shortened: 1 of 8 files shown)


Napokue (approved):
LGTM, thank you! Question: do all shader languages support push constants,  I believed it is a Vulkan-only thing? Or are we artificially creating "push constants" for the other languages that do not have it?

kvark:
Great question! Only Vulkan supports that natively, but I think we need that in the IR because we are going to see a lot of SPIRV-in used around, and push constants are used a lot. For now, I think we can rely on the backends failing in the presence of push constants, and wgpu-core failing the validation if they aren't supported. But later we can consider having some `Capability` bitflags in the IR itself to identify clearly what it requires.

Napokue:
Agreed, and yes I think we really need to include the concept of push constants in Naga.

PR #432, opened 2021-02-07
[spv-in] Implement OpCompositeInsert

### LINKED ISSUE #413: Support OpCompositeInsert in SPIRV frontend
It's a bit painful to construct an expression like this, but we have to do it.
Blocks #410

kvark:
Fixes #413
Another step closer to #410

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -491,6 +491,62 @@ impl<I: Iterator<Item = u32>> Parser<I> {
         Ok(())
     }
 
+    fn insert_composite(
+        &self,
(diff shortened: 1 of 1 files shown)


Napokue (approved):
LGTM, thank you! Besides your Array todo, the Matrix type also support CompositeInsert. Maybe something for later?

kvark:
The matrix type is actually handled in the PR:
```rust
             crate::TypeInner::Vector { size, .. }
            | crate::TypeInner::Matrix { columns: size, .. } => {
```
I'm not sure what to do with Array, they are extremely awkward, but I think it's ok for now.

PR #433, opened 2021-02-07
Move snapshot inputs and outputs around

grovesNL:
This PR moves snapshot inputs to `in` and outputs to `out` as discussed on Matrix


kvark (approved):
Thank you!
I think Makefile also needs to be updated

kvark (approved):

PR #437, opened 2021-02-08
[spv-in] support projection sampling and sampling from arrays

### LINKED ISSUE #419: Array layer support in SPIR-V frontent image ops
In SPIR-V, the texture coordinates must have the array layer as the last component, and it can be `f32`, where IR expects `i32`. The proper logic would need to extract that array index and convert it.

Note: this is about all the `Expression::Image*` operations in the IR.

### LINKED ISSUE #430: Implement OpImageSampleProjDrefImplicitLod in spv-in
Blocks #410

kvark:
Fixes #430
Fixes #419


Napokue (approved):
LGTM, thank you!

PR #438, opened 2021-02-08
[spv-in] rewrite comparison sampling logic

kvark:
The second commit is refactoring stuff into a module. I think only the first commit needs to be reviewed (separately).
This PR gets us to validate all of Veloren's SPIR-V shaders!


Napokue (approved):
LGTM, thank you! Really like the separation of the image logic, as it was growing rather large to be included in the parser!

kvark:
This moved out roughly 1/5th of the logic. There are still blocks to move out for future refactoring work ;)
Thank you for the review!

PR #440, opened 2021-02-10
Refactor and fix the global usage code

### LINKED ISSUE #390: Aggregate interfaces of called functions

kvark:
Fixes #390, will allow us to switch on `EXPERIMENTAL_TRANSLATION` for the shadow example on Metal in wgpu-rs.


Napokue (approved):
LGTM, thank you!

kvark:
I realized there is more work to be done to fix the shadow example on Apple: we need to route the arguments into the helper functions we call.
But that's ok, the PR can still be landed.

PR #441, opened 2021-02-10
[glsl-in] Fix StorageClass for texture and sampler

pjoe:
Use `Handle` storage class for texture and sampler

### DIFF
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -1087,7 +1087,7 @@ pomelo! {
             }).unwrap_or(StorageQualifier::StorageClass(StorageClass::Private));
 
             match storage {
-                StorageQualifier::StorageClass(class) => {
+                StorageQualifier::StorageClass(storage_class) => {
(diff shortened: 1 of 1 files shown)


src/front/glsl/parser.rs
```diff
@@ -1100,6 +1100,16 @@ pomelo! {
                     for (id, initializer) in d.ids_initializers {
                         let init = initializer.map(|init| extra.solve_constant(init.expression)).transpose()?;
 
+                        // use StorageClass::Handle for texture and sampler uniforms
+                        let class = if storage_class == StorageClass::Uniform {
+                            match &extra.module.types[d.ty].inner {
```
kvark, line 1105: please no matching on references

kvark (commented):
one small nit!

src/front/glsl/parser.rs
pjoe: hmmm, think I was struggling with the syntax, and along the way the compiler suggested me to add the `&` ... still a noob :)

kvark (approved):

Issue #409, opened 2021-02-03
[meta] Process Dota2's SPIRV to MSL


kvark:
We'd want to run Dota2 with [gfx-portability](https://github.com/gfx-rs/portability) on Metal in a way that doesn't involve SPIRV-Cross at all. Will file separate issues for the SPV-in and MSL-out until we fully cover this path.

Test set - [dota2-shaders.zip](https://github.com/gfx-rs/naga/files/6333187/dota2-shaders.zip)
My bash script to bulk-convert them:
```bash
NAGA_DIR=<insert naga path here>
NAGA_BIN=${NAGA_DIR}/target/debug/convert
(cd ${NAGA_DIR} && cargo build --features spv-in,msl-out,serialize)
for file in *.spv ; do \
	echo "Translating" ${file}; \
	$NAGA_BIN ${file} >${file}.txt
	$NAGA_BIN ${file} ${file}.metal
	#tail -n +5 ${file} | xcrun -sdk macosx metal -x metal - -o /dev/null; \
done
```
I launch it as "sh ../naga.sh" from the folder with SPVs.

kvark:
Here is my small script:
```bash
NAGA=../../naga/target/debug/examples/convert
(cd ../../naga && cargo build --example convert --features spv-in,msl-out)
for file in spirv/*.spv ; do \
	echo "Translating" ${file}; \
	$NAGA ${file} ${file}.metal
	#xcrun -sdk macosx metal -mmacosx-version-min=10.11 ${file}.metal - -o /dev/null; \
done
```

kvark:
Here is the first log, up to shader-229, which hangs because of another problem that I'm about to address.
```
[2021-02-09T15:33:08Z ERROR naga::back::msl] Missing binding for Some("")
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Validation', examples/convert.rs:193:65
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnsupportedBinaryOp(LogicalOr)', examples/convert.rs:193:65
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnsupportedCapability(ClipDistance)', examples/convert.rs:95:64
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnsupportedInstruction(Type, ConstantFalse)', examples/convert.rs:95:64
[...]

PR #442, opened 2021-02-10
SPV-in capabilities, dynamic vector access, boolean constants

kvark:
These are all the things we have in #409 so far.


Napokue (approved):
LGTM, thanks! Does this run one of the Dota shaders already, or not yet?

kvark:
I haven't actually tested the Dota shaders again. I need to resolve this issue with MSL backend exponential complexity first.

PR #443, opened 2021-02-11
[msl] Propagate entry point inputs and resources into called functions

kvark:
Fixes the shadow snapshot!


kvark:
The new CI job here is super helpful to show that the produced MSL is correct :tada:

PR #444, opened 2021-02-11
Separate validation job on macOS

kvark:


kvark:
I need to merge this in order for GH to accept the new workflows. Will be playing with it right after.

PR #447, opened 2021-02-11
[glsl-in] Support multiple entrypoints

pjoe:
Handle multiple entry points.

Add insta test for this.

### DIFF
--- a/examples/convert.rs
+++ b/examples/convert.rs
@@ -106,8 +106,7 @@ fn main() {
             let input = fs::read_to_string(&args[1]).unwrap();
             naga::front::glsl::parse_str(
                 &input,
-                "main",
-                naga::ShaderStage::Vertex,
+                vec![("main".to_string(), naga::ShaderStage::Vertex)],
                 Default::default(),
             )
             .unwrap_pretty()
@@ -117,8 +116,7 @@ fn main() {
             let input = fs::read_to_string(&args[1]).unwrap();
             naga::front::glsl::parse_str(
                 &input,
-                "main",
-                naga::ShaderStage::Fragment,
+                vec![("main".to_string(), naga::ShaderStage::Fragment)],
                 Default::default(),
             )
             .unwrap_pretty()
@@ -128,8 +126,7 @@ fn main() {
             let input = fs::read_to_string(&args[1]).unwrap();
             naga::front::glsl::parse_str(
                 &input,
-                "main",
-                naga::ShaderStage::Compute,
+                vec![("main".to_string(), naga::ShaderStage::Compute)],
                 Default::default(),
             )
             .unwrap_pretty()
--- a/examples/convert.rs
+++ b/examples/convert.rs
@@ -104,30 +104,42 @@ fn main() {
         #[cfg(feature = "glsl-in")]
         "vert" => {
             let input = fs::read_to_string(&args[1]).unwrap();
+            let mut entry_points = naga::FastHashMap::default();
+            entry_points.insert("main".to_string(), naga::ShaderStage::Vertex);
             naga::front::glsl::parse_str(
                 &input,
-                vec![("main".to_string(), naga::ShaderStage::Vertex)],
-                Default::default(),
+                &naga::front::glsl::Options {
+                    entry_points,
+                    defines: Default::default(),
+                },
             )
             .unwrap_pretty()
         }
         #[cfg(feature = "glsl-in")]
         "frag" => {
             let input = fs::read_to_string(&args[1]).unwrap();
+            let mut entry_points = naga::FastHashMap::default();
+            entry_points.insert("main".to_string(), naga::ShaderStage::Fragment);
             naga::front::glsl::parse_str(
                 &input,
-                vec![("main".to_string(), naga::ShaderStage::Fragment)],
-                Default::default(),
+                &naga::front::glsl::Options {
+                    entry_points,
+                    defines: Default::default(),
+                },
             )
             .unwrap_pretty()
         }
         #[cfg(feature = "glsl-in")]
         "comp" => {
             let input = fs::read_to_string(&args[1]).unwrap();
+            let mut entry_points = naga::FastHashMap::default();
+            entry_points.insert("main".to_string(), naga::ShaderStage::Compute);
             naga::front::glsl::parse_str(
                 &input,
-                vec![("main".to_string(), naga::ShaderStage::Compute)],
-                Default::default(),
+                &naga::front::glsl::Options {
+                    entry_points,
+                    defines: Default::default(),
+                },
             )
             .unwrap_pretty()
         }
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -10,8 +10,7 @@ use crate::{
 pub struct Program {
     pub version: u16,
     pub profile: Profile,
-    pub shader_stage: ShaderStage,
-    pub entry: Option<String>,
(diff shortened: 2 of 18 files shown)


tests/snapshots.rs
```diff
@@ -233,3 +233,54 @@ fn convert_spv(name: &str) {
...
+        entry_points,
+        Default::default(),
+    )
+    .unwrap();
+    naga::proc::Validator::new().validate(&module).unwrap();
+
+    #[cfg(feature = "spv-out")]
+    {
+        if language.contains(Language::SPIRV) {
+            check_output_spv(&module, name, &params);
```
kvark, line 256: Here is a controversial proposal for you: let's make WGSL a "blessed" frontend.
The problem here is that, if we are going to have this sort of snapshot testing for all frontends, then we'll have way too much backend coverage. Its not bad in itself, but the dis-balance between frontend and backend coverage is bad.
So we need to be more selective. Most frontend testing should simply be `front -> IR` snapshots, like the current SPIR-V frontend snapshot. Only a few would need to go into all backends. We could select them arbitrarily, or we can say that all WGSL tests go this way. I think it's a fine direction because of the following reasons:
  1. WGSL is super close to our IR, so we aren't going to get into a situation where the backends only test specific subset of IR that the blessed frontend produces.
  2. it's consistent. We aren't going to ask ourselves this question in the future if we just make this call now.

src/front/glsl/functions.rs
```diff
@@ -275,4 +275,28 @@ impl Program {
...
+        } else {
+            None
+        };
+        let name = f.name.clone().unwrap();
+        if let Some(stage) = stage {
+            self.module.entry_points.insert(
+                (stage, name),
+                EntryPoint {
+                    early_depth_test: None,
+                    workgroup_size: [0; 3], //TODO
```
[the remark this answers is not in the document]
pjoe: Yeah, or maybe passed in. Anyway in this PR this code is just moved out from `parser.rs`

tests/snapshots.rs
pjoe: Heh, was actually thinking along those lines as well :)

Initially using the full `spv-out > spirv-val`, to make sure everything is good.

pjoe: Hmm after thinking some more: as glsl -> spv is likely to be a main use case (at least until wgsl takes over the world), so I'd like to at least have the option to do the full glsl -> spirv-val in our auto tests. Would also allow things like refactoring glsl-in to be done with more confidence :)

Anyway lemme know what you think.

kvark: ok, sounds reasonable. So let's have GLSL->IR->SPV snapshots

pjoe: Ok, for now made glsl-in tests always do IR and spv

examples/convert.rs
```diff
@@ -106,8 +106,7 @@ fn main() {
             let input = fs::read_to_string(&args[1]).unwrap();
             naga::front::glsl::parse_str(
                 &input,
-                "main",
-                naga::ShaderStage::Vertex,
+                vec![("main".to_string(), naga::ShaderStage::Vertex)],
```
kvark, line 109: time to introduce a `struct Options`!

src/front/glsl/functions.rs
```diff
@@ -281,4 +281,28 @@ impl Program {
         f.fill_global_use(self.module.global_variables.len(), &self.module.functions);
         f
     }
+
+    pub fn declare_function(&mut self, f: Function) {
+        let stage = if let Some(ref name) = f.name {
+            self.entry_points
```
kvark, line 287: Can `entry_points` have the same name for multiple stages? I think IR would allow that, but GLSL is totally free to forbid this. If we don't allow it in GLSL, let's make `entry_points` a `FastHashMap<Name, ShaderStage>`?

src/front/glsl/functions.rs
```diff
@@ -281,4 +281,28 @@ impl Program {
...
+                .iter()
+                .find_map(|(n, s)| if *name == *n { Some(*s) } else { None })
+        } else {
+            None
+        };
+        let name = f.name.clone().unwrap();
```
kvark, line 293: since we unwrap here, the `f.name` check above is unnecessary

src/front/glsl/variables.rs
```diff
@@ -21,19 +21,12 @@ impl Program {
         match name {
             "gl_Position" => {
                 #[cfg(feature = "glsl-validate")]
-                match self.shader_stage {
-                    ShaderStage::Vertex | ShaderStage::Fragment { .. } => {}
-                    _ => {
-                        return Err(ErrorKind::VariableNotAvailable(name.into()));
-                    }
+                if !self.has_stage(ShaderStage::Vertex) {
```
kvark, line 24: I don't think we ever need to call `has_stage`. Why are we checking it? If no entry points use this global variable, it's totally fine for it to stay in the IR.

src/front/glsl/ast.rs
```diff
@@ -44,6 +42,10 @@ impl Program {
         }
     }
 
+    pub fn has_stage(&self, stage: ShaderStage) -> bool {
```
kvark, line 45: we should probably not have this function

tests/in/quad-glsl.param.ron
```diff
@@ -0,0 +1,8 @@
+(
+	spv_version: (1, 0),
+	spv_capabilities: [ Shader ],
+	mtl_bindings: {
```
kvark, line 4: not really needed since we aren't generating MSL from this

kvark (commented):
Nice!

src/front/glsl/variables.rs
pjoe: I think this was done in order to be able to give errors, if e.g. a fragment shader references a builtin that is only available in vertex stage. But when supporting multiple stages this gets less clear.

kvark: This is caught by IR validation already, you don't need to to worry about this in the frontend.

examples/convert.rs
pjoe: Ok, you thinking `Vec<StageOptions>` or `GlslParserOptions` as general options struct for glsl parser?

src/front/glsl/functions.rs
pjoe: I think that would be weird in glsl, as in source that would be multiple functions with same name. Will change to `FastHashMap`

examples/convert.rs
kvark: Just a single `struct Options` that's passed into the frontend by reference, similar to how SPV-out and MSL-out are structured.

src/front/glsl/functions.rs
```diff
@@ -281,4 +281,29 @@ impl Program {
...
+
+    pub fn declare_function(&mut self, f: Function) -> Result<(), ErrorKind> {
+        let name = f
+            .name
+            .clone()
+            .ok_or_else(|| ErrorKind::SemanticError("Unnamed function".into()))?;
+        let stage = self
+            .entry_points
+            .iter()
+            .find_map(|(n, s)| if *name == *n { Some(*s) } else { None });
```
kvark, line 293: no need to `find`, since it's a hashmap now. You can just do `self.entry_points.get(&name)`

kvark (approved):
One last tiny thing!

src/front/glsl/functions.rs
pjoe: damn, you spot everything 😄

PR #449, opened 2021-02-11
[wgsl] support 'elseif' construct

kvark:
Apparently, it's a thing!
We need to consider uplifting this to the IR, but so far it looks to be too specific to WGSL.


Napokue (approved):
LGTM, thank you! This construct is pretty normal I think. I have encountered it in multiple programming languages. However, I think most if not all else if constructs can be changed to a switch construct. And I think switch constructs are even more normal than this else if construct across languages.

kvark:
My main motivation is to not impose the heap allocation on every `If`. I.e. if it's `If { cases: Vec<(Expression, Statement), other :Statement }` they we'd have a perf penalty for each and every If statement.

PR #450, opened 2021-02-11
[spv-out] support Switch statements

kvark:
Tested locally on the following example:
```rust
[[stage(vertex)]]
fn main() {
	var x: i32 = 4;
	const y: f32 = 6.0;
	var z: f32;
	switch(x) {
		case 4: { z = 4.0; fallthrough;}
		case 5: { z = 5.0; }
		default: { z = 1.0; }
	}
}
```
Worked at first try!

### DIFF
--- a/src/back/spv/instructions.rs
+++ b/src/back/spv/instructions.rs
@@ -707,6 +707,26 @@ pub(super) fn instruction_branch_conditional(
     instruction
 }
 
+pub struct Case {
+    pub value: Word,
+    pub label_id: Word,
+}
+
+pub(super) fn instruction_switch(
+    selector_id: Word,
+    default_id: Word,
+    cases: &[Case],
+) -> Instruction {
+    let mut instruction = Instruction::new(Op::Switch);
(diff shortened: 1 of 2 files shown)


Napokue (approved):
LGTM, thank you! So as far as I understand the code, each case is its own block that will be merged with the block that is after the switch case (`case_finish_id` I presume?).

The only thing I do not see implemented, but that can be my bad (and I missed it), is the case of `fallthrough`.

Last question: does WGSL implicitly `break` its cases? I have worked with languages where `breaks` are required to specify, if it is not specified it would implicitly be a `fallthrough`.

kvark:
@Napokue it's amazing that you caught that! I was thinking about fallthrough, but instead of checking for it I implemented everything as fallthrough :sweat_smile: 
Now it's fixed.

> Last question: does WGSL implicitly break its cases? I have worked with languages where breaks are required to specify, if it is not specified it would implicitly be a fallthrough.

Correct! Implicit fallthrough was considered a dangerous practice.

Issue #380, opened 2021-01-27
Consider codespan for the frontend error reporting


kvark:
https://github.com/brendanzab/codespan

grovesNL:
As mentioned on Matrix, I'd be happy to prototype this

PR #451, opened 2021-02-12
[wgsl] track number of bytes consumed per token

grovesNL:
This isn't a very useful change on its own, but we need to track the number of bytes consumed per token in order to support #380 or other later contextual errors that need to know the byte span of relevant tokens.

Alternatively we could derive it by measuring the change in `self.input.len()` whenever `self.input` is changed, but counting the bytes directly as they're consumed felt like the most straightforward path.

If these changes seem useful, I was thinking we could land this PR on its own to avoid diverging too much from `master` in my branch.

### DIFF
--- a/src/front/wgsl/lexer.rs
+++ b/src/front/wgsl/lexer.rs
@@ -58,101 +58,97 @@ fn consume_number(input: &str) -> (Token, &str) {
 }
 
 fn consume_token(mut input: &str) -> (Token<'_>, &str) {
-    loop {
-        input = input.trim_start();
-        let mut chars = input.chars();
-        let cur = match chars.next() {
-            Some(c) => c,
-            None => return (Token::End, input),
-        };
-        break match cur {
-            ':' => {
-                input = chars.as_str();
-                if chars.next() == Some(':') {
-                    (Token::DoubleColon, chars.as_str())
-                } else {
-                    (Token::Separator(cur), input)
-                }
+    let mut chars = input.chars();
+    let cur = match chars.next() {
+        Some(c) => c,
+        None => return (Token::End, input),
+    };
+    match cur {
+        ':' => {
+            input = chars.as_str();
+            if chars.next() == Some(':') {
+                (Token::DoubleColon, chars.as_str())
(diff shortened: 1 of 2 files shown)


kvark:
@grovesNL do we have a good idea about how the alternative would work for codespan?
I feel that using the `self.input.len()` could be simpler, and could zero out chances that we make mistakes in the implementation. I.e. this PR is 170 lines of fairly logic-heavy code, so that's good to have as a basleine to compare the alternative to.

grovesNL:
@kvark sounds good, let's try the `self.input.len()` approach. Both approaches will wok fine with codespan – they're just two different way to measure how many bytes were consumed by each token. I'll update this PR with the `self.input.len()` approach and keep these changes in another branch for comparison

grovesNL:
@kvark just updated this branch. The number of bytes read is given by `let _bytes_read = original_len - self.input.len();` but we don't return it from the lexer yet

src/front/wgsl/lexer.rs
```diff
@@ -197,16 +199,16 @@ impl<'a> Lexer<'a> {
...
             Err(Error::Unexpected(token, description))
         }
     }
 
     pub(super) fn skip(&mut self, what: Token<'_>) -> bool {
-        let (token, rest) = consume_token(self.input);
-        if token == what {
-            self.input = rest;
+        if self.peek() == what {
+            let _ = self.next();
```
grovesNL, line 211: This currently consumes the token twice but we could fix this before merging

kvark: sounds good, let's try to fix this!

kvark (approved):
This looks like a trivial change if you consider the refactoring of the loop (moving from `consume_token` higher up) to be independent. I like it!

src/front/wgsl/lexer.rs
grovesNL: I wasn't sure what to call it but we have `peek_token_and_rest` now – let me know if this is clear enough

kvark (approved):

PR #457, opened 2021-02-13
[spv-out] make instruction constructions to be methods

kvark:


Napokue (approved):
Thank you! I have not reviewed every single line, but we have already discussed this change some time ago, and I totally agree with this approach!

The only concern I had back then was about the `return` instruction, as it is also a keyword in Rust, but I see you have solved it by renaming it to `return_void` which is totally acceptable for me.

Thank you for this change, as it requires a lot less typing of `super::instructions::XXX`.

PR #458, opened 2021-02-13
[spv-in] patch function calls before deriving the global use

kvark:
We had a bug in there, since global use resolution needs proper function indices, and ours were unpatched.


Napokue (approved):
LGTM, thank you!

PR #460, opened 2021-02-13
[spv-out] support select expressions

acowley:


acowley:
I couldn't determine how I should add tests for this. Guidance there would be appreciated!

src/back/spv/instructions.rs
```diff
@@ -727,6 +727,22 @@ pub(super) fn instruction_switch(
     instruction
 }
 
+pub(super) fn instruction_select(
+    result_type_id: Word,
+    id: Word,
+    condition_id: Word,
+    accept_id: Word,
+    reject_id: Word
```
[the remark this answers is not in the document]
acowley: Fixed!

src/back/spv/instructions.rs
```diff
@@ -727,6 +727,22 @@ pub(super) fn instruction_switch(
     instruction
 }
 
+pub(super) fn instruction_select(
```
kvark, line 730: this needs to be rebased now, we made all these to be constructor methods

kvark (approved):

kvark:
> I couldn't determine how I should add tests for this. Guidance there would be appreciated!

We have a number of snapshot tests. We don't want to have a separate test for each new kind of expression, but rather have fewer tests that cover large group of expressions. If you have a WGSL shader using the select (and perhaps as switch?), could you provide it here? Maybe it's a good candidate for inclusion into snapshots.

src/back/spv/instructions.rs
acowley: I have performed the rebase.

acowley:
I haven't made a new snapshot yet. Do I need to worry about MSL output?

kvark (approved):
Great, thank you!

kvark:
> I haven't made a new snapshot yet. Do I need to worry about MSL output?

We have a lot of instructions that aren't covered yet. I'm hoping to get #464 guiding the process of adding more snapshots, so it's not pressing for your PR.
If you end up adding a snapshot, you can do it in a way that doesn't produce MSL output (see the `Language` bitflags). If you include MSL, the CI will make sure that it's at least valid.

PR #461, opened 2021-02-13
Support constants in the naming processor

kvark:
Previously, the MSL backend was just embedding all the constants in-place, which resulted in lot of repetition whenever composite constants were involved. Now we are declaring them separately and using throughout the code.
We may consider at some point to still in-line the scalar constants in the code, but I don't think it changes the core of this PR.

### DIFF
--- a/src/proc/namer.rs
+++ b/src/proc/namer.rs
@@ -5,6 +5,7 @@ pub type EntryPointIndex = u16;
 
 #[derive(Debug, Eq, Hash, PartialEq)]
 pub enum NameKey {
+    Constant(Handle<crate::Constant>),
     GlobalVariable(Handle<crate::GlobalVariable>),
     Type(Handle<crate::Type>),
     StructMember(Handle<crate::Type>, u32),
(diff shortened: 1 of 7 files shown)


src/proc/namer.rs
```diff
@@ -91,6 +88,52 @@ impl Namer {
...
+                            value: crate::ScalarValue::Uint(v),
+                        } => write!(temp, "const_{}u", v),
+                        crate::ConstantInner::Scalar {
+                            width: _,
+                            value: crate::ScalarValue::Float(v),
+                        } => write!(
+                            temp,
+                            "const_{}_{}f",
+                            v.abs().trunc(),
+                            (v.fract() * 100.0) as i8
```
grovesNL, line 119: The naming output seems a bit off when using this, e.g. https://github.com/gfx-rs/naga/pull/461/files#diff-332e4d2098a2ab03271f1d2d73b2ac4b8b8612fa8510c95573e7fea94a46ba65R48

Could we do something like use `f64`'s `Display` impl and substitute `.` with `_`, `-` with `n`?

grovesNL (approved):

src/proc/namer.rs
kvark: is `const_0_5f` wrong? I thought that's the best we can do

grovesNL: I expected something like `const_0_05f = 0.05` there (an extra zero), especially because we also have `const_0_50f = 0.5` above

kvark: oh snap, you are right, that's totally wrong!

PR #463, opened 2021-02-13
[wgsl] restrict type casts to matching container types

### LINKED ISSUE #454: Constructing a vec with the wrong number of arguments in WGSL creates the wrong type
```wgsl
[[stage(vertex)]]
fn vs_main() {
    var x: vec2<f32> = vec2<f32>(1.0);
}
```

This WGSL snippet is accepted and results in the following SPIR-V:

```spirv
  %float = OpTypeFloat 32
%float_1 = OpConstant %float 1
%v2float = OpTypeVector %float 2
%_ptr_Function_v2float = OpTypePointer Function %v2float
%vs_main = OpFunction %void None %9
     %10 = OpLabel
      %x = OpVariable %_ptr_Function_v2float Function
     %11 = OpBitcast %float %float_1
           OpStore %x %11
           OpReturn
           OpFunctionEnd
```

This results in the following validation error:

```
VALIDATION [UNASSIGNED-CoreValidation-Shader-InconsistentSpirv (7060244)] : Validation Error: [ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object 0: handle = 0x1fa5a369398, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x6bbb14 | SPIR-V module 
not valid: OpStore Pointer <id> '5[%x]'s type does not match Object <id> '11[%11]'s type.
  OpStore %x %11
```

I'm not fluent in SPIR-V, but I assume what's happening is that `vec2<f32>(1.0)` in WGSL is converted to a `OpTypeFloat` of value `1` instead of a `OpTypeVector`, and trying to store this in a `OpTypeVector` variable then results in a mismatch.

The [type constructor section](https://gpuweb.github.io/gpuweb/wgsl.html#type-constructor-expr) in the spec lists no single-value constructor for vecs, so I assume this should be denied instead.

<details>
<summary>In the IR, the <code>vec2&lt;f32&gt;(1.0)</code> expression seems to also be represented as just a float constant.</summary>

```ron
(
    header: (
        version: (1, 0, 0),
        generator: 0,
    ),
    types: [
        (
            name: None,
            inner: Vector(
                size: Bi,
                kind: Float,
                width: 4,
            ),
        ),
    ],
    constants: [
        (
            name: None,
            specialization: None,
            inner: Scalar(
                width: 4,
                value: Float(1.0),
            ),
        ),
    ],
    global_variables: [],
    functions: [],
    entry_points: {
        (Vertex, "vs_main"): (
            early_depth_test: None,
            workgroup_size: (0, 0, 0),
            function: (
                name: Some("vs_main"),
                arguments: [],
                return_type: None,
                global_usage: [],
                local_variables: [
                    (
                        name: Some("x"),
                        ty: 1,
                        init: None,
                    ),
                ],
                expressions: [
                    Constant(1),// [0]
                    As(
                        expr: 1,
                        kind: Float,
                        convert: true,
                    ),// [1]
                    LocalVariable(1),
                ],
                body: [
                    Store(
                        pointer: 3,
                        value: 2,
                    ),// [0]
                    Return(
                        value: None,
                    ),
                ],
            ),
        ),
    },
)
```
</detail>

Systemcluster:
Restricts type casts to types of the same container.

resolves #454

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -844,7 +844,6 @@ impl Parser {
                         ctx.constants,
                     )?;
                     let kind = inner.scalar_kind();
-                    let ty = ctx.types.fetch_or_append(crate::Type { name: None, inner });
 
                     lexer.expect(Token::Paren('('))?;
                     let mut components = Vec::new();
@@ -856,13 +855,24 @@ impl Parser {
                     }
                     lexer.expect(Token::Paren(')'))?;
                     if components.is_empty() {
-                        crate::Expression::As {
-                            expr: last_component,
-                            kind: kind.ok_or(Error::BadTypeCast(word))?,
-                            convert: true,
+                        let last_component_inner = ctx.resolve_type(last_component)?;
+                        match (&inner, last_component_inner) {
+                            (crate::TypeInner::Scalar { .. }, crate::TypeInner::Scalar { .. })
+                            | (crate::TypeInner::Matrix { .. }, crate::TypeInner::Matrix { .. })
+                            | (crate::TypeInner::Vector { .. }, crate::TypeInner::Vector { .. }) => {
+                                crate::Expression::As {
+                                    expr: last_component,
+                                    kind: kind.ok_or(Error::BadTypeCast(word))?,
+                                    convert: true,
+                                }
+                            }
+                            _ => {
+                                return Err(Error::BadTypeCast(word));
+                            }
                         }
                     } else {
                         components.push(last_component);
+                        let ty = ctx.types.fetch_or_append(crate::Type { name: None, inner });
                         crate::Expression::Compose { ty, components }
                     }
                 }
--- a/tests/out/shadow.msl.snap
+++ b/tests/out/shadow.msl.snap
@@ -37,11 +37,9 @@ typedef float type7;
 
 typedef metal::float2 type8;
 
-typedef int type9;
+typedef metal::float3 type9;
 
-typedef metal::float3 type10;
-
-typedef bool type11;
+typedef bool type10;
 
 type7 fetch_shadow(
     type6 light_id,
@@ -57,7 +55,7 @@ type7 fetch_shadow(
 }
 
 struct fs_mainInput {
-    type10 in_normal_fs [[user(loc0)]];
+    type9 in_normal_fs [[user(loc0)]];
     type2 in_position_fs [[user(loc1)]];
 };
 
@@ -73,7 +71,7 @@ fragment fs_mainOutput fs_main(
     type5 sampler_shadow [[sampler(0)]]
 ) {
     fs_mainOutput output;
-    type10 color1 = type10(0.05, 0.05, 0.05);
+    type9 color1 = type9(0.05, 0.05, 0.05);
     type6 i = 0u;
     bool loop_init = true;
     while(true) {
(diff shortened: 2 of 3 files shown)


src/front/wgsl/mod.rs
```diff
@@ -856,10 +859,29 @@ impl Parser {
                     }
                     lexer.expect(Token::Paren(')'))?;
                     if components.is_empty() {
-                        crate::Expression::As {
-                            expr: last_component,
-                            kind: kind.ok_or(Error::BadTypeCast(word))?,
-                            convert: true,
+                        let last_component_inner = ctx.resolve_type(last_component)?;
+                        match (inner, last_component_inner) {
+                            (crate::TypeInner::Scalar { .. }, crate::TypeInner::Scalar { .. })
```
kvark, line 864: the type cast should only really work on scalars, vectors, and matrices. We can error on anything else

kvark (commented):
Thank you for addressing this! Just a few changes needed

src/front/wgsl/mod.rs
```diff
@@ -844,7 +844,10 @@ impl Parser {
                         ctx.constants,
                     )?;
                     let kind = inner.scalar_kind();
-                    let ty = ctx.types.fetch_or_append(crate::Type { name: None, inner });
+                    let ty = ctx.types.fetch_or_append(crate::Type {
+                        name: None,
+                        inner: inner.clone(),
```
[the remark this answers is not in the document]
Systemcluster: Moving this to after the component resolution changes the order the types are resolved, so it would need a change in the msl snapshots it seems.

src/front/wgsl/mod.rs
```diff
@@ -856,13 +855,25 @@ impl Parser {
...
-                        crate::Expression::As {
-                            expr: last_component,
-                            kind: kind.ok_or(Error::BadTypeCast(word))?,
-                            convert: true,
+                        let last_component_inner = ctx.resolve_type(last_component)?;
+                        match (&inner, last_component_inner) {
+                            (crate::TypeInner::Scalar { .. }, crate::TypeInner::Scalar { .. })
+                            | (crate::TypeInner::Matrix { .. }, crate::TypeInner::Matrix { .. })
+                            | (crate::TypeInner::Vector { .. }, crate::TypeInner::Vector { .. }) => {
+                                ctx.types.fetch_or_append(crate::Type { name: None, inner });
```
kvark, line 863: this isn't necessary. If the snapshot is changed, it's totally fine, as long as the produced code is correct. Our CI will validate that it's at least valid.

kvark (changes requested):

src/front/wgsl/mod.rs
Systemcluster: Without `fetch_or_append` there, the msl shader output differs from the snapshots in the tests. The typedef names would change and, in particular, a `typedef int` would be missing in `shadow.msl`, which I don't think is correct?

Example:

```msl
   34    34 │
   35       │-typedef int type9;
         35 │+typedef metal::float3 type9;
   36    36 │
   37       │-typedef metal::float3 type10;
   38       │-
   39       │-typedef bool type11;
         37 │+typedef bool type10;
   40    38 │
```

Systemcluster: I pushed a new version with adjusted snapshots

kvark: The CI will yell at you if the typedefs are wrong, so just give it a try

kvark (approved):

PR #465, opened 2021-02-13
ImageStore statement in IR

kvark:


Napokue (approved):
LGTM, thank you! 

What did change that you removed the `crate::` or `spirv::` in some files? I believe we once talked about this and you said that you have a preference over using the full qualifier instead of shortening it.

Diggsey:
I tested using `ImageStore` on DX12 and got the following error:
> ERROR gfx_backend_dx12::device > D3DCompile error 80004005: C:\Users\diggs\Projects\lbfs\Shader@0x000002148BA21E10(14,5-53): error X3025: l-value specifies const object

I'm unsure if this is an issue with naga or with spirv-cross though (currently attempting to test in vulkan to see if the spv output works when used directly)

Diggsey:
The `textureStore` issue appears to be a problem with `spirv-cross`, as I can use the .spv with vulkan directly and it works.

However, I found an issue with `textureLoad`, where naga appears to be generating invalid SPIRV. It seems any usage of this function generates an .spv file which I cannot disassemble with `spriv-dis` due to the following error:

> error: 101: Id 14 is defined more than once

If I try to *use* the shader, I get a segfault.

kvark:
Thanks @Diggsey ! I'll look into this and make sure the generated SPIR-V is correct. If you have a shader in mind that would use select/switch/ imageStore, I'd be happy to add a snapshot test for it.

Diggsey:
@kvark Here's the code I was testing: it's basically the bare minimum texture load/store example, copying one image to another.

```wgsl

// CS inputs
[[builtin(global_invocation_id)]] var<in> global_id: vec3<u32>;

// CS bindings
[[group(0), binding(0)]]
var bound_src: [[access(read)]] texture_storage_2d<rgba32float>;
[[group(0), binding(1)]]
var bound_dst: [[access(write)]] texture_storage_2d<rgba32float>;

fn compute() {
    var coord: vec2<i32> = vec2<i32>(global_id.x, global_id.y);
    var value: vec4<f32> = textureLoad(bound_src, coord);
    textureStore(bound_dst, coord, value);
}
```

kvark:
@Diggsey thank you!
I was looking for more of a real-world example, actually. But I'll be happy to test your shader code locally, at least.

@Napokue thank you for the review!
> What did change that you removed the crate:: or spirv:: in some files? I believe we once talked about this and you said that you have a preference over using the full qualifier instead of shortening it.

It's the new warning I enabled about trivial casts. In the places I fixed the thing behind `crate::xxx` was already imported into the scope, so `crate::` was unnecessary.
Our code is not fully consistent on the code style, so I wrote https://github.com/gfx-rs/naga/wiki/Code-style to help.

src/back/spv/writer.rs
```diff
@@ -1510,6 +1594,44 @@ impl Writer {
...
+                    ir_function,
+                    coordinate,
+                    array_index,
+                    block,
+                    function,
+                )?;
+
+                let id = self.generate_id();
+                let mut instruction =
+                    Instruction::image_read(id, result_type_id, image_id, coordinate_id);
```
Diggsey, line 1616: It looks like `id` and `result_type_id` are transposed here, and that's causing the invalid output.

kvark: this is very helpful, thank you!

Diggsey:
When I made that change locally, my example ran successfully, so toy example though it may be, I think that's all that's required to implement cool things like Lattice-Boltzmann fluid simulation :)

kvark:
wow nice!
there are still fixes needed though, I'm seeing:
> error: line 40: Target of NonWritable decoration is invalid: must point to a storage image, uniform block, or storage buffer
  %bound_src = OpVariable %_ptr_UniformConstant_29 UniformConstant

Diggsey:
Weird, I'm running this exact code and it's working fine? 😕

kvark:
Ok, it's all fixed now
Note that your example should use 
> var coord: vec2<i32> = vec2<i32>(vec2<u32>(global_id.x, global_id.y));

kvark:
spirv-cross just ignores some of the flags, it's not validating, hence it seems to work for you.
anyway, I fixed everything. Thank you for quick testing! It's really making the development much easier ❤️

PR #466, opened 2021-02-14
Remove the Header from IR

kvark:
This is a cleanup PR. We had the the `Header` basically copied from SPIR-V when we were just starting.
Now it's clear that it doesn't make much sense. For example, "generator" is a property of the tool (i.e. self), not the shader itself. And the "version" doesn't make sense if our IR is still in active development.


kvark:
Note: I wasn't able to switch some of the snapshots to SPIR-V 1.3, because our `validate-spv` scripts aren't able to preserve the SPIR-V version, unfortunately, because `spirv-as` screws us over.

grovesNL:
> `validate-spv` scripts aren't able to preserve the SPIR-V version

If we need to avoid `spirv-as` we could probably include the original SPIR-V binary in or alongside the text snapshot and use the binary directly

src/back/msl/writer.rs
```diff
@@ -1339,7 +1338,10 @@ impl<W: Write> Writer<W> {
                 let separator = separate(last_used_global == Some(handle));
                 write!(self.out, "{}", INDENT)?;
                 tyvar.try_fmt(&mut self.out)?;
-                resolved.try_fmt_decorated(&mut self.out, separator)?;
+                if var.binding.is_some() {
```
grovesNL, line 1341: Is this change intentionally included in this PR?

grovesNL (approved):

src/back/msl/writer.rs
kvark: not really, I'll move it out

kvark:
> If we need to avoid spirv-as we could probably include the original SPIR-V binary in or alongside the text snapshot and use the binary directly

Another option we should try is forcefully extracting the SPIR-V version from the disassembly. After all, it's just a pattern match.

Issue #265, opened 2020-11-09
[spv-in] SPIRV Parsing Failures for BVE-Reborn and Rend3


cwfitzgerald:
Here all the current problems preventing bve-reborn and rend3 from getting full shader validation. I have updated both to the latest wgpu master, so I may have issues that were already fixed but not yet pulled all the way through.

### Rend3:

This is a bindless based renderer, so uses more advanced features.

- `UnsupportedInstruction(ExecutionMode, String)`
  [Shader Playground](http://shader-playground.timjones.io/e1732368c06338ab32096f943c7e8dc7)
  [SPIR-V](https://github.com/gfx-rs/naga/files/5511895/pre_cull.zip)
- `UnsupportedInstruction(EntryPoint, String)`
  [Shader Playground](http://shader-playground.timjones.io/096642565325c39bacb16b7984803edc)
  [SPIR-V](https://github.com/gfx-rs/naga/files/5511913/blit.zip)
- `UnsupportedCapability(ShaderNonUniform)`
  [Shader Playground](http://shader-playground.timjones.io/c1d45c1960ca7dd40234eaaae3caf0ad)
  [SPIR-V](https://github.com/gfx-rs/naga/files/5511941/depth.zip)
  This also has `RuntimeDescriptorArray` and `SampledImageArrayNonUniformIndexing` which are likely unsupported.

### BVE-Reborn

More traditional renderer:

- `UnsupportedInstruction(ExecutionMode, String)`
  [Shader Playground](http://shader-playground.timjones.io/12bf63f7b226a52271c3c20c3d67fd81)
  [SPIR-V](https://github.com/gfx-rs/naga/files/5512112/opaque.zip)

### Other Info

All other errors I believe to be duplicates of the above.

Systemcluster:
Two of these shaders (`pre_cull.spv` and `opaque.spv`) currently generate the following error:

```
...
[2021-02-14T17:54:15Z DEBUG naga::front::spv]           Return [1]
[2021-02-14T17:54:15Z DEBUG naga::front::spv::function] FunctionEnd
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidId(24)', examples\convert.rs:95:64
```

When `lookup_function.lookup(fun_id)` is called from `patch_function_calls` it's empty, so the lookup fails.
https://github.com/gfx-rs/naga/blob/7a124550d4f75a7089d1fe9d415a87e042fd88a8/src/front/spv/mod.rs#L1615

[...]

PR #467, opened 2021-02-14
[spv-in] support string instruction

Systemcluster:
Add support for the `OpString` instruction.

`shaderc` adds a `OpString` instruction with the source filename to generated SPIR-V, which causes otherwise well-validating shaders to not parse. (Ref. #265) 

This is a dummy implementation that basically ignores the result of the instruction. This is valid here, because [the spec](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpString) explicitly allows ignoring it, and other instructions where the result may be used (`OpLine` and `OpSource`) either don't parse yet (`OpLine`) or are ignored as well (`OpSource`).

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -1643,6 +1643,7 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                 Op::MemoryModel => self.parse_memory_model(inst),
                 Op::EntryPoint => self.parse_entry_point(inst),
                 Op::ExecutionMode => self.parse_execution_mode(inst),
+                Op::String => self.parse_string(inst),
                 Op::Source => self.parse_source(inst),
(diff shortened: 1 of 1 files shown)


src/front/spv/mod.rs
```diff
@@ -111,6 +111,7 @@ pub enum ModuleState {
     MemoryModel,
     EntryPoint,
     ExecutionMode,
+    String,
```
kvark, line 114: this shouldn't be a separate module state, it's the same state as `Source` in the spec

kvark (changes requested):
Thank you!
Just one implementation note to fix.
In the future, we'll want to have something in `Module` to store the sources and strings like this.

kvark (approved):
Thank you!

PR #468, opened 2021-02-14
Specify line endings in ron snapshot tests

Systemcluster:
On Windows, `ron::ser::PrettyConfig::default()` defaults to `\r\n` line endings, which caused the snapshot tests to fail (and print the whole file in the error log) because the existing snapshot is saved with `\n` line endings.

Explicitly specify `\n` to be the line endings to fix this.


kvark (approved):
Sweet, thank you!

PR #472, opened 2021-02-14
[glsl-out] use existing Analysis instead of implementing Visitor

kvark:
Based on #445 
Part of the larger move to use the intermediate processing artifacts in the backends.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -45,8 +45,8 @@ pub use features::Features;
 
 use crate::{
     proc::{
-        CallGraph, CallGraphBuilder, Interface, NameKey, Namer, ResolveContext, ResolveError,
-        Typifier, Visitor,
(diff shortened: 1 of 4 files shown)


src/back/glsl/mod.rs
```diff
@@ -341,7 +341,10 @@ impl<'a, W: Write> Writer<'a, W> {
     ///
     /// # Panics
     /// Might panic if the module is invalid
-    pub fn write(&mut self) -> Result<FastHashMap<String, TextureMapping>, Error> {
+    pub fn write(
+        &mut self,
+        analysis: &Analysis,
```
JCapucho, line 346: I think it would be better to store `analysis` in the `Writer` itself since we will use it in others parts (for example when writing functions)

JCapucho (approved):
Just one small change that i want to get your opinion on, but everything else looks good to me

Issue #452, opened 2021-02-12
Expression ownership


kvark:
It appears that our SSA expressions are different from SPIR-V in that they lack the "ownership" semantics. In SPIR-V an op belongs to a block, but can be used in other blocks that are dominated by this one.
I think we need some mechanism here in the IR in order to make the whole thing sound. Without this, doing transformations becomes a bit confusing.

kvark:
We need some kind of association between an expression and the owner block.
Ideas:
  1. Keep the ownership implicit, based on the first block that uses an expression.
    - it's not clear if this actually expresses the semantic we need, i.e. what if there is an expression we want to be available to children blocks but not used here?
    - not correct by construction, requires heavy validation.
  1. Put blocks into an `Arena`, and have each expression pointing to a `Handle<Block>`.
    - still doesn't let us easily answer a question "which expressions are introduced in the block?".
    - doesn't make IR correct by construction, puts more pressure on the validator.
  1. Have a `Vec<Handle<Expression>>` per block, listing expressions that are introduced here.
    - also not correct by construction, requires serious validation.
    - 2x more heap allocations for blocks.
  1. Move the expression arenas into blocks. Each block has a separate `Arena<Expression>`, but they are addressed by a pair of `(block_depth, Handle<Expression>)` in all the contexts.
    - prevents passes on the expressions that don't care about the block structure, such as the `Typifier` pass.
    - 2x more heap allocations for blocks.

kvark:
I've been thinking about this. I believe it's hugely important as it affects the IR in a somewhat significant way, and there is a cost of delaying with this change. So it's somewhat critical atm.

So (1) is an instant no-go, because the expressions immediately used by a statement may either be used previously, or depend on other expressions. In any case, reconstructing the ownership from this is non-trivial, and we don't want backend and processors to deal with ti.

(2) seems awkward because we generally expect things in arenas to be re-usable. I.e. we expect more than a single unique handle living, otherwise the point of `Arena` is not fully utilized. Also, this approach doesn't allow scanning the statements efficiently, since we don't know which expressions they are introducing.

(3) and (4) are somewhat similar. I've identified the following issues with (4):
  - each reference to a handle takes more space, and is more verbose to work with. This may affect performance, since handles are most actively processed.
  - if anything goes wrong, debugging the "relative" handles that point nowhere is going to be very tough.

---
With this in mind, I think we should go with (3) and evolve it in the following directions:
[...]

PR #474, opened 2021-02-15
[glsl-out]: bake frequently used expressions

JCapucho:
Uses the results of analysis to bake frequently used expressions and avoid exponential explosions when writing

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -45,6 +45,7 @@ pub use features::Features;
 
 use crate::{
     proc::{
+        analyzer::{Analysis, FunctionInfo},
         CallGraph, CallGraphBuilder, Interface, NameKey, Namer, ResolveContext, ResolveError,
         Typifier, Visitor,
     },
@@ -171,6 +172,8 @@ struct FunctionCtx<'a, 'b> {
     expressions: &'a Arena<Expression>,
     /// A typifier that has already resolved all expressions in the function being written
     typifier: &'b Typifier,
+    /// Information about the current function being processed
+    info: &'b FunctionInfo,
 }
 
 impl<'a, 'b> FunctionCtx<'a, 'b> {
@@ -193,6 +196,20 @@ impl<'a, 'b> FunctionCtx<'a, 'b> {
             FunctionType::Function(handle) => &names[&NameKey::FunctionArgument(handle, arg)],
             FunctionType::EntryPoint(_) => unreachable!(),
         }
+    }
+
+    /// Helper function that returns wether the instruction should be baked or not
+    fn should_bake_expression(&self, expr: Handle<Expression>) -> bool {
+        let min_ref_count = match self.expressions[expr] {
+            Expression::Constant(_)
+            | Expression::FunctionArgument(_)
+            | Expression::GlobalVariable(_)
+            | Expression::LocalVariable(_)
+            | Expression::Load { .. } => !0,
+            _ => 2,
+        };
+
+        self.info[expr].ref_count > min_ref_count
     }
 }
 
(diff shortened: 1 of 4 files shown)


kvark:
Will need to be rebased on #472 when merged

src/back/glsl/mod.rs
```diff
@@ -844,6 +878,31 @@ impl<'a, W: Write> Writer<'a, W> {
             writeln!(self.out, ";")?
         }
 
+        // Bake expressions, this prevents exponential explosions
+        for expr in func
+            .expressions
```
kvark, line 883: I don't think we can go through all function expressions and create local variables for them in one place. For example, if an expression is introduced within the loop (related to #452), the relevant temporary variable needs to be within the loop.

src/back/glsl/mod.rs
```diff
@@ -514,8 +542,9 @@ impl<'a, W: Write> Writer<'a, W> {
     /// - If type is either a image or sampler
     /// - If it's an Array with a [`ArraySize::Constant`](crate::ArraySize::Constant) with a
     /// constant that isn't [`Uint`](crate::ConstantInner::Uint)
-    fn write_type(&mut self, ty: Handle<Type>) -> BackendResult {
-        match self.module.types[ty].inner {
+    /// - If the type is a struct and the `handle` is None
+    fn write_type(&mut self, ty: &TypeInner, handle: Option<Handle<Type>>) -> BackendResult {
```
kvark, line 546: Technically this should accept either a handle, or a reference to the type inner, so it could work with something like the thing that `get_handle()` returns: `Result<Handle<crate::Type>, &crate::TypeInner>`
That would allow call sites to skip doing the silly thing as `(&self.module.types[base].inner, Some(base))` (that would simply be `Ok(base)`. For clearest semantics we could define our own enum instead of using the `Result`:
```rust
enum ExpressionType<'a> {
  Handle(Handle<Expression>),
  Value(&'a TypeInner),
}
```

src/back/glsl/mod.rs
```diff
@@ -1168,20 +1227,30 @@ impl<'a, W: Write> Writer<'a, W> {
     ///
     /// # Notes
     /// Doesn't add any newlines or leading/trailing spaces
-    fn write_expr(&mut self, expr: Handle<Expression>, ctx: &FunctionCtx<'_, '_>) -> BackendResult {
+    fn write_expr(
+        &mut self,
+        expr: Handle<Expression>,
+        ctx: &FunctionCtx<'_, '_>,
+        bake: bool,
```
kvark, line 1234: I don't fully understand how this `bake` bool interops with `should_bake_expression` function. What if we don't have `bake` here and instead just always check for `should_bake_expression`? Would that introduce any issues?

JCapucho: Yes because when writing the baked expressions it would write a reference to itself
```glsl
vec4 _naga_expression_x = _naga_expression_x;
```

src/back/glsl/mod.rs
JCapucho: I've been thinking about this and working around this now seems to be a waste of time since the issue is so high priority and we would have to change this later, I think it's better to wait for us to solve the ownership issue and return to this later

kvark: This code is not going to be correct in most programs though. I.e. Anything that's in the loop can't be baked here. So it doesn't appear to be reasonable to try landing this as a temporary solution even. Maybe it's not that high a priority? Currently the problem is that some IRs may cause the backend to explode, but at least it's correct all the time.

JCapucho: I'm not saying to land this, I'm saying we should close this PR, figure out the owner ship problem and then fix the backends

kvark: yep, sounds reasonable!
I'm going to start working on the ownership rewrite tonight.

PR #476, opened 2021-02-15
[wgsl] support nested type parameters

### LINKED ISSUE #415: WGSL: Array of Vec seems to cause a parser error
For example: if I have a type alias as follows:

```
type PixelArr = array<vec4<f32>>;
```

I get an error as follows:

> Failed to parse WGSL code: error while parsing WGSL in scopes [TypeDecl, TypeDecl] at line .. pos ..: unexpected token ShiftOperation('>'), expected paren

Currently, I have a workaround for this by using an intermediate type alias:

```
type Pixel = vec4<f32>;
type PixelArr = array<Pixel>;
```

However, it should ideally still be possible to specify `array<vec4<f32>>`

Systemcluster:
Adds a ~~context to the lexer to keep track of the type parameter nesting level~~ `next_generic` method to the lexer that skips parsing shift operations.

This could be kept fairly simple ~~and limited to the lexer~~ because WGSL doesn't allow operations inside type parameters.

Tested on this shader sample:

```wgsl
  var x: array<vec2<f32>>;
  var y: array<array<vec2<f32>>>;

  var a: i32 = 12 << 1;
  
  if (a < 10 >> 3) {
    a = 10;
  }

  var b: i32 = 10 < 12 >> 1;
```

Resolves #415

### DIFF
--- a/src/front/wgsl/lexer.rs
+++ b/src/front/wgsl/lexer.rs
@@ -57,7 +57,7 @@ fn consume_number(input: &str) -> (Token, &str) {
     }
 }
 
-fn consume_token(mut input: &str) -> (Token<'_>, &str) {
+fn consume_token(mut input: &str, generic: bool) -> (Token<'_>, &str) {
     let mut chars = input.chars();
     let cur = match chars.next() {
         Some(c) => c,
@@ -84,9 +84,9 @@ fn consume_token(mut input: &str) -> (Token<'_>, &str) {
         '<' | '>' => {
             input = chars.as_str();
             let next = chars.next();
-            if next == Some('=') {
+            if next == Some('=') && !generic {
                 (Token::LogicalOperation(cur), chars.as_str())
-            } else if next == Some(cur) {
+            } else if next == Some(cur) && !generic {
                 (Token::ShiftOperation(cur), chars.as_str())
             } else {
                 (Token::Paren(cur), input)
@@ -174,7 +174,20 @@ impl<'a> Lexer<'a> {
     pub(super) fn next(&mut self) -> Token<'a> {
         let original_len = self.input.len();
         loop {
-            let (token, rest) = consume_token(self.input);
+            let (token, rest) = consume_token(self.input, false);
+            self.input = rest;
+            if token != Token::Trivia {
+                let _bytes_read = original_len - self.input.len();
+                return token;
+            }
+        }
+    }
+
+    #[must_use]
+    pub(super) fn next_generic(&mut self) -> Token<'a> {
+        let original_len = self.input.len();
+        loop {
+            let (token, rest) = consume_token(self.input, true);
             self.input = rest;
             if token != Token::Trivia {
                 let _bytes_read = original_len - self.input.len();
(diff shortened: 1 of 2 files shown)


grovesNL:
Thanks! Could this be simplified if we did it in the parser instead? IMO it would be nice to avoid tracking state in the lexer, and potentially avoid some amount of extra work/improve error handling if we do it in the parser (i.e. we'd only try to handle nested types when we're currently trying to parse a type).

We also might need some clarification in the WGSL specification based on https://github.com/gfx-rs/naga/issues/415#issuecomment-774584572

Systemcluster:
The spec has example code [with e.g. `array<vec2<f32>>`](https://gpuweb.github.io/gpuweb/wgsl.html#module-scope-variables) in lots of different places, so in its current state this should be supported to conform. I'd also be somewhat opposed to changing the spec to disallow `>>` in type expressions just to slightly simplify parsing.

Tracking this in the parser instead of the lexer should be possible, I'll see how that would look.

Systemcluster:
Updated the PR to keep state tracking out of the lexer and use the information already available in the parser.

It's actually *much* simpler this way and avoids tracking state at all.

src/front/wgsl/lexer.rs
```diff
@@ -215,6 +228,32 @@ impl<'a> Lexer<'a> {
         }
     }
 
+    pub(super) fn expect_generic(&mut self, expected: Token<'a>) -> Result<(), Error<'a>> {
```
kvark, line 231: can we refactor this into `fn expect_generic(&mut self, paren: char) -> ...`? Since it's only called on `Token::Paren`, specifying it seems redundant

src/front/wgsl/lexer.rs
```diff
@@ -215,6 +228,32 @@ impl<'a> Lexer<'a> {
...
+        let token = self.next_generic();
+        if token == expected {
+            Ok(())
+        } else {
+            let description = match expected {
```
kvark, line 236: this whole match should be removed given that we know it's a paren

src/front/wgsl/lexer.rs
Systemcluster: In this form it's consistent with `fn expect(&mut self, expected: Token<'a>)`, analogous to `next` and `next_generic`, and the implementation details are internal to the lexer with the only difference being if `true` or `false` are passed to `consume_token`; don't you think it's cleaner this way?

src/front/wgsl/lexer.rs
Systemcluster: Alternatively we could extract the description match into a separate function here to avoid the duplication. I've pushed that version first, let me know what you think. Otherwise I'll change it to your recommendation.

src/front/wgsl/lexer.rs
kvark: It's a lot more code. We don't want to be generic where we don't have to. This could be a very specific thing called `expect_generic_paren` for clarity.

Systemcluster: It was only a lot more code because the match was duplicated from `expect` :) It should look much better now I think.

kvark: Are you anticipating `expect_generic` to be used for anything *but* parens?
If not, then `expect_generic` doesn't need to work with any `Token`, and consequently the `describe_expect` isn't going to be needed (although I have nothing against it)

Systemcluster: I don't think it would be necessary for anything else. I just prefer to keep a consistent API 😄 I'll change it to `expect_generic_paren` for clarity then, it can always be changed in the future if needed.

kvark: thank you! I don't want us to over-engineer this.
General rule is start to abstract once you have 2-3 repeating cases, not 1.

kvark (approved):
Thank you! "+32 -10" looks great to me :)

PR #477, opened 2021-02-16
Set default SPIR-V version in convert example

Systemcluster:
After #466 the convert example panicked with `target SPIRV-0.0 is not supported` in the `spv-out` path since it was set to `(0, 0)` with the default constructor.


kvark (approved):
Nice!

PR #478, opened 2021-02-16
Miscellaneous fixes for parsing Dota shaders

kvark:
More steps towards #409 
The `Interface` optimization is particularly interesting. When traversing the expressions, we only want to visit each expression once, and only when looking at the block that introduces it. If we had the ownership established in the IR like #452 suggests, we'd not need to keep the bitmask here.


Napokue (approved):
Thank you, changes look good to me. I like the optimization that you have made in the Interface!

PR #482, opened 2021-02-16
Custom debug implementations

### LINKED ISSUE #481: Custom implementation of Debug for Module
I think there is a real opportunity here to debug-print the module using expression trees. It would make it much nicer to look at. Besides, what is currently printed is pretty much equivalent to the serialization implementaiton, so we aren't going to lose this view, it was redundant in fact.

kvark:
Fixes #481

I'm not totally sure it's worth it. There is quite a bit of code involved, and there is going to be a problem if expressions start getting repeated a lot...


kvark:
I believe this is inferior to #502

PR #488, opened 2021-02-17
Add MIT license

### LINKED ISSUE #298: MIT License isn't present in project
The project was changed to be dual licensed under MIT or Apache 2.0 with https://github.com/gfx-rs/naga/commit/70054b7693f67e1cb110f8bd041b5599f6f02072, but no MIT license was actually attached to the project. It is noted as having an MIT license in the projects Cargo.toml, but I think the actual license needs to be in the project too

Systemcluster:
Resolves #298

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,0 +1,19 @@
+Copyright (c) [yyyy] [name of copyright owner]
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


LICENSE-MIT
```diff
@@ -0,0 +1,19 @@
+Copyright (c) [yyyy] [name of copyright owner]
```
Systemcluster, line 1: This was kept as-is and not filled in to match the content of `LICENSE-APACHE`

kvark (approved):
thank you!

PR #491, opened 2021-02-17
Move global use flags into FunctionInfo

### LINKED ISSUE #486: Move global usage flags to the Analysis
This is one of the things that belong to "artifacts" of the IR analysis. We used to keep it in the IR itself, but we can now move it to `FunctionInfo` instead.
The tricky part here is that figuring out the global use may require recursive descent into the expression tree, while currently Analysis is only produced by a linear scan.

### LINKED ISSUE #487: Global usage flags are not correctly derived
Related to #486
I don't think we correctly identify things like `buffer.array[a+4] = b`. Here, the LHS expression is `Access`, not the global itself. We need to make it so `buffer` is identified as being written to, and `a` with `b` are read.

kvark:
Closes #486
Fixes #487


src/proc/analyzer.rs
```diff
@@ -52,14 +66,35 @@ pub struct SamplingKey {
...
     pub control_flags: ControlFlags,
     pub ref_count: usize,
+    assignable_global: Option<Handle<crate::GlobalVariable>>,
 }
 
 pub struct FunctionInfo {
     pub control_flags: ControlFlags,
+    /// Vector of global variable usages.
+    ///
+    /// Each item corresponds to a global variable in the module.
```
JCapucho, line 76: nit: This comment seems to be out of place

src/back/glsl/mod.rs
```diff
@@ -405,18 +405,21 @@ impl<'a, W: Write> Writer<'a, W> {
...
-        for (handle, global) in self
-            .module
-            .global_variables
-            .iter()
-            .zip(&self.entry_point.function.global_usage)
-            .filter_map(|(global, usage)| Some(global).filter(|_| !usage.is_empty()))
-        {
+        for (handle, global) in self.module.global_variables.iter() {
+            let usage = ep_info[handle];
+            if usage.is_empty() {
```
JCapucho, line 419: nit: inlining the usage seems better since it won't pollute the namespace and overall just looks better (in my opininion)

JCapucho (commented):
Good job, gave it a light review, nothing seems wrong just some small nits and it could use some tests.

I think removing the `GlobalUse` from the IR makes the snapshots more readable so it's also a nice plus

kvark:
@JCapucho noted that we need a way to test the global usage now that it's a part of `Analysis`.
I'll be working on it now, merging in the meantime.

PR #495, opened 2021-02-18
[wgsl] allow empty for statement headers, and other compliance additions

Systemcluster:
The first commit allows for statement header components to be optional.

This parses now correctly:

```wgsl
for(;;) {
   ...
}
```

The second commit limits the allowed initializer contents to `variable_statement | assignment_statement | func_call_statement` instead of allowing any statement.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -62,6 +62,8 @@ pub enum Error<'a> {
     BadTypeCast(&'a str),
     #[error(transparent)]
     InvalidResolve(ResolveError),
+    #[error("invalid statement {0:?}, expected {1}")]
+    InvalidStatement(crate::Statement, &'a str),
     #[error("unknown import: `{0}`")]
     UnknownImport(&'a str),
     #[error("unknown storage class: `{0}`")]
@@ -1816,7 +1818,19 @@ impl Parser {
                 let initialization = if lexer.skip(Token::Separator(';')) {
                     None
                 } else {
-                    self.parse_statement(lexer, context.reborrow(), is_uniform_control_flow)?
+                    let statement =
+                        self.parse_statement(lexer, context.reborrow(), is_uniform_control_flow)?;
+                    match statement {
+                        Some(crate::Statement::Store { .. })
+                        | Some(crate::Statement::Call { .. }) => statement,
+                        Some(other) => {
+                            return Err(Error::InvalidStatement(
+                                other,
+                                "variable, assignment or function call",
+                            ))
+                        }
+                        None => None,
+                    }
                 };
                 let condition = if lexer.skip(Token::Separator(';')) {
                     None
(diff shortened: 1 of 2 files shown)


src/front/wgsl/mod.rs
```diff
@@ -1813,16 +1815,37 @@ impl Parser {
                 let mut block = Vec::new();
 
                 lexer.expect(Token::Paren('('))?;
-                let initialization =
-                    self.parse_statement(lexer, context.reborrow(), is_uniform_control_flow)?;
-                let condition = self.parse_general_expression(lexer, context.as_expression())?;
+                let initialization = if lexer.peek() != Token::Separator(';') {
```
kvark, line 1818: we shouldn't ever use the `peek()` followed by `next()`. Instead, you can do `let initialization = if lexer.skip(..) { None } else { ... parse_statement() };`

src/front/wgsl/mod.rs
```diff
@@ -1843,19 +1866,26 @@ impl Parser {
...
-                let mut body = [crate::Statement::If {
-                    condition,
-                    accept: Vec::new(),
-                    reject: [crate::Statement::Break].to_vec(),
-                }]
-                .to_vec();
+                let mut body = Vec::new();
+                if let Some(condition) = condition {
+                    body.push(crate::Statement::If {
+                        condition: context.expressions.append(crate::Expression::Unary {
```
kvark, line 1881: why is this better than the previous variant? It looks like we are adding an extra expression `Unary` without actually reducing any other parts

kvark (commented):
Thank you for addressing this!
A few small concerns about the code

src/front/wgsl/mod.rs
Systemcluster: This way it conforms literally to the loop transformation example [here](https://gpuweb.github.io/gpuweb/wgsl.html#for-statement). I wasn't sure about this either (- how normative are the examples?) - I'll happily remove that commit if preferred.

kvark: These IRs are equivalent anyway, we aren't required to be literal here. If the difference is 1 expression, we should pick IR that has less of them.

src/front/wgsl/mod.rs
```diff
@@ -1813,16 +1815,38 @@ impl Parser {
...
-                if let Token::Word(ident) = lexer.peek() {
+                let initialization = if lexer.skip(Token::Separator(';')) {
+                    None
+                } else {
+                    match self.parse_statement(
+                        lexer,
+                        context.reborrow(),
+                        is_uniform_control_flow,
+                    )? {
+                        statement @ Some(crate::Statement::Store { .. }) => statement,
```
kvark, line 1826: it seems to me that the fancy pattern matching here is not necessary.
Instead we could have simpler code by just doing `let statement = self.parse_statement(...); match statement {... }`

kvark (approved):
One last nit, and we are good to go!

kvark (approved):

PR #496, opened 2021-02-18
Remove the call graph processor

JCapucho:
Uses the `GlobalUse` in glsl-out to check for function availability


src/back/glsl/mod.rs
```diff
@@ -126,14 +123,16 @@ impl fmt::Display for Version {
...
     pub version: Version,
-    /// The name and stage of the entry point
+    /// The stage of the entry point
+    pub shader_stage: ShaderStage,
+    /// The name of the entry point
     ///
     /// If no entry point that matches is found a error will be thrown while creating a new instance
     /// of [`Writer`](struct.Writer.html)
-    pub entry_point: (ShaderStage, String),
+    pub entry_point: &'a str,
```
kvark, line 135: I'm not sure this is needed to borrow. This makes it impossible to deserialize `Options`, and we'd ideally want to use it directly in those `.param.ron` files.

src/proc/analyzer.rs
```diff
@@ -91,6 +91,9 @@ impl FunctionInfo {
     pub fn expression_count(&self) -> usize {
         self.expressions.len()
     }
+    pub fn global_uses(&self) -> &[GlobalUse] {
```
kvark, line 94: Can we perhaps expose this as a more high-level methiod? Something like
```rust
impl FunctionInfo {
  pub fn dominates_global_use(&self, other: &Self) -> bool {...}
}
```

kvark (commented):
Looks good overall, just a few notes to address!

src/back/glsl/mod.rs
JCapucho: It's possible to deserialize a struct with borrows

kvark: My concern here, and it's a minor-ish concern, is that you have to carry that `'options` lifetime through the code. It's polluting the definitions, and it doesn't get us anything. So the old choice with `String` was both simpler and more in-line with other backends.

kvark (approved):
Thank you!

PR #497, opened 2021-02-19
Expression Ownership: A `Block` lists the expressions to emit

### LINKED ISSUE #452: Expression ownership

kvark:
Closes #452
TL;DR: every block carries a list of expression handles that it introduces into the scope for all other blocks dominated by it.
There is validation in place. The idea is that it should be straightforward for the front-ends to collect this info on the fly, but it's impossible to derive from a module in a general sense.


kvark:
Found an issue with assignments:
```diff
--- a/tests/out/skybox.msl.snap
+++ b/tests/out/skybox.msl.snap
@@ -50,12 +50,12 @@ vertex vs_mainOutput vs_main(
     type4 tmp1_;
     type4 tmp2_;
     type unprojected;
+    type _expr24 = metal::float4(((static_cast<float>(tmp1_) * const_4f) - const_1f), ((static_cast<float>(tmp2_) * const_4f) - const_1f), const_0f, const
_1f);
     tmp1_ = (static_cast<int>(in_vertex_index) / const_2i);
     tmp2_ = (static_cast<int>(in_vertex_index) & const_1i);
-    type expr24 = metal::float4(((static_cast<float>(tmp1_) * const_4f) - const_1f), ((static_cast<float>(tmp2_) * const_4f) - const_1f), const_0f, const_
1f);
-    unprojected = (r_data.proj_inv * expr24);
+    unprojected = (r_data.proj_inv * _expr24);
     output.out_uv = (metal::transpose(metal::float3x3(metal::float3(r_data.view[0].x, r_data.view[0].y, r_data.view[0].z), metal::float3(r_data.view[1].x, r_data.view[1].y, r_data.view[1].z), metal::float3(r_data.view[2].x, r_data.view[2].y, r_data.view[2].z))) * metal::float3(unprojected.x, unprojected.y, unprojected.z));
-    output.out_position = expr24;
+    output.out_position = _expr24;
     return output;
 }
```
Basically, if there is a `Store` statement, followed by any statements using the result of it, we can't declare this expression (that we stored into) as being owned by the block... This is a fundamental problem with the semantic of ownership as implemented in this PR, so I need to figure out how to proceed.

Basically, it shows the fact that expressions and statements are tightly connected, and we can't separate them even at the block level.

kvark:
Closing in favor of #528

PR #507, opened 2021-02-21
Add struct instantiation support to wgsl-in

lachlansneff:
Add the ability to parse instantiations of user-defined types to wgsl. Previously, this errored with "unknown type".﻿


src/lib.rs
```diff
@@ -337,7 +337,7 @@ pub struct Type {
 }
 
 /// Enum with additional information, depending on the kind of type.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
```
kvark, line 340: We don't want this struct to be clone-able. Let's find another way of getting the logic you are implementing here

lachlansneff: Hmm, okay.

lachlansneff: Fixed.

src/front/wgsl/mod.rs
```diff
@@ -1520,8 +1520,67 @@ impl Parser {
...
-        })
+            name => {
+                if let Some(&handle) = self.lookup_type.get(name) {
+                    let ty = &type_arena[handle];
+                    if ty.name.as_deref() != Some(name) {
+                        return Err(Error::UnknownType(word));
+                    }
+
+                    // This should work for both user-defined structs and type redefinitions.
+                    let inner = match &ty.inner {
```
kvark, line 1531: Sorry, I should have clarified. We shouldn't clone here at all.
That is somewhat the reason I don't want to see `derive(Clone)` on the type inner - so that there is a barrier to write the logic like this. See my other comments for how to address this properly.

src/front/wgsl/mod.rs
```diff
@@ -838,14 +838,14 @@ impl Parser {
                 if let Some(expr) = self.parse_function_call_inner(lexer, word, ctx.reborrow())? {
                     expr
                 } else {
-                    let inner = self.parse_type_decl_impl(
+                    let ty = self.parse_type_decl_impl(
```
kvark, line 841: We should just call `parse_type_decl_name` here, which gives us `Handle`. Then we can do `type_arena[ty].inner` for that match below

kvark (changes requested):

src/front/wgsl/mod.rs
lachlansneff: Ah, I see, sorry didn't make that connection. Will fix later this week (currently in midterms).

src/front/wgsl/mod.rs
```diff
@@ -857,25 +858,41 @@ impl Parser {
...
-                                    convert: true,
-                                }
+                        // The issue here is that both `ctx.types[handle]` and `last_component_inner` borrow ctx (and one is mutable).
+                        // This isn't pretty.
+
+                        let last_component_ty = {
+                            let last_component_inner = ctx.resolve_type(last_component)?;
+                            match last_component_inner {
+                                crate::TypeInner::Scalar { .. } => Some(0),
+                                crate::TypeInner::Matrix { .. } => Some(1),
```
kvark, line 887: why is `Matrix` associated with `1`?

Cargo.toml
```diff
@@ -20,10 +20,11 @@ pomelo = { version = "0.1.4", optional = true }
 thiserror = "1.0.21"
 serde = { version = "1.0", features = ["derive"], optional = true }
 petgraph = { version ="0.5", optional = true }
+pp-rs = { git = "https://github.com/Kangz/glslpp-rs", rev = "4f2f72a", optional = true }
```
kvark, line 23: what happened here? maybe we are missing a proper rebase?

kvark (changes requested):

Cargo.toml
lachlansneff: Hmm, I may have screwed up the rebase. Let me try that again,

src/front/wgsl/mod.rs
lachlansneff: I just picked random numbers here, so we can check if the types are the same without running into borrow checking issues.

Cargo.toml
lachlansneff: This is in master already.

src/front/wgsl/mod.rs
```diff
@@ -876,25 +877,38 @@ impl Parser {
...
-                                    kind: kind.ok_or(Error::BadTypeCast(word))?,
-                                    convert: true,
-                                }
+                        // The issue here is that both `ctx.types[handle]` and `last_component_inner` borrow ctx (and one is mutable).
+                        // This isn't pretty.
+
+                        let last_component_ty = {
+                            let last_component_inner = ctx.resolve_type(last_component)?;
+                            match last_component_inner {
+                                crate::TypeInner::Scalar { .. } => Some(0),
```
kvark, line 886: can we make this nicer a bit? i.e. something like this defined nearby
```rust
impl crate::TypeInner {
  fn value_dimensions(&self) -> Option<NonZeroU8> {..}
}
```

src/front/wgsl/mod.rs
```diff
@@ -876,25 +877,38 @@ impl Parser {
...
                             }
+                        };
+                        let ty_type = match &ctx.types[handle].inner {
+                            crate::TypeInner::Scalar { .. } => Some(0),
+                            crate::TypeInner::Matrix { .. } => Some(1),
+                            crate::TypeInner::Vector { .. } => Some(2),
+                            _ => None,
+                        };
+
+                        match (ty_type, last_component_ty) {
```
kvark, line 899: we should just do `if ty_type.is_some() && ty_type == last_component_ty` instead of the match here

tests/in/struct-instantiation.wgsl
```diff
@@ -0,0 +1,10 @@
+
+struct Foo {
```
kvark, line 2: let's not put it here. Instead, add this to `src/front/wgsl/tests.rs`

kvark (changes requested):

kvark:
You need to do `cargo insta review` after `cargo test` to add these reftest changes

kvark (approved):

kvark:
Looks like we have a problem with the generated MSL (see the failed "validation-macos" job):
> Validating tests/out/shadow.msl.snap
<stdin>:1:1: error: expected unqualified-id

kvark:
@lachlansneff we need this more then ever now. Could we revive this PR please?

lachlansneff:
Yes sorry, I'm happy to try this again. I should be able to get back to it
this evening.

kvark:
@lachlansneff ping!

lachlansneff:
@kvark I'm so sorry. I've had a really busy few weeks, I'll try to get back to this and the stencil8 PR as soon as I can. If you need them really soon, I'm happy to hand them over.

kvark:
Np, we can wait :)

src/front/wgsl/mod.rs
```diff
@@ -1092,32 +1093,31 @@ impl Parser {
...
-                            )
-                            | (
-                                &crate::TypeInner::Vector { .. },
-                                &crate::TypeInner::Vector { .. },
-                            ) => crate::Expression::As {
-                                expr: last_component,
-                                kind: kind.ok_or(Error::BadTypeCast(word))?,
-                                convert: true,
-                            },
+                        ctx.resolve_type(last_component)?;
```
kvark, line 1096: why are we not using the result of this?

tests/out/shadow.spvasm.snap
```diff
@@ -22,12 +22,12 @@ OpMemberName %17 1 "pos"
 OpMemberName %17 2 "color"
 OpName %19 "Lights"
 OpMemberName %19 0 "data"
-OpName %24 "c_ambient"
```
kvark, line 25: I am confused as to why IR has changed here. The code doesn't look like it's necessarily changing the IR.
You can get the IR by running it this way:
```bash
cargo run --features wgsl-in -- tests/in/shadow.wgsl >shadow.txt
```

kvark:
@lachlansneff there is a few questions in the last review + needs a rebase now that cargo-insta is gone

src/front/wgsl/mod.rs
lachlansneff: I left a comment in the rebased code. Short version is that `resolve_type` mutably borrows ctx, but we need an immutable borrow of it, if we're going to borrow it to look up `handle` as well.

kvark (approved):
Thank you!
Going to play with it a bit before merging

kvark:
@lachlansneff I was looking for the reason why the IR changed, and it looks like the code registered some of the types that it didn't have to. This isn't a big problem in itself, but I tried to think about ways to preserve the old behavior and emit less types. I came up with using the `TypeResolution` enum that is already exposed by the typifier, and it brings back the test differences.

Another commit here is using the new syntax for 2 of the tests.

lachlansneff:
Thank you for putting up with me through this! I'm sure it would've taken you less time to do it yourself then you spent coaching me through it, and I'm grateful.

PR #509, opened 2021-02-21
[wgsl] Add parser error snapshot

grovesNL:
This starts to track parser errors for the WGSL frontend. The snapshots are inline (instead of separate files like shader snapshots) because the parser errors should generally remain small.

The benefits of error snapshots are basically the same as shader snapshots – it lets us verify that the error is returned correctly and that we avoid regressions as parts move around internally.


kvark (approved):
Is this inspired by Rust by any chance?

grovesNL:
Yeah there are a few compilers (including rustc) that use error snapshots like this

PR #510, opened 2021-02-22
Rewrite the type validation

kvark:
Instead of traversing the types from top to bottom, enforcing the `HOST_SHARED` flag as we go, we are now collecting the type capabilities bottom to top (naturally), and comparing against the needs for global variables and such.
The PR also adds extends the checked properties into `SIZED`, and has a general `DATA` bit for the non-handle types.

This helps to validate cases like https://github.com/gfx-rs/wgpu/issues/1223


Napokue (approved):
Love it, LGTM! If I understand it correctly, we are now going from the most outer type (a type that is not a dependency on another type), and go back through the dependency tree in reverse?

kvark:
Correct! This matches the way we traverse expressions as well.

PR #511, opened 2021-02-22
Remove mutation from the Call expression

kvark:
First step of https://github.com/gfx-rs/naga/issues/452#issuecomment-782784985

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -276,6 +276,8 @@ pub struct Writer<'a, W> {
     entry_point_idx: crate::proc::EntryPointIndex,
     /// Used to generate a unique number for blocks
     block_id: IdGenerator,
+    /// Set of expressions that have associated temporary variables
+    cached_expressions: FastHashMap<Handle<Expression>, String>,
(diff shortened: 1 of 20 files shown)


src/back/glsl/mod.rs
```diff
@@ -1682,14 +1693,9 @@ impl<'a, W: Write> Writer<'a, W> {
...
             }
-            // A `Call` is written `name(arguments)` where `arguments` is a comma separated expressions list
-            Expression::Call {
-                function,
-                ref arguments,
-            } => {
-                write!(self.out, "{}(", &self.names[&NameKey::Function(function)])?;
-                self.write_slice(arguments, |this, _, arg| this.write_expr(*arg, ctx))?;
-                write!(self.out, ")")?
+            Expression::Call(_function) => {
```
JCapucho, line 1696: nit: Since we won't use this could we use `_`

JCapucho (commented):
Overall looks good but we should have some documentation about what the barrier statements do

Napokue (approved):
LGTM!

PR #512, opened 2021-02-22
[spv-out] optimize if/else branching

kvark:
Now we aren't generating a "Label + Branch" pair for any empty block to branch on within if/else.


Napokue (approved):
LGTM, thank you!

PR #513, opened 2021-02-22
[glsl-in] Add initial support for glslpp-rs

pjoe:
Updates glsl-in to use [glslpp-rs](https://github.com/Kangz/glslpp-rs) for lexer and pre-processor


pjoe:
WASM size seems to go from ~559KB > 569KB (uncompressed) when using glslpp-rs. This seems worthwhile to me for getting a proper preprocessor :)

src/front/glsl/lex_pp.rs
```diff
@@ -0,0 +1,253 @@
...
+                match t {
+                    Ok(t) => Some(t),
+                    Err((err, loc)) => {
+                        meta.line = loc.line as usize;
+                        meta.chars.start = loc.pos as usize;
+                        //TODO: proper location end
+                        meta.chars.end = loc.pos as usize + 1;
+                        return Some(Token::Unknown((meta, format!("{:?}", err))));
+                    }
+                }
```
JCapucho, line 47: You could write this as
```rust
let pp_token = self.pp_buf.pop_front().or_else(|| {
   Some(match self.pp.next()? {
     Ok(t) => t,
     Err((err, loc)) => {
         meta.line = loc.line as usize;
         meta.chars.start = loc.pos as usize;
         //TODO: proper location end
         meta.chars.end = loc.pos as usize + 1;
         Token::Unknown((meta, format!("{:?}", err)))
     }
   })
});
```
You could also add a `?` at the end and you just need to wrap the result in `Some` instead of doing a `map`

JCapucho (commented):
Very exciting work, did a small review and left a comment about some styling issues, there also some issues that will block this PR but we have time to fix them:
- We can't rely on `glslpp-rs` until it has been published to `crates.io` because crates can't pull in git dependencies so we wouldn't be able to publish `naga`
- The location problems

pjoe:
@JCapucho great comments :)

Will try and figure out publishing to crates.io with @Kangz.

Location issue is already tracked here: https://github.com/Kangz/glslpp-rs/issues/2

src/front/glsl/lex_pp.rs
pjoe: nice, was struggling a bit with this part - had actually tried using `or_else` but of course `return` inside there wouldn't work.

pjoe: hmmm actually this won't work, as `Token::Unknown((meta, format!("{:?}", err)))` is a parser token, not a pp token :S

JCapucho: you're right completely missed the return still something like this is in my opinion better
```rust
    let pp_token = match self.pp_buf.pop_front() {
        Some(t) => t,
        None => match self.pp.next()? {
            Ok(t) => t,
            Err((err, loc)) => {
                meta.line = loc.line as usize;
                meta.chars.start = loc.pos as usize;
                //TODO: proper location end
                meta.chars.end = loc.pos as usize + 1;
                return Some(Token::Unknown((meta, format!("{:?}", err))));
            }
        }
    };
```
you wouldn't need the `map` with the closure this way and is more compact and readable in my opinion

Cargo.toml
```diff
@@ -20,10 +20,11 @@ pomelo = { version = "0.1.4", optional = true }
 thiserror = "1.0.21"
 serde = { version = "1.0", features = ["derive"], optional = true }
 petgraph = { version ="0.5", optional = true }
+pp-rs = { git = "https://github.com/Kangz/glslpp-rs", branch = "main", optional = true }
```
kvark, line 23: we need to specify the exact revision instead of the branch. Since we don't vendor a lockfile, the branch can move.

src/front/glsl/lex_pp.rs
```diff
@@ -0,0 +1,242 @@
+use super::{parser::Token, token::TokenMetadata, types::parse_type};
```
kvark, line 1: does this module need to be renamed? I understand that it's rewritten, but the `pp` suffix doesn't seem to be needed if we don't have `lex.rs` any more

src/front/glsl/lex_pp.rs
```diff
@@ -0,0 +1,242 @@
...
+use crate::FastHashMap;
+use pp_rs::{
+    pp::Preprocessor,
+    token::{Punct, Token as PPToken, TokenValue},
+};
+use std::collections::VecDeque;
+
+//#[derive(Clone, Debug)]
```
kvark, line 9: do we need this comment?

src/front/glsl/lex_pp.rs
```diff
@@ -0,0 +1,242 @@
...
+pub struct LexerPP<'a> {
+    pp: Preprocessor<'a>,
+    pp_buf: VecDeque<PPToken>,
```
kvark, line 12: should this be named `tokens`?

src/front/glsl/lex_pp.rs
```diff
@@ -0,0 +1,242 @@
...
+        let pp_token = match self.pp_buf.pop_front() {
+            Some(t) => t,
+            None => match self.pp.next()? {
+                Ok(t) => t,
+                Err((err, loc)) => {
+                    meta.line = loc.line as usize;
+                    meta.chars.start = loc.pos as usize;
+                    //TODO: proper location end
+                    meta.chars.end = loc.pos as usize + 1;
+                    return Some(Token::Unknown((meta, format!("{:?}", err))));
```
kvark, line 41: any reason we can't just propagate the PP error up? I.e. instead of doing `format!("{:?}", err))` could we just do put `err` directly in there (as a `#[source]`)?

src/front/glsl/lex_pp.rs
```diff
@@ -0,0 +1,242 @@
...
+                Punct::LeftAngle => Token::LeftAngle(meta),
+                Punct::RightAngle => Token::RightAngle(meta),
+                Punct::Semicolon => Token::Semicolon(meta),
+                Punct::Comma => Token::Comma(meta),
+                Punct::Colon => Token::Colon(meta),
+                Punct::Dot => Token::Dot(meta),
+                Punct::Equal => Token::Equal(meta),
+                Punct::Bang => Token::Bang(meta),
+                Punct::Minus => Token::Dash(meta),
+                Punct::Tilde => Token::Tilde(meta),
```
kvark, line 137: it seems rather awkward that most (if not all) token variants come with `meta`. Can we instead just place them side by side? I.e. `(Token, Meta)` or a structure with fields, etc

src/front/glsl/lex_pp.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+            _ => Token::Unknown((meta, format!("{:?}", pp_token.value))),
+        })
+    }
+}
+
+impl<'a> Iterator for LexerPP<'a> {
+    type Item = Token;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.next()
```
kvark, line 161: why do we need `fn next()` as an internal method if we are implementing the iterator anyway? Can we instead just move the logic in here?

src/front/glsl/lex_pp.rs
```diff
@@ -0,0 +1,242 @@
+use super::{parser::Token, token::TokenMetadata, types::parse_type};
+use crate::FastHashMap;
+use pp_rs::{
+    pp::Preprocessor,
+    token::{Punct, Token as PPToken, TokenValue},
+};
+use std::collections::VecDeque;
+
+//#[derive(Clone, Debug)]
+pub struct LexerPP<'a> {
```
kvark, line 10: similarly, if we don't have the regular `Lexer` any more, we can just call this struct `Lexer` instead of `LexerPP`.

src/front/glsl/lex_pp.rs
pjoe: this is (unfortunately) entirely controlled by pomelo

Cargo.toml
pjoe: would this work for now, or do we need it on crates.io?

kvark: we don't need it on crates until we need to publish Naga-0.4

pjoe:
Added issue for publishing glslpp-rs to crates.io: https://github.com/Kangz/glslpp-rs/issues/26

Cargo.toml
pjoe: added specific rev, resolving for now

src/front/glsl/lex_pp.rs
pjoe: Not quite sure I follow? Return type for iterator is `Option<Token>`, so wasn't sure how to propagate the error, except turning it into a string and wrapping in `Token::Unknown`

kvark: I'm talking about `Token::Unknown(meta, glslpp::Error)` variant. Can we do it?

src/front/glsl/lex_pp.rs
pjoe: Resolving for now - until we remove pomelo :)

src/front/glsl/lex_pp.rs
pjoe: ahh yeah that might work ... though would make parser directly dependent on glslpp 🤔 Is that what `#[source]` could help with?

kvark: The `glsl-in` enables `glslpp` right now in this PR, is this not the goal? So we already depend on glslpp here.

pjoe: Yeah, was more thinking on source level to have glslpp entirely encapsulated in `lex.rs`, but that is probably over complicating things at this point. I'll try making it work with passing the error on.

Kangz:
And it is published now!

kvark (approved):
ftr, publishing it was not a blocker here

PR #514, opened 2021-02-22
[wgsl] Track token span for unexpected errors

grovesNL:
These changes should be non-functional but will enable #380 for `Error::Unexpected`. Other error variants will also be able to include these token spans after this change, but I thought it might be useful to try it with `Unexpected` first to keep the diff small.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -14,10 +14,12 @@ use crate::{
 };
 
 use self::lexer::Lexer;
-use std::num::NonZeroU32;
+use std::{num::NonZeroU32, ops::Range};
 use thiserror::Error;
 
 const SPACE: &str = "                                                                                                    ";
+
+type TokenSpan<'a> = (Token<'a>, Range<usize>);
 
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub enum Token<'a> {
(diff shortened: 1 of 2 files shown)


src/front/wgsl/mod.rs
```diff
@@ -1199,20 +1201,20 @@ impl Parser {
                 let mut ready = true;
                 loop {
                     match lexer.next() {
```
kvark, line 1203: should we just put `.0` at the end here?

src/front/wgsl/mod.rs
```diff
@@ -792,13 +794,13 @@ impl Parser {
     ) -> Result<Handle<crate::Expression>, Error<'a>> {
         self.scopes.push(Scope::PrimaryExpr);
         let expression = match lexer.next() {
```
kvark, line 796: similarly, here

src/front/wgsl/mod.rs
```diff
@@ -2151,8 +2153,8 @@ impl Parser {
 
         // read items
         match lexer.next() {
```
kvark, line 2155: and here

src/front/wgsl/mod.rs
grovesNL: We need to capture the span in the `other` pattern in these cases, so we couldn't use `lexer.next().0` directly. Alternatively we could store `lexer.next()` temporarily (to use for `other`) and match on the `temp.0` part only

kvark: How about this - we return the same thing we used to return - the `Token`, but also the `Lexer` knows about the span of the last token it produced. So the `other` branch can always ask the lexer on what the current span is, but other code paths are unaffected.

grovesNL: Yeah sure we could do that. I personally prefer to group token and span together because we'll eventually want to use the spans in lots of places (e.g. errors using spans variable names, spans of function parameter names, the start of a scope, etc.) but `lexer.previous_span()` seems less convenient to use. It would also be nice to avoid the extra state in the lexer but it's not a big deal.

kvark: ok, sounds reasonable. Let's proceed with your approach, just need a rebase

grovesNL:
@kvark this is rebased now

kvark (approved):

PR #517, opened 2021-02-23
Validate signed integer array size constants

### LINKED ISSUE #516: Validation of constant sized array types: signed or unsigned
The [validator requires](https://github.com/gfx-rs/naga/blob/master/src/proc/validator.rs#L465) that a constant-sized array's size be a `uint`. This requires a `u` suffix on a literal used for this purpose (e.g. `array<vec2<f32>, 3u>`, but the tint compiler rejects a program with a `u` suffix on an array size. All [examples in the spec](https://gpuweb.github.io/gpuweb/wgsl.html#type-declarations) do not include the suffix.

We could change the array type parsing to produce a `UInt` when the literal fits, or we could say that spec and tint should require the `u` suffix, or perhaps another option. I'm happy to help with a change if a change is wanted.

acowley:
Until we have some amount of type inference, an integer literal used as a constant array size must have a `u` suffix. This is inconsistent with the surface syntax presented in the spec, and is a minor inconvenience.

For now, allow a signed integer array size to pass validation so that user code need not change when type inference deduces values of the correct type.

Fixes #516

### DIFF
--- a/src/proc/layouter.rs
+++ b/src/proc/layouter.rs
@@ -73,6 +73,14 @@ impl Layouter {
                                 width: _,
                                 value: crate::ScalarValue::Uint(value),
                             } => value as u32,
+                            // Accept a signed integer size to avoid
+                            // requiring an explicit uint
+                            // literal. Type inference should make
+                            // this unnecessary.
+                            crate::ConstantInner::Scalar {
+                                width: _,
+                                value: crate::ScalarValue::Sint(value),
+                            } => value as u32,
                             ref other => unreachable!("Unexpected array size {:?}", other),
                         },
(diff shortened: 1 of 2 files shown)


src/proc/layouter.rs
```diff
@@ -73,6 +73,14 @@ impl Layouter {
                                 width: _,
                                 value: crate::ScalarValue::Uint(value),
                             } => value as u32,
+                            // Accept a signed integer size to avoid
```
kvark, line 76: no need to explain it both here and in the validator. Only the latter is source of truth about what's possible.

kvark (approved):

kvark:
Thank you!

PR #519, opened 2021-02-23
[spv-in] fix OpString and OpLine parsing

### LINKED ISSUE #518: SpirV with debug information can't be parsed
I'm using shaderc (runtime) compiled spirv shaders in my wgpu based project for a while. With a recent wgpu-rs update it stopped working because wgpu now relies on Naga to extract entry points (even if no validation is active). Turns out that Naga currently can't parse spirv code if it contains debug information.

Repro steps:
* I'm using this simple glsl vertex shader as input
```
#version 450

out gl_PerVertex { vec4 gl_Position; };

const vec2 triPositions[3] = vec2[3](vec2(-1.0, -3.0), vec2(-1.0, 1.0), vec2(3.0, 1.0));

void main() { gl_Position = vec4(triPositions[gl_VertexIndex], 1.0, 1.0); }
```
* Compile it with debug information using shaderc to spirv (I have not yet tried if this is also the case with other compilers!)
  * Debug information enabled with [this](https://github.com/google/shaderc/blob/main/libshaderc/include/shaderc/shaderc.h#L300)
  * **Yields [this output](https://github.com/gfx-rs/naga/files/6031764/screentri.zip) which completely embeds the source shader**
* `cargo run --all-features --example convert -- c:\dev\naga\repro\screentri.spv test.ron`
* test.ron output looks like this
```
(
    types: [],
    constants: [
        (
            name: None,
            specialization: None,
            inner: Scalar(
                width: 4,
                value: Uint(0),
            ),
        ),// [0]
        (
            name: None,
            specialization: None,
            inner: Scalar(
                width: 4,
                value: Uint(1),
            ),
        ),// [1]
        (
            name: None,
            specialization: None,
            inner: Scalar(
                width: 4,
                value: Uint(2),
            ),
        ),// [2]
        (
            name: None,
            specialization: None,
            inner: Scalar(
                width: 4,
                value: Uint(3),
            ),
        ),
    ],
    global_variables: [],
    functions: [],
    entry_points: {},
)
```

kvark:
Fixes #518

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -599,6 +599,12 @@ impl<I: Iterator<Item = u32>> Parser<I> {
             log::debug!("\t\t{:?} [{}]", inst.op, inst.wc);
 
             match inst.op {
+                Op::Line => {
+                    inst.expect(4)?;
+                    let _file_id = self.next()?;
+                    let _row_id = self.next()?;
+                    let _col_id = self.next()?;
+                }
                 Op::Undef => {
                     inst.expect(3)?;
                     let _result_type_id = self.next()?;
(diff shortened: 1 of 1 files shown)


Napokue (approved):
LGTM! 

What I wonder is if we some time later want to parse the debug information, having it behind a flag would be great. Those instructions that can be skipped, and if they can be skipped, the logic that otherwise is getting executed can also be skipped. Like the spv-out does with the debug flag.

This is not something for now, but I was just thinking about it.

kvark:
I think that makes sense, yeah. Not sure how much time we'd win, but we can try to measure.

PR #521, opened 2021-02-24
Make the conversion tool to be the default binary target

kvark:
It's really painful (for me, at least) to always type `--example convert` whenever I run Naga for local testing. Unfortunately, there are issues preventing us from getting the same functionality without it being an example:
  - we'd like to read the parameters from RON, but having RON as a dependency (even an optional one) for Naga doesn't sound right at all
  - we'd like to see the console output via `env_logger`, but adding this dependency to Naga is also not desired

So this PR suggests us to eat the cost and accept the reduced functionality, while getting the usage more convenient.
Ideally, I'd love Rust to support binary-only dependencies, which would address this properly.


Napokue:
I have looked through the PR, it looks good for me, but I do not have enough experience in Rust to know the implications that these changes bring with it.

Personally, I use an IDE that have so called "build configurations" and I can create as many configurations that I like. However, the features I have is not something we can account for others have/want to have.

So for me still trying to participate in this PR: What is the downside with making the Convert a binary target, and what is the upside? As I see in the updated read me, the command only changed from `cargo run --package naga --features XXX XXX --example convert XXX` (the only I use in my build configurations) to `cargo run --features XXX XXX -- XXX`.

kvark:
The downsides are:
  - inability to read/write RON modules
  - inability to read RON parameters (instead, rely more on command line)
  - inability to output any logs to console (we don't want to link to env_logger)

pjoe:
Sounds sensible to me :) Just being able to `cargo run` feels more intuitive to me.
👍

Gordon-F:
How about transfer the current repo into the `workspace` and move `converter` into its own crate? Like `gfx` [hal-examples](https://github.com/gfx-rs/gfx/tree/master/examples) did. I don't know any downsides, except you can't pass any features in workspace root.

grovesNL (approved):
In the longer term it might be useful to consider making this binary a proper CLI, e.g. `naga-cli` that can be installed with `cargo install`. I guess we'd have to use a workspace for that

Napokue (approved):
Agreed with @grovesNL, LGTM!

kvark:
I agree it would be nice to `cargo install`, and in fact this PR allows one to do this, since cargo install includes the binary targets.
So let's proceed. Thank you for all the feedback!

Issue #520, opened 2021-02-23
Shader parsing gets stuck in `naga_traverse`


Wumpf:
Have not identified what it is exactly that it gets stuck on. Debugger shows a long recursive callstack in `naga_traverse`. Waited for a couple of minutes and it did not run into stackoverflow (so who knows maybe it just works if one waits long enough ;-))

Repro:
`cargo run --all-features --example convert -- c:\dev\naga\repro\shader.comp.spv test.ron `
[shader.comp.zip](https://github.com/gfx-rs/naga/files/6032082/shader.comp.zip)
Original shader combines in a few includes but [this](https://github.com/gfx-rs/naga/files/6032102/extrapolate_velocity.comp.txt) should give a good idea of the main structure

This is a somewhat repetitive compute shader. Compiled with `shaderc`

kvark:
[flow.Compute-main.dot.txt](https://github.com/gfx-rs/naga/files/6036683/flow.Compute-main.dot.txt)

PR #524, opened 2021-02-26
[spv] fix unsigned selector and bitcast

kvark:
Fixes the issue found in #520 related to the switch selection.
Complements #523 
cc @MatusT

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1555,10 +1555,6 @@ impl Writer {
                 kind,
                 convert,
             } => {
-                if !convert {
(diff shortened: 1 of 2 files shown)


Napokue (approved):
LGTM, thank you!

Question: Is the sint required for the IR only, or is it something SPIR-V requires? I guess it is the first one, as I have read in the SPIR-V spec that OpSwitch also accepts uint. I am just curious why this was causing problems 🤔

kvark:
IR validation currently requires this. SPIR-V doesn't distinguish between sint/uint types, it reinterprets them freely.

PR #528, opened 2021-02-27
Explicitly emit expressions by `Statement::Emit`

### LINKED ISSUE #452: Expression ownership

kvark:
Closes #452 
Closes #497 
Closes #474 

Similar to #497, but more flexible, efficient, and less intrusive.
*Flexibility*: emitting is done as a statement, so one can emit after `Statement::Store` to pick up the value. This was the major blocker for #497 .
*Efficiency*: whole ranges of expressions are emitted, instead of lists of handles. Besides, there is a group of expressions that is considered in scope from the start of a function.
*Intrusiveness*: this PR doesn't change `type Block`, working with the statements is unchanged, so things like SPIR-V control flow processing don't even need to care about the change.

TODO:
- [x] IR
- [x] validation
- [x] spv-in
- [x] wgsl-in
- [ ] ~~glsl-in~~ (leaving for a follow-up)
- [x] spv-out
- [x] msl-out
- [x] glsl-out


kvark:
(copying from the matrix for our records)
GLSL-in shoud roughly following `wgsl-in`:
1. Generate `Expression::Load` whenever we operate on a value of a variable (both locals and globals)
2. Track the expressions we produce (possibly using `Emitter` struct that's currently shared between spv-in and wgsl-in):
- flush (into `Statement::Emit`) before adding a statement using any of the expressions
- exclude the `Expression::Constant` and `Expression::Call` by flushing before we add them, and restarting the emitter afterwards

cc @pjoe @JCapucho halp plz

kvark:
Reviewing this is hard, and nobody has volunteered so far.
I take the responsibility for risking here, don't want to have this waiting longer.

Gordon-F:
This PR was introduced funny semicolon hell in glsl shaders (wgpu-rs boids, for example).

It would be nice if we also enable GLSL target for wgsl boid snapshot testing to prevent an error like this in the future.

```glsl
void main() {
    vec2 vPos;
    vec2 vVel;
    vec2 cMass;
    vec2 cVel;
    vec2 colVel;
    int cMassCount = 0;
    int cVelCount = 0;
    vec2 pos1;
    vec2 vel1;
    uint i = 0u;
    ;
    ;
    if((gl_GlobalInvocationID[0] >= 1500u)) {
        return;
    }
    ;
    vPos = _group_0_binding_1.particles[gl_GlobalInvocationID[0]].pos;
    ;
    vVel = _group_0_binding_1.particles[gl_GlobalInvocationID[0]].vel;
    ;
    cMass = vec2(0.0, 0.0);
    ;
    cVel = vec2(0.0, 0.0);
    ;
    colVel = vec2(0.0, 0.0);
    while(true) {
        ;
        if((i >= 1500u)) {
            break;
        }
        ;
        if((i == gl_GlobalInvocationID[0])) {
            continue;
        }
        ;
        pos1 = _group_0_binding_1.particles[i].pos;
        ;
        vel1 = _group_0_binding_1.particles[i].vel;
        ;
        if((distance(pos1, vPos) < _group_0_binding_0.rule1Distance)) {
            ;
            cMass = (cMass + pos1);
            ;
            ;
            cMassCount = (cMassCount + 1);
        }
        ;
        if((distance(pos1, vPos) < _group_0_binding_0.rule2Distance)) {
            ;
            colVel = (colVel - (pos1 - vPos));
        }
        ;
        if((distance(pos1, vPos) < _group_0_binding_0.rule3Distance)) {
            ;
            cVel = (cVel + vel1);
            ;
            ;
            cVelCount = (cVelCount + 1);
        }
        ;
        ;
        i = (i + 1u);
    }
    ;
    ;
    if((cMassCount > 0)) {
        ;
        ;
        cMass = ((cMass * (1.0 / float(cMassCount))) - vPos);
    }
    ;
    ;
    if((cVelCount > 0)) {
        ;
        ;
        cVel = (cVel * (1.0 / float(cVelCount)));
    }
    ;
    vVel = (((vVel + (cMass * _group_0_binding_0.rule1Scale)) + (colVel * _group_0_binding_0.rule2Scale)) + (cVel * _group_0_binding_0.rule3Scale));
    ;
    ;
    vVel = (normalize(vVel) * clamp(length(vVel), 0.0, 0.1));
    ;
    vPos = (vPos + (vVel * _group_0_binding_0.deltaT));
    ;
    ;
    if((vPos[0] < -1.0)) {
        ;
        vPos[0] = 1.0;
    }
    ;
    ;
    if((vPos[0] > 1.0)) {
        ;
        vPos[0] = -1.0;
    }
    ;
    ;
    if((vPos[1] < -1.0)) {
        ;
        vPos[1] = 1.0;
    }
    ;
    ;
    if((vPos[1] > 1.0)) {
        ;
        vPos[1] = -1.0;
    }
    ;
    _group_0_binding_2.particles[gl_GlobalInvocationID[0]].pos = vPos;
    ;
    _group_0_binding_2.particles[gl_GlobalInvocationID[0]].vel = vVel;
    return;
}
```

kvark:
@Gordon-F 
I'm aware of the semicolon issue. It's fairly easy to fix, I just didn't want to do it right away.
Problem is that the GLS: backend does indentation before any statement, but `Emit` can end up generating no sub-statements. So a solution would be to indent for each statement variant independently. If you want to solve this, please receive our big gratitude!

Note that we have the skybox GLSL snapshot, and it also shows this, so this isn't about preventing error like this in the future.

PR #529, opened 2021-02-27
[wgsl-in] Use codespan to report errors

### LINKED ISSUE #380: Consider codespan for the frontend error reporting

grovesNL:
Fixes #380 (at least for wgsl-in)

Only `Unexpected` includes span information for now. We can expand this to include all the other variants without much trouble but we could do it in separate PRs.

`StringErrorBuffer` is currently only used in `emit_to_string`, so we could make the fields of `ParseError` public and move `StringErrorBuffer` inside the tests module if we don't want to expose `emit_to_string` outside of tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 bitflags = "1"
 bit-set = "0.5"
+codespan-reporting = { version = "0.11.0", optional = true }
 fxhash = "0.2"
 log = "0.4"
 num-traits = "0.2"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -12,6 +12,7 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 bitflags = "1"
 bit-set = "0.5"
+codespan-reporting = "0.11.0"
```
kvark, line 15: this should be an optional dependency, enabled by `wgsl-in`

kvark (commented):
Neat! Just one thing here

kvark (approved):

PR #535, opened 2021-03-02
[glsl-out] Set workgroup size for compute shaders

Gordon-F:


kvark (approved):
Looks great! We just need to update the snapshots.
For this, after you ran `cargo test --all-features`, you need to follow-up with `cargo insta review` (uses "cargo-insta" that needs to be installed first)

kvark (approved):

PR #538, opened 2021-03-03
[wgsl-in] Expose location of parser errors

### LINKED ISSUE #532: Structured WGSL parsing errors
WebGPU API has the following structure:
```
[Serializable]
interface GPUCompilationMessage {
    readonly attribute DOMString message;
    readonly attribute GPUCompilationMessageType type;
    readonly attribute unsigned long long lineNum;
    readonly attribute unsigned long long linePos;
};
```
We'd need to map our `wgsl-in` errors into it, ideally.
cc @grovesNL

grovesNL:
Fixes #532 by exposing the line number and column of the first span in the error. The first span is assumed to be the most significant because we could potentially have many important spans in a single error.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -16,7 +16,7 @@ use crate::{
 use self::lexer::Lexer;
 use codespan_reporting::{
     diagnostic::{Diagnostic, Label},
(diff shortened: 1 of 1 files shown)


kvark (approved):
Nice, thank you!

PR #540, opened 2021-03-04
[wgsl-in] Add descriptions for more errors

grovesNL:
Add some initial descriptions for `BadInteger`, `BadFloat`, `BadScalarWidth`, and `BadAccessor`. I removed some of the fields that weren't being used in the new descriptions – we could expand these over time as we improve the descriptions/notes.

### DIFF
--- a/src/front/wgsl/lexer.rs
+++ b/src/front/wgsl/lexer.rs
@@ -1,4 +1,5 @@
 use super::{conv, Error, Token, TokenSpan};
+use std::ops::Range;
 
 fn _consume_str<'a>(input: &'a str, what: &str) -> Option<&'a str> {
     if input.starts_with(what) {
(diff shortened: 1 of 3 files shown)


src/front/wgsl/lexer.rs
```diff
@@ -1,4 +1,5 @@
 use super::{conv, Error, Token, TokenSpan};
+use std::ops::Range;
```
kvark, line 2: this may be a little tricky given that `crate::arena::Range` is often imported in other modules

src/front/wgsl/mod.rs
```diff
@@ -78,16 +116,16 @@ impl<'a> Error<'a> {
...
     Unexpected(TokenSpan<'a>, &'a str),
-    #[error("unable to parse `{0}` as integer: {1}")]
-    BadInteger(&'a str, std::num::ParseIntError),
-    #[error("unable to parse `{1}` as float: {1}")]
-    BadFloat(&'a str, std::num::ParseFloatError),
-    #[error("unable to parse `{0}{1}{2}` as scalar width: {3}")]
-    BadScalarWidth(&'a str, char, &'a str, std::num::ParseIntError),
-    #[error("bad field accessor `{0}`")]
-    BadAccessor(&'a str),
+    #[error("")]
```
kvark, line 121: Do we even need to implement `Error` here?

kvark (approved):
Thank you! Looking forward to see those new errors:)

src/front/wgsl/mod.rs
grovesNL: Yeah this is just temporary. If we don't have a full error message implemented for either of these variants, we currently fall back to the messages listed here. We can remove `Error` here once they've all been updated.

PR #541, opened 2021-03-04
[glsl-in] Expression ownership

pjoe:
Just copy pasted info for easy ref:

GLSL-in shoud roughly following wgsl-in:

1. Generate Expression::Load whenever we operate on a value of a variable (both locals and globals)
2. Track the expressions we produce (possibly using Emitter struct that's currently shared between spv-in and wgsl-in):
- flush (into Statement::Emit) before adding a statement using any of the expressions
- exclude the Expression::Constant and Expression::Call by flushing before we add them, and restarting the emitter afterwards


JCapucho:
Closing this since #889 already implements this for the new parser

PR #545, opened 2021-03-04
[glsl-out] Remove unnecessary parentheses in binary expression

Gordon-F:
Slightly improve readability

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1551,7 +1551,10 @@ impl<'a, W: Write> Writer<'a, W> {
                     None
                 };
 
-                write!(self.out, "{}(", function.unwrap_or(""))?;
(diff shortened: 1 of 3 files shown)


src/back/glsl/mod.rs
```diff
@@ -1551,7 +1551,10 @@ impl<'a, W: Write> Writer<'a, W> {
                     None
                 };
 
-                write!(self.out, "{}(", function.unwrap_or(""))?;
```
kvark, line 1554: what happens if we have `(4+5) * 6` today? would this now become `4 + 5 * 6`?

Gordon-F: Nice catch! Thank you!

Maybe I should find another way how we can improve readability :smile:

PR #546, opened 2021-03-05
First-class pointer semantics

### LINKED ISSUE #542: Real SSA based on pointer semantics for variables
Right now, the `LocalVariable` and `GlobalVariable` aren't really pointers. They are "transparent" values, they materialize only when `Load` is issued. This harms the type checking, which isn't able to properly assess the `Access` and `AccessIndex` semantics. It will be future-proof to have real pointers, such that `Load` expects one.

kvark:
Fixes #542

We talked about this with @jrmuizel and decided to try out an idea (after a series of other ideas that didn't work well - https://github.com/gfx-rs/naga/issues/542#issuecomment-790674102) where the type system has *another* pointer type for implicit value types. So it can only point to scalars or vectors. It allows us to keep the typifier roughly doing the same, while getting all the benefit from the real pointer type checks. This PR implements the approach, and I think it's good to land 🚀 

TL;DR: this PR allows the type system to differentiate between pointers (which have to be used only in `Access`, `AccessIndex`, and `Load` expressions, or the `Store` statement) and values. All the non-Handle variables are pointers now, aligning with SPIR-V.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ thiserror = "1.0.21"
 serde = { version = "1.0", features = ["derive"], optional = true }
 petgraph = { version ="0.5", optional = true }
 pp-rs = { git = "https://github.com/Kangz/glslpp-rs", rev = "4f2f72a", optional = true }
+#env_logger = "0.8" # uncomment temporarily for developing with `convert`
 
 [features]
 default = []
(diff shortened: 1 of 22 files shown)


Napokue (approved):
LGTM, I like this new concept that has been introduced in this PR!

PR #547, opened 2021-03-05
[glsl-out] Set precision for image type and fix structure type in statement

Gordon-F:
Before:
```glsl
Light_block_2 {
    mat4x4 proj;
    vec4 pos;
    vec4 color;
} _expr29 = _group_0_binding_1.data[i];
```

After:
```glsl
Light _expr29 = _group_0_binding_1.data[i];
```


kvark (approved):
Nice!

PR #549, opened 2021-03-07
Switch entry points from map to a vec

kvark:
The idea of having a map was to make it conflict-less by construction. However, it turned out to be too inconvenient for all involved, and also inconsistent towards everything else in the API. Making it a Vec seems much nicer now.


Napokue (approved):
LGTM, thank you!

PR #554, opened 2021-03-08
[glsl-out] Workaround lod array shadow as textureGrad

Gordon-F:
I've spied a solution from [SPIRV-Cross](https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L6699) :smile:

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1340,12 +1340,24 @@ impl<'a, W: Write> Writer<'a, W> {
             } => {
                 //TODO: handle MS
 
+                // textureLod on sampler2DArrayShadow and samplerCubeShadow does not exist in GLSL.
+                // To emulate this, we will have to use textureGrad with a constant gradient of 0.
+                let workaround_lod_array_shadow_as_grad =
+                    array_index.is_some() && depth_ref.is_some();
+
                 //Write the function to be used depending on the sample level
                 let fun_name = match level {
                     crate::SampleLevel::Auto | crate::SampleLevel::Bias(_) => "texture",
-                    crate::SampleLevel::Zero | crate::SampleLevel::Exact(_) => "textureLod",
+                    crate::SampleLevel::Zero | crate::SampleLevel::Exact(_) => {
+                        if workaround_lod_array_shadow_as_grad {
+                            "textureGrad"
+                        } else {
+                            "textureLod"
+                        }
+                    }
                     crate::SampleLevel::Gradient { .. } => "textureGrad",
                 };
+
                 write!(self.out, "{}(", fun_name)?;
 
                 // Write the image that will be used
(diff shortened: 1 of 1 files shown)


src/back/glsl/mod.rs
```diff
@@ -1341,11 +1341,21 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                let fun_name = match level {
+                let mut fun_name = match level {
                     crate::SampleLevel::Auto | crate::SampleLevel::Bias(_) => "texture",
                     crate::SampleLevel::Zero | crate::SampleLevel::Exact(_) => "textureLod",
                     crate::SampleLevel::Gradient { .. } => "textureGrad",
                 };
+
+                // textureLod on sampler2DArrayShadow and samplerCubeShadow does not exist in GLSL.
+                // To emulate this, we will have to use textureGrad with a constant gradient of 0.
+                let workaround_lod_array_shadow_as_grad =
```
kvark, line 1352: This should only be active when we actually want to use the `textureLod`. Currently in this code, it may end up replacing `texture` with `textureGrad` for no reason.

src/back/glsl/mod.rs
```diff
@@ -1399,6 +1413,10 @@ impl<'a, W: Write> Writer<'a, W> {
                     }
                 }
 
+                if fun_name == "textureGrad" && workaround_lod_array_shadow_as_grad {
```
kvark, line 1416: don't need to check the function name if we know the workaround is active

src/back/glsl/mod.rs
```diff
@@ -1385,7 +1395,11 @@ impl<'a, W: Write> Writer<'a, W> {
                     // Auto needs no more arguments
                     crate::SampleLevel::Auto => (),
                     // Zero needs level set to 0
-                    crate::SampleLevel::Zero => write!(self.out, ", 0")?,
+                    crate::SampleLevel::Zero => {
+                        if !workaround_lod_array_shadow_as_grad {
```
kvark, line 1399: shouldn't this condition wrap the whole `match level`?

kvark (changes requested):

Gordon-F:
@kvark Thank you for your review! I think I fixed all the requested changes.

src/back/glsl/mod.rs
kvark: It would be even better if we don't compare strings here. Strings are the product of the backend, it shouldn't feed them it its own logic, which operates on the stronger IR. So instead of comparing to string, this should do:
```rust
let workaround_lod_array_shadow_as_grad = array_index.is_some() && depth_ref.is_some();
              let mut fun_name = match level {
                  crate::SampleLevel::Auto | crate::SampleLevel::Bias(_) => "texture",
                  crate::SampleLevel::Zero | crate::SampleLevel::Exact(_) => {
if workaround_lod_array_shadow_as_grad { "textureGrad" } else { "textureLod" },
                  crate::SampleLevel::Gradient { .. } => "textureGrad",
              };
```
And then check for `workaround_lod_array_shadow_as_grad` in the another match we do

kvark (commented):
Thank you!
One last thing :)

src/back/glsl/mod.rs
```diff
@@ -1381,21 +1393,25 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                    crate::SampleLevel::Exact(expr) | crate::SampleLevel::Bias(expr) => {
-                        write!(self.out, ", ")?;
-                        self.write_expr(expr, ctx)?;
-                    }
-                    crate::SampleLevel::Gradient { x, y } => {
-                        write!(self.out, ", ")?;
-                        self.write_expr(x, ctx)?;
-                        write!(self.out, ", ")?;
-                        self.write_expr(y, ctx)?;
+                if workaround_lod_array_shadow_as_grad {
```
kvark, line 1396: sorry, this is incorrect. `workaround_lod_array_shadow_as_grad` is only valid for `SampleLevel::Zero` and `SampleLevel::Exact` variants

Gordon-F: Thank you! Should be fixed now 😄

Issue #553, opened 2021-03-08
[glsl-out] Return uniform reflection info to users


Gordon-F:
Currently, `glsl-out` `Writer` return only `TextureMapping` information. But for targets that don't have `EXPLICIT_LAYOUTS_IN_SHADER` feature, this is not enough. We should also return uniforms information.

For example (wgpu-rs [cube shader](https://github.com/gfx-rs/wgpu-rs/blob/master/examples/cube/shader.wgsl)):
```glsl
#version 300 es

precision highp float;

struct Locals {
    mat4x4 transform;
};

in vec4 _location_0;

in vec2 _location_1;

out vec2 _location_0_vs;

uniform Locals_block_0 {
    mat4x4 transform;
} _group_0_binding_0;

void main() {
    _location_0_vs = _location_1;
    gl_Position = (_group_0_binding_0.transform * _location_0);
    return;
}
```

On `gfx-rs` side, we get a uniform index block, but we don't know the real uniform name. Only IR name (`r_locals` in the current example).
```rust
n::BindingRegister::UniformBuffers => unsafe {
              let index = gl.get_uniform_block_index(program, name).unwrap();
              gl.uniform_block_binding(program, index, slot as _);
          },
```

Blocker for `WebGL 2.0` and `OpenGL ES 3.0` targets.

PR #555, opened 2021-03-08
[glsl-out] Return reflection info to users

Gordon-F:
Fixes https://github.com/gfx-rs/naga/issues/553

Tested on `cube` and `shadow` without `EXPLICIT_LAYOUTS_IN_SHADER` feature.


kvark (approved):
Sweet!

PR #556, opened 2021-03-09
Relax array and block structure validation

kvark:
This helps to work with SPIR-V shaders that have interface blocks, including the `gl_ClipDistance[]` arrays in them.


Napokue (approved):
LGTM, thank you!

PR #557, opened 2021-03-09
[spv-in] function calls and variable types as pointers

kvark:
The function call patching was wrong because it didn't work with forward declarations.

The variables as pointers thing is a required follow up to #546: the SPIR-V frontend was still considering everything to be non-pointer, unconditionally. This prevented use cases where a function argument is a pointer, and this is what `glslang` always generates (including wgpu-rs water example shaders). So this PR makes pointers really first-class in the frontend, at the cost of slightly increased complexity in the type logic.

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -1754,6 +1754,16 @@ impl<I: Iterator<Item = u32>> Parser<I> {
             }?;
         }
 
+        log::info!("Patching...");
+        // patch all the function calls
+        for (_, fun) in module.functions.iter_mut() {
+            self.patch_function_calls(fun)?;
+        }
+        for ep in module.entry_points.iter_mut() {
+            self.patch_function_calls(&mut ep.function)?;
+        }
+        self.lookup_function.clear();
+
         // Check all the images and samplers to have consistent comparison property.
         for (handle, flags) in self.handle_sampling.drain() {
             if !image::patch_comparison_type(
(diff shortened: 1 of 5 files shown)


src/front/spv/mod.rs
```diff
@@ -694,10 +693,12 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                     let base_id = self.next()?;
                     log::trace!("\t\t\tlooking up expr {:?}", base_id);
                     let mut acex = {
-                        let expr = self.lookup_expression.lookup(base_id)?;
+                        // the base type has to be a pointer, so we derefernce it here for the walthrough
```
Napokue, line 696: Nit: `walthrough` -> `walkthrough`? I think in technical terms: `traverse` or `traversal` would fit a little better.

src/front/spv/mod.rs
```diff
@@ -1754,6 +1758,16 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
 
+        log::info!("Patching...");
+        // patch all the function calls
+        for (_, fun) in module.functions.iter_mut() {
+            self.patch_function_calls(fun)?;
+        }
+        for ep in module.entry_points.iter_mut() {
+            self.patch_function_calls(&mut ep.function)?;
+        }
+        self.lookup_function.clear();
```
Napokue, line 1769: I always wondered why there are `clear` function calls used. Is this better that reinitializing the collection?

Napokue (approved):
Outside a very small comment, LGTM!

PR #558, opened 2021-03-09
More error information for uniformity analysis

kvark:
This is a big update to #445. Error reporting was really harsh, since it wasn't telling the user what part of the code requires uniformity, and what part doesn't let it happen. This PR adds this information to the errors. It also fixes the logic somewhere, since the water example no longer triggers an uniformity error.


kvark:
Did a thorough self-review and noticed a few things :) Going to proceed with gfx-13 tag after this.

PR #561, opened 2021-03-10
[glsl-out] Properly write array type

### LINKED ISSUE #560: [glsl-out] Properly write array type
Current implementation caused validation error with `glslangValidator`.

```glsl
readonly buffer Lights_block_1 {
    Light[] data;
} _group_0_binding_1;
```

Should be: `Light data[];`

Gordon-F:
Fix #560

Also this PR fix:
1. Remove redundant struct write
2. Enable `glsl-out` test with `shadow` shader

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -416,7 +416,16 @@ impl<'a, W: Write> Writer<'a, W> {
                 ref members,
             } = ty.inner
             {
-                self.write_struct(handle, members)?
+                // No needed to write a struct that also should be written as a global variable
+                let is_global_struct = self
+                    .module
+                    .global_variables
+                    .iter()
+                    .any(|e| e.1.ty == handle);
+
+                if !is_global_struct {
+                    self.write_struct(false, handle, members)?
+                }
             }
         }
 
@@ -571,11 +580,10 @@ impl<'a, W: Write> Writer<'a, W> {
             )?,
             // glsl has no pointer types so just write types as normal and loads are skipped
             TypeInner::Pointer { base, .. } => self.write_type(base)?,
-            // Arrays are written as `base[size]`
-            TypeInner::Array { base, size, .. } => {
-                let ty_name = &self.names[&NameKey::Type(base)];
-
-                write!(self.out, "{}", ty_name)?;
+            // GLSL arrays are written as `type name[size]`
+            // Current code is written arrays only as `[size]`
+            // Base `type` and `name` should be written outside
+            TypeInner::Array { base: _, size, .. } => {
                 write!(self.out, "[")?;
 
                 // Write the array size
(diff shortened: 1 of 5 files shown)


src/back/glsl/mod.rs
```diff
@@ -597,36 +605,12 @@ impl<'a, W: Write> Writer<'a, W> {
...
             // glsl structs are written as just the struct name if it isn't a block
-            //
-            // If it's a block we need to write `block_name { members }` where `block_name` must be
-            // unique between blocks and structs so we add `_block_ID` where `ID` is a `IdGenerator`
-            // generated number so it's unique and `members` are the same as in a struct
             TypeInner::Struct { block, ref members } => {
                 // Get the struct name
                 let name = &self.names[&NameKey::Type(ty)];
 
                 if block {
```
kvark, line 612: we should probably just match it here, i.e. `TypeInner::Struct { block: true, ref members }` instead of `if/else` inside

src/back/glsl/mod.rs
```diff
@@ -416,7 +416,16 @@ impl<'a, W: Write> Writer<'a, W> {
                 ref members,
             } = ty.inner
             {
-                self.write_struct(handle, members)?
+                // No needed to write a struct that also should be written as a global variable
```
kvark, line 419: could you clarify, why do we write the struct *there* in the first place if we can write it here? Doing everything in this loop seems a bit simpler.

tests/snapshots.rs
```diff
@@ -259,7 +259,7 @@ fn convert_wgsl_collatz() {
 #[cfg(feature = "wgsl-in")]
 #[test]
 fn convert_wgsl_shadow() {
-    convert_wgsl("shadow", Targets::SPIRV | Targets::METAL);
+    convert_wgsl("shadow", Targets::SPIRV | Targets::METAL | Targets::GLSL);
```
kvark, line 262: This is amazing!

src/back/glsl/mod.rs
Gordon-F: The first loop goes through all structs, the second one goes through all global variables. 

For example:
```wgsl
// Type {
//     name: Some(
//         "Light",
//     ),
//     inner: Struct {
//         block: true,
//         members: [
//             StructMember {
//                 name: Some(
//                     "proj",
//                 ),
//                 span: None,
//                 ty: [1],
//             },
//         ],
//     },
// },
[[block]]
struct Light {
    proj: mat4x4<f32>;
};

// Type {
//     name: Some(
//         "Lights",
//     ),
//     inner: Struct {
//         block: true,
//         members: [
//             StructMember {
//                 name: Some(
//                     "data",
//                 ),
//                 span: None,
//                 ty: [3],
//             },
//         ],
//     },
// },
[[block]]
struct Lights {
    data: [[stride(96)]] array<Light>;
};

// GlobalVariable {
//     name: Some(
//         "s_lights",
//     ),
//     class: Storage,
//     binding: Some(
//         Resource {
//             group: 0,
//             binding: 1,
//         },
//     ),
//     ty: [4],
//     init: None,
//     interpolation: None,
//     storage_access: LOAD,
// },
[[group(0), binding(1)]]
var<storage> s_lights: [[access(read)]] Lights;
```

Output GLSL:
```glsl
struct Light {
    mat4x4 proj;
};

readonly buffer Lights_block_0 {
    Light data[];
} _group_0_binding_1;
```

Both structs (`Light` and `Lights`) are `block`, but we don't need to write redundant struct for `Lights`, because it also written as buffer.

kvark (commented):
Talked about this on Matrix, and this PR will need to be followed up.
It doesn't allow the structure type to be used for anything else, other than the global buffer resource.

PR #563, opened 2021-03-11
Uniformity Analysis - take 3

kvark:
Fixes #562
The main feature of this iteration is that function uniformity is explicitly treated differently from the expression uniformity. We have a separate type, and it integrates nicely into the `process_block` logic, which used to return a tuple.
Another important change is that all the expression checking is now done at `Emit` time as opposed to the time where an expression is used.
The quad shader is modified to do the discard now. It's doing it legally, but the old analyzer failed to understand this.

PR #564, opened 2021-03-12
Uniformity take 4: strongly consider the requirements of emits only

kvark:
See previous iteration in #563 
The problem was that it made the requirements to be inherited. So a program like this would not validate properly:
```
  const color: vec4<f32> = textureSample(u_texture, u_sampler, v_uv); // requires uniform
  if (color.a == 0.0) {
    discard;
  }
  const premultiplied: vec4<f32> = color.a * color; // doesn't require uniform, but it depends on an expression that does
  o_color = premultiplied;
```
Mechanically, since the `color` was emitted in uniform control flow, we can use it in the dependencies even if they are not uniform. This PR refactors the whole thing to take this into account. Also introduces a neat `UniformityRequirement` enum that we can return in errors for better clarity.


Napokue (approved):
Changes look good to me, thanks! Although already discussed in Matrix, I approve the code, but I am no where knowledgeable enough about this subject of uniformity. Pure implementation details I cannot give a review for.

PR #569, opened 2021-03-13
[glsl-out] fix cached expressions

### LINKED ISSUE #565: GLSL backend misses some of the emits
The quad shader translates into:
```rust
void main() {
    if((texture(_group_0_binding_0, vec2(_location_0_vs))[3] == 0.0)) {
        discard;
    }
    _location_0 = texture(_group_0_binding_0, vec2(_location_0_vs));
    _location_0 = (texture(_group_0_binding_0, vec2(_location_0_vs))[3] * texture(_group_0_binding_0, vec2(_location_0_vs)));
    return;
}
```
This is because of this check in the emitting code:
```rust
if let Ok(ty_handle) = ctx.typifier.get_handle(handle) {
```

kvark:
Fixes #565

The problem with emitting here was that it only worked on expressions that have registered types. A lot of expressions don't.

Related, this PR also includes a small change in all the frontends to force the caching of `Derivative` expression, so that it gets tied to the region and doesn't slip deeper where the control flow may be less uniform.

Finally, `cached_expressions` needed to be reset between the functions, so that we don't hit it accidentally.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -539,17 +539,17 @@ impl<'a, W: Write> Writer<'a, W> {
         self.collect_reflection_info()
     }
 
-    /// Helper method used to write non image/sampler types
+    /// Helper method used to write value types
     ///
     /// # Notes
     /// Adds no trailing or leading whitespace
     ///
     /// # Panics
-    /// - If type is either a image or sampler
+    /// - If type is either a image, a sampler, a pointer, or a struct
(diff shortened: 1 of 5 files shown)


JCapucho (approved):
LGTM
It would be better if we had `cached_expressions` in the `FunctionCtx` (since we wouldn't have to clear the map) but then there would be more allocations so I'm happy to keep it has is

kvark:
Yep, let's keep this on the radar!

PR #572, opened 2021-03-14
[spv-out] implement derivatives

lain-dono:
Basic implementation of derivatives (`dpdx`, `dpdy` and `fwidth`).


kvark (approved):
Thank you!

Issue #589, opened 2021-03-17
[glsl-out] Naga generates shaders with no matching outputs


Gordon-F:
wgpu-rs [cube shader](https://github.com/gfx-rs/wgpu-rs/blob/master/examples/cube/shader.wgsl):

```glsl
#version 320 es

precision highp float;

struct VertexOutput {
    vec2 tex_coord;
    vec4 position;
};

uniform Locals_block_0 {
    mat4x4 transform;
} _group_0_binding_0;

in vec4 _in_location_0;
in vec2 _in_location_1;
out vec2 _out_location_0;

void main() {
    vec4 position1 = _in_location_0;
    vec2 tex_coord1 = _in_location_1;
    VertexOutput out1;
    out1.tex_coord = tex_coord1;
    out1.position = (_group_0_binding_0.transform * position1);
    _out_location_0 = out1.tex_coord;
    gl_Position = out1.position;
    return;
}
```

```glsl
#version 320 es

precision highp float;

struct VertexOutput {
    vec2 tex_coord;
    vec4 position;
};

uniform highp sampler2D _group_0_binding_1;

in vec2 _in_location_0;
out vec4 _out_location_0;

void main() {
    VertexOutput in1 = VertexOutput(_in_location_0, gl_FragCoord);
    vec4 _expr5 = texture(_group_0_binding_1, vec2(in1.tex_coord));
    _out_location_0 = _expr5;
    return;
}
```

```log
[190.865418 INFO](Device::create_render_pipeline wrapper | Device::create_render_pipeline)(no module):  Compiled shader 3
[190.866476 INFO](Device::create_render_pipeline wrapper | Device::create_render_pipeline)(no module):  Linked program 1
[190.866579 ERROR](Device::create_render_pipeline wrapper | Device::create_render_pipeline)(no module):         Log: error: fragment shader input `_in_location_0' has no matching output in the previous stage
```

kvark:
Fixed by #590

PR #590, opened 2021-03-17
[glsl-out] Explicitly set location bindings

Gordon-F:
We already known location index. Should help with #589

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -811,8 +811,22 @@ impl<'a, W: Write> Writer<'a, W> {
                 }
 
                 // Write the storage class
-                write!(self.out, "{} ", if output { "out" } else { "in" })?;
-
+                if self.options.version >= Version::Embedded(310)
+                    || self.options.version >= Version::Desktop(410)
+                {
+                    if let Binding::Location(index, ..) = *binding {
+                        write!(
+                            self.out,
+                            "layout(location = {}) {} ",
+                            index,
+                            if output { "out" } else { "in" }
+                        )?;
+                    } else {
+                        write!(self.out, "{} ", if output { "out" } else { "in" })?;
+                    }
+                } else {
+                    write!(self.out, "{} ", if output { "out" } else { "in" })?;
+                }
                 // Write the type
                 // `write_type` adds no leading or trailing spaces
                 self.write_type(ty)?;
--- a/tests/out/quad-Fragment.glsl.snap
+++ b/tests/out/quad-Fragment.glsl.snap
@@ -13,8 +13,8 @@ struct VertexOutput {
 
 uniform highp sampler2D _group_0_binding_0;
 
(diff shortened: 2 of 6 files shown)


src/back/glsl/mod.rs
```diff
@@ -811,8 +811,22 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                            self.out,
+                            "layout(location = {}) {} ",
+                            index,
+                            if output { "out" } else { "in" }
+                        )?;
+                    } else {
+                        write!(self.out, "{} ", if output { "out" } else { "in" })?;
+                    }
+                } else {
+                    write!(self.out, "{} ", if output { "out" } else { "in" })?;
```
kvark, line 828: Would you be interested to fix this properly for earlier GLSL versions? We basically need to have the naming scheme where the outputs from the previous stage are named the same as the inputs from the next stage.
I.e. these would be `_location_vs_fs` for VS outputs *and* FS inputs.

src/back/glsl/mod.rs
```diff
@@ -811,8 +811,22 @@ impl<'a, W: Write> Writer<'a, W> {
                 }
 
                 // Write the storage class
-                write!(self.out, "{} ", if output { "out" } else { "in" })?;
-
+                if self.options.version >= Version::Embedded(310)
+                    || self.options.version >= Version::Desktop(410)
+                {
+                    if let Binding::Location(index, ..) = *binding {
```
kvark, line 817: we already check for the location above:
> Some(binding @ &Binding::Location(..)) => binding,

Let's do something like this:
```
Some(binding @ &Binding::Location(location, _)) => (binding, location),
```

kvark (changes requested):
Looks great, thank you for fixing this!
We need more GLSL output snapshots. If any of them break, we should see it in CI, where we have the jobs validating outputs.

src/back/glsl/mod.rs
Gordon-F: Thank you for fix suggestions. Currently, I trying to fix this error for earlier GLSL versions. Just make a PR for a quick fix, because it is a blocker for `glsl-out`.

kvark: no problem, I can follow up with this.

kvark (approved):

Gordon-F:
> We need more GLSL output snapshots. If any of them break, we should see it in CI, where we have the jobs validating outputs.

@kvark All current snapshots passed `glslangValidator` validation, because it's a program linking errror. Maybe it can validate full pipeline (vertex -> fragment) ?

If it can't, maybe we can use `llvmpipe` for shader compilation/program linking and catch an error like this one? This looks like a little bit crazy idea 😄

kvark:
You are correct. Maybe it's one of those things that we just can't check. Hopefully, this class of errors is small.

PR #591, opened 2021-03-17
[glsl-out] make varying names to match

kvark:
Follow-up to #590 
r? @Gordon-F


Gordon-F:
@kvark Awesome! Thank you! Tested with `cube` example on `WebGL`/`OpenGL ES`. All works fine.

PR #596, opened 2021-03-18
First bits of expression validation

kvark:
Another step towards #59
We can close the issue as soon as at least the listed sub-issues are addressed, and consider the rest of the missing logic to be bugs.
With this PR, adding more expression validation code should be straightforward.


Napokue (approved):
LGTM, thanks! Fairly straightforward code!

kvark:
Thank you for the review!
there is a lot more to write here, but let's land things gradually

PR #599, opened 2021-03-19
Validate Compose, Unary, Binary, Select, and Relational

kvark:
More steps towards #59

### DIFF
--- a/src/proc/validator.rs
+++ b/src/proc/validator.rs
@@ -173,6 +173,14 @@ pub enum ExpressionError {
     InvalidPointerType(Handle<crate::Expression>),
     #[error("Array length of {0:?} can't be done")]
     InvalidArrayType(Handle<crate::Expression>),
+    #[error("Compose type {0:?} doesn't exist")]
+    ComposeTypeDoesntExist(Handle<crate::Type>),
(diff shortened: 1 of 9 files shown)


src/proc/validator.rs
```diff
@@ -1005,15 +1140,222 @@ impl Validator {
...
+                            Sk::Bool => false,
+                        },
+                        _ => false,
+                    },
+                    Bo::Multiply => {
+                        let kind_match = match left_inner.scalar_kind() {
+                            Some(Sk::Uint) | Some(Sk::Sint) | Some(Sk::Float) => true,
+                            Some(Sk::Bool) | None => false,
+                        };
+                        //TODO: should we be more restrictive here? I.e. expect scalar only to the left.
```
Napokue, line 1174: I would say yes, because it can be crucial for the back-ends processing. However, not really needed for now I guess. It is better to push this validation forward first, and iron out the minor things later.

Napokue (approved):
LGTM, thanks! We are one step closer!

src/proc/validator.rs
kvark: Agreed on both accounts!

PR #600, opened 2021-03-20
Big Validation re-organization

kvark:
This PR makes uniformity analysis optional, and it moves both the analysis and the validator into new home.
They become intertwined: validation *produces* the analysis, and the analysis is used by backends on the validated modules. So really the analysis need to be done as a part of validation, right in the same `match` arms. This is something to follow-up on.

The original `validator.rs` module was 1827 LOC, now it's split into 3 sub-modules around 500 LOC each. Looks like a textbook example on modularization now :)


src/valid/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+use crate::{
+    arena::{Arena, Handle},
+    proc::Typifier,
+    FastHashSet,
+};
+use bit_set::BitSet;
+use thiserror::Error;
+
+//TODO: analyze the model at the same time as we validate it,
+// merge the corresponding matches over expressions and statements.
```
Napokue, line 15: So to be clear: this is now done in two passes? First analyzing then validating?

Napokue (approved):
LGTM, thank you! I didn't really review the code, as it was only re-organized. I think the only real change was "Analysis" -> "ModuleInfo", which is fine by me.

src/valid/mod.rs
kvark: correct. The PR doesn't change this part of the logic, it just re-groups it closer so that we can merge it some time later

PR #601, opened 2021-03-21
Validate structure layouts

kvark:
I was hoping that introducing these errors would prevent people from using unaligned types in WGSL. However, it looks like the spec is not saying anything about `f32` or `vec3<f32>` in https://gpuweb.github.io/gpuweb/wgsl.html#storage-class-constraints, or I'm just missing it.
In any way, this code can be extended when we have all the restrictions on the table.


Napokue (approved):
LGTM, thanks!

PR #602, opened 2021-03-22
Avoid re-typification in the backends

kvark:
This is a continuation of the ideas in #600. We need to merge the analysis into validation.
This PR does it for functions, and allows them to re-use the type results instead of having the backends to regenerate it.

Edit:
This is now complete (and needs another look for the added commits). I really like the scheme we ended up with here:
  - the `TypeResolution` is exposed, and it's a nicer thing to work with, versus separate `get_handle`, `get`, and `try_get` methods on a `Typifier`.
  - the type is now a part of `ExpressionInfo`, as opposed to be coming from a different source. Everything is in one place now.
  - the typifier doesn't have this double-identity issue where it needs to both be growable (for frontends) and immutable (for backends). The validation now produces type information, and the frontends growable semantics is properly moved to the frontend helpers (in `src/front/mod.rs`).
  - we do less work in the backends now, and we don't need to handle the error conditions for the work we've already done. Previously, the `enum Error` size in MSL was bound by the `TypifyError`, which was harmful (for the stack size issues...).

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -44,7 +44,7 @@
 pub use features::Features;
 
 use crate::{
-    proc::{EntryPointIndex, NameKey, Namer, ResolveContext, Typifier, TypifyError},
+    proc::{EntryPointIndex, NameKey, Namer, TypeResolution, TypifyError},
     valid::{FunctionInfo, ModuleInfo},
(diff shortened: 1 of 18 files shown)


Napokue:
So far, LGTM!

Napokue (approved):
Thank you! I think this is a great direction you are taking for the validation/analyzer merger!

PR #603, opened 2021-03-23
Move Layouter into ModuleInfo

kvark:
Continuation of #602 
The type layout is now a part of `ModuleInfo`.

I switched spv-out to use it instead of creating a separate layout object, and I faced a problem wrt how things are organized there. With lazy type serialization, it would take a lot of boilerplate to propagate the layouter to all the things - basically to all the `write_` methods, since most of them call `get_type_id`.

Instead, I did what I wanted to do for some time and refactored the type serialization more drastically. Now all the "handle" types are serialized at the start. There is just a trick we were missing to make it work:
  - first, all the scalar constants are serialized
  - then gather the types corresponding to structures
  - then all the types
  - then all the composite constants

I also made it so the `LookupType::Local` is populated based on `make_local`, which reduces the code complexity a bit.

Reviewer note: most of the LOCs changed stem from the reftest changes in SPIR-V, since we are serializing types in a different order now. I wonder if we could generate IDs in such a way that these kind of changes are minimized?.. (cc @MatusT )
A good addition to the LOCs volume is from the `Layouter` itself, which is now a part of IR snapshots (and it's good!).

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -260,7 +260,6 @@ pub struct Writer {
     // so it may require us to duplicate the type!
     struct_type_handles: crate::FastHashMap<Handle<crate::Type>, crate::StorageAccess>,
     gl450_ext_inst_id: Word,
-    layouter: Layouter,
     temp_chain: Vec<Word>,
 }
 
@@ -291,7 +290,6 @@ impl Writer {
             cached: CachedExpressions::default(),
             struct_type_handles: crate::FastHashMap::default(),
             gl450_ext_inst_id,
-            layouter: Layouter::default(),
             temp_chain: Vec::new(),
         })
     }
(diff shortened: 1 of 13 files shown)


src/back/spv/writer.rs
```diff
@@ -716,10 +709,28 @@ impl Writer {
...
+            match self.lookup_type.entry(LookupType::Local(local)) {
+                // if it's already known as local, re-use it
+                Entry::Occupied(e) => {
+                    let id = *e.into_mut();
+                    self.lookup_type.insert(LookupType::Handle(handle), id);
+                    return Ok(id);
+                }
+                // also register the type as "local", to avoid duplication
+                Entry::Vacant(e) => {
+                    let id = self.id_count + 1; // has to match the next ID
```
Napokue, line 727: Is there a reason to not use `self.generate_id()` here? Wouldn't this give a collision for the first next id that is generated via `self.generate_id()`, as `self.id_count` is not incremented here, but will be at line 732?

Napokue (approved):
LGTM, thanks! Only got one question that I commented. I like the way you split up the sequence of type initialization for all the types, instead of the way it used to be that it would've been "discovered" during writing.

src/back/spv/writer.rs
kvark: fixed now with `IdGenerator` struct

PR #606, opened 2021-03-23
[glsl-out] Add samplers to reflection info

Gordon-F:


kvark (approved):
Looks reasonable, thank you!

PR #607, opened 2021-03-24
Validate image expressions

kvark:
Even more steps towards #59 
The only expression type left is `Math`. It's a big one, and also, interestingly, the one that motivated me to dive into the validation in the first place (more precisely, the "Frexp" operation is tricky and would benefit from type checks).


Napokue (approved):
LGTM, thank you!

PR #608, opened 2021-03-24
Make layout validation optional, disable the struct size check

kvark:
The struct size check was just plain broken, and if I fix it, we'd start failing on the boids example...
Disabling now until https://github.com/gpuweb/gpuweb/issues/1558 is cleared.


Napokue (approved):
LGTM, thank you!

PR #609, opened 2021-03-25
SPIRV frontend fixes for image operands and shifts

kvark:
The image operands were incorrect. We are expected to read 1 mask followed by the parameters, and instead we used to read each bit separately.


Napokue (commented):
LGTM, thanks!

Napokue (approved):
(didn't choose "approve" 😅)

PR #610, opened 2021-03-25
Fix expression validation check

kvark:
Looks like the expression validation check was flipped, unfortunately.
This should be merged after #609, which addresses the correctness of the image array index.


Napokue (approved):
LGTM, thanks!

PR #612, opened 2021-03-25
Align the block decoration with SPIR-V, require storage buffers

### LINKED ISSUE #611: Refactor the Block attribute
Currently, the `[[block]]` means "this is a host-shared struct", while instead we need it to mean "this is a top-level struct that is also host-shared". See https://github.com/gpuweb/gpuweb/pull/1559

kvark:
Closes #611
The side effect of this change is that now we require SPV-1.3 or `SPV_KHR_storage_buffer_storage_class`. Apparently, this is universally supported, and it's a good thing. You can see how the code size is reduced in the PR.

### DIFF
--- a/src/back/spv/mod.rs
+++ b/src/back/spv/mod.rs
@@ -1,3 +1,6 @@
+/*! Standard Portable Intermediate Representation (SPIR-V) backend
+!*/
+
 mod helpers;
 mod instructions;
 mod layout;
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1,5 +1,6 @@
-/*! Standard Portable Intermediate Representation (SPIR-V) backend !*/
-use super::{Instruction, LogicalLayout, Options, PhysicalLayout, WriterFlags};
+use super::{
+    helpers::map_storage_class, Instruction, LogicalLayout, Options, PhysicalLayout, WriterFlags,
+};
 use crate::{
     arena::{Arena, Handle},
     proc::{Layouter, TypeResolution},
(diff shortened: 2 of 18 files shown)


src/back/spv/writer.rs
```diff
@@ -2345,6 +2322,15 @@ impl Writer {
...
     ) -> Result<(), Error> {
+        let has_storage_buffers = ir_module
+            .global_variables
+            .iter()
+            .any(|(_, var)| var.class == crate::StorageClass::Storage);
+        if self.physical_layout.version < 0x10300 && has_storage_buffers {
+            // enable the storage buffer class on < SPV-1.3
+            Instruction::extension("SPV_KHR_storage_buffer_storage_class")
+                .to_words(&mut self.logical_layout.extensions);
+        }
```
Napokue, line 2333: For now this is okay, maybe if we later get more use cases of checking the SPV version, and adding extensions, we can do it in a more elegant way. 😄

src/back/spv/mod.rs
```diff
@@ -1,3 +1,8 @@
+/*! Standard Portable Intermediate Representation (SPIR-V) backend
+ *
+ * Generates SPV-1.3 bytecode.
```
Napokue, line 3: This still stands? Can't we potentially generate SPV-1.5 now too? Just wondering and giving a thought.

Napokue (approved):
LGTM, thanks! I have got 2 little things that are not blocking. Address them how you like.

src/back/spv/mod.rs
kvark: gonna fix this now!

src/back/spv/writer.rs
kvark: Possibly, yes. This PR doesn't actually change the number of checks, it just moves this check into a better place :)

Issue #620, opened 2021-03-27
Some operations are sign-agnostic in SPIR-V


kvark:
Looking at the current code for `map_binary_operator`:
```rust
        Op::IAdd | Op::FAdd => Ok(BinaryOperator::Add),
        Op::ISub | Op::FSub => Ok(BinaryOperator::Subtract),
        Op::IMul | Op::FMul => Ok(BinaryOperator::Multiply),
        Op::UDiv | Op::SDiv | Op::FDiv => Ok(BinaryOperator::Divide),
        Op::UMod | Op::SMod | Op::FMod => Ok(BinaryOperator::Modulo),
        // Relational and Logical Instructions
        Op::IEqual | Op::FOrdEqual | Op::FUnordEqual => Ok(BinaryOperator::Equal),
        Op::INotEqual | Op::FOrdNotEqual | Op::FUnordNotEqual => Ok(BinaryOperator::NotEqual),
```
It becomes evident that `OpIAdd, OpISub, OpIMul, OpIEqual, and OpINotEqual` are valid to use even if on one side it's signed, and on the other side - unsigned.

This seems like a royal PITA. Even if we detect that the types have a different sign, even if we bitcast one of them to match the other, we'd still not know what sign the considers of this SSA ID expect. In fact, there might be different consumers expecting different things. For example, the `OpIAdd` can be followed by `OpUDiv` *and* `OPSDiv` on the same result value...

kvark:
This is actually maybe not a problem at all at this stage. It turns out the failures I saw were about the `Equal` operations we inserted as boilerplate, not `OpIEqual` of SPIR-V inputs.

kvark:
Now I'm really hitting this: http://shader-playground.timjones.io/3f4cba62bdbe8cd640644a8e8c4c20a8
Thinking if a hack needs to be put in place to see if one of the expressions being a constant, and adjust it to the other one.

kvark:
GLSL's `foo++` translates to `OpIAdd %uint %10 %int_1` (effectively uint + int)
but `foo+=1` translates to `OpIAdd %uint %11 %uint_1` (so uint + uint)

We can work around this in `OpIAdd`/`OpISubtract`, but I really don't want to make us fully conformant to the crazy SPIR-V spec here, since it would require this workaround to be pervasive, everywhere.

PR #626, opened 2021-03-29
[spv-in] derivatives and fixed dynamic vector access

kvark:
Related to #620 (fixes the original problem, but the issue is larger as filed).


Napokue (approved):
LGTM, thanks!

PR #629, opened 2021-03-30
[wgsl-in] ability to take a pointer to something

### LINKED ISSUE #627: Taking address of a variable in WGSL
Related to #625 (`modf` and a few others require a pointer).
Blocks https://github.com/gfx-rs/wgpu-rs/pull/763

kvark:
Fixes #627 
Very basic, there is no dereference, etc, but enough to be able to call `frexp` and friends.


kvark:
Ok, this appears to work for `water` example so far 🎉

PR #638, opened 2021-04-01
[msl] use uint behind metal namespace only

kvark:
May help https://github.com/gfx-rs/wgpu/issues/1299


kvark:
Confirmed the fix!

PR #639, opened 2021-04-01
[spv-in] pre-emit constant expressions

kvark:
If we emit them on the fly, we hit a validation error, since frontends aren't expected to emit constant expressions.


kvark:
There is very little code here, all the changes are from the snapshots

PR #662, opened 2021-04-04
fixed bug where metal backend would make all switch cases fallthrough

adamnemecek:


kvark (approved):
Wow, this is a great catch!

PR #663, opened 2021-04-05
New struct layout IR

### LINKED ISSUE #657: Respect the structure layout
This affects multiple frontends and backends. We aren't currently respecting the layout when generating code, and we aren't filling up the align/size when building IR from SPIR-V.

### LINKED ISSUE #658: Cover Layouter in unit tests
See #654

kvark:
Fixes #657 
Closes #658

The main idea here is that "align: Option" with "size: Option" (the data we have in `StructMember`) is unnecessary verbose for the IR. It should be resolved on WGSL side, and IR can have something simple and more specific, like the `span`.

Draft because I haven't figured out what to do with `vec3`. In SPIR-V, some of the offsets I'm seeing imply that it takes 12 bytes, but in MSL it has to take 16 bytes. Need a bit more investigation before we proceed.


kvark:
Here is another interesting issue I'm seeing with real-world SPIR-V:
```
               OpMemberDecorate %_struct_1165 0 Offset 304
               OpMemberDecorate %_struct_1165 1 Offset 448
               OpMemberDecorate %_struct_1165 2 Offset 460
               OpMemberDecorate %_struct_1165 3 Offset 464
               OpMemberDecorate %_struct_1165 4 Offset 468
               OpMemberDecorate %_struct_1165 5 Offset 472
```
Basically, the first offset is not zero. This is not compatible with the IR's "span" that this PR brings.

kvark:
So the problem we are trying to solve is: how to specify structure layout in IR. In text-based languages, it's always implicit: every type defines its size/align, and the struct layout becomes derived from the members. In SPIR-V, howerver, every struct member is annotated with an "offset" in bytes.

What this PR does right now, is adding a "span" to each member. It seemed like a good approach to me because it's more composable (i.e. in many cases you can switch two members and still have valid IR without recomputing anything), and it allows us to control the whole structure size better (the span of the last member would affect the structure size, while this is not an option with "offset" kind of decorations).

One of the problems faced here is that in SPIR-V the offset can sometimes start with non-zero. I can think of the following ways to proceed (please don't treat this as a complete list!):
  1. add `start_offset` field to `TypeInner::Struct` variant, and respect it in the backends.
  2. change "span" into "offset" to stay close to SPIR-V. The composability would take a hit, and it's not clear what to do with the struct size then.
  3. have something like "pad_before: u32" for each member. There are many options here: "pad_before"/"pad_after", "pad"/"align".

kvark:
The last commit implements (2). The part about `enum StructLevel` is a little bit weird and may need to iterate, but the offsets look fine.

grovesNL (approved):

PR #667, opened 2021-04-07
Panic to prevent arena Handle overflow

zakarumych:
It seems impossible practically, but possible intentionally.
When arena size becomes `u32::MAX+1` on append the new handle returned contains 0 in NonZeroU32 which is UB.

### DIFF
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -188,6 +188,10 @@ impl<T> Arena<T> {
 
     /// Adds a new value to the arena, returning a typed handle.
     pub fn append(&mut self, value: T) -> Handle<T> {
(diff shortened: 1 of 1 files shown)


src/arena.rs
```diff
@@ -188,6 +188,10 @@ impl<T> Arena<T> {
 
     /// Adds a new value to the arena, returning a typed handle.
     pub fn append(&mut self, value: T) -> Handle<T> {
+        debug_assert!(self.data.len() <= u32::MAX as usize);
+        if self.data.len() as u32 == u32::MAX {
```
kvark, line 192: let's instead change `new_unchecked` into just `new`, and have `expect` there. Less unsafe code!

kvark (approved):

PR #668, opened 2021-04-07
Metal inlined samplers

### LINKED ISSUE #665: MSL immutable sampler support
This is required for Vulkan usage, including Dota2. In particular, in some cases there is no other way than making a sampler inlined.

kvark:
Closes #665

### DIFF
--- a/src/back/msl/mod.rs
+++ b/src/back/msl/mod.rs
@@ -23,19 +23,30 @@ For the result type, if it's a structure, we re-compose it with a temporary valu
 holding the result.
 !*/
 
-use crate::{arena::Handle, valid::ModuleInfo, FastHashMap};
+use crate::{
+    arena::{Arena, Handle},
+    valid::ModuleInfo,
+    FastHashMap,
+};
 use std::fmt::{Error as FmtError, Write};
 
 mod keywords;
+mod sampler;
 mod writer;
 
 pub use writer::Writer;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum BindSamplerTarget {
+    Resource(u8),
+    Inline(Handle<sampler::InlineSampler>),
+}
 
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct BindTarget {
     pub buffer: Option<u8>,
     pub texture: Option<u8>,
-    pub sampler: Option<u8>,
+    pub sampler: Option<BindSamplerTarget>,
     pub mutable: bool,
 }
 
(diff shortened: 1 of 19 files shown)


src/back/msl/writer.rs
```diff
@@ -1322,60 +1318,64 @@ impl<W: Write> Writer<W> {
...
-                INDENT,
-                sampler.compare_func.as_str()
+                "{}{}::compare_func::{},",
+                level,
+                NAMESPACE,
+                sampler.compare_func.as_str(),
             )?;
         }
-        writeln!(self.out, "{}coord::normalized);", INDENT)?;
+        writeln!(self.out, "{}{}::coord::normalized", NAMESPACE, level)?;
```
grovesNL, line 1378: `NAMESPACE` and `level` should be switched here

tests/in/collatz.param.ron
```diff
@@ -4,7 +4,5 @@
 	spv_capabilities: [ Shader ],
 	spv_debug: true,
 	spv_adjust_coordinate_space: false,
-	mtl_bindings: {
-		(stage: Compute, group: 0, binding: 0): (buffer: Some(0), mutable: true),
-	}
+	msl_custom: false,
```
grovesNL, line 7: nit: it could be useful to keep the `binding_map` instead of using the fake bindings for MSL validation when possible. e.g. just in case there are subtle interactions between valid bindings and the rest of the shader that won't appear when we use `user` attributes instead

grovesNL (approved):
Looks great, thank you!

src/back/msl/mod.rs
```diff
@@ -90,28 +101,43 @@ enum LocationMode {
...
+            inline_samplers: Arena::new(),
             spirv_cross_compatibility: false,
             fake_missing_bindings: true,
+        }
+    }
+}
+
+// A subset of options that are meant to be changed per pipeline.
+#[derive(Debug, Clone)]
+pub struct SubOptions {
```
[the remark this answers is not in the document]
kvark: definitely a better name!

src/back/msl/writer.rs
```diff
@@ -1316,6 +1322,63 @@ impl<W: Write> Writer<W> {
...
 
+    fn write_inline_sampler(
+        &mut self,
+        sampler: &sm::InlineSampler,
+        stage: crate::ShaderStage,
+        handle: Handle<crate::GlobalVariable>,
+    ) -> Result<(), Error> {
+        write!(
+            self.out,
+            "constexpr sampler ism_{}_{:?}(",
```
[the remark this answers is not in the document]
kvark: fortunately, this code is gone :)

src/back/msl/writer.rs
kvark: just noticed this as well :)

tests/in/collatz.param.ron
kvark: I'm trying to think of what those interactions could be, and unable to come up with anything. We can enable more tests to use the full `Options` struct as a follow-up?

PR #671, opened 2021-04-08
[wgsl-in] Decorations are now called attributes.

jimblandy:
Renames:

Error::UnknownDecoration => Error::UnknownAttribute
TypeDecoration => TypeAttributes
Scope::Decoration => Scope::Attribute

various argument names, string literals, comments


kvark (approved):
Nice, thank you!

PR #676, opened 2021-04-08
[msl] Inline some of the types

kvark:
Fixes #666
For global image variables (Metal textures), the type needs to depend on the storage access. One way to get that working is to always inline the type, since this is where we know about the storage access unconditionally. I figured it's going to be a general improvement for the produced code, since previously it was annoying to look up "typeN" for anything trivial.

PR #677, opened 2021-04-08
[msl] inline some of the constants

### LINKED ISSUE #656: Gather constant refcount info

kvark:
Closes #656
Technically, it doesn't add refcounts. The original motivation there was we could omit constants we don't use. This PR achieves the same effect, and even better. There was also an issue about some constants depending on the other, which made the idea difficult to implement in a useful way.

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -229,6 +229,42 @@ impl<'a> TypedGlobalVariable<'a> {
     }
 }
 
+struct ConstantContext<'a> {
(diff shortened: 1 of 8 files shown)


kvark:
Goes in-line (with pun intended!) with #676

PR #678, opened 2021-04-08
[spv-in] Copy less bytes in `parse_function`

LingMan:
The initial intention was to make the code a little more elegant and match on
the Vec itself instead of its length, however there's also a - probably
minuscule - performance advantage:

`Vec::remove` internally copies the removed element onto the stack. In this case
that's a `StructMember` weighting 48 bytes. The `Binding` we are actually
interested in is only 8 bytes tough.

Cloning just the `Binding` saves us from copying 40 bytes for nothing.


kvark (approved):
Thank you!
I'm surprised our linter doesn't yell at this.

PR #679, opened 2021-04-08
[spv-out]: Add some `use` declarations to the writer.

jimblandy:
(Obviously subjective, but I figured I'd see what folks think.)

This adds six `use` declarations and eliminates ~300 identifiers.

I tried to avoid cases where SPIR-V and Naga use similar terminology. For example,
`BuiltIn` remains `spirv::BuiltIn` or `crate::BuiltIn`.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -790,6 +790,8 @@ impl Writer {
             }
         }
 
+        use spirv::Decoration;
+
         let instruction = match ty.inner {
             crate::TypeInner::Scalar { kind, width } => self.write_scalar(id, kind, width),
             crate::TypeInner::Vector { size, kind, width } => {
@@ -846,7 +848,7 @@ impl Writer {
                 if let Some(array_stride) = stride {
                     self.annotations.push(Instruction::decorate(
                         id,
-                        spirv::Decoration::ArrayStride,
+                        Decoration::ArrayStride,
                         &[array_stride.get()],
                     ));
                 }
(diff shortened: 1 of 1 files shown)


kvark (changes requested):
Sorry to put the red light here.

The general way Naga is written so far is that, of all the IR, only "arena" sub-module types are directly imported at the module levels. Everything else is spelled explicitly, like `crate::ScalarKind`. The validation and processor includes are fine, they aren't the IR.

This is written like this because it's hard to draw the line on which IR types are used, and which aren't (and thus do the fully qualified name). Then we'd end up with a constant game of adjusting the headers, like [in WebRender](https://searchfox.org/mozilla-central/rev/3a798ef9252896fb389679f06dd3203169565af0/gfx/wr/webrender/src/renderer/mod.rs#37-98) and mixing fully-qualified types with non-qualified. We've been there, and that was rather sad.

What naga does today is much more strict. You know what you import in any module, because it's always: IR's arena types, a few processor/validation types, and maybe backend-specific things. Then if you have a function that uses some IR types a lot, like matching over `crate::Expression`, you import it in as a short name, i.e. `use crate::Expression as E`. So all IR imports are short and local.

Note: if we apply this rule to some more places where the matching happens, we can save the same 300 chars, or more, easily, without changing the code style for the whole project.

I know this is a bit controversial, and I don't want to sound like the other ways of organizing the code are bad. We have our tastes, and this approach just worked very well so far. So please let's keep it?

jimblandy:
I'm happy to go along with whatever the conventions are. Those seem fine to me.

I've revised the PR to add `use` directives for `spirv` module types only. Is this better?

jimblandy:
Oops - I missed some IR `use` declarations. I'll fix those...

jimblandy:
After some discussion, we agreed to just add some local `use` declarations for `spirv`'s `Decoration` and `Builtin` to the functions where they're used repeatedly.

kvark (approved):
Thank you!

PR #682, opened 2021-04-09
SPV-in workaround for row-major matrices

### LINKED ISSUE #675: Parse SPIR-V row-major matrices
Our IR is column-major only. There is currently code in SPV-in `fn parse_type_matrix` to check for this:
```rust
if let Some(Decoration {
            matrix_major: Some(ref major),
            ..
        }) = decor
        {
            match *major {
                Majority::Column => (),
                Majority::Row => return Err(Error::UnsupportedRowMajorMatrix),
            }
        }
```

However, it never fires up, because the decoration is actually expected on a struct member instead of the type.
The same applies to matrix stride. They both are not correctly handled in SPV-in.
Here is an example dump from 215.spv:
```
               OpMemberDecorate %_struct_1504 0 RowMajor
               OpMemberDecorate %_struct_1504 0 Offset 0
               OpMemberDecorate %_struct_1504 0 MatrixStride 16
               OpMemberDecorate %_struct_1504 1 Offset 304
               OpMemberDecorate %_struct_1504 2 Offset 320
               OpMemberDecorate %_struct_1504 3 Offset 400
               OpMemberDecorate %_struct_1504 4 Offset 704
```
This is blocking #409

I think the idea on how this should be handled is to try to avoid translating it, unless necessary.
I.e. vector multiplication by a matrix can just switch the order in order to account for this, and it's possible that this is enough.

kvark:
Fixes #675 
I admit, this is a real PITA, and this code reflects it.

The workaround I wanted to implement was to issue an `Load` into the matrix if we see it being row-major, followed by `Transpose`, then pretending that it was column-major all this time.
The problem is that there are many ways to cross that "ima matrix" barrier:
  - `OpAccessChain` doesn't have intermediate IDs, so we can't do it iteratively
  - it may start high up and end on a matrix, or on an element of it
  - it turns out, an array of matrices, when being a struct field, can also have this row-major decoration

If anybody can have a better idea, I'd love to consider it!

### DIFF
--- a/src/front/spv/error.rs
+++ b/src/front/spv/error.rs
@@ -54,6 +54,5 @@ pub enum Error {
     InvalidEdgeClassification,
     FunctionCallCycle(spirv::Word),
     // incomplete implementation error
-    UnsupportedRowMajorMatrix,
     UnsupportedMatrixStride(spirv::Word),
 }
--- a/src/front/spv/function.rs
+++ b/src/front/spv/function.rs
@@ -63,6 +63,9 @@ impl<I: Iterator<Item = u32>> super::Parser<I> {
     }
 
     pub(super) fn parse_function(&mut self, module: &mut crate::Module) -> Result<(), Error> {
+        self.lookup_expression.clear();
+        self.lookup_load_override.clear();
+
         let result_type_id = self.next()?;
         let fun_id = self.next()?;
         let _fun_control = self.next()?;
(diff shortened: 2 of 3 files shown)


Napokue (approved):
LGTM, thanks! It is quite some complexity to "only' support basically a matrix agnostically. I have no idea how to improve upon this.

Issue #683, opened 2021-04-09
Incorrect loop continue parsing in SPIR-V


kvark:
Seeing this IR coming from SPIR-V:
```
                Loop {
                    body: [
                        Emit(
                            [96..96],
                        ),
                        Emit(
                            [98..98],
                        ),
                        Emit(
                            [100..100],
                        ),
                        Emit(
                            [102..103],
                        ),
                        If {
                            condition: [103],
                            accept: [
                                Emit(
                                    [104..152],
                                ),
                                Store {
                                    pointer: [95],
                                    value: [151],
                                },
                                Store {
                                    pointer: [97],
                                    value: [149],
                                },
                                Store {
                                    pointer: [99],
                                    value: [145],
                                },
                                Store {
                                    pointer: [101],
                                    value: [152],
                                },
                            ],
                            reject: [],
                        },
                    ],
                    continuing: [
                        Emit(
                            [104..152],
                        ),
                        Store {
                            pointer: [95],
                            value: [151],
                        },
                        Store {
                            pointer: [97],
                            value: [149],
                        },
                        Store {
                            pointer: [99],
                            value: [145],
                        },
                        Store {
                            pointer: [101],
                            value: [152],
                        },
                    ],
                },
```
There is no exit condition here. I think the SPV-in CFG incorrectly cloned some block of statements, instead of having the `Break`, which is meant to be here (behind the condition).

See full shader - 
[236.spv.zip](https://github.com/gfx-rs/naga/files/6283184/236.spv.zip)

Flow graph:
![Vertex](https://user-images.githubusercontent.com/107301/114130143-eef9d480-98cd-11eb-8d79-61bd3d6b0a62.png)

kvark:
Hmm, strange. I'm still seeing this in the produced IR:
```
                Loop {
                    body: [
                        Emit(
                            [105..105],
                        ),
                        Emit(
                            [107..107],
                        ),
                        Emit(
                            [109..109],
                        ),
                        Emit(
                            [111..112],
                        ),
                        If {
                            condition: [112],
                            accept: [
                                Emit(
                                    [113..161],
                                ),
                                Store {
                                    pointer: [104],
                                    value: [160],
                                },
                                Store {
                                    pointer: [106],
                                    value: [158],
                                },
                                Store {
                                    pointer: [108],
                                    value: [154],
                                },
                                Store {
                                    pointer: [110],
                                    value: [161],
                                },
                            ],
                            reject: [
                                Break,
                            ],
[...]

PR #687, opened 2021-04-09
spv-in: special-case a loop with true/false node == merge node

MatusT:
Fixes: https://github.com/gfx-rs/naga/issues/683

It seems that in real-world SPIR-V shaders (not trivially understood from a specification) there can be a case in CFG where edge to merge node can be the same branch as true or false edge from a loop block. Therefore, this block will be classified as something else rather than `break block` as it would need to be to produce a `break` statement.

### DIFF
--- a/src/front/spv/flow.rs
+++ b/src/front/spv/flow.rs
@@ -338,13 +338,24 @@ impl FlowGraph {
                         condition,
                         true_id,
                         false_id,
-                    } => body.push(crate::Statement::If {
-                        condition,
-                        accept: self
-                            .naga_traverse(self.block_to_node[&true_id], Some(merge_node_index))?,
-                        reject: self
-                            .naga_traverse(self.block_to_node[&false_id], Some(merge_node_index))?,
-                    }),
+                    } => {
+                        let true_node_index = self.block_to_node[&true_id];
+                        let false_node_index = self.block_to_node[&false_id];
+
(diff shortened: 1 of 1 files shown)


src/front/spv/flow.rs
```diff
@@ -338,13 +338,42 @@ impl FlowGraph {
...
                         false_id,
-                    } => body.push(crate::Statement::If {
-                        condition,
-                        accept: self
-                            .naga_traverse(self.block_to_node[&true_id], Some(merge_node_index))?,
-                        reject: self
-                            .naga_traverse(self.block_to_node[&false_id], Some(merge_node_index))?,
-                    }),
+                    } => {
+                        let true_node_index = self.block_to_node[&true_id];
```
kvark, line 342: let's move the condition to inside the `If` constructor:
```rust
                           body.push(crate::Statement::If {
                                condition,
                                accept: if true_node_index == merge_node_index {
                                        vec![crate::Statement::Break]
                                } else {
                                   self.naga_traverse(
                                    true_node_index,
                                    Some(merge_node_index),
                                  )?
                                },
                                reject: if false_node_index == merge_node_index {
                                        vec![crate::Statement::Break]
                                } else {
                                   self.naga_traverse(
                                    false_node_index,
                                    Some(merge_node_index),
                                )?
                                },
                            })
```

kvark (approved):

Issue #578, opened 2021-03-15
Detect loops in CFG of SPIR-V


kvark:
See #570
If we can easily detect loops there, we should do it. Anything is better than a stack overflow.

Napokue:
How exactly do you want to detect loops? I guess we won't want to signal the user after one loop? Is there some way to know that code might potentially go over the stack?

kvark:
My hope was just to mark the visited nodes in each CFG traversal, so that we don't visit any node twice.
cc @MatusT

kvark:
Fixed by #688

PR #688, opened 2021-04-09
spv-in: path forward

### LINKED ISSUE #613: SPIR-V frontend emits some expressions multiple times
[108.spv.zip](https://github.com/gfx-rs/naga/files/6209447/108.spv.zip)

See "1070" for example, which is emitted at least 3 times within the same function.

MatusT:
Fixes #613, #691, #578, #683
Makes obsolete: #690 

A lot of stuff is borrowed/inspired by Tint.

Changes
- Added postorder traversal
- Added constructs based on postorder traversal
- Add HashSet of all points traversal should be stopped
- Reworked when/where to stop for if selection
- Slightly improved CFG Visualization for better debugging
- Adds loop detection
- Replaces of cloning blocks with move

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ pomelo = { version = "0.1.4", optional = true }
 thiserror = "1.0.21"
 serde = { version = "1.0", features = ["derive"], optional = true }
 petgraph = { version ="0.5", optional = true }
+rose_tree = { version ="0.2", optional = true }
 pp-rs = { version = "0.1", optional = true }
 #env_logger = "0.8" # uncomment temporarily for developing with the binary target
 
@@ -36,7 +37,7 @@ glsl-out = ["petgraph"]
 msl-out = []
 serialize = ["serde"]
 deserialize = ["serde"]
-spv-in = ["petgraph", "spirv"]
+spv-in = ["petgraph", "spirv", "rose_tree"]
 spv-out = ["spirv"]
 wgsl-in = ["codespan-reporting"]
 wgsl-out = []
--- a/src/front/spv/flow.rs
+++ b/src/front/spv/flow.rs
@@ -25,13 +25,23 @@ type BlockNodeIndex = NodeIndex<u32>;
 /// Internal representation of a CFG constisting of function's basic blocks.
 type ControlFlowGraph = petgraph::Graph<ControlFlowNode, ControlFlowEdgeType, Directed, u32>;
 
+type ConstructNodeIndex = NodeIndex<u32>;
+
+type ConstructsGraph = rose_tree::RoseTree<ConstructNode>;
+
 /// Control flow graph (CFG) containing relationships between blocks.
 pub(super) struct FlowGraph {
     ///
     flow: ControlFlowGraph,
 
+    ///
+    constructs: ConstructsGraph,
+
     /// Block ID to Node index mapping. Internal helper to speed up the classification.
     block_to_node: FastHashMap<BlockId, BlockNodeIndex>,
+
+    /// Reversed post-order traversal of the `flow` graph.
+    traversal_order: Vec<BlockNodeIndex>,
 }
 
 impl FlowGraph {
@@ -39,7 +49,13 @@ impl FlowGraph {
     pub(super) fn new() -> Self {
         Self {
             flow: ControlFlowGraph::default(),
+            constructs: ConstructsGraph::new(ConstructNode {
+                ty: ConstructType::Function,
+                ..Default::default()
+            })
+            .0,
             block_to_node: FastHashMap::default(),
+            traversal_order: Vec::new(),
         }
     }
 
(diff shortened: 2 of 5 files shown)


src/front/spv/flow.rs
```diff
@@ -217,33 +217,34 @@ impl FlowGraph {
...
                     false_id,
                 } => {
                     let true_node_index = self.block_to_node[&true_id];
                     let false_node_index = self.block_to_node[&false_id];
-                    let merge_node_index = self.block_to_node[&node.merge.unwrap().merge_block_id];
+                    let merge_node_index =
+                        self.block_to_node[&self.flow[node_index].merge.unwrap().merge_block_id];
 
-                    let mut result = node.block.clone();
+                    let mut result: crate::Block = self.flow[node_index].block.drain(..).collect();
```
kvark, line 247: Is there a guarantee that `node_index` is only ever going to be processed once?
If it's the case, can we just do something like `let node = self.flow.remove(node_index)` here, and then move the `block` out?
If it's not the case, then I don't understand what's going to happen here: we may try to drain a block that's already empty.

If neither of those is a concern, and the way it's done currently is right, we should do `mem::replace(&mut self.flow[node_index].block, Vec::new())` instead of this `drain` + `collect` pair, since the replace would avoid another heap allocation.

MatusT: As I have written in https://github.com/gfx-rs/naga/issues/691#issuecomment-821100226 I believe we can assume that each node should be processed only once.

I alway forget the trick with mem::replace :) my rust-fu is still weak It seems.

kvark: Could you clarify why we aren't just removing the node upon traversal?
```
let node = self.flow.remove(node_index);
```

kvark:
I'm worried about the correctness here. If we extract the statements from a node block, what would stop us from accidentally going through the same node again?

src/front/spv/flow.rs
MatusT: I am not sure removing is entirely correct. We visit each node once for cloning statements(so really I use visit word for the purpose of dding statements to IR). But we might still need the nodes for other stuff. For example switch cases ask about accesibility to other cases.

src/front/spv/flow.rs
```diff
@@ -217,33 +217,35 @@ impl FlowGraph {
...
+        &mut self,
         node_index: BlockNodeIndex,
         stop_node_index: Option<BlockNodeIndex>,
     ) -> Result<crate::Block, Error> {
         if stop_node_index == Some(node_index) {
             return Ok(vec![]);
         }
 
-        let node = &self.flow[node_index];
+        let terminator = self.flow[node_index].terminator.clone();
```
kvark, line 233: we need something like `let (node_block, terminator) = self.flow[node_index].consume();`, so that we can have some condition inside it to check if the node was already consumed previously

kvark:
Seeing many failures there. Attaching some of the shaders - 
[dota2-cfg-issue.zip](https://github.com/gfx-rs/naga/files/6333096/dota2-cfg-issue.zip)

src/front/spv/flow.rs
```diff
@@ -217,61 +519,100 @@ impl FlowGraph {
...
+                    let premerge_index: Option<NodeIndex> = {
+                        let second_head = self.flow[true_node_index]
+                            .position
+                            .max(self.flow[false_node_index].position);
+                        let end_first_clause = self.traversal_order[second_head - 1];
+
+                        self.flow
+                            .edges_directed(end_first_clause, Direction::Outgoing)
+                            .find(|e| {
+                                (*e.weight() == ControlFlowEdgeType::Forward
```
kvark, line 564: this is asking for a match

Cargo.toml
```diff
@@ -24,6 +24,7 @@ pomelo = { version = "0.1.4", optional = true }
 thiserror = "1.0.21"
 serde = { version = "1.0", features = ["derive"], optional = true }
 petgraph = { version ="0.5", optional = true }
+rose_tree = { version ="0.2", optional = true }
```
kvark, line 27: is it necessary?  depending on both petgraph and rese_tree seems strange

kvark (commented):
Are we close yet? :)

Cargo.toml
MatusT: rose_tree is just petgraph's extension that has a concept of a parent. It's directly built on petgraph so it shouldn't really add much. Makes some stuff a little nicer.

kvark:
@MatusT please rebase and optionally refactor that `match` (in my last comment)

src/front/spv/flow.rs
MatusT: like this?

MatusT:
> @MatusT please rebase and optionally refactor that `match` (in my last comment)

shipit

kvark (approved):
Great!

Issue #670, opened 2021-04-07
Update interpolation qualifiers


kvark:
See https://github.com/gpuweb/gpuweb/pull/1605

jimblandy:
I can work on this.

PR #689, opened 2021-04-09
Implement WGSL `interpolate` syntax from gpuweb PR #1605.

jimblandy:
Fixes: #670

This PR has separate commits for mechanical or logically independent changes, to make reviewing easier.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -48,9 +48,10 @@ use crate::{
     valid::{FunctionInfo, ModuleInfo},
     Arena, ArraySize, BinaryOperator, Binding, BuiltIn, Bytes, ConservativeDepth, Constant,
     ConstantInner, DerivativeAxis, Expression, FastHashMap, Function, GlobalVariable, Handle,
-    ImageClass, Interpolation, LocalVariable, Module, RelationalFunction, ScalarKind, ScalarValue,
-    ShaderStage, Statement, StorageAccess, StorageClass, StorageFormat, StructMember, Type,
-    TypeInner, UnaryOperator,
+    ImageClass, Interpolation, LocalVariable, Module, RelationalFunction, Sampling, ScalarKind,
+    ScalarValue, ShaderStage, Statement, StorageAccess, StorageClass, StorageFormat, StructMember,
+    Type, TypeInner, UnaryOperator,
+
 };
 use features::FeaturesManager;
 use std::{
@@ -232,8 +233,10 @@ impl IdGenerator {
 /// Helper wrapper used to get a name for a varying
 ///
 /// Varying have different naming schemes depending on their binding:
-/// - Varyings with builtin bindings get the from [`glsl_built_in`](glsl_built_in)
-/// - Varyings with location bindings are named `_location_X` where `X` is the location
+/// - Varyings with builtin bindings get the from [`glsl_built_in`](glsl_built_in).
+/// - Varyings with location bindings are named `_S_location_X` where `S` is a
+///   prefix identifying which pipeline stage the varying connects, and `X` is
+///   the location.
 struct VaryingName<'a> {
     binding: &'a Binding,
     stage: ShaderStage,
@@ -793,16 +796,22 @@ impl<'a, W: Write> Writer<'a, W> {
                 }
             }
             _ => {
-                let (location, interpolation) = match binding {
-                    Some(&Binding::Location { location, interpolation }) => (location, interpolation),
+                let (location, interpolation, sampling) = match binding {
+                    Some(&Binding::Location { location, interpolation, sampling }) => (location, interpolation, sampling),
                     _ => return Ok(()),
                 };
+
                 // Write the interpolation modifier if needed
                 //
-                // We ignore all interpolation modifiers that aren't used in input globals in fragment
-                // shaders or output globals in vertex shaders
+                // We ignore all interpolation and auxiliary modifiers that aren't used in fragment
+                // shaders' input globals or vertex shaders' output globals.
+                let emit_interpolation_and_auxiliary = match self.options.shader_stage {
+                    ShaderStage::Vertex => output,
+                    ShaderStage::Fragment => !output,
+                    _ => false,
+                };
                 if let Some(interp) = interpolation {
-                    if self.options.shader_stage == ShaderStage::Fragment {
+                    if emit_interpolation_and_auxiliary {
                         write!(self.out, "{} ", glsl_interpolation(interp))?;
                     }
                 }
@@ -811,13 +820,24 @@ impl<'a, W: Write> Writer<'a, W> {
                 if self.options.version.supports_explicit_locations() {
                     write!(
                         self.out,
-                        "layout(location = {}) {} ",
-                        location,
-                        if output { "out" } else { "in" }
-                    )?;
-                } else {
-                    write!(self.out, "{} ", if output { "out" } else { "in" })?;
+                        "layout(location = {}) ",
+                        location)?;
                 }
+
+                // Write the sampling auxiliary qualifier.
+                //
+                // Before GLSL 4.2, the `centroid` and `sample` qualifiers were required to appear
+                // immediately before the `in` / `out` qualifier, so we'll just follow that rule
+                // here, regardless of the version.
+                if let Some(sampling) = sampling {
+                    if emit_interpolation_and_auxiliary {
+                        write!(self.out, "{} ", glsl_sampling(sampling))?;
+                    }
+                }
+
+                // Write the input/output qualifier.
+                write!(self.out, "{} ", if output { "out" } else { "in" })?;
+
                 // Write the type
                 // `write_type` adds no leading or trailing spaces
                 self.write_type(ty)?;
@@ -825,7 +845,7 @@ impl<'a, W: Write> Writer<'a, W> {
                 // Finally write the global name and end the global with a `;` and a newline
                 // Leading space is important
                 let vname = VaryingName {
-                    binding: &Binding::Location { location, interpolation: None },
+                    binding: &Binding::Location { location, interpolation: None, sampling: None },
                     stage: self.entry_point.stage,
                     output,
                 };
@@ -2201,8 +2221,15 @@ fn glsl_interpolation(interpolation: Interpolation) -> &'static str {
         Interpolation::Perspective => "smooth",
         Interpolation::Linear => "noperspective",
         Interpolation::Flat => "flat",
-        Interpolation::Centroid => "centroid",
-        Interpolation::Sample => "sample",
+    }
+}
+
+/// Return the GLSL auxiliary qualifier for the given sampling value.
+fn glsl_sampling(sampling: Sampling) -> &'static str {
+    match sampling {
+        Sampling::Center => "",
+        Sampling::Centroid => "centroid",
+        Sampling::Sample => "sample",
     }
 }
 
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -831,7 +831,9 @@ impl<'a, W: Write> Writer<'a, W> {
                 // here, regardless of the version.
                 if let Some(sampling) = sampling {
                     if emit_interpolation_and_auxiliary {
-                        write!(self.out, "{} ", glsl_sampling(sampling))?;
+                        if let Some(qualifier) = glsl_sampling(sampling) {
+                            write!(self.out, "{} ", qualifier)?;
+                        }
                     }
                 }
 
@@ -2225,11 +2227,11 @@ fn glsl_interpolation(interpolation: Interpolation) -> &'static str {
 }
 
 /// Return the GLSL auxiliary qualifier for the given sampling value.
-fn glsl_sampling(sampling: Sampling) -> &'static str {
+fn glsl_sampling(sampling: Sampling) -> Option<&'static str> {
     match sampling {
-        Sampling::Center => "",
-        Sampling::Centroid => "centroid",
-        Sampling::Sample => "sample",
+        Sampling::Center => None,
+        Sampling::Centroid => Some("centroid"),
+        Sampling::Sample => Some("sample"),
     }
 }
 
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -242,7 +242,7 @@ struct VaryingName<'a> {
 impl fmt::Display for VaryingName<'_> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self.binding {
-            Binding::Location(location, _) => {
+            Binding::Location { location, .. } => {
                 let prefix = match (self.stage, self.output) {
                     (ShaderStage::Compute, _) => unreachable!(),
                     // pipeline to vertex
@@ -794,7 +794,7 @@ impl<'a, W: Write> Writer<'a, W> {
             }
             _ => {
                 let (location, interpolation) = match binding {
-                    Some(&Binding::Location(location, interpolation)) => (location, interpolation),
+                    Some(&Binding::Location { location, interpolation }) => (location, interpolation),
                     _ => return Ok(()),
                 };
                 // Write the interpolation modifier if needed
@@ -825,7 +825,7 @@ impl<'a, W: Write> Writer<'a, W> {
                 // Finally write the global name and end the global with a `;` and a newline
                 // Leading space is important
                 let vname = VaryingName {
-                    binding: &Binding::Location(location, None),
+                    binding: &Binding::Location { location, interpolation: None },
                     stage: self.entry_point.stage,
                     output,
                 };
--- a/src/front/wgsl/conv.rs
+++ b/src/front/wgsl/conv.rs
@@ -44,9 +44,15 @@ pub fn map_interpolation(word: &str) -> Result<crate::Interpolation, Error<'_>> 
     match word {
         "linear" => Ok(crate::Interpolation::Linear),
         "flat" => Ok(crate::Interpolation::Flat),
-        "centroid" => Ok(crate::Interpolation::Centroid),
-        "sample" => Ok(crate::Interpolation::Sample),
         "perspective" => Ok(crate::Interpolation::Perspective),
+        _ => Err(Error::UnknownAttribute(word)),
+    }
+}
+
+pub fn map_sampling(word: &str) -> Result<crate::Sampling, Error<'_>> {
+    match word {
+        "centroid" => Ok(crate::Sampling::Centroid),
+        "sample" => Ok(crate::Sampling::Sample),
         _ => Err(Error::UnknownAttribute(word)),
     }
 }
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -507,6 +507,10 @@ impl BindingParser {
         match (self.location, self.built_in, self.interpolation, self.sampling) {
             (None, None, None, None) => Ok(None),
             (Some(location), None, interpolation, sampling) => {
+                // Before handing over the completed `Module`, we call
+                // `apply_common_default_interpolation` to ensure that the interpolation and
+                // sampling have been explicitly specified on all vertex shader output and fragment
+                // shader input user bindings, so leaving them potentially `None` here is fine.
                 Ok(Some(crate::Binding::Location { location, interpolation, sampling }))
             }
             (None, Some(bi), None, None) => Ok(Some(crate::Binding::BuiltIn(bi))),
@@ -2769,6 +2773,7 @@ impl Parser {
                         log::error!("Reached the end of file, but scopes are not closed");
                         return Err(Error::Other.as_parse_error(lexer.source));
                     };
+                    module.apply_common_default_interpolation();
                     return Ok(module);
                 }
             }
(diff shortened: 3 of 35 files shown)


src/front/wgsl/mod.rs
```diff
@@ -452,24 +454,35 @@ impl BindingParser {
                 lexer.expect(Token::Paren('('))?;
                 let raw = lexer.next_ident()?;
                 self.interpolation = Some(conv::map_interpolation(raw)?);
-                lexer.expect(Token::Paren(')'))?;
+                match lexer.next() {
```
kvark, line 457: we can write it simpler:
```rust
if lexer.skip(Token::Separator(',')) {
...
}
```

src/valid/interface.rs
```diff
@@ -215,6 +215,11 @@ impl VaryingContext<'_> {
                 if !needs_interpolation && interpolation.is_some() {
```
kvark, line 215: `needs_interpolation` has to change now, since WGSL demands the vertex outputs to have matching interpolation, and we try to follow WGSL in the IR

src/valid/interface.rs
```diff
@@ -215,6 +215,11 @@ impl VaryingContext<'_> {
...
                     return Err(VaryingError::InvalidInterpolation);
                 }
+                // It doesn't make sense to specify a sampling when
```
kvark, line 218: nice to see this described!

tests/in/interpolate.wgsl
```diff
@@ -0,0 +1,12 @@
+struct FragmentInput {
+  [[location(0), interpolate(flat)]] flat : f32;
```
kvark, line 2: let's have some integer attributes as well

tests/out/interpolate.ron.snap
```diff
@@ -0,0 +1,128 @@
+---
```
kvark, line 1: I wonder if we can disable the IR snapshot and instead enable the MSL/SPV/GLSL backends, thus getting better coverage, and implicitly still testing IR

src/front/wgsl/mod.rs
```diff
@@ -452,24 +454,35 @@ impl BindingParser {
...
         }
         Ok(())
     }
 
     fn finish<'a>(self) -> Result<Option<crate::Binding>, Error<'a>> {
-        match (self.location, self.built_in, self.interpolation) {
-            (None, None, None) => Ok(None),
-            (Some(loc), None, interpolation) => {
-                Ok(Some(crate::Binding::Location(loc, interpolation)))
+        match (self.location, self.built_in, self.interpolation, self.sampling) {
```
kvark, line 475: There is a discussion to be had on whether the IR can accept `None` for interpolation where it's expected to be, i.e. whether IR should have a notion of the default interpolation. I think the IR is stronger if it doesn't have defaults. 

That is, whenever IR sees that an interpolation would be needed (i.e. vertex outputs and fragment inputs), it should expect it to be `Some` and error otherwise. That means, in turn that all the frontends would need to resolve their "default" semantics, and it makes sense to me given that it's *their* default to resolve. I.e. the defaults can be different in frontends. For example SPIR-V probably always defaults to "Perspective", while WGSL defaults to "Flat" for integers:
> User-defined IO of scalar or vector integer type is always [[interpolate(flat)]] and, therefore, must not be specified in a WGSL program.

Resolving the default interpolation may be a little bit annoying, since it's probably the code that all frontends would need to use, similarly: scan all entry points, scan all the IO structures they use, check all the locations on them. So I think we'd need a small processor, e.g. `src/proc/interpolator.rs`, that frontends would run before producing a module.

src/lib.rs
```diff
@@ -465,7 +478,11 @@ pub enum Binding {
     /// Built-in shader variable.
     BuiltIn(BuiltIn),
     /// Indexed location.
-    Location(u32, Option<Interpolation>),
+    Location {
+        location: u32,
+        interpolation: Option<Interpolation>,
+        sampling: Option<Sampling>
```
kvark, line 484: if I understand correctly, this has to be specified if `interpolation` is either `Some(Perspective)` or `Some(Linear)`, correct?

kvark (commented):
This is just beautiful, thank you!
I have a few notes to discuss.

src/front/wgsl/mod.rs
jimblandy: Ah, nice.

src/valid/interface.rs
jimblandy: Okay, so this entails finding the matching declaration in the next / previous pipeline stage and checking them against each other, right?

kvark: This code doesn't know about which shaders this will connect to. It's going to be `wgpu` that validates the pipeline interface. But this code can at least check that vertex outputs and fragment inputs *have* interpolation. Right now, `needs_interpolation` is only true for fragment inputs

jimblandy: Ah, okay. That makes more sense now.

tests/in/interpolate.wgsl
jimblandy: What sorts of attributes do you mean, exactly? Do you just mean, we should be mixing in other sorts of attributes in the list?

tests/out/interpolate.ron.snap
jimblandy: Seems easy, I'll give it a try.

src/lib.rs
jimblandy: Yes. An earlier version of this PR had the `Perspective` and `Linear` variants of `Interpolation` carry the `Sampling` values, but it was pretty unwieldy for the front ends. This code would be one of the sites to do the default resolution you referred to earlier, correct?

kvark: Yes. I was just thinking if it makes sense to have `interpolation: Option<(Interpolation, Sampling)>`, and the downside would be that the frontends would have to specify the `Sampling` for `Flat`, which wouldn't be nice.
So I think what you have here is better.

tests/in/interpolate.wgsl
kvark: The test you added only uses `f32` type. I'm just suggesting to have a few members of this struct to be `i32` or `u32` instead.

jimblandy: Oh - of course. I had forgotten entirely about the right side of the colon...

jimblandy:
294c9e7 incorporates all the review comments except the suggestion that the IR require interpolation defaults to be resolved by the front ends.

kvark (approved):

jimblandy:
The only CI failure remaining in 68be808 is a problem with the MSL generated for `interpolation.wgsl`: if I understand correctly, it's unhappy that the vertex shader didn't actually return a position. I'll work that out tomorrow.

kvark:
Thank you! Please let me know how you'd like this to be merged.
If we aren't squashing, then we'd strongly prefer each commit it be working on its own. I suspect this isn't the case here.

jimblandy:
@kvark You approved the changes—are you satisfied with landing this without a change to have the front ends resolve interpolation defaults? I can do that in a follow-up, if you like.

kvark:
To clarify, I approved everything that was in here, but fixing the frontends (at least WGSL and SPIR-V) appear to be essential for this PR.

jimblandy:
Each commit compiles and passes `cargo test`, `make validate-spv validate-glsl`, and `cargo clippy --all-features -D warnings`. Once I fix the MSL validation, they should all pass CI. I think the history will be a bit clearer if we don't squash.

kvark:
Could you clarify the frontends situation? Are SPV and WGSL fixes going to be a part of this PR?

jimblandy:
> Could you clarify the frontends situation? Are SPV and WGSL fixes going to be a part of this PR?

Yes, I will do SPV, WGSL, and GLSL. I haven't looked at any MSL documentation yet, but if it's clear enough, then I can do that too.

I read [your earlier comment](https://github.com/gfx-rs/naga/pull/689#issuecomment-818421904) to mean these front-end fixes were a prerequisite for getting this merged, and my thumbs-up was meant to say, "Sure, I'll do that next."

jimblandy:
The current push has some [significant changes](https://github.com/gfx-rs/naga/pull/689/files#diff-cc33bd3c8c18000d3ee7b6f5578e8fa07ffd08d87b0aa3d9525a107937fb3d3fR112-R141) to the GLSL generation, to take the version into account. (This addresses some problems revealed by having the snapshot tests generate more than just the IR, so, good call there.)

But I wonder if it shouldn't be an error if the selected GLSL version can't express what's requested in the IR. @kvark, what do you think?

kvark:
> But I wonder if it shouldn't be an error if the selected GLSL version can't express what's requested in the IR. @kvark, what do you think?

Should be an error. What else can we do?

jimblandy:
> Should be an error. What else can we do?

Okay, I agree. I'll fix that.

src/back/glsl/mod.rs
```diff
@@ -106,6 +107,38 @@ impl Version {
...
+    /// Checks if this version supports a qualifier for the given `Interpolation`.
+    fn supports_interpolation(self, interpolation: Interpolation) -> bool {
+        use Interpolation::*;
+        match interpolation {
+            Perspective => true,                   // `smooth`: the original
+            Linear => match self {                 // `noperspective`
+                Version::Desktop(_) => true,       // added in GLSL 1.30 / OpenGL 3.0
+                Version::Embedded(_) => false,     // absent as of OpenGL ES 3.2
+            }
+            Flat => match self {                   // `flat`
```
kvark, line 120: so it's true everywhere, right? Do we still need this `match`?

src/back/glsl/mod.rs
```diff
@@ -106,6 +107,38 @@ impl Version {
...
+            }
+        }
+    }
+
+    /// Return true if this version supports a qualifier for the given `Sampling`.
+    fn supports_sampling(self, sampling: Sampling) -> bool {
+        use Sampling::*;
+        match sampling {
+            Center => true,                        // original behavior
+            Centroid => match self {               // `centroid`
```
kvark, line 132: same here

src/front/wgsl/conv.rs
```diff
@@ -44,13 +44,19 @@ pub fn map_interpolation(word: &str) -> Result<crate::Interpolation, Error<'_>>
     match word {
         "linear" => Ok(crate::Interpolation::Linear),
         "flat" => Ok(crate::Interpolation::Flat),
-        "centroid" => Ok(crate::Interpolation::Centroid),
-        "sample" => Ok(crate::Interpolation::Sample),
         "perspective" => Ok(crate::Interpolation::Perspective),
         _ => Err(Error::UnknownAttribute(word)),
```
kvark, line 48: we'd need better errors here, but this can be a follow-up

src/front/wgsl/mod.rs
```diff
@@ -498,16 +504,17 @@ impl BindingParser {
...
+            (None, None, None, None) => Ok(None),
```
kvark, line 508: We need to have a pass, after the module is mostly read, that would fill out the defaults where they are expected.
And we'll need to put a comment here saying that this is handled in this pass.

tests/out/interpolate.msl
```diff
@@ -0,0 +1,56 @@
...
+    out.perspective_sample = 2744.0;
+    const auto _tmp = out;
+    return main1Output { _tmp.position, _tmp.flat, _tmp.linear, _tmp.linear_centroid, _tmp.linear_sample, _tmp.perspective, _tmp.perspective_centroid, _tmp.perspective_sample };
+}
+
+
+struct main2Input {
+    metal::uint flat [[user(loc0)]];
+    float linear [[user(loc1)]];
+    metal::float2 linear_centroid [[user(loc2)]];
```
kvark, line 44: we still need to implement the qualifiers in MSL backend, it looks like

tests/out/interpolate.Fragment.glsl
```diff
@@ -0,0 +1,28 @@
...
+    vec3 linear_sample;
+    vec4 perspective;
+    float perspective_centroid;
+    float perspective_sample;
+};
+
+flat layout(location = 0) in uint _vs2fs_location0;
+layout(location = 1) in float _vs2fs_location1;
+layout(location = 2) centroid in vec2 _vs2fs_location2;
+layout(location = 3) in vec3 _vs2fs_location3;
```
kvark, line 19: hmm, why is there no `noperspective` on any of them?

tests/out/interpolate.spvasm.snap
```diff
@@ -0,0 +1,217 @@
+---
+source: tests/snapshots.rs
```
kvark, line 2: please remove all the ".snap" files

kvark (changes requested):
Looks close to be ready! A few notes now left

src/back/glsl/mod.rs
jimblandy: All of this needs to use `FeatureManager` instead of this sort of ad-hoc glue.

tests/out/interpolate.Fragment.glsl
jimblandy: The version of this PR rebased on #722 has `noperspective` as required. The omission here was caused by the initial bogus behavior where GLSL output just omitted qualifiers not supported in the requested version. With #722, requesting unsupported qualifiers causes an error, and the snapshot tests request a GLSL version that does support `noperspective`.

jimblandy:
This PR is now dependent on #722.

kvark:
Needs a rebase :)

jimblandy:
Still working on making sure the interpolation for all vertex shader outputs and fragment shader inputs is resolved. All our languages have the same default: perspective-correct center-sampled interpolation, so I think I can take a simpler approach than what I'd been trying. I'm expecting to finish this up tomorrow.

kvark:
Note that WGSL defaults to Flat for integer varyings, unlike other APIs.

jimblandy:
> Note that WGSL defaults to Flat for integer varyings, unlike other APIs.

I looked through the other APIs and here's what I came up with:

- SPIR-V:"Fragment shader inputs that are signed or unsigned integers, integer vectors, or any double-precision floating-point type must be decorated with Flat."

- GLSL: "Fragment shader inputs that are, or contain, integral or double-precision floating-point types must be qualified with the interpolation qualifier flat."

- MSL: "`center_perspective` is the default sampling and interpolation attribute, with the following exceptions: ... For an integer variable, the only valid sampling and interpolation attribute is `flat`."

So the variation seems to be in whether integral types default to, or must be specified as, flat.

kvark:
@jimblandy yes, right. It's certain that all the languages *have to* accept Flat there and only Flat.
I wonder what the SPV/GLSL validation says about the interpolation attributes being missing. Would it consider them being Perspective and then fail validation? Or something else?

Regardless though, since we *assume* that frontends are valid, we can safely assume that None == Flat there. You can ignore my note about WGSL default, it's not changing anything for us here.

jimblandy:
I believe all commits are CI-clean.

src/proc/mod.rs
```diff
@@ -212,7 +213,131 @@ impl crate::Binding {
...
+    /// binding that can vary, and everything else either defaults to flat, or
+    /// requires an explicit flat qualifier/attribute/what-have-you.
+    ///
+    /// [`Binding`s]: super::Binding
+
+    pub fn apply_common_default_interpolation(&mut self) {
+        use super::{Binding, ScalarKind, Type, TypeInner};
+
+        fn type_is_variable(ty: Handle<Type>, arena: &Arena<Type>) -> bool {
+            match &arena[ty].inner {
```
kvark, line 252: why not `match arena[ty].inner`? A reference doesn't give us anything here

src/proc/mod.rs
```diff
@@ -212,7 +213,131 @@ impl crate::Binding {
     pub fn to_built_in(&self) -> Option<crate::BuiltIn> {
         match *self {
             Self::BuiltIn(bi) => Some(bi),
-            Self::Location(..) => None,
+            Self::Location { .. } => None,
+        }
+    }
+}
+
+impl super::Module {
```
kvark, line 221: Let's move this to `src/proc/interpolator.rs` (or another name, but having a separate module).

src/proc/mod.rs
```diff
@@ -212,7 +213,131 @@ impl crate::Binding {
...
+    /// perspective-correct, center-sampled interpolation is the default for any
+    /// binding that can vary, and everything else either defaults to flat, or
+    /// requires an explicit flat qualifier/attribute/what-have-you.
+    ///
+    /// [`Binding`s]: super::Binding
+
+    pub fn apply_common_default_interpolation(&mut self) {
+        use super::{Binding, ScalarKind, Type, TypeInner};
+
+        fn type_is_variable(ty: Handle<Type>, arena: &Arena<Type>) -> bool {
```
kvark, line 251: I don't think we actually need to detect this. All the varyings we care about are strictly scalars and vectors. So we'd just need to check the I/O structs, reach the leaf nodes in them (i.e. non-structs), and place the defaults according to the `ScalarKind`.

src/proc/mod.rs
```diff
@@ -212,7 +213,131 @@ impl crate::Binding {
...
+        {
+            if let Some(binding) = binding.as_mut() {
+                default_binding(binding, ty, types);
+            } else if let TypeInner::Struct { ref members, .. } = types[ty].inner {
+                let new_ty_bindings = members.iter()
+                    .map(|member| {
+                        // unwrap: all members of structs used as `EntryPoint`
+                        // arguments or return values must have bindings.
+                        let mut binding_copy = member.binding.as_ref().unwrap().clone();
+                        default_binding(&mut binding_copy, member.ty, types);
```
kvark, line 299: Do we actually need to resolve it here? That forces us to carry around `Vec<Binding>`. Instead, since we are iterating the members anyway (when going through the deferred list), we could then call `default_binding` in place.

So we really need just a set of Type handles relevant to I/O structs here. It could be a `BitSet`, where bit N corresponds to `type_handle.index() == N`.

tests/out/interpolate.msl
kvark: looks like this is still a TODO? I'm not seeing the interpolation qualifiers here

jimblandy:
Oh, right - we'd decided to add MSL as part of this PR. I'll add that.

src/proc/mod.rs
jimblandy: Ah, okay. Clearly it's only the leaves that can have meaningful bindings.

src/proc/mod.rs
jimblandy: The deferred set will always need to somehow record how each member is to be defaulted. If we have mutable access to the `StructMember`, then by definition we don't have any access to its type at all, but we need to check the type to decide how to default it.

jimblandy: Some members need to be defaulted to perspective interpolation, others to flat, based on their type. So the result of the scan needs to be recording information about individual members; a `BitSet` of type indices isn't enough.

kvark: > If we have mutable access to the StructMember, then by definition we don't have any access to its type at all

I don't understand this part. If you have `member: StructMember`, then `module.types[member.ty]` would give you a type that you can check to figure out the default interpolation

kvark: My apologies, I understand it now. We can come back to this later.
It's a bit unfortunate that this processor would have a vec of a vec of things, but it's not a big concern, really.

jimblandy: Actually, I found a nice solution for this. No allocations at all, just some surprising mutations.

src/proc/interpolator.rs
```diff
@@ -0,0 +1,119 @@
...
+                TypeInner::Struct { members: ref mut m, .. } => {
+                    // To choose the right interpolations for `members`, we must consult other
+                    // elements of `types`. But both `members` and the types it refers to are stored
+                    // in `types`, and Rust won't let us mutate one element of the `Arena`'s `Vec`
+                    // while reading others.
+                    //
+                    // So, temporarily swap the member list out its type, assign appropriate
+                    // interpolations to its members, and then swap the list back in.
+                    use std::mem::replace;
+                    let mut members = replace(m, vec![]);
```
kvark, line 54: aha! Glad to see you found a way out of this :)

src/proc/interpolator.rs
```diff
@@ -0,0 +1,119 @@
...
+                        _ => unreachable!("ty must be a struct"),
+                    };
+                }
+
+                // Some interpolatable type.
+                //
+                // GLSL has 64-bit floats, but it won't interpolate them. WGSL and MSL only have
+                // 32-bit floats. SPIR-V has 16- and 64-bit float capabilities, but Vulkan is vague
+                // about what can and cannot be interpolated.
+                TypeInner::Scalar { kind: ScalarKind::Float, width: 4 } |
```
kvark, line 74: why are we checking for `width: 4` specifically in these cases? I imagine, once we have other widths, we'd need to change it anyway, and right now this processor is not the right place to validate the width, anyway.

kvark (approved):

kvark:
Thank you for wrapping this up! There is one thing that I think we should address (about `width: 4`) but otherwise it's as good as it gets, very high quality work here 🎖️

src/proc/interpolator.rs
kvark: that's an interesting discovery! I wonder what other backends think of this.

PR #690, opened 2021-04-10
[spv-in] stop loop body search on a continue label instead of the merge

### LINKED ISSUE #683: Incorrect loop continue parsing in SPIR-V

kvark:
Fixes #683
Tested on Dota2 - seems to work.

### DIFF
--- a/src/front/spv/error.rs
+++ b/src/front/spv/error.rs
@@ -51,7 +51,6 @@ pub enum Error {
     BadString,
     IncompleteData,
     InvalidTerminator,
-    InvalidEdgeClassification,
     FunctionCallCycle(spirv::Word),
     // incomplete implementation error
     UnsupportedMatrixStride(spirv::Word),
--- a/src/front/spv/flow.rs
+++ b/src/front/spv/flow.rs
@@ -322,15 +322,12 @@ impl FlowGraph {
             },
             Some(ControlFlowNodeType::Loop) => {
                 let merge_node_index = self.block_to_node[&node.merge.unwrap().merge_block_id];
-                let continuing: crate::Block = {
-                    let continue_edge = self
-                        .flow
-                        .edges_directed(node_index, Direction::Outgoing)
-                        .find(|&ty| *ty.weight() == ControlFlowEdgeType::ForwardContinue)
-                        .unwrap();
-
-                    self.flow[continue_edge.target()].block.clone()
-                };
+                let continue_node_index = self
+                    .flow
+                    .edges_directed(node_index, Direction::Outgoing)
+                    .find(|&ty| *ty.weight() == ControlFlowEdgeType::ForwardContinue)
+                    .unwrap()
+                    .target();
 
                 let mut body = node.block.clone();
                 match node.terminator {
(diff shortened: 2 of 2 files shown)


MatusT:
This fix might work for Dota but it is not correct IMHO. Anything should really stop at the merge block.
The actual problem might be in the classification meaning that some blocks that should be recognised as `Continue blocks` are not.
When you look at the specification it says: 
```Continue Block: A block containing a branch to an OpLoopMerge instruction’s Continue Target.```
And as you can see in the code: 
```Some(ControlFlowNodeType::Continue) => Ok(node.block.clone())```
Continue blocks do not traverse further! And that makes complete sense, since:
1. Continuing blocks (the continue blocks point to) have a back-edge(as the only type of block in the CFG) which would cause loop behaviour.
2. It's their end. Continuing block is cloned/drained at the loop creation.

kvark:
> This fix might work for Dota but it is not correct IMHO. Anything should really stop at the merge block.

It's not that we don't want to stop at the merge block. It's that we *do* want to also stop at the continuing block, because otherwise we are including its contents multiple times. I was asking you about turning a single `merge_id: Option<>` into plural `merge_ids: Vec<>` (in the chat a while ago) for this reason: so that we can stop on either the merge block or continuing block.

I wasn't sure if we'd always traverse into the continuing block, and we actually do, at least on Dota shaders, hence just having `merge_id = Some(continuing_id)` works. But I understand that this may be missing other cases.

> Based on this I think it is a good idea to look at the produced CFG and see if there are some non-continue blocks pointing to continuing blocks.

Please check the flow graph in the description of #683, it's as simple as it can get.

MatusT:
> It's not that we don't want to stop at the merge block. It's that we _do_ want to also stop at the continuing block, because otherwise we are including its contents multiple times. I was asking you about turning a single `merge_id: Option<>` into plural `merge_ids: Vec<>` (in the chat a while ago) for this reason: so that we can stop on either the merge block or continuing block.

That's why I wrote why we always stop at continue block. With correctly classified CFG we currently always stop at continue block. You can only get to continuing(back) block through Loop or Continue block. So there is no need to add additional stop for it.

> Please check the flow graph in the description of #683, it's as simple as it can get.

I did. And the result is fine now. I believe we fixed the problematic stuff already. With std::mem::replace no loops are happening, no block is cloned or visited twiced.

MatusT:
> It's not that we don't want to stop at the merge block. It's that we _do_ want to also stop at the continuing block, because otherwise we are including its contents multiple times. I was asking you about turning a single `merge_id: Option<>` into plural `merge_ids: Vec<>` (in the chat a while ago) for this reason: so that we can stop on either the merge block or continuing block.

That's why I wrote why we always stop at continue block. With correctly classified CFG we currently always stop at continue block. You can only get to continuing(back) block through Loop or Continue block. So there is no need to add additional stop for it.

> Please check the flow graph in the description of #683, it's as simple as it can get.

I did. And the result is fine now. I believe we fixed the problematic stuff already. With std::mem::replace no loops are happening, no block is cloned or visited twiced.

kvark:
Ok, thank you. Let's come back to this after we figure out #688

JCapucho:
Closing this since #1294 completely changes the way block parsing is done and this PR is working on something that no longer exists.

Issue #685, opened 2021-04-09
[SPV -> Metal] Output metal shader doesn't call a function that takes `OpTypePointer`s correctly


expenses:
I have the following glsl shader:
```glsl
#version 450

vec2 lerp(vec2 a, vec2 b, vec2 f)
{
    return a + (b-a)*f;
}

void main() {
    gl_Position = vec4(lerp(vec2(-1), vec2(1), vec2(0.5)), 0, 1);
}
```

which gets converted via `glslc` into the following SPIR-V (cropped for brevity):

```
...
%_ptr_Function_v2float = OpTypePointer Function %v2float
...
%lerp_vf2_vf2_vf2_ = OpFunction %v2float None %9
          %a = OpFunctionParameter %_ptr_Function_v2float
          %b = OpFunctionParameter %_ptr_Function_v2float
          %f = OpFunctionParameter %_ptr_Function_v2float
         %14 = OpLabel
         %15 = OpLoad %v2float %a
         %16 = OpLoad %v2float %b
         %17 = OpLoad %v2float %a
         %18 = OpFSub %v2float %16 %17
         %19 = OpLoad %v2float %f
         %20 = OpFMul %v2float %18 %19
         %21 = OpFAdd %v2float %15 %20
               OpReturnValue %21
               OpFunctionEnd
```

when running naga, this gets transformed into the following MSL shader:
```metal
constexpr constant metal::float2 const_type1_ = {-1.0, -1.0};
constexpr constant metal::float2 const_type1_1 = {1.0, 1.0};
constexpr constant metal::float2 const_type1_2 = {0.5, 0.5};

metal::float2 lerpvf2vf2vf2_(
    thread metal::float2* a,
    thread metal::float2* b,
    thread metal::float2* f
) {
    return a + ((b - a) * f);
}

void main1(
    thread gl_PerVertex& _
) {
    metal::float2 param;
    metal::float2 param1;
    metal::float2 param2;
    param = const_type1_;
    param1 = const_type1_1;
    param2 = const_type1_2;
    metal::float2 _e17 = lerpvf2vf2vf2_(param, param1, param2);
    _.gl_Position = metal::float4(_e17.x, _e17.y, 0.0, 1.0);
    return;
}
```

This fails to be converted to AIR because `lerpvf2vf2vf2_` is called incorrectly.
```
hmm.metal:25:25: error: invalid operands to binary expression ('long' and 'metal::float2 *' (aka 'float2 *'))
[...]

PR #692, opened 2021-04-10
[Metal] Determine whether a load expression should dereference a pointer

expenses:
This is one half of a fix for https://github.com/gfx-rs/naga/issues/685.

Now, instead of the incorrect
```metal
metal::float2 lerpvf2vf2vf2_(
    thread metal::float2* a,
    thread metal::float2* b,
    thread metal::float2* f
) {
    return a + ((b - a) * f);
}
```

we get

```metal
metal::float2 lerpvf2vf2vf2_(
    thread metal::float2* a,
    thread metal::float2* b,
    thread metal::float2* f
) {
    return *a + ((*b - *a) * *f);
}
```

The function is still called the wrong way though.

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -668,7 +668,17 @@ impl<W: Write> Writer<W> {
                 write!(self.out, "{}", name)?;
             }
             crate::Expression::Load { pointer } => {
-                //write!(self.out, "*")?;
+                let should_dereference = match context.info[pointer].ty {
+                    TypeResolution::Handle(handle) => {
+                        let ty = &context.module.types[handle];
+                        matches!(ty.inner, crate::TypeInner::Pointer { .. })
+                    }
+                    TypeResolution::Value(_) => false,
+                };
+
+                if should_dereference {
+                    write!(self.out, "*")?;
+                }
                 self.put_expression(pointer, context, is_scoped)?;
             }
             crate::Expression::ImageSample {
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -668,12 +668,12 @@ impl<W: Write> Writer<W> {
                 write!(self.out, "{}", name)?;
             }
             crate::Expression::Load { pointer } => {
-                let should_dereference = match context.info[pointer].ty {
-                    TypeResolution::Handle(handle) => {
-                        let ty = &context.module.types[handle];
-                        matches!(ty.inner, crate::TypeInner::Pointer { .. })
-                    }
-                    TypeResolution::Value(_) => false,
+                let should_dereference = match &context.function.expressions[pointer] {
+                    crate::Expression::AccessIndex { .. } => false,
+                    crate::Expression::Access { .. } => false,
+                    crate::Expression::LocalVariable { .. } => false,
+                    crate::Expression::GlobalVariable { .. } => false,
+                    _ => true,
                 };
 
                 if should_dereference {
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -668,7 +668,7 @@ impl<W: Write> Writer<W> {
                 write!(self.out, "{}", name)?;
             }
             crate::Expression::Load { pointer } => {
-                let should_dereference = match &context.function.expressions[pointer] {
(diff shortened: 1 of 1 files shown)


src/back/msl/writer.rs
```diff
@@ -668,7 +668,17 @@ impl<W: Write> Writer<W> {
                 write!(self.out, "{}", name)?;
             }
             crate::Expression::Load { pointer } => {
-                //write!(self.out, "*")?;
+                let should_dereference = match context.info[pointer].ty {
```
kvark, line 671: We need to do something different here. In the IR, the `pointer` *has* to be a pointer already, so this needs to always dereference, unconditionally. However, I imagine that if we see that the expression is `Access` or `AccessIndex`, then we know that there was already a root of this access chain somewhere, and we inserted `->` in the beginning (#699), so we don't need to have a dereference here.

kvark (changes requested):

src/back/msl/writer.rs
expenses: Oh right, I see! Yeah I thought that dereferencing all the time was wrong because it breaks `main2Output`:
```metal

struct main2Output {
    metal::float4 gl_Position1 [[position]];
    float gl_PointSize1 [[point_size]];
    type5 gl_ClipDistance1 [[clip_distance]];
};
vertex main2Output main2(
) {
    gl_PerVertex _ = {};
    main1(_);
    const auto _tmp = type9 {*_.gl_Position, *_.gl_PointSize, {*_.gl_ClipDistance[0]}};
    return main2Output { _tmp.gl_Position1, _tmp.gl_PointSize1, {_tmp.gl_ClipDistance1[0]} };
}
```
but that just means we need more changes.

expenses: Instead we want something like:
```metal
struct main2Output {
    metal::float4 gl_Position1 [[position]];
    float gl_PointSize1 [[point_size]];
    type3 gl_ClipDistance1 [[clip_distance]];
};
vertex main2Output main2(
) {
    gl_PerVertex _owned = {};
    auto _ = &_owned;
    main1(*_);
    const auto _tmp = type7 {(*_).gl_Position, (*_).gl_PointSize, {(*_).gl_ClipDistance[0]}};
    return main2Output { _tmp.gl_Position1, _tmp.gl_PointSize1, {_tmp.gl_ClipDistance1[0]} };
}
```

expenses: Or actually just
```metal
vertex main2Output main2(
) {
    gl_PerVertex _owned = {};
    thread gl_PerVertex& _ = _owned;
    main1(_);
    const auto _tmp = type7 {_.gl_Position, _.gl_PointSize, {_.gl_ClipDistance[0]}};
    return main2Output { _tmp.gl_Position1, _tmp.gl_PointSize1, {_tmp.gl_ClipDistance1[0]} };
}
```

expenses: Oh no, nevermind, simply
```metal
ertex main2Output main2(
) {
    gl_PerVertex _ = {};
    main1(_);
    const auto _tmp = type7 {_.gl_Position, _.gl_PointSize, {_.gl_ClipDistance[0]}};
    return main2Output { _tmp.gl_Position1, _tmp.gl_PointSize1, {_tmp.gl_ClipDistance1[0]} };
}
```
works.

expenses: It shouldn't be dereferenced in the case of `LocalVariable` and `GlobalVariable` either (unless the IR is wrong) because otherwise in the collatz msl we get:
```metal
metal::uint collatz_iterations(
    metal::uint n_base
) {
    metal::uint n;
    metal::uint i = 0u;
    n = n_base;
    while(true) {
        if (*n <= 1u) {
            break;
        }
        if ((*n % 2u) == 0u) {
            n = *n / 2u;
        } else {
            n = (3u * *n) + 1u;
        }
        i = *i + 1u;
    }
    return *i;
}
```
which is pretty clearly wrong.

expenses: After #693 is merged, `FunctionArgument`s are all passed by reference, and shouldn't be dereferenced either.

expenses: We might not need this after the changes to #693.

PR #693, opened 2021-04-10
[Metal] Change the function argument pointer type to use `&` instead of `*`

expenses:
This ~~, alongside https://github.com/gfx-rs/naga/pull/692, fully~~ fixes https://github.com/gfx-rs/naga/issues/685. ~~It feels a bit like a hacky solution though and I'm not sure if it's going to work in all cases.~~

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -1284,11 +1284,25 @@ impl<W: Write> Writer<W> {
                     }
                     let fun_name = &self.names[&NameKey::Function(function)];
                     write!(self.out, "{}(", fun_name)?;
+
+                    let function_arguments =
+                        &context.expression.module.functions[function].arguments;
+
                     // first, write down the actual arguments
-                    for (i, &handle) in arguments.iter().enumerate() {
+                    for (i, (&handle, argument)) in
+                        arguments.iter().zip(function_arguments).enumerate()
+                    {
                         if i != 0 {
                             write!(self.out, ", ")?;
                         }
+
+                        let ty = &context.expression.module.types[argument.ty];
+
+                        // If the function expects a pointer as an argument, take a reference.
+                        if matches!(ty.inner, crate::TypeInner::Pointer { .. }) {
+                            write!(self.out, "&")?;
+                        }
+
                         self.put_expression(handle, &context.expression, true)?;
                     }
                     // follow-up with any global resources used
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -91,7 +91,7 @@ impl<'a> Display for TypeContext<'a> {
                     Some(name) => name,
                     None => return Ok(()),
                 };
-                write!(out, "{} {}*", class_name, sub)
+                write!(out, "{} {}&", class_name, sub)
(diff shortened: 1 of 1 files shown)


expenses:
This seems to increase the stack size of `put_block` to `13632` up from the last recorded size of `12736`.

src/back/msl/writer.rs
```diff
@@ -1284,11 +1284,25 @@ impl<W: Write> Writer<W> {
...
+                    for (i, (&handle, argument)) in
+                        arguments.iter().zip(function_arguments).enumerate()
+                    {
                         if i != 0 {
                             write!(self.out, ", ")?;
                         }
+
+                        let ty = &context.expression.module.types[argument.ty];
+
+                        // If the function expects a pointer as an argument, take a reference.
```
kvark, line 1301: Is there a reason for this to be handled specifically for function parameters? I'd expect it to be handled more generally by `put_expression` instead.

kvark (changes requested):

src/back/msl/writer.rs
expenses: That probably would be better, but I'm not sure how that should be done. In the trace I just have

```
[TRACE naga::back::msl::writer] statement[1] Call { function: [1], arguments: [[15], [16], [17]], result: Some([18]) }
[TRACE naga::back::msl::writer] expression [15] = LocalVariable([1])
[TRACE naga::back::msl::writer] expression [16] = LocalVariable([2])
[TRACE naga::back::msl::writer] expression [17] = LocalVariable([3])
```
and I can't see how a `LocalVariable` can know that it's meant to be referenced with `&` with the information it has available.

kvark: Perhaps the trace log is misleading? Our validation strictly requires the types of arguments to match, see https://github.com/gfx-rs/naga/blob/98252cf5d29790f85cc4397e0cf7a5b6cfaac614/src/valid/function.rs#L202
So the arguments passed into a call (of a function that expects pointers) are pointers, given that the validation succeeded.

expenses:
The fix for this was actually pretty easy; MSL casts to references but not pointers so we just have to make the function take those instead.

kvark:
Why are we coercing pointers to references, exactly?
This may be the way to go, I just want to understand what are the blockers for treating pointers as real pointers.

expenses:
> Why are we coercing pointers to references, exactly?
> This may be the way to go, I just want to understand what are the blockers for treating pointers as real pointers.

As far as I can tell, references are the way to go in C++, and pointers should only be used when necessary. Using references for metal makes a ton of stuff easier, such as not requiring `&` when passing arguments and not needing the changes from https://github.com/gfx-rs/naga/pull/697 etc.

The one downside is that it makes the semantics a little different from the IR, as you're never having to deref using `*` in `Expression::Load` etc.

kvark:
I suppose it's fine, given that we already generate a ton of references in the boilerplate code.

kvark (approved):

kvark:
I ran the Dota shader conversion with this and didn't see any issues.

Issue #694, opened 2021-04-10
[Metal] Assignments to pointer function arguments aren't done correctly


expenses:
Here's another one in the same vein as https://github.com/gfx-rs/naga/issues/685. With this glsl shader compiled with `glslc`:
```glsl
#version 440

layout(location = 0) in vec2 uv;
layout(location = 0) out vec4 colour;

float hash(vec2 p) {
      p = fract(p*vec2(931.733,354.285));
      p += dot(p,p+39.37);
      return fract(p.x*p.y);
}

void main() {
	colour = vec4(hash(uv));
}
```
we get the following metal function (when compiled on a branch with https://github.com/gfx-rs/naga/pull/692 and https://github.com/gfx-rs/naga/pull/693 merged):
```metal
constexpr constant metal::float2 const_type1_ = {931.7329711914063, 354.2850036621094};

float hashvf2_(
    thread metal::float2* p
) {
    p = metal::fract(*p * const_type1_);
    float _e20 = metal::dot(*p, *p + metal::float2(39.369998931884766, 39.369998931884766));
    p = *p + metal::float2(_e20, _e20);
    return metal::fract(p[0u] * p[1u]);
}
```

which errors on compilation with:
```metal
3d_test.metal:9:9: error: assigning to 'metal::float2 *' (aka 'float2 *') from incompatible type 'metal::float2' (aka 'float2')
    p = metal::fract(*p * const_type1_);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
3d_test.metal:11:12: error: assigning to 'metal::float2 *' (aka 'float2 *') from incompatible type 'metal::float2' (aka 'float2')
    p = *p + metal::float2(_e20, _e20);
        ~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
3d_test.metal:12:12: error: cannot initialize return object of type 'float' with an rvalue of type 'metal::float2' (aka 'float2')
    return metal::fract(p[0u] * p[1u]);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~
3 errors generated.
```

`p = ...` needs to become `*p = ...` and `p[...]` needs to become `(*p)[...]`. However, if we do that, we're altering the value of the pointer that's passed in, so the semantics would be different from the glsl (but maybe not the spir-v?).

expenses:
Would be fixed by https://github.com/gfx-rs/naga/pull/693

PR #697, opened 2021-04-10
[Metal] More dereferencing

expenses:
Fixes https://github.com/gfx-rs/naga/issues/694


expenses:
We might not need this after the changes to #693.

PR #698, opened 2021-04-10
Fix validation for `textureDimensions(t)` call

### LINKED ISSUE #696: Validation error on `textureDimensions(t)` call
Based on the WGSL spec I understand that the following pseudo code should work and return texture dimensions:

```
[[group(0), binding(1)]]
var t: texture_2d<f32>;
...
// ... and then in a vertex stage:
const size = textureDimensions(t);
```

However wgpu master with naga reports a validation error:

```
Caused by:
    Entry point main at Vertex is invalid
    Expression [7] is invalid
    Unable to operate on image class Sampled { kind: Float, multi: false }
```

Is this expected behavior or a potential bug?

simast:
Fixes #696

Based on the [spec for textureDimensions()](https://gpuweb.github.io/gpuweb/wgsl/#texturedimensions) - it should be valid for all image classes without mip level param.

### DIFF
--- a/src/valid/expression.rs
+++ b/src/valid/expression.rs
@@ -592,10 +592,10 @@ impl super::Validator {
                         };
                         let good = match query {
                             crate::ImageQuery::NumLayers => arrayed,
+                            crate::ImageQuery::Size { level: None } => true,
                             crate::ImageQuery::Size { level: Some(_) }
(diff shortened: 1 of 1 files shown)


kvark (approved):
Looks good, thank you for fixing this!

Issue #695, opened 2021-04-10
[spv-in/mtl-out] handle `float length(float x)`


expenses:
`glsl` let's you call `length()` on a scalar (for some reason) which just returns the same scalar. SPIR-V does the same with:

```
%21 = OpLoad %float %20
%22 = OpExtInst %float %1 Length %21
```

Metal is a bit more sensible and doesn't have a `metal::length(scalar)` function, so at the moment any code that does the above will fail to compile. I've come up with a solution for the metal backend where the function is just removed if called on a scalar, but because multiple backends are likely to run into this it might be better to handle this while parsing the SPIR-V.

kvark:
Interesting. We need to make a call on whether the IR allows this or not.
To complete the investigation, let's check what WGSL and HLSL expect.

expenses:
`float distance(float a, float b)` has the same problem:
```
msl/4tdGRfImage.spv.metal:82:9: error: call to 'distance' is ambiguous
    if (metal::distance(uv[0u], 0.5 + ((metal::sin((uv[1u] - (_.iTime * 0.5)) * 10.0) * 0.05000000074505806) * uv[1u])) > (metal::sin(uv[1u]) * 0.30000001192092896)) {
        ^~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/metal/macos/lib/clang/31001.160/include/metal/metal_geometric:481:17: note: candidate function
METAL_FUNC half distance(half2 x, half2 y)
                ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/metal/macos/lib/clang/31001.160/include/metal/metal_geometric:512:17: note: candidate function
METAL_FUNC half distance(half3 x, half3 y)
                ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/metal/macos/lib/clang/31001.160/include/metal/metal_geometric:543:17: note: candidate function
METAL_FUNC half distance(half4 x, half4 y)
                ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/metal/macos/lib/clang/31001.160/include/metal/metal_geometric:648:18: note: candidate function
METAL_FUNC float distance(float2 x, float2 y)
                 ^
[...]

PR #706, opened 2021-04-11
[Metal] Handle `length(float)` and `distance(float, float)`

expenses:
Closes https://github.com/gfx-rs/naga/issues/695. It could probably be a little tidier though, I'm open to suggestions.
```glsl
#version 450

layout(location = 0) in float a;
layout(location = 1) in float b;
layout(location = 2) in float c;

void main() {
    float l = length(a);
    float d = distance(b, c);
    float ds = dot(b, c);
    colour = vec4(l, d, ds, 1.0);
}
```
->
```metal
void main1(
    thread float const& a,
    thread float const& b,
    thread float const& c,
    thread metal::float4& colour
) {
    float l;
    float d;
    float ds;
    l = metal::abs(a);
    d = metal::abs(b - c);
    ds = b * c;
    colour = metal::float4(l, d, ds, 1.0);
    return;
}
```

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -887,14 +887,13 @@ impl<W: Write> Writer<W> {
             } => {
                 use crate::MathFunction as Mf;
 
-                let ignore = fun == Mf::Length
-                    && match &context.info[arg].ty {
-                        TypeResolution::Handle(handle) => {
-                            let ty = &context.module.types[*handle];
-                            matches!(ty.inner, crate::TypeInner::Scalar { .. })
-                        }
-                        TypeResolution::Value(ty) => matches!(ty, crate::TypeInner::Scalar { .. }),
-                    };
+                let scalar_argument = match &context.info[arg].ty {
+                    TypeResolution::Handle(handle) => {
+                        let ty = &context.module.types[*handle];
+                        matches!(ty.inner, crate::TypeInner::Scalar { .. })
+                    }
+                    TypeResolution::Value(ty) => matches!(ty, crate::TypeInner::Scalar { .. }),
+                };
 
                 let fun_name = match fun {
                     // comparison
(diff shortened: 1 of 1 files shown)


expenses:
I'm unsure how to fix the clippy warnings - nothing I do seems to make it happy.

src/back/msl/writer.rs
```diff
@@ -944,8 +953,16 @@ impl<W: Write> Writer<W> {
                     Mf::ReverseBits => "reverse_bits",
                 };
 
-                write!(self.out, "{}::{}", NAMESPACE, fun_name)?;
-                self.put_call_parameters(iter::once(arg).chain(arg1).chain(arg2), context)?;
+                if fun == Mf::Distance && scalar_argument {
+                    write!(self.out, "{}::abs(", NAMESPACE)?;
+                    self.put_expression(arg, context, true)?;
```
expenses, line 958: I suppose these expressions should be wrapped in brackets actually

src/back/msl/writer.rs
```diff
@@ -887,6 +887,14 @@ impl<W: Write> Writer<W> {
             } => {
                 use crate::MathFunction as Mf;
 
+                let scalar_argument = match &context.info[arg].ty {
```
kvark, line 890: let's rewrite this as:
```rust
let scalar_argument = match context.resolve_type(arg) {
  crate::TypeInner::Scalar { .. } => true,
  _ => false,
};
```

src/back/msl/writer.rs
kvark: I don't think we have any way to guarantee this, other than passing `false` here and of the other argument

kvark (changes requested):
Thank you!
Needs a few touches before we can land it

src/back/msl/writer.rs
```diff
@@ -887,6 +887,11 @@ impl<W: Write> Writer<W> {
             } => {
                 use crate::MathFunction as Mf;
 
+                let scalar_argument = match context.resolve_type(arg) {
```
kvark, line 890: this should have "*", since otherwise you are matching a reference to the type

expenses: Oh interesting! I've never had to do that before.

kvark: yes, because we force-enable this requirement on clippy, and by default Rust tries to be too smart and derefefence for you. We don't want to have it too smart, we value explicitness more.

kvark:
Stack size expectations need to be adjusted

kvark (approved):

PR #709, opened 2021-04-11
[Metal] Add metal constants to keywords

expenses:
These are also from `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/metal/macos/lib/clang/31001.160/include/metal/metal_types`. The only ones I've actually seen 'in the wild' are `auto` and `MAXFLOAT`, but we might as well have them all.


kvark (commented):
Are the constants really conflicting with our variable names?

expenses:
> Are the constants really conflicting with our variable names?

Well, like I said `MAXFLOAT` is in one case. But this is a lot of constants, and if you want to add only a few of them and wait and see if other people have conflicts then that would make sense as well.

kvark:
I mean, what happens if you define something with this name?
My expectation is that it's fine, since the metal constant is in `metal::`, so it wouldn't conflict with anything.
I.e. it's not really a reserved word, it's just a constant in the namespace.

expenses:
> I mean, what happens if you define something with this name?
> My expectation is that it's fine, since the metal constant is in `metal::`, so it wouldn't conflict with anything.
> I.e. it's not really a reserved word, it's just a constant in the namespace.

This is odd to me too, but as far as I'm aware, these are all the root namespace. `MAXFLOAT`, [`NAN` and `INFINITY`](https://github.com/gfx-rs/naga/commit/a8f1022b125712a09f713f8dad0bd089b054013f) are atleast.

kvark:
Do you have a compile error that happens if you use this identifier?

expenses:
> Do you have a compile error that happens if you use this identifier?

Yep:
```glsl
#version 450

void main() {
    float MAXFLOAT = 5.0;
    colour = vec4(MAXFLOAT, MAXFLOAT * 2.0, 1.0, 1.0);
}
```
->
```metal
#include <metal_stdlib>
#include <simd/simd.h>


void main1(
    thread metal::float4& colour
) {
    float MAXFLOAT;
    MAXFLOAT = 5.0;
    colour = metal::float4(MAXFLOAT, MAXFLOAT * 2.0, 1.0, 1.0);
    return;
}
```
-> 
```
max.metal:8:11: error: expected unqualified-id
    float MAXFLOAT;
          ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/metal/macos/lib/clang/31001.160/include/metal/metal_types:211:18: note: expanded from macro 'MAXFLOAT'
#define MAXFLOAT __FLT_MAX__
                 ^
<built-in>:295:21: note: expanded from here
#define __FLT_MAX__ 3.40282347e+38F
                    ^
max.metal:9:14: error: expression is not assignable
    MAXFLOAT = 5.0;
    ~~~~~~~~ ^
2 errors generated.
```

kvark:
Ah, those are defines, ugh, C++ never stops surprising.

kvark (approved):

PR #716, opened 2021-04-13
Impl `Hash` for `msl::Options`

expenses:
This is needed in order to use it in a pipeline cache for https://github.com/gfx-rs/gfx/pull/3719. Most of the changes are trivial, but of note is the fact that we have to use a `BTreeMap` instead of a `FastHashMap` for the `BindingsMap` as `HashMap` doesn't (and can't, efficiently) implement `Hash`. We also have to use a custom `Hash` implementation for `InlineSampler` as it uses `f32`s.


expenses:
We might not need this, as naga is super fast already: https://github.com/gfx-rs/gfx/pull/3719#issuecomment-819078231

kvark:
I did prototype a similar change before, and I got worried that deriving so much stuff would have negative effect on the build times. Would you mind taking some measurements on how much time it takes to build Naga with Metal backend and all serialization before and after this change?

Another thing is adding the `ordered_float` dependency. We are keen on minimizing the dependencies. Would it be an option here to change `f32` to something, like wrap `f32` into a structure, or encode it in fixed-point format? For example, https://github.com/gfx-rs/gfx/blob/3da35c4339513de6a69e7794d42a63203f5820e5/src/hal/src/image.rs#L404

expenses:
I ran `cargo +nightly build --release -Z timings --features msl-out` and build times do appear to be a bit longer:
3f57e4db5adfd794d170843d7f1b8845dac1609c: 6.23s for naga
this branch: 6.79s for naga

kvark (approved):

kvark:
Given that this is required for caching (or persistently caching on disk, to be precise), seems like we can eat that cost in build times.
Let's resolve this clippy warning (i.e. you can silence it locally), and proceed then.

expenses:
It's not _required_ for hashing, we could easily just use a binary archive alone as a pipeline cache. Having a SPIR-V -> MSL is a nice extra that _very slightly_ improves shader compile times (by around a second for https://github.com/gfx-rs/gfx/pull/3719#issuecomment-819078231). We could very easily do without it or add it later.

kvark:
@expenses please clarify which parts of the PR are needed for *any* kind of caching versus parts that are needed for SPV -> MSL cache specifically.

expenses:
You can do Binary Archive caching without this PR, it is only needed for the SPV -> MSL cache.

kvark:
Building
```
cargo +nightly build --release -Z timings --features msl-out,serialize
```
I'm seeing quite a significant difference. Old: 7.7s, New 9.6s...

kvark:
So my conclusion so far is that the compile times suffer from deriving the traits on `Arena`, as opposed to anything happening in the `sampler` sub-module of the MSL backend. What I tried therefore is to just manually derive everything on the `Options` instead of changing the `Arena`. See https://github.com/kvark/naga/commit/1c29085e3f12eb8b06c03b09103ec81f12c5a085

The results are promising - 7.4s build time with the same command line. So it's basically free!
I have a few ideas how we can reduce the code size here.

kvark:
I tried to just replace `Arena` by `Vec` in https://github.com/kvark/naga/commit/df5a4f086740968259e3f9d724f803f7a0d85927, and the times show as 7.6s. It's auto-derived to the same level as your PR.

We have the following choices to proceed:
  1. do this ^ change and land
  2. introduce a new feature in `naga`, something like `derive`, which would add those derives on the `Arena`. Presumably, it's going to be enabled by gfx-rs's pipeline caching feature but likely disabled by `wgpu` for the nearest future.
  3. not doing anything, rely purely on binary archives

@expenses do you have a feeling, or a preference, on which path would be the best to proceed?

expenses:
> I tried to just replace `Arena` by `Vec` in [kvark@df5a4f0](https://github.com/kvark/naga/commit/df5a4f086740968259e3f9d724f803f7a0d85927), and the times show as 7.6s. It's auto-derived to the same level as your PR.
> 
> We have the following choices to proceed:
> 
>     1. do this ^ change and land
> 
>     2. introduce a new feature in `naga`, something like `derive`, which would add those derives on the `Arena`. Presumably, it's going to be enabled by gfx-rs's pipeline caching feature but likely disabled by `wgpu` for the nearest future.
> 
>     3. not doing anything, rely purely on binary archives
> 
> 
> @expenses do you have a feeling, or a preference, on which path would be the best to proceed?

If 1) is not a problem for you then let's go with that. Otherwise 2) is not bad either.

kvark:
Let's go with (1) then

expenses:
Damn, we're getting an error because the skybox examples has this line:
https://github.com/gfx-rs/naga/blob/2eb3e81e19630410caa10821fcfcaba4f98602ce/tests/in/skybox.param.ron#L13
What's the solution here? Do inline sampler indexes just start with 1?

kvark:
> What's the solution here? Do inline sampler indexes just start with 1?

We just need to switch it to 0 now in your PR.
It was the `Handle` before your change, and handles start with 1, so that `Option<Handle>` is free.

kvark (approved):

PR #717, opened 2021-04-13
Makefile: Be more flexible in removing YAML headers from snapshots.

jimblandy:
The snapshot files in tests/out start with headers like this:

    ---
    source: tests/snapshots.rs
    expression: dis
    ---
    ; SPIR-V
    ; Version: 1.0
    ; Generator: rspirv
    ...

The Makefile's `validate-foo` targets trim off those lines with `tail -n +5`
before submitting their test to the validators.

However, some versions of the `insta` crate seem to generate headers with an
extra blank line, like this:

    ---
    source: tests/snapshots.rs
    expression: dis

    ---

This makes the `validate-foo` targets unhappy.

This commit changes the `validate-foo` targets to use `sed` to cope with both
kinds of headers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,34 +26,38 @@ clean:
 %.png: %.dot
 	dot -Tpng $< -o $@
 
+# A command to remove the '---' YAML document start markers from the snapshots
+# before passing their contents to a validator.
+TRIM=sed -e '1d' -e '2,/---/d'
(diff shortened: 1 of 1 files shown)


kvark (approved):

kvark:
Thank you!

PR #720, opened 2021-04-14
Remove cargo-insta

### LINKED ISSUE #568: Consider removing cargo-insta in favor of just Git
As long as the snapshots are regenerated automatically, there is not much that cargo-insta brings on the table. We already see the snapshot differences in git, so the consent is there.

Things to do instead:
  - an easy way to regenerate the snapshots. Maybe it's just `make snapshots`, or maybe it's a part of `cargo test`. Personally, I find the idea of `cargo test` writing any files to be weird, so the makefile command seems fine. But we have cargo test today, so keeping it this way is not bad either.
  - add a CI check to ensure the snapshots are up to date.
  - remove cargo-insta dependency.

kvark:
Closes #568 

I think this turned out to be OK.

I wanted to have `make snapshots` instead of this being a test, but this would require the library to have extra dependencies for `rspirv`, `ron`, and `diff`... So we'll get this once Rust allows binary-specific dependencies.


grovesNL (approved):
Thanks, looks good to me!

PR #730, opened 2021-04-19
Add wgsl-out into test infrastructure

Gordon-F:


.github/workflows/validation-linux.yml
```diff
@@ -2,9 +2,10 @@ name: validation-linux
 on:
   pull_request:
     paths:
-    - 'tests/out/*.spvasm.snap'
-    - 'tests/out/*.glsl.snap'
-    - 'tests/out/*.dot.snap'
+    - 'tests/out/*.spvasm'
```
Gordon-F, line 5: Looks like test infrastructure been broken for a while after https://github.com/gfx-rs/naga/pull/720

kvark: ahh, thank you for spotting this!

kvark (approved):
Very nice!

PR #731, opened 2021-04-19
[glsl-in] add support for array variables

mockersf:
I tried uncommenting the first line in https://github.com/gfx-rs/naga/blob/9cd6fd9c205a57824644d0baedc6c15997be1e36/tests/parse.rs#L39-L44 and fixed the first failure

This PR add parsing for https://github.com/gfx-rs/naga/blob/9cd6fd9c205a57824644d0baedc6c15997be1e36/tests/cases/glsl_constant_expression.vert#L5

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -155,35 +155,37 @@ impl<'a> Program<'a> {
                 width,
             } => columns as u8 * rows as u8 * width,
             crate::TypeInner::Pointer { .. } => {
-                Err(ErrorKind::NotImplemented("type size of pointer"))?
+                return Err(ErrorKind::NotImplemented("type size of pointer"))
             }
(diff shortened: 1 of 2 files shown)


kvark (approved):
please make clippy happy :)

mockersf:
done!

kvark:
> I tried uncommenting the first line in 

Is it still failing with this PR?

mockersf:
seems https://github.com/gfx-rs/naga/blob/9cd6fd9c205a57824644d0baedc6c15997be1e36/tests/cases/glsl_constant_expression.vert#L8
automatically adds an arg to the function: https://github.com/gfx-rs/naga/blob/9cd6fd9c205a57824644d0baedc6c15997be1e36/src/front/glsl/variables.rs#L19

but not completely and that makes it fail. I'm not yet up to what's happening there, will try to fix it next

Issue #358, opened 2021-01-17
No span information in errors


sotrh:
The frontend parsers currently don't include any span information when an error occurs. This makes it difficult to determine what's going wrong if the shader is particularly complex. I suggest modifying the parsers (at least the text based ones) to keep track of span information while parsing so that errors could use it.

Doing this with pomelo would likely be extremely difficult and could be a reason to remove it as per #327

kvark:
WGSL frontend has some information about the span, but I think we need something more robust and general, to be used in all frontends. Any ideas how this should be implemented?

sotrh:
I'm actually messing around with a prototype of a lexer that includes span information. It's a non trivial process as Rust uses UTF-8 instead of ASCII. The struct I'm working on looks something like this:

```rust
/// Represents the substring of a `&str`
#[derive(Debug, Clone, Copy)]
pub struct Span<'a> {
    data: &'a str,
    pub(crate) first_byte: usize,
    pub(crate) last_byte: usize,
}
```
I've coded up an example of how I would solve the problem generally though applying it to naga would likely be different. https://github.com/sotrh/span-lexer

kvark:
I'd assume that for Naga frontents span is just `Range<usize>`, and the input string is known by the parser.

sotrh:
I've updated the `Span` struct to include the line number and column number to aid debugging, but ya basically.

ElectronicRU:
> I'd assume that for Naga frontents span is just `Range<usize>`, and the input string is known by the parser.

For spv inputs, it's probably useful to interpret OpLine instructions and report that as an error source, if present.
I'd like to help with this, though I'm not terribly sure how to begin. Where should debug info live? Some kind of sparse array in the arena? Types and Constants and Expressions themselves? What do we do when we deserialize from RON?

jimblandy:
These are good questions. There are a lot of ways this could go.

- **integrated**: each type - `Statement`, `Expression`, `Type`, etc. - gets a `span` field.
- **parallel**: add  arrays to `Module` that parallel the `Arena`s, indexed by `Handle<Expression>`, `Handle<Type>`, etc. that hold span information.
- **parallel separate**: parallel arrays, but in a type separate from `Module`.

### integrated

The **integrated** approach is pretty easy to understand. If you've got the data structure in hand, you have its span. Hardly needs documentation.

[...]

PR #734, opened 2021-04-20
Swizzle expression

### LINKED ISSUE #681: Swizzle expression in IR
The reason we don't have `Expression::Swizzle` today is because it's not required for anything. However, without swizzling we end up naively generating code like this:
```cpp
metal::float3(_expr23.x, _expr23.y, _expr23.z)
```
Given that both our input languages and output languages support swizzling (on vectors, at least), it seems more trouble than it's worth (i.e. if we are trying to keep the IR minimal) to avoid them in the IR. Let's have them?

kvark:
Closes #681
As you can see, the produced output is much better now.

@grovesNL there is one controversial thing here - removal of the invalid access test in WGSL. We were generally trying to have a single place where inputs are validated. So WGSL code only has to validate the state that the IR validation doesn't cover. The access expressions are fully covered now, so technically WGSL should leave it to IR for validation.
However, the IR validation errors are nearly not as nice, and this is a general problem we have that needs more thought. We can consider #358 to be this issue.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -379,55 +379,49 @@ impl<'a> ExpressionContext<'a, '_, '_> {
 }
 
 enum Composition {
-    Single(crate::Expression),
-    Multi(crate::VectorSize, Vec<Handle<crate::Expression>>),
+    Single(u32),
+    Multi(crate::VectorSize, [crate::SwizzleComponent; 4]),
 }
 
 impl Composition {
     //TODO: could be `const fn` once MSRV allows
-    fn letter_pos(letter: char) -> u32 {
+    fn letter_component(letter: char) -> Option<crate::SwizzleComponent> {
+        use crate::SwizzleComponent as Sc;
         match letter {
-            'x' | 'r' => 0,
-            'y' | 'g' => 1,
-            'z' | 'b' => 2,
-            'w' | 'a' => 3,
-            _ => !0,
+            'x' | 'r' => Some(Sc::X),
+            'y' | 'g' => Some(Sc::Y),
+            'z' | 'b' => Some(Sc::Z),
+            'w' | 'a' => Some(Sc::W),
+            _ => None,
+        }
+    }
+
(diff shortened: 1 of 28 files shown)


grovesNL (approved):
Looks good to me! Yeah I think we'll be able to restore the validation once we pass span information through to the validator

src/front/wgsl/mod.rs
```diff
@@ -379,55 +379,49 @@ impl<'a> ExpressionContext<'a, '_, '_> {
...
-        name_span: Range<usize>,
-    ) -> Result<crate::Expression, Error> {
+    fn extract_impl(name: &str, name_span: Range<usize>) -> Result<u32, Error> {
         let ch = name
             .chars()
             .next()
             .ok_or_else(|| Error::BadAccessor(name_span.clone()))?;
-        let index = Self::letter_pos(ch);
-        if index >= base_size as u32 {
-            return Err(Error::BadAccessor(name_span));
```
jimblandy, line 410: So is the idea that the expression validator now checks that the accessors make sense for the input vector type, so we don't need to check for that here?

jimblandy (commented):
Looks good - just some suggestions.

src/lib.rs
```diff
@@ -679,6 +679,21 @@ pub enum ImageQuery {
     NumSamples,
 }
 
+/// Component selection for a vector swizzle.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+pub enum SwizzleComponent {
```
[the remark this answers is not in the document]
kvark: Good idea! We do this for `VectorSize`, and we should definitely do it here.

src/front/wgsl/mod.rs
kvark: right, we don't want to implement the same check in multiple places if this can be avoided

kvark:
Thanks @jimblandy ! Everything should be addressed now with #750

PR #735, opened 2021-04-20
Use fmt::Writer instead of io::Writer for glsl and hlsl backend

Gordon-F:
As discussed in https://github.com/gfx-rs/naga/pull/727.

`wgsl-out` will be fixed in another PR.


kvark (approved):
Thank you!

Issue #736, opened 2021-04-20
[Metal] `center_perspective` shouldn't be applied to uniform structs


expenses:
https://github.com/gfx-rs/naga/commit/9cd6fd9c205a57824644d0baedc6c15997be1e36 has broken the following glsl:
```glsl
#version 450

layout(location = 0) out vec4 colour;

layout(set = 0, binding = 0) uniform Glob {
    vec4 z;
} glob;

void main() {
    colour = glob.z;
}
```
->
```metal
#include <metal_stdlib>
#include <simd/simd.h>

struct Glob {
    metal::float4 z;
};

void main1(
    thread metal::float4& colour,
    constant Glob& glob
) {
    colour = glob.z;
    return;
}

struct main2Output {
    metal::float4 member [[color(0)]];
};
fragment main2Output main2(
  constant Glob& glob [[user(fake0), center_perspective]]
) {
    metal::float4 colour = {};
    main1(colour, glob);
    return main2Output { colour };
}
```
->
```
testfrag.metal:20:38: error: 'center_perspective' attribute only applies to non-static data members
  constant Glob& glob [[user(fake0), center_perspective]]
                                     ^
1 error generated.
```

expenses:
I don't know the metal binding model, but it seems like we can't apply any interpolation to these structs. I think the best solution would be to not use an interpolation at all when `fake_missing_bindings` is on so that it becomes `[[user(fake0)]]`.

kvark:
Thank you for filing this! I also discovered this yesterday. The bug crawled in because our CI was not triggering correctly (we broke it), see https://github.com/gfx-rs/naga/pull/730#discussion_r615597893. Good thing @Gordon-F discovered this early!
We'll try to address this ASAP. For the meantime, the MSL will need to be produced by SPIRV-Cross.

expenses:
I have a branch here: https://github.com/gfx-rs/naga/compare/master...expenses:fake-missing-bindings-no-interpolation?expand=1 that fixes this in the way I suggest, but I'm not sure if that's the best way to go.

PR #737, opened 2021-04-20
[Metal] Don't use an interpolation on fake bindings

expenses:
Fixes https://github.com/gfx-rs/naga/issues/736.


kvark (approved):
Thank you! I didn't realize this is only a problem with fake bindings. In this case, it's not blocking any releases or wgpu work.

PR #738, opened 2021-04-20
[Metal] Impl `Expression::Splat`

expenses:
As an example of how this is used, this now means that the following can compile:
```glsl
#version 450

layout(location = 0) flat in uvec4 xyz;

layout(location = 0) out uvec4 colour;

void main() {
    colour = xyz >> 2;
}
```
->
```metal
#include <metal_stdlib>
#include <simd/simd.h>


void main1(
    thread metal::uint4& colour,
    thread metal::uint4 const& xyz
) {
    colour = xyz >> as_type<uint4>(int4(2));
    return;
}

struct main2Input {
    metal::uint4 xyz1 [[user(loc0), flat]];
};
struct main2Output {
    metal::uint4 member [[color(0)]];
};
fragment main2Output main2(
  main2Input varyings [[stage_in]]
) {
    metal::uint4 colour = {};
    metal::uint4 xyz = {};
    const auto xyz1 = varyings.xyz1;
    xyz = xyz1;
    main1(colour, xyz);
    return main2Output { colour };
}
```
(The `int4` is `glslc` being bad)

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -608,8 +608,17 @@ impl<W: Write> Writer<W> {
                 };
                 write!(self.out, "{}", coco)?;
             }
-            crate::Expression::Splat { size: _, value } => {
+            crate::Expression::Splat { size, value } => {
+                let scalar_kind = match *context.resolve_type(value) {
+                    crate::TypeInner::Scalar { kind, .. } => kind,
+                    _ => unreachable!()
+                };
+                let scalar = scalar_kind_string(scalar_kind);
+                let size = vector_size_string(size);
+
+                write!(self.out, "{}{}(", scalar, size)?;
                 self.put_expression(value, context, is_scoped)?;
+                write!(self.out, ")")?;
             }
             crate::Expression::Compose { ty, ref components } => {
                 let inner = &context.module.types[ty].inner;
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -611,13 +611,13 @@ impl<W: Write> Writer<W> {
             crate::Expression::Splat { size, value } => {
                 let scalar_kind = match *context.resolve_type(value) {
                     crate::TypeInner::Scalar { kind, .. } => kind,
-                    _ => unreachable!()
+                    _ => return Err(Error::Validation)
                 };
                 let scalar = scalar_kind_string(scalar_kind);
                 let size = vector_size_string(size);
 
                 write!(self.out, "{}{}(", scalar, size)?;
-                self.put_expression(value, context, is_scoped)?;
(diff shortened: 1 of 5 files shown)


src/back/msl/writer.rs
```diff
@@ -608,8 +608,17 @@ impl<W: Write> Writer<W> {
                 };
                 write!(self.out, "{}", coco)?;
             }
-            crate::Expression::Splat { size: _, value } => {
+            crate::Expression::Splat { size, value } => {
+                let scalar_kind = match *context.resolve_type(value) {
+                    crate::TypeInner::Scalar { kind, .. } => kind,
+                    _ => unreachable!()
```
expenses, line 614: I'm not sure if we want to handle the case where `value` _isn't_ a scalar.

kvark:
What error are you getting if this PR isn't implemented?

expenses:
> What error are you getting if this PR isn't implemented?

```
testfrag.metal:9:21: error: as_type cast from scalar 'int' to vector 'uint4' (vector of 4 'unsigned int' values) of different size
    colour = xyz >> as_type<uint4>(2);
                    ^~~~~~~~~~~~~~~~~
1 error generated.
```

Changing `as_type` to `static_cast` also works, but this is more in line with the IR.

kvark:
That seems like an unrelated issue. It shouldn't be generating `as_type<uint4>(2)` at all, ever. It should instead generate `as_type<uint4>(2u)`.

expenses:
`as_type<uint4>(2u)` still fails. It's `as_type<uint4>(2)` instead because of the way `glslc` outputs SPIR-V.  The reason it's not doing `static_cast` is because `Expression::As` has `convert` set to false. We could change the IR so that it's set to true when doing a splat like here.

kvark:
ok, so splat isn't happening automatically for the shifts, I see.

src/back/msl/writer.rs
kvark: let's return `Error::Validation` instead of panicking here

src/back/msl/writer.rs
```diff
@@ -608,8 +608,17 @@ impl<W: Write> Writer<W> {
...
+                };
+                let scalar = scalar_kind_string(scalar_kind);
+                let size = vector_size_string(size);
+
+                write!(self.out, "{}{}(", scalar, size)?;
                 self.put_expression(value, context, is_scoped)?;
```
kvark, line 620: should now have `true` as the last parameter, no need to inherit `is_scoped`

kvark (changes requested):
This is almost ready!

kvark (approved):
Thank you!

Issue #710, opened 2021-04-11
[Metal] A `packed_float3` needs to be casted to a `float3` before being multiplied with a `float3x3` matrix


expenses:
Yeah, so this is a wierd one.
```glsl
#version 450

struct Ray {
    vec3 dir;
    vec3 origin;
};

Ray ray = {vec3(0.0, 1.0, 0.0), vec3(0.0)};
mat3 rot = mat3(1.0);

layout(location = 0) out vec4 colour;

void main() {
    colour = vec4(rot * ray.dir, 1.0);
}
```
becomes (with https://github.com/gfx-rs/naga/pull/707):
```metal
struct Ray {
    packed_float3 dir;
    packed_float3 origin;
};
constant metal::float3 const_type1_ = {0.0, 1.0, 0.0};
constant metal::float3 const_type1_1 = {0.0, 0.0, 0.0};
constant Ray const_Ray = {const_type1_, const_type1_1};
constant metal::float3 const_type1_2 = {1.0, 0.0, 0.0};
constant metal::float3 const_type1_3 = {0.0, 0.0, 1.0};
constant metal::float3x3 const_type3_ = {const_type1_2, const_type1_, const_type1_3};

void main1(
    thread Ray& ray,
    thread metal::float3x3& rot,
    thread metal::float4& colour
) {
    ray = const_Ray;
    rot = const_type3_;
    metal::float3 _e19 = rot * ray.dir;
    colour = metal::float4(_e19.x, _e19.y, _e19.z, 1.0);
    return;
}
```
which fails with:
```
rot.metal:22:30: error: cannot convert between vector and non-scalar values ('metal::float3x3' (aka 'matrix<float, 3, 3>') and 'packed_float3' (packed vector of 3 'float' values))
    metal::float3 _e19 = rot * ray.dir;
                         ~~~ ^ ~~~~~~~
...
```
wrapping `ray.dir` like `metal::float3(ray.dir)` works, but it's pretty odd that you have to do this cast in the first place.

expenses:
Also happens with a `float3x4` and a `float3x2` but not with a `float3` or another `packed_float3`.

kvark:
uh, thank you for finding this!

PR #740, opened 2021-04-20
[Metal] Wrap a `packed_float3` with `float3` when loading from struct

expenses:
This closes https://github.com/gfx-rs/naga/issues/710. I don't think there's a good expression to check if an expression is packed, so we just wrap all vectors that could be. We need to handle both `matrix * vec` and `vec * matrix`. Here's an example with both of these:
```glsl
#version 450

struct Ray {
    vec3 dir;
    vec3 origin;
};

Ray ray = {vec3(0.0, 1.0, 0.0), vec3(0.0)};
mat3 rot = mat3(1.0);

void main() {
    vec3 a = ray.dir * rot;
    vec3 b = rot * ray.dir;
    colour = vec4(a.xy, b.yz);
}
```
Without these changes we get:
```
testfrag.metal:24:17: error: cannot convert between vector and non-scalar values ('packed_float3' (packed vector of 3 'float' values) and 'metal::float3x3'
      (aka 'matrix<float, 3, 3>'))
    a = ray.dir * rot;
        ~~~~~~~ ^ ~~~
...
testfrag.metal:25:13: error: cannot convert between vector and non-scalar values ('metal::float3x3' (aka 'matrix<float, 3, 3>') and 'packed_float3'
      (packed vector of 3 'float' values))
    b = rot * ray.dir;
        ~~~ ^ ~~~~~~~
...
```


expenses:
Re: stack size, dropping `left_type` and `right_type` does seem to reduce the stack, despite them just being references. Clippy doesn't like dropping references though.

src/back/msl/writer.rs
```diff
@@ -815,8 +815,21 @@ impl<W: Write> Writer<W> {
...
                 };
-                let kind = context
-                    .resolve_type(left)
+
+                let left_type = context.resolve_type(left);
+                let right_type = context.resolve_type(right);
+
+                let is_matrix_mul_tri_vec =
+                    op == crate::BinaryOperator::Multiply &&
+                    matches!(left_type, &crate::TypeInner::Matrix { ..}) &&
```
kvark, line 824: can we please not use `matches!`? a regular match would take the same amount of code, and involve less sugar

src/back/msl/writer.rs
```diff
@@ -829,9 +842,23 @@ impl<W: Write> Writer<W> {
                     if !is_scoped {
```
kvark, line 842: would it make sense to do `else if op == crate::BinaryOperator::Mul` and put the logic there instead?

src/back/msl/writer.rs
```diff
@@ -829,9 +842,23 @@ impl<W: Write> Writer<W> {
...
                         write!(self.out, "(")?;
                     }
-                    self.put_expression(left, context, false)?;
+                    // Because packed vectors such as `packed_float3` cannot be directly multipied by
+                    // matrices, we them in the non-packed variant first.
+                    if is_tri_vec_mul_matrix {
```
kvark, line 847: could use a helper function `self.put_mul_expression` or something, to ease on the stack size

kvark (commented):
Thank you for suggesting a fix!
I'm worried that there are other places where `packed_float3` would not be expected.

I imagine there could be a more general fix that, which would activate in a place where we access `packed_float3` specifically, as opposed to any operations it takes part in.

A primitive version of this could be done in `crate::Expression::Load { pointer }` arm:
  1. we'd check the type of pointer, and proceed if it is `AccessIndex` only
  2. we'd check the member type to be `vec3`
  3. we'd check the next member offset to see if this one is packet
  4. we'd wrap the result in `metal::float3` if we reached through all those conditions

This approach still wouldn't handle the case where a member pointer is accessed by pointer, but we can follow-up with this - it's a very niche case today.

expenses:
> A primitive version of this could be done in `crate::Expression::Load { pointer }` arm:
> 
>     1. we'd check the type of pointer, and proceed if it is `AccessIndex` only
> 
>     2. we'd check the member type to be `vec3`
> 
>     3. we'd check the next member offset to see if this one is packet
> 
>     4. we'd wrap the result in `metal::float3` if we reached through all those conditions
> 
> 
> This approach still wouldn't handle the case where a member pointer is accessed by pointer, but we can follow-up with this - it's a very niche case today.

I'll try this before making some changes to this PR.

src/back/msl/writer.rs
```diff
@@ -689,10 +689,58 @@ impl<W: Write> Writer<W> {
                 write!(self.out, "{}", name)?;
             }
             crate::Expression::Load { pointer } => {
-                // We don't do any dereferencing with `*` here as pointer arguments to functions
-                // are done by `&` references and not `*` pointers. These do not need to be
-                // dereferenced.
-                self.put_expression(pointer, context, is_scoped)?;
+                // Because packed vectors such as `packed_float3` cannot be directly multipied by
```
kvark, line 692: I suppose this would equally apply to `packed_int3` and friends? So the `wrap_float_scalar_kind` needs to be renamed, and the comments adjusted

src/back/msl/writer.rs
```diff
@@ -689,10 +689,58 @@ impl<W: Write> Writer<W> {
...
+                // matrices, we wrap them with `float3` on load.
+                let wrap_float_scalar_kind = match context.function.expressions[pointer] {
+                    crate::Expression::AccessIndex { base, index } => {
+                        let ty = match context.resolve_type(base) {
+                            &crate::TypeInner::Pointer { base, .. } => {
+                                &context.module.types[base].inner
+                            }
+                            ty => ty,
```
kvark, line 700: I assume this path is not really expected here, but we just don't want to bother with any early exits. Maybe worth leaving a small comment?

src/back/msl/writer.rs
```diff
@@ -689,10 +689,58 @@ impl<W: Write> Writer<W> {
...
+                        match *ty {
+                            crate::TypeInner::Struct { ref members, .. } => {
+                                let member = &members[index as usize];
+                                let ty = &context.module.types[member.ty].inner;
+
+                                let last_offset =
+                                    member.offset + ty.span(&context.module.constants);
+                                let is_tight = match members.get(index as usize + 1) {
+                                    Some(next) => next.offset == last_offset,
+                                    None => false,
```
kvark, line 711: we should do `span == last_offset` in this case, where `span` comes from `crate::TypeInner::Struct { ref members, span, .. }`.
We could also refactor it into:
```rust
let next_offset = match members.get(index as usize + 1) {
   Some(next) => next.offset,
   None => span,
};
```
And then use `if next_offset == last_offset`

src/back/msl/writer.rs
```diff
@@ -689,10 +689,58 @@ impl<W: Write> Writer<W> {
...
+                        let ty = match context.resolve_type(base) {
+                            &crate::TypeInner::Pointer { base, .. } => {
+                                &context.module.types[base].inner
+                            }
+                            ty => ty,
+                        };
+                        match *ty {
+                            crate::TypeInner::Struct { ref members, .. } => {
+                                let member = &members[index as usize];
+                                let ty = &context.module.types[member.ty].inner;
```
kvark, line 705: let's not shadow the same-typed variable. This one could be `member_ty` or something

src/back/msl/writer.rs
```diff
@@ -689,10 +689,58 @@ impl<W: Write> Writer<W> {
...
+
+                                let last_offset =
+                                    member.offset + ty.span(&context.module.constants);
+                                let is_tight = match members.get(index as usize + 1) {
+                                    Some(next) => next.offset == last_offset,
+                                    None => false,
+                                };
+
+                                match *ty {
```
kvark, line 714: ideally, we'd have the detection code for this case moved out and shared between here and the place that makes a decision to use `packed_`

kvark (commented):
Great stuff, thank you for prototyping this approach!
I think it's landable, but a few things would be good to get right from the start.

expenses:
The skybox snapshot is failing with the following:
```diff
 --- a/tests/out/skybox.msl
+++ b/tests/out/skybox.msl
@@ -3,7 +3,7 @@
 
 struct VertexOutput {
     metal::float4 position;
-    metal::float3 uv;
+    packed_float3 uv;
 };
```
I'm not sure, but I think this change might be incorrect as I thought that shader input/outputs were tightly packed and didn't have the same 16 byte boundary thing that uniforms do.

src/back/msl/writer.rs
```diff
@@ -311,6 +311,32 @@ fn separate(need_separator: bool) -> &'static str {
     }
 }
 
+fn should_pack_struct_member(
+    members: &[crate::StructMember],
+    span: u32,
+    index: usize,
+    module: &crate::Module,
+) -> bool {
```
kvark, line 319: maybe it can return `Option<(crate::VectorSize, crate::ScalarKind)>` for convenience?

kvark (commented):
Thank you, looking very good now!

kvark:
> I'm not sure, but I think this change might be incorrect as I thought that shader input/outputs were tightly packed (in vulkan/opengl atleast) and didn't have the same 16 byte boundary thing that uniforms do.

The inputs/outputs layouts don't matter, they aren't observable. So let's just update the test expectation and ship it? :)

kvark (approved):

PR #742, opened 2021-04-20
Improve wgsl-out

Gordon-F:
Improve `wgsl-out` backend. A lot of code has been copy-pasted from `glsl-out` and `msl-out`.

Now it can produce simple shader:

In:
```wgsl
// vertex
let c_scale: f32 = 1.2;

struct VertexOutput {
  [[location(0)]] uv : vec2<f32>;
  [[builtin(position)]] position : vec4<f32>;
};

[[stage(vertex)]]
fn main([[location(0)]] pos : vec2<f32>, [[location(1)]] uv : vec2<f32>) -> VertexOutput {
  return VertexOutput(uv, vec4<f32>(c_scale * pos, 0.0, 1.0));
}
```

Out:
```wgsl
let c_scale: f32 = 1.2;

struct VertexOutput {
    [[location(0)]] uv: vec2<f32>;
};

[[stage(vertex)]]
fn main([[location(0)]] pos: vec2<f32>, [[location(1)]] uv: vec2<f32>) -> VertexOutput {
    return VertexOutput(uv1, vec4<f32>(c_scale * pos, 0.0, 1.0));

}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,5 +61,5 @@ validate-dot: $(SNAPSHOTS_OUT)/*.dot
 validate-wgsl: $(SNAPSHOTS_OUT)/*.wgsl
 	@set -e && for file in $^ ; do \
 		echo "Validating" $${file#"$(SNAPSHOTS_OUT)/"};	\
-		cargo run --bin convert --features wgsl-in $${file} >/dev/null; \
+		cargo run --bin convert --features wgsl-in $${file}; \
 	done
--- a/bin/convert.rs
+++ b/bin/convert.rs
@@ -188,8 +188,10 @@ fn main() {
         None => {
             if info.is_some() {
                 println!("Validation successful");
+                return;
+            } else {
+                std::process::exit(!0);
             }
-            return;
         }
     };
     if output_path == "-" {
--- a/bin/convert.rs
+++ b/bin/convert.rs
@@ -269,7 +269,7 @@ fn main() {
         "wgsl" => {
             use naga::back::wgsl;
 
-            let wgsl = wgsl::write_string(&module).unwrap_pretty();
+            let wgsl = wgsl::write_string(&module, info.as_ref().unwrap()).unwrap_pretty();
             fs::write(output_path, wgsl).unwrap();
         }
         other => {
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -1,24 +1,58 @@
+// TODO: temp
+#![allow(dead_code)]
 use super::Error;
-use crate::FastHashMap;
 use crate::{
-    back::wgsl::keywords::RESERVED, Function, Module, ShaderStage, StructLevel, TypeInner,
+    back::wgsl::keywords::RESERVED,
+    proc::{EntryPointIndex, TypeResolution},
+    valid::{FunctionInfo, ModuleInfo},
+    Arena, BinaryOperator, Binding, Constant, Expression, Function, GlobalVariable, Handle,
+    ImageClass, ImageDimension, Module, ScalarKind, ShaderStage, Statement, StorageFormat,
+    StructLevel, Type, TypeInner,
 };
 use crate::{
     proc::{NameKey, Namer},
     StructMember,
 };
-use std::io::Write;
+use crate::{FastHashMap, VectorSize};
+use std::fmt::Write;
 
-const _INDENT: &str = "    ";
+const INDENT: &str = "    ";
 
 /// Shorthand result used internally by the backend
 type BackendResult = Result<(), Error>;
 
-enum Decoration {
-    VertexStage,
-    FragmentStage,
-    ComputeStage { workgroup_size: [u32; 3] },
+/// WGSL attribute
+/// https://gpuweb.github.io/gpuweb/wgsl/#attributes
+enum Attribute {
+    Binding(u32),
     Block,
+    BuiltIn(crate::BuiltIn),
+    Group(u32),
+    Location(u32),
+    Stage(ShaderStage),
+    WorkGroupSize([u32; 3]),
+}
+
+/// Stores the current function type (either a regular function or an entry point)
+///
+/// Also stores data needed to identify it (handle for a regular function or index for an entry point)
+// TODO: copy-paste from glsl-out
+enum FunctionType {
+    /// A regular function and it's handle
+    Function(Handle<Function>),
+    /// A entry point and it's index
+    EntryPoint(EntryPointIndex),
+}
+
+/// Helper structure that stores data needed when writing the function
+// TODO: copy-paste from glsl-out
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -61,5 +61,5 @@ validate-dot: $(SNAPSHOTS_OUT)/*.dot
 validate-wgsl: $(SNAPSHOTS_OUT)/*.wgsl
 	@set -e && for file in $^ ; do \
 		echo "Validating" $${file#"$(SNAPSHOTS_OUT)/"};	\
-		cargo run --bin convert --features wgsl-in $${file} >/dev/null; \
+		cargo run --bin convert --features wgsl-in $${file} -- --validation-only; \
```
kvark, line 64: I think we need to approach this differently. It should always be validating, but we don't need to print the contents out. Instead, we can have a special target name, say "-", which means the module will be printed without any files written. That would allow us to avoid the flag.

src/back/wgsl/writer.rs
```diff
@@ -1,24 +1,54 @@
...
-enum Decoration {
+/// WGSL attribute
+/// https://gpuweb.github.io/gpuweb/wgsl/#attributes
+enum Attribute {
+    Block,
+    Location(u32),
+    BuiltIn(crate::BuiltIn),
     VertexStage,
     FragmentStage,
     ComputeStage { workgroup_size: [u32; 3] },
```
kvark, line 31: these don't need to be within the same attribute
instead, let's have:
```
Stage(crate::ShaderStage),
WorkGroupSize([u32; 3]),
```
WGSL allows you to specify them separately or together, as you wish. The backend should probably just slap them separately all the time.

src/back/wgsl/writer.rs
```diff
@@ -122,23 +213,395 @@ impl<W: Write> Writer<W> {
...
+        &mut self,
+        module: &Module,
+        stmt: &Statement,
+        func_ctx: &FunctionCtx<'_>,
+        indent: usize,
+    ) -> BackendResult {
+        match *stmt {
+            Statement::Emit(ref range) => {
+                for handle in range.clone() {
+                    let min_ref_count = func_ctx.expressions[handle].bake_ref_count();
```
kvark, line 301: if you notice that there are re-usable parts possible between MSL/GLSL/WGSL, let's make a note somewhere to try moving them out into a common area

src/back/wgsl/writer.rs
```diff
@@ -122,23 +213,395 @@ impl<W: Write> Writer<W> {
...
+        BuiltIn::FrontFacing => "front_facing",
+        BuiltIn::FragDepth => "frag_depth",
+        BuiltIn::LocalInvocationId => "local_invocation_id",
+        BuiltIn::LocalInvocationIndex => "local_invocation_index",
+        BuiltIn::GlobalInvocationId => "global_invocation_id",
+        BuiltIn::WorkGroupId => "workgroup_id",
+        BuiltIn::WorkGroupSize => "workgroup_size",
+        BuiltIn::SampleIndex => "sample_index",
+        BuiltIn::SampleMask => "sample_mask",
+        _ => todo!("built_in_str {:?}", built_in),
```
kvark, line 579: not all builtins are going to be mappable to WGSL, so we should make this return an `Error`

kvark (commented):
Amazing work!

Makefile
Gordon-F: We need validation info for `wgsl-out` shaders, but without printing module to stdout. Maybe we should find a better name for this flag.

src/back/wgsl/writer.rs
Gordon-F: I am planning to use a list of attributes as described in [spec](https://gpuweb.github.io/gpuweb/wgsl/#attributes), like you suggest.

src/back/wgsl/writer.rs
Gordon-F: I already face it when I tried to convert `spv` to `wgsl`. In `spv` we have `gl_PointSize`, for example. And I can't find anything about it in WGSL Spec.  What should we do in this case?

Gordon-F:
Also, I don't know how to defeat `namer`. For example:

In:
```wgsl
return VertexOutput(uv, vec4<f32>(c_scale * pos, 0.0, 1.0));
```

Out
```
return VertexOutput(uv1, vec4<f32>(c_scale * pos, 0.0, 1.0));
---------------------^
Another variable!
```

`NameKey::FunctionArgument(handle, pos)` returns unique name, because we already have `uv` inside `VertexOutput` struct.

Makefile
kvark: I addressed this in https://github.com/gfx-rs/naga/pull/744

src/back/wgsl/writer.rs
kvark: Like I said, we should return an error variant in this case

src/back/wgsl/writer.rs
```diff
@@ -79,25 +123,75 @@ impl<W: Write> Writer<W> {
...
+        // Write function arguments
+        for (index, arg) in func.arguments.iter().enumerate() {
+            // Write argument attribute if a binding is present
+            if let Some(ref binding) = arg.binding {
+                self.write_attribute(map_binding_to_attribute(binding))?;
+                write!(self.out, " ")?;
+            }
+            // TODO: unnamed args?
+            // Write argument name
+            write!(self.out, "{}: ", arg.name.as_ref().unwrap())?;
```
kvark, line 144: this should use `self.names[key]`, where key depends on whether this is a function or an entry point (can be passed as an argument to `write_function`)

kvark:
@Gordon-F what's the status?

Gordon-F:
@kvark Should be ready to merge. I think all suggestions are addressed. If everything is ok, we can merge it or PR will become huge 😄

bin/convert.rs
```diff
@@ -188,8 +188,10 @@ fn main() {
         None => {
             if info.is_some() {
                 println!("Validation successful");
+                return;
+            } else {
+                std::process::exit(!0);
```
kvark, line 193: nice!

src/back/wgsl/writer.rs
```diff
@@ -122,23 +247,528 @@ impl<W: Write> Writer<W> {
...
+                        format!("<{}>", storage_format_str(storage_format)),
+                    ),
+                };
+                let ty_str = format!(
+                    "texture_{}{}{}{}{}",
+                    class_str, multisampled_str, dim_str, arrayed_str, scalar_str
+                );
+                write!(self.out, "{}", ty_str)?;
+            }
+            _ => todo!("write_value_type {:?}", inner),
```
kvark, line 344: we should return something like `Error::Unimplemented` for all these cases

kvark (approved):
Amazing work, thank you!

PR #749, opened 2021-04-22
[spv-out] image queries

### LINKED ISSUE #646: Implement image queries in SPIR-V backend
`Expression::ImageQuery { .. }` support is missing

kvark:
Closes #646
The focus of this PR was to implement image queries on SPIR-V backend. It ended up being quite nuanced, because Vulkan packs the array index together with the dimensions, and also our Cube semantics is slightly different, but I'm happy with the result code.
This change renames "image-copy" test into just "image" and adds an entry point there that checks all kinds of queries on sampled textures (only sampled so far). It uncovered a few bugs we had in the MSL backend as well, which are fixed here.


src/back/spv/writer.rs
```diff
@@ -1894,12 +1899,16 @@ impl Writer {
...
-                    let index_id = self.get_index_constant(0, &ir_module.types)?;
-                    let sub_instruction =
-                        Instruction::vector_extract_dynamic(result_type_id, sub_id, id, index_id);
-                    block.body.push(sub_instruction);
+                    block.body.push(Instruction::composite_extract(
+                        result_type_id,
+                        sub_id,
+                        id,
+                        &[0],
+                    ));
```
Napokue, line 1911: Not really a pressing issue, but we aren't reusing this ID, and are using exactly the same code underneath here.

src/back/spv/writer.rs
```diff
@@ -2098,9 +2109,145 @@ impl Writer {
...
+                            Ic::Storage(_) => (spirv::Op::ImageQuerySize, None),
+                            _ => {
+                                let level_id = match level {
+                                    Some(expr) => self.cached[expr],
+                                    None => self.get_index_constant(0, &ir_module.types)?,
+                                };
+                                (spirv::Op::ImageQuerySizeLod, Some(level_id))
+                            }
+                        };
+                        let id_fat = self.id_gen.next();
```
Napokue, line 2164: Maybe I am missing something, but what does `fat` mean in this context?

Napokue (approved):
LGTM, thanks! Only got 2 small comments that are not really pressing.

src/back/spv/writer.rs
kvark: Please clarify. Which ID we can re-use? What is the code underneath?

src/back/spv/writer.rs
kvark: renaming now and adding a comment

src/back/spv/writer.rs
Napokue: Sorry for not clarifying. I cannot send the link with linenumber now (Github mobile app doesn't support it).

It is at line 2074. That is exactly the same code. Even with the old code (the vector extract dynamic) it was also duplicated.

I was suggesting that maybe it could be combined, but I am not sure if that is either possible or feasible to do in this situation.

src/back/spv/writer.rs
Napokue: Thank you!

src/back/spv/writer.rs
kvark: By "combined" do you mean moving the call to `composite_extract` to some place that is shared, or actually associating the same ID with the result and using this ID in multiple places?

Napokue: What I mean was both things that you are pointing out, as I believed that it was the same thing and thus can be shared, both by code and by id. But I think it is not shared, is it?

kvark: There is a few `composite_extract` things: one in `AccessIndex`, one in `write_texture_coordinates`, one in `ImageLoad`, one in `ImageSample`, and one in the `ImageQuery`. I don't think there is a good way to share any of them.

Napokue: Sure, that is also fine by me, I was just wondering. Thanks for this rather lengthy discussion.

src/back/spv/writer.rs
```diff
@@ -2098,9 +2109,147 @@ impl Writer {
...
+                            Id::D2 | Id::Cube => 2,
+                            Id::D3 => 3,
+                        };
+                        let extended_size_type_id = {
+                            let array_coords = if arrayed { 1 } else { 0 };
+                            let vector_size = match dim_coords + array_coords {
+                                2 => Some(crate::VectorSize::Bi),
+                                3 => Some(crate::VectorSize::Tri),
+                                4 => Some(crate::VectorSize::Quad),
+                                _ => None,
```
jimblandy, line 2141: This should be `unreachable!`, it seems to me, and the `Some` should be in the `LocalType::Value` constructor.

jimblandy (approved):
Looks good to me, just two suggestions.

src/back/spv/writer.rs
kvark: It's valid to be `None` here, which happens if you query a 1D image size (non-arrayed)

src/back/msl/writer.rs
```diff
@@ -510,9 +515,13 @@ impl<W: Write> Writer<W> {
...
             crate::ImageDimension::Cube => {
-                write!(self.out, "int(")?;
+                write!(self.out, "int3(")?;
                 self.put_image_query(image, "width", level, context)?;
-                write!(self.out, ").xxx")?;
+                write!(self.out, ", ")?;
+                self.put_image_query(image, "width", level, context)?;
+                write!(self.out, ", ")?;
+                self.put_image_query(image, "width", level, context)?;
+                write!(self.out, ")")?;
```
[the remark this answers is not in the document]
kvark: late night fixes... should be better now

PR #756, opened 2021-04-23
[spv-in] fix integer constants parsing

### LINKED ISSUE #755: Signed integers come as unsigned from SPIR-V
>     program_source:46:39: error: constant expression evaluates to 4294967295 which cannot be narrowed to type 'int'
    constant metal::int2 const_type11_ = {4294967295, 0};
cc @expenses

kvark:
Fixes  #755


kvark:
This isn't too pretty, but it works.

Issue #760, opened 2021-04-23
[Metal] Swizzles don't scope their input


expenses:
```glsl
#version 450

layout(location = 0) out vec4 fragColor;

void main()
{
    vec3 rd = vec3(1.0);
    float t = 1.0;

    vec2 param = (rd*t).xy;

    fragColor = vec4(param, param);
}
```
->
```
void main1(
    thread metal::float4& fragColor
) {
    metal::float3 rd;
    float t;
    metal::float2 param;
    rd = const_type1_;
    t = 1.0;
    param = rd * t.xy;
    metal::float2 _e14 = param;
    metal::float2 _e15 = param;
    fragColor = metal::float4(_e14.x, _e14.y, _e15.x, _e15.y);
    return;
}
```
->
```
testfrag.metal:14:19: error: member reference base type 'float' is not a structure or union
    param = rd * t.xy;
                 ~^~~
1 error generated.
```

PR #761, opened 2021-04-23
[Metal] Set `is_scoped` to false when evaluating a swizzle subexpression

expenses:
Fixes https://github.com/gfx-rs/naga/issues/760.


kvark (approved):
Nice!

PR #762, opened 2021-04-23
Rename the binary to just naga

kvark:
Otherwise, when users install `naga` they get a strange "convert" binary in their path.


grovesNL (approved):

memoryruins:
I ran into this earlier while installing and testing out naga cli; this is a nice change, thanks!

PR #763, opened 2021-04-23
[glsl-in] small usability fixes

JCapucho:
- Export some type to allow for error handling for the user
- Load local variables
- Fix type constructors with a single argument
- Add `textureLod` function
- Add support for the `in` keyword in function parameters


kvark (approved):
I thought this frontend is non-functional right now. Are you running it on anything? It would be good to start re-enabling some of the tests.

Issue #741, opened 2021-04-20
[Metal] functions cannot return arrays


expenses:
Functions in metal cannot return raw arrays, see:
```glsl
#version 450

vec3[8] hello_world() {
    vec3[8] xyz;
    return xyz;
}

layout(location = 0) out vec4 colour;

void main() {
    vec3[8] xyz = hello_world();
    colour = vec4(xyz[7], 1.0);
}
```
->
```metal
...
typedef metal::float3 type3[8u];

type3 hello_world(
) {
    type3 xyz;
    return xyz;
}

void main1(
    thread metal::float4& colour
) {
    type3 xyz1;
    type3 _e9 = hello_world();
    for(int _i=0; _i<8; ++_i) xyz1[_i] = _e9[_i];
    metal::float3 _e11 = xyz1[7];
    colour = metal::float4(_e11.x, _e11.y, _e11.z, 1.0);
    return;
}
...
```
->
```
testfrag.metal:6:18: error: function cannot return array type 'type3' (aka 'metal::float3 [8]')
type3 hello_world(
                 ^
testfrag.metal:9:12: error: cannot initialize return object of type 'int' with an lvalue of type 'type3' (aka 'metal::float3 [8]')
    return xyz;
           ^~~
```

They can return arrays in structs though, so I think the correct transformation would be something like this (which compiles):
```metal
...
struct type3 {
    metal::float3 inner[8u];
};

type3 hello_world(
) {
    type3 xyz;
    return xyz;
}

void main1(
    thread metal::float4& colour
) {
    type3 xyz1;
    type3 _e9 = hello_world();
    for(int _i=0; _i<8; ++_i) xyz1.inner[_i] = _e9.inner[_i];
    metal::float3 _e11 = xyz1.inner[7];
    colour = metal::float4(_e11.x, _e11.y, _e11.z, 1.0);
    return;
}
...
```

PR #764, opened 2021-04-23
[msl-out] wrap arrays in structs so that they can be returned by functions

expenses:
This fixes https://github.com/gfx-rs/naga/issues/741. Unfortunately it currently breaks the `quad-vert.msl` snapshot with this error:
```
Validating quad-vert.msl
<stdin>:40:30: error: type 'type6' is not valid for attribute 'clip_distance'
    type6 gl_ClipDistance1 [[clip_distance]];
                             ^~~~~~~~~~~~~
<stdin>:42:8: error: invalid return type 'main2Output' for vertex function
vertex main2Output main2(
       ^
<stdin>:54:86: error: type 'type6' does not provide a subscript operator
    const auto _tmp = type10 {v_uv, _.gl_Position, _.gl_PointSize, {_.gl_ClipDistance[0]}};
                                                                    ~~~~~~~~~~~~~~~~~^~
3 errors generated.
```
This right solution to fix this test would be for `gl_ClipDistance` to be treated as a `float` instead of a `float[1]` in `msl-out`.

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -548,39 +548,6 @@ impl<W: Write> Writer<W> {
         Ok(())
     }
 
-    fn put_initialization_component(
-        &mut self,
-        component: Handle<crate::Expression>,
-        context: &ExpressionContext,
-    ) -> Result<(), Error> {
-        // we can't initialize the array members just like other members,
-        // we have to unwrap them one level deeper...
-        let component_res = &context.info[component].ty;
-        if let crate::TypeInner::Array {
-            size: crate::ArraySize::Constant(const_handle),
-            ..
-        } = *component_res.inner_with(&context.module.types)
-        {
-            //HACK: we are forcefully duplicating the expression here,
-            // it would be nice to find a more C++ idiomatic solution for initializing array members
-            let size = context.module.constants[const_handle]
-                .to_array_length()
-                .unwrap();
-            write!(self.out, "{{")?;
-            for j in 0..size {
-                if j != 0 {
-                    write!(self.out, ",")?;
-                }
-                self.put_expression(component, context, false)?;
-                write!(self.out, ".{}[{}]", WRAPPED_ARRAY_FIELD, j)?;
-            }
-            write!(self.out, "}}")?;
-        } else {
-            self.put_expression(component, context, true)?;
-        }
-        Ok(())
-    }
-
     fn put_expression(
         &mut self,
         expr_handle: Handle<crate::Expression>,
@@ -724,7 +691,7 @@ impl<W: Write> Writer<W> {
                             if i != 0 {
                                 write!(self.out, ", ")?;
                             }
-                            self.put_initialization_component(component, context)?;
+                            self.put_expression(component, context, true)?;
                         }
                         write!(self.out, "}}")?;
                     }
@@ -1173,7 +1140,9 @@ impl<W: Write> Writer<W> {
                             let comma = if is_first { "" } else { "," };
                             is_first = false;
                             let name = &self.names[&NameKey::StructMember(result_ty, index as u32)];
-                            // logic similar to `put_initialization_component`
+                            // HACK: we are forcefully deduplicating the expression here
+                            // to convert from a wrapped struct to a raw array, e.g.
+                            // `float gl_ClipDistance1 [[clip_distance]] [1];`.
                             if let crate::TypeInner::Array {
                                 size: crate::ArraySize::Constant(const_handle),
                                 ..
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -17,6 +17,7 @@ use std::{
 const NAMESPACE: &str = "metal";
 const INDENT: &str = "    ";
 const BAKE_PREFIX: &str = "_e";
+const WRAPPED_ARRAY_FIELD: &str = "inner";
 
 #[derive(Clone)]
 struct Level(usize);
@@ -594,8 +595,31 @@ impl<W: Write> Writer<W> {
         log::trace!("expression {:?} = {:?}", expr_handle, expression);
         match *expression {
             crate::Expression::Access { base, index } => {
+                let accessing_wrapped_array = {
+                    let ty = match &context.info[base].ty {
+                        TypeResolution::Value(value) => value,
+                        TypeResolution::Handle(handle) => &context.module.types[*handle].inner,
+                    };
+
+                    match ty {
+                        crate::TypeInner::Array { .. } => true,
+                        crate::TypeInner::Pointer {
+                            base: pointer_base,
+                            class,
+                        } => match (&context.module.types[*pointer_base].inner, class) {
+                            (crate::TypeInner::Array { .. }, crate::StorageClass::Function) => true,
+                            _ => false,
+                        },
+                        _ => false,
+                    }
+                };
+
                 self.put_expression(base, context, false)?;
-                write!(self.out, "[")?;
+                if accessing_wrapped_array {
+                    write!(self.out, ".{}[", WRAPPED_ARRAY_FIELD)?;
+                } else {
+                    write!(self.out, "[")?;
+                }
                 self.put_expression(index, context, true)?;
                 write!(self.out, "]")?;
             }
@@ -1370,9 +1394,9 @@ impl<W: Write> Writer<W> {
                                 .unwrap();
                             write!(self.out, "{}for(int _i=0; _i<{}; ++_i) ", level, size)?;
                             self.put_expression(pointer, &context.expression, true)?;
-                            write!(self.out, "[_i] = ")?;
+                            write!(self.out, ".{}[_i] = ", WRAPPED_ARRAY_FIELD)?;
                             self.put_expression(value, &context.expression, true)?;
-                            writeln!(self.out, "[_i];")?;
+                            writeln!(self.out, ".{}[_i];", WRAPPED_ARRAY_FIELD)?;
                         }
                         None => {
                             write!(self.out, "{}", level)?;
@@ -1493,7 +1517,7 @@ impl<W: Write> Writer<W> {
                         access: crate::StorageAccess::empty(),
                         first_time: false,
                     };
-                    write!(self.out, "typedef {} {}", base_name, name)?;
+
                     match size {
                         crate::ArraySize::Constant(const_handle) => {
                             let coco = ConstantContext {
@@ -1502,10 +1526,17 @@ impl<W: Write> Writer<W> {
                                 names: &self.names,
                                 first_time: false,
                             };
-                            writeln!(self.out, "[{}];", coco)?;
+
+                            writeln!(self.out, "struct {} {{", name)?;
+                            writeln!(
+                                self.out,
+                                "{}{} {}[{}];",
+                                INDENT, base_name, WRAPPED_ARRAY_FIELD, coco
+                            )?;
+                            writeln!(self.out, "}};")?;
                         }
                         crate::ArraySize::Dynamic => {
-                            writeln!(self.out, "[1];")?;
+                            writeln!(self.out, "typedef {} {}[1];", base_name, name)?;
                         }
                     }
                 }
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -576,10 +576,10 @@ impl<W: Write> Writer<W> {
                             base: pointer_base,
                             class,
                         } => match context.module.types[pointer_base].inner {
-                            crate::TypeInner::Array { .. } => {
-                                class == crate::StorageClass::Function
-                                    || class == crate::StorageClass::Private
-                            }
+                            crate::TypeInner::Array {
+                                size: crate::ArraySize::Constant(_),
+                                ..
+                            } => true,
                             _ => false,
                         },
                         _ => false,
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -578,7 +578,7 @@ impl<W: Write> Writer<W> {
                     write!(self.out, ",")?;
                 }
                 self.put_expression(component, context, false)?;
-                write!(self.out, ".inner[{}]", j)?;
+                write!(self.out, ".{}[{}]", WRAPPED_ARRAY_FIELD, j)?;
             }
             write!(self.out, "}}")?;
         } else {
@@ -614,8 +614,8 @@ impl<W: Write> Writer<W> {
                         crate::TypeInner::Pointer {
                             base: pointer_base,
                             class,
-                        } => match &context.module.types[pointer_base].inner {
-                            &crate::TypeInner::Array { .. } => {
+                        } => match context.module.types[pointer_base].inner {
+                            crate::TypeInner::Array { .. } => {
                                 class == crate::StorageClass::Function
                                     || class == crate::StorageClass::Private
                             }
@@ -1194,7 +1194,11 @@ impl<W: Write> Writer<W> {
                                     if j != 0 {
                                         write!(self.out, ",")?;
                                     }
-                                    write!(self.out, "{}.{}.inner[{}]", tmp, name, j)?;
+                                    write!(
+                                        self.out,
+                                        "{}.{}.{}[{}]",
+                                        tmp, name, WRAPPED_ARRAY_FIELD, j
+                                    )?;
                                 }
                                 write!(self.out, "}}")?;
                             } else {
@@ -2016,11 +2020,11 @@ impl<W: Write> Writer<W> {
                         {
                             continue;
                         }
-                        let array_len = match &module.types[ty].inner {
+                        let array_len = match module.types[ty].inner {
                             crate::TypeInner::Array {
                                 size: crate::ArraySize::Constant(handle),
                                 ..
-                            } => module.constants[*handle].to_array_length(),
+                            } => module.constants[handle].to_array_length(),
                             _ => None,
                         };
                         let resolved = options.resolve_local_binding(binding, out_mode)?;
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -566,8 +566,7 @@ impl<W: Write> Writer<W> {
                     match *context.info[base].ty.inner_with(&context.module.types) {
                         crate::TypeInner::Array { .. } => true,
                         crate::TypeInner::Pointer {
-                            base: pointer_base,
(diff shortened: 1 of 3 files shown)


expenses:
> This right solution to fix this test would be for `gl_ClipDistance` to be treated as a `float` instead of a `float[1]` in `msl-out`.

I guess we need some kind of type override system for this?

kvark:
> This right solution to fix this test would be for gl_ClipDistance to be treated as a float instead of a float[1] in msl-out.

Are you

expenses:
> Are you sure this is the right solution? I think there may be multiple clip distances, technically, so if we implement this workaround, we'll be back at square one with regards to general support for this.
> Instead, the subject of this PR sounds more reasonable. I do wonder if we should simply wrap _all_ of the arrays...


Yeah I just realised that table 5.3 of https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf states that it can be an array of floats. I'll have a go at using the array type in the output struct so instead of `type6 gl_ClipDistance1 [[clip_distance]];` we have `float[1] gl_ClipDistance1 [[clip_distance]];`

expenses:
> Instead, the subject of this PR sounds more reasonable. I do wonder if we should simply wrap _all_ of the arrays...

To clarify, the reason we can't do is is because a wrapped array doesn't count for `[[clip_distance]]`:
```
<stdin>:40:30: error: type 'type6' is not valid for attribute 'clip_distance'
    type6 gl_ClipDistance1 [[clip_distance]];
                             ^~~~~~~~~~~~~
```

expenses:
I think what I've got here works pretty well now. Just FYI, what `spirv-cross` does in these situations is that it creates a template struct like:
```metal

template<typename T, size_t Num>
struct spvUnsafeArray
{
    T elements[Num ? Num : 1];
    
    thread T& operator [] (size_t pos) thread
    {
        return elements[pos];
    }
    constexpr const thread T& operator [] (size_t pos) const thread
    {
        return elements[pos];
    }
    
    device T& operator [] (size_t pos) device
    {
        return elements[pos];
    }
    constexpr const device T& operator [] (size_t pos) const device
    {
        return elements[pos];
    }
    
    constexpr const constant T& operator [] (size_t pos) const constant
    {
        return elements[pos];
    }
    
    threadgroup T& operator [] (size_t pos) threadgroup
    {
        return elements[pos];
    }
    constexpr const threadgroup T& operator [] (size_t pos) const threadgroup
    {
        return elements[pos];
    }
};
```
and returns that in functions instead:
```metal

static inline __attribute__((always_inline))
spvUnsafeArray<float, 1> xyz()
{
    spvUnsafeArray<float, 1> y;
    return y;
}
```
This means that they don't have to do `.inner` at all.

src/back/msl/writer.rs
```diff
@@ -124,7 +125,20 @@ impl<'a> Display for TypeContext<'a> {
                     vector_size_string(size),
                 )
             }
-            crate::TypeInner::Array { .. } | crate::TypeInner::Struct { .. } => unreachable!(),
+            crate::TypeInner::Array { base, .. } => {
+                let sub = Self {
+                    arena: self.arena,
+                    names: self.names,
+                    handle: base,
```
kvark, line 132: could we do `let sub = Self { handle: base, .. *self };`?

src/back/msl/writer.rs
```diff
@@ -594,8 +608,28 @@ impl<W: Write> Writer<W> {
...
             crate::Expression::Access { base, index } => {
+                let accessing_wrapped_array =
+                    match *context.info[base].ty.inner_with(&context.module.types) {
+                        crate::TypeInner::Array { .. } => true,
+                        crate::TypeInner::Pointer {
+                            base: pointer_base,
+                            class,
+                        } => match context.module.types[pointer_base].inner {
+                            crate::TypeInner::Array { .. } => {
+                                class == crate::StorageClass::Function
```
kvark, line 619: why are we limiting this only to Function and Private classes?

src/back/msl/writer.rs
```diff
@@ -594,8 +608,28 @@ impl<W: Write> Writer<W> {
...
+                            }
+                            _ => false,
+                        },
+                        _ => false,
+                    };
+
                 self.put_expression(base, context, false)?;
-                write!(self.out, "[")?;
+                if accessing_wrapped_array {
+                    write!(self.out, ".{}[", WRAPPED_ARRAY_FIELD)?;
```
kvark, line 629: nit: let's make this purely additive:
```rust
                if accessing_wrapped_array {
                    write!(self.out, ".{}", WRAPPED_ARRAY_FIELD)?;
                }
                write!(self.out, "[")?;
```

src/back/msl/writer.rs
```diff
@@ -1160,7 +1194,11 @@ impl<W: Write> Writer<W> {
                                     if j != 0 {
                                         write!(self.out, ",")?;
                                     }
-                                    write!(self.out, "{}.{}[{}]", tmp, name, j)?;
+                                    write!(
```
kvark, line 1197: wait, do we need this section at all? It's only there because in this code:
```rust
Foo foo = { a ,b , c};
```
This code doesn't work if `b` is an array... But if we have it wrapped inside a structure, then we should no longer need a workaround!

src/back/msl/writer.rs
```diff
@@ -1967,17 +2012,27 @@ impl<W: Write> Writer<W> {
                             names: &self.names,
                             usage: GlobalUse::empty(),
                             access: crate::StorageAccess::empty(),
-                            first_time: false,
+                            first_time: true,
```
kvark, line 2015: why is this `true` now?

kvark (commented):
Amazing stuff, I'm excited to see this happening!
I'd prefer to make this work with `inner` instead of defining this complex template in the code like SPIRV-Cross does.

tests/out/quad-vert.msl
```diff
@@ -35,7 +37,7 @@ struct main2Output {
     metal::float2 member [[user(loc0), center_perspective]];
     metal::float4 gl_Position1 [[position]];
     float gl_PointSize1 [[point_size]];
-    type6 gl_ClipDistance1 [[clip_distance]];
+    float gl_ClipDistance1 [[clip_distance]] [1];
```
kvark, line 40: this seems wrong. Why is it not written as `float gl_ClipDistance1[1] [[clip_distance]];`?

expenses: Nope, the compiler doesn't like that:
```
testfrag.metal:40:34: error: 'clip_distance' attribute cannot be applied to types
    float gl_ClipDistance1 [1] [[clip_distance]];
                                 ^
testfrag.metal:42:8: error: invalid return type 'main2Output' for vertex function
vertex main2Output main2(
       ^
testfrag.metal:40:5: note: field of illegal type 'float [1]' declared here
    float gl_ClipDistance1 [1] [[clip_distance]];
    ^
2 errors generated.
```

src/back/msl/writer.rs
expenses: Because we want to have `float` here, instead of `type6` which is the alias that gets used every time after the first.

src/back/msl/writer.rs
expenses: It's not quite that simple unfortunately. In the quad-vert code at the moment we have:
```metal
struct gl_PerVertex {
    metal::float4 gl_Position;
    float gl_PointSize;
    type6 gl_ClipDistance;
    type6 gl_CullDistance;
};
struct type10 {
    metal::float2 member;
    metal::float4 gl_Position1;
    float gl_PointSize1;
    type6 gl_ClipDistance1;
};
...
struct main2Output {
    metal::float2 member [[user(loc0), center_perspective]];
    metal::float4 gl_Position1 [[position]];
    float gl_PointSize1 [[point_size]];
    float gl_ClipDistance1 [[clip_distance]] [1];
};
vertex main2Output main2(
  main2Input varyings [[stage_in]]
) {
    gl_PerVertex _ = {};
    ...
    const auto _tmp = type10 {v_uv, _.gl_Position, _.gl_PointSize, {_.gl_ClipDistance.inner[0]}};
    return main2Output { _tmp.member, _tmp.gl_Position1, _tmp.gl_PointSize1, {_tmp.gl_ClipDistance1.inner[0]} };
}
```
If we remove the hack from `put_initialization_component` we get:
```metal
const auto _tmp = type10 {v_uv, _.gl_Position, _.gl_PointSize, _.gl_ClipDistance};
```
and that compiles fine, but doing a similar thing to `put_return_value` results in:
```metal
return main2Output { _tmp.member, _tmp.gl_Position1, _tmp.gl_PointSize1, _tmp.gl_ClipDistance1 };
```
which fails to compile with:
```

File: src/back/msl/writer.rs (hunk 2 above)
Comment (expenses) [reply src/back/msl/writer.rs:580]: I wasn't sure what other classes count. I don't think we want to be doing `.inner` on storage or uniform arrays though.

Comment (kvark) [reply src/back/msl/writer.rs:580]: Supposing I have a struct
```rust
struct Foo {
  array: array<f32; 3>;
};
```
So if this structure is a uniform/storage buffer, then we don't have `inner`.
But if it's a copy we made into function/private storage, it does have inner? It seems to me that we'd not even be able to re-use this structure when copying it, would we?

src/back/msl/writer.rs
kvark: what happens if we let it use `type6`?

expenses: We get the error:
```
testfrag.metal:40:30: error: type 'type6 [1]' is not valid for attribute 'clip_distance'
    type6 gl_ClipDistance1 [[clip_distance]] [1];
                             ^~~~~~~~~~~~~
testfrag.metal:42:8: error: invalid return type 'main2Output' for vertex function
vertex main2Output main2(
       ^
2 errors generated.
```

kvark: ok, cool, thanks!

src/back/msl/writer.rs
expenses: I'm not entirely sure what you mean. If you could come up with a test shader for this then I would be grateful.
Something like:
```glsl
struct Foo {
  float[3] array;
};

Foo foo() {
    Foo foo = {{1.0, 1.5, 2.5}};
    return foo;
}
```
works fine though:
```metal
struct type2 {
    float inner[3u];
};
struct Foo {
    type2 array;
};
constant type2 const_type2_ = {1.0, 1.5, 2.5};
constant Foo const_Foo = {const_type2_};

Foo foo(
) {
    Foo foo1;
    foo1 = const_Foo;
    return foo1;
}

void main1(
    thread metal::float4& colour
) {
    Foo foo2;
    Foo _e14 = foo();
    foo2 = _e14;
    colour = float4(foo2.array.inner[2]);
    return;
}
```

kvark: In your case, what happens if `Foo` is used as uniform/storage buffer? If you have a pointer to it, it will not be in Private/Function storage class, so the code in question will not trigger.

expenses: Ah, good point. I've fixed this now.

kvark:
Are we all set?

kvark (approved):

tests/out/quad-vert.msl
```diff
@@ -49,6 +51,6 @@ vertex main2Output main2(
     a_uv = a_uv1;
     a_pos = a_pos1;
     main1(v_uv, a_uv, _, a_pos);
-    const auto _tmp = type10 {v_uv, _.gl_Position, _.gl_PointSize, {_.gl_ClipDistance[0]}};
-    return main2Output { _tmp.member, _tmp.gl_Position1, _tmp.gl_PointSize1, {_tmp.gl_ClipDistance1[0]} };
+    const auto _tmp = type10 {v_uv, _.gl_Position, _.gl_PointSize, _.gl_ClipDistance};
+    return main2Output { _tmp.member, _tmp.gl_Position1, _tmp.gl_PointSize1, {_tmp.gl_ClipDistance1.inner[0]} };
```
kvark, line 55: wait, so here we can't just do `_tmp.gl_ClipDistance1` for the last argument?

expenses: Nope, unfortunately not. Not `gl_ClipDistance1.inner` either.

expenses:
> Are we all set?

I think so! I hope this doesn't break any vertex shaders.

tests/out/quad-vert.msl
kvark: hmm, I'm concerned now, a little bit. So when we are initializing the I/O struct, we can't just pass `_tmp.gl_ClipDistance1` because it's an array. But how would this work in other places? I.e. if you have a user function returning a user struct, then passing `{ xxx.inner[0], xxx.inner[1] }` would just fail horribly, since the only field of the target struct is `inner`.

expenses: Afaik, only having a single pair of those `{` brackets when constructing a struct is okay if it only has one field. I think more testing would be nice, but I haven't found a (trivial) example that this breaks. Something like the following, for example, is fine:
```glsl
#version 450

vec3[8] hello_world() {
    vec3[8] xyz;
    return xyz;
}

struct X {
  vec3[8] yyy;
};

vec3[8] hmm2() {
  X x;
  x.yyy = hello_world();
  x.yyy[1] = vec3(3.0);
  return x.yyy;
}

void main() {
    vec3[8] xyz = hmm2();
    colour = vec4(xyz[7], 1.0);
}
```
->
```metal
#include <metal_stdlib>
#include <simd/simd.h>

struct type3 {
    metal::float3 inner[8u];
};
struct X {
    type3 yyy;
};
constant metal::float3 const_type1_ = {3.0, 3.0, 3.0};

type3 hello_world(
) {
    type3 xyz;
    return xyz;
}

type3 hmm2_(
) {
    X x;
    type3 _e13 = hello_world();
    for(int _i=0; _i<8; ++_i) x.yyy.inner[_i] = _e13.inner[_i];
    x.yyy.inner[1] = const_type1_;
    return x.yyy;
}

void main1(
    thread metal::float4& colour
) {
    type3 xyz1;
    type3 _e13 = hmm2_();
    for(int _i=0; _i<8; ++_i) xyz1.inner[_i] = _e13.inner[_i];
    metal::float3 _e15 = xyz1.inner[7];
    colour = metal::float4(_e15.x, _e15.y, _e15.z, 1.0);
    return;
}

struct main2Output {
    metal::float4 member [[color(0)]];
};
fragment main2Output main2(
) {
    metal::float4 colour = {};
    main1(colour);
    return main2Output { colour };
}
```

kvark: I mean, specifically for arrays.
```rust
struct Foo {
  xxx: array<f32; 3>;
};
fn bar() -> Foo {
  return Foo(array<f32;3>(1.0, 2.0, 3.0));
}
```
My understanding is that this PR right now is not going to handle this. Ideally, we'd want it to generate something like
```rust
const auto _tmp = Foo { .. };
return { { _tmp.xxx[0], _tmp.xxx[1], _tmp.xxx[2] } };
```
This will fail because the produced `Foo` would be declared with the `inner` in it, but the `return` code doesn't use it.

expenses:
```rust
struct Foo {
  xxx: array<f32, 3>;
};
fn bar() -> Foo {
  return Foo(array<f32, 3>(1.0, 2.0, 3.0));
}
```
fails for me with `error: invalid type cast to `Foo``. Looks like struct composition isn't implemented yet/ working?

kvark: Sorry about this! Filed #792
Struct composition works fine for 2+ arguments:
```rust
struct Foo {
  xxx: array<f32, 3>;
  bar: i32;
};
fn bar() -> Foo {
  return Foo(array<f32, 3>(1.0, 2.0, 3.0), 2);
}
```

expenses:
```rust
struct Foo {
  xxx: array<f32, 3>;
  bar: i32;
};
fn bar() -> Foo {
  return Foo(array<f32, 3>(1.0, 2.0, 3.0), 2);
}

[[stage(vertex)]]
fn foo([[builtin(vertex_index)]] vi: u32) -> [[builtin(position)]] vec4<f32> {
	let foo = bar();
	return vec4<f32>(foo.xxx[0], foo.xxx[1], foo.xxx[2], 1.0);
}
```
translates to:
```metal
#include <metal_stdlib>
#include <simd/simd.h>

struct type1 {
    float inner[3];
};
struct Foo {
    type1 xxx;
    int bar;
};

Foo bar1(
) {
    return Foo {type1 {1.0, 2.0, 3.0}, 2};
}

struct fooInput {
};
struct fooOutput {
    metal::float4 member [[position]];
};
vertex fooOutput foo(
  metal::uint vi [[vertex_id]]
) {
    Foo _e1 = bar1();
    return fooOutput { metal::float4(_e1.xxx.inner[0], _e1.xxx.inner[1], _e1.xxx.inner[2], 1.0) };
}
```
and that compiles fine.

kvark: Ok, I don't understand how `type1 {1.0, 2.0, 3.0}` part works :(
Type1 is a structure with 1 element. So shouldn't it yell at us for providing 3 elements instead of 1?

expenses: Yeah I'm not 100% sure either, but what seems to happen is that (as I wrote in https://github.com/gfx-rs/naga/pull/764#discussion_r620511225):

> Afaik, only having a single pair of those `{` brackets when constructing a struct is okay if it only has one field

expenses: Same thing is allowed in C:
```c
struct Arr {
    float inner[4];
};

int main() {
    struct Arr arr = { 1.0, 1.5, 2.0, 2.5 };
}
```

expenses: Uh oh, in both MSL and C it lets you skip fields in the constructor though so
```metal
Foo bar1(
) {
    return Foo {type1 {1.0}, 2};
}
```
also works?? With no warnings or errors 😟 ??

kvark: What's going to happen if the array only has one element? How is the compiler going to figure out if the initializer is for this field or for the `inner` itself? And initializers don't even have to be provided for all the fields. Ugh, C is weird :(

kvark:
I'm royally confused about how C arrays work, and that is going to make the maintenance of this code to be difficult.
But you've really done a great job implementing it, and it addresses a very important issue we have, and there isn't any cleaner solution that we know. So let's proceed!

PR #770, opened 2021-04-24
More wgsl-out improvements

Gordon-F:
1. Return Err instead of panic for unimplemented cases
2. Implement writing for Scalar, Array, Matrix
3. Share some code between backends

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1836,30 +1836,7 @@ impl<'a, W: Write> Writer<'a, W> {
                 if function.is_some() {
                     write!(self.out, ",")?
                 } else {
-                    write!(
-                        self.out,
-                        " {} ",
-                        match op {
-                            BinaryOperator::Add => "+",
-                            BinaryOperator::Subtract => "-",
-                            BinaryOperator::Multiply => "*",
-                            BinaryOperator::Divide => "/",
-                            BinaryOperator::Modulo => "%",
-                            BinaryOperator::Equal => "==",
-                            BinaryOperator::NotEqual => "!=",
-                            BinaryOperator::Less => "<",
-                            BinaryOperator::LessEqual => "<=",
-                            BinaryOperator::Greater => ">",
-                            BinaryOperator::GreaterEqual => ">=",
-                            BinaryOperator::And => "&",
-                            BinaryOperator::ExclusiveOr => "^",
-                            BinaryOperator::InclusiveOr => "|",
-                            BinaryOperator::LogicalAnd => "&&",
-                            BinaryOperator::LogicalOr => "||",
-                            BinaryOperator::ShiftLeft => "<<",
-                            BinaryOperator::ShiftRight => ">>",
-                        }
-                    )?;
+                    write!(self.out, " {} ", super::binary_operation_str(op))?;
                 }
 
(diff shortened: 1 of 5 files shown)


src/back/wgsl/writer.rs
```diff
@@ -463,30 +517,7 @@ impl<W: Write> Writer<W> {
                 self.write_expr(module, left, func_ctx)?;
 
                 // TODO: glsl-out copy-paste
```
ColdIce1605, line 519: I think this todo can be removed.

src/back/glsl/mod.rs
```diff
@@ -1836,30 +1836,7 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                            BinaryOperator::And => "&",
-                            BinaryOperator::ExclusiveOr => "^",
-                            BinaryOperator::InclusiveOr => "|",
-                            BinaryOperator::LogicalAnd => "&&",
-                            BinaryOperator::LogicalOr => "||",
-                            BinaryOperator::ShiftLeft => "<<",
-                            BinaryOperator::ShiftRight => ">>",
-                        }
-                    )?;
+                    write!(self.out, " {} ", crate::back::binary_operation_str(op))?;
```
kvark, line 1839: nit: maybe refer to this as `super::` instead?

src/back/mod.rs
```diff
@@ -33,3 +33,53 @@ impl crate::Expression {
         }
     }
 }
+
+/// Helper function that returns the string corresponding to the [`BinaryOperator`](crate::BinaryOperator)
+/// # Notes
+/// Used by `glsl-out`, `msl-out`, `wgsl-out`.
+fn binary_operation_str(op: crate::BinaryOperator) -> &'static str {
+    match op {
+        crate::BinaryOperator::Add => "+",
```
kvark, line 42: let's put `use crate::BinaryOperator as Bo` at the top of the function

src/back/mod.rs
```diff
@@ -33,3 +33,53 @@ impl crate::Expression {
...
+        crate::VectorSize::Bi => "2",
+        crate::VectorSize::Tri => "3",
+        crate::VectorSize::Quad => "4",
+    }
+}
+
+/// Helper function that returns the string corresponding to the [`ImageDimension`](crate::ImageDimension)
+/// # Notes
+/// Used by `msl-out`, `wgsl-out`.
+fn image_dimension_str(dim: crate::ImageDimension) -> &'static str {
```
kvark, line 78: let's not have this shared, it seems too language specific

kvark (commented):
Nice work!

Gordon-F:
@kvark Is there some way to disable clippy warning only for some features? To avoid warning like this one: `function is never used: binary_operation_str`

kvark:
@Gordon-F see https://github.com/gfx-rs/naga/blob/77a64da189e6ed2c4631ca03cb01b8d8124c4dad/src/front/mod.rs#L22

Gordon-F:
> @Gordon-F see
> 
> https://github.com/gfx-rs/naga/blob/77a64da189e6ed2c4631ca03cb01b8d8124c4dad/src/front/mod.rs#L22

Thank you! PR is ready to merge.

kvark (approved):
Thank you!

PR #776, opened 2021-04-25
[glsl-in]: Glsl new parser

JCapucho:
It's very much a work in progress for now just opening this to get some comments as  I go

Replaces `pomelo` with a hand written parser


JCapucho:
Small update: I finally managed to figure out declarations, things should be easier from here (I hope)

Cargo.toml
```diff
@@ -20,18 +20,18 @@ fxhash = "0.2"
 log = "0.4"
 num-traits = "0.2"
 spirv = { package = "spirv_headers", version = "1.5", optional = true }
-pomelo = { version = "0.1.4", optional = true }
 thiserror = "1.0.21"
 serde = { version = "1.0", features = ["derive"], optional = true }
 petgraph = { version ="0.5", optional = true }
 rose_tree = { version ="0.2", optional = true }
-pp-rs = { version = "0.1", optional = true }
+# TODO: Make another release of glslpp-rs
```
kvark, line 27: it's OK to depend on a git revision (but needs either a particular revision, or a tag) until we need to publish Naga itself

src/front/glsl/ast.rs
```diff
@@ -1,142 +1,150 @@
...
+            //TODO: better error report
+            Err(error) => Err(ErrorKind::SemanticError(
+                format!("Can't resolve type: {:?}", error).into(),
+            )),
+            Ok(()) => Ok(context.typifier.get(handle, &self.module.types)),
+        }
     }
 
-    pub fn resolve_type(
+    pub fn resolve_handle(
```
kvark, line 114: We shouldn't need to have this as a separate method. Instead, we could add `fn get_resolution()` method to `Typifier`, and return `TypeResolution` here in this method. Then everything else can be done on it:
  - if you need a `TypeInner`, then call `resolution.inner_with(types)`
  - if you need `Handle<Type>`, then call `resolution.handle().unwrap()`

src/front/glsl/ast.rs
```diff
@@ -257,64 +314,209 @@ impl Context {
...
+                    let (op, fun) = match equals {
+                        true => (BinaryOperator::Equal, RelationalFunction::All),
+                        false => (BinaryOperator::NotEqual, RelationalFunction::Any),
+                    };
+
+                    let expr = self
+                        .expressions
+                        .append(Expression::Binary { op, left, right });
+
+                    if left_is_vector && right_is_vector {
```
kvark, line 365: If only one of them is a vector, is this valid in GLSL? If it's valid, I assume we'd need to `Splat` the scalar one first?

src/front/glsl/ast.rs
```diff
@@ -257,64 +314,209 @@ impl Context {
...
+#[derive(Debug, Clone)]
+pub struct Expr {
+    pub kind: ExprKind,
+    pub meta: TokenMetadata,
+}
+
+#[derive(Debug, Clone)]
+pub enum ExprKind {
+    Access {
+        base: Box<Expr>,
```
kvark, line 440: ooh, that's a bit heavy-handed. Many boxes!
Perhaps, GLSL frontend doesn't need to be as fast as the others, but everywhere else we stay away from such boxes.
Perhaps, we can instead just use the arena here? I.e. that would be `Handle<Expr>` instead of `Box<Expr>`. Although, I'm not sure what purpose would `enum ExprKind` have in this case - it would look very similar to `enum Expression`.

kvark (commented):
Amazing work! First review notes are below

Cargo.toml
JCapucho: I'm trying to get the branch merged so that we can keep this pointing to the main repo or potentially a new release, also I tried to add a rev but cargo complained because of the branch

src/front/glsl/ast.rs
JCapucho: I'm planning to replace the boxes with an arena just haven't gotten around to it (also I wonder if it would be possible to make the lowering in a loop trough the arena instead of recursive).

The `ExprKind` exists to handle stores, LHS expressions and in the future postfix and prefix Inc/Dec.

JCapucho: This part now uses a `Arena` there's still an annoying clone in the lowering part (It's possible to remove but it would require some organizational changes that I think should be done in another PR) but it's a shallow clone and should much better than the previous implementation.

src/front/glsl/ast.rs
JCapucho: hmm I'm not sure that would work since the frontend would panic on things like float literals.

Ideally I would like to not have this at all because right now this is used in `FunctionSignature` to support overloading but this is really error prone because we must ensure that all the types that are equal must have the same handle.

jakobhellermann:
I tried using this branch for [bevy's](https://github.com/bevyengine/bevy) glsl-to-spirv compilation, and while the parsing worked, using the `naga::valid::Validator` failed on the vertex attributes with `ValidationError::GlobalVariable { handle: [1], name: "Vertex_Position", error: InvalidUsage }` because [their storage type is `Function`](https://github.com/gfx-rs/naga/blob/b191c982ec146f97d3f495f7a8b7af648c189f0d/src/valid/interface.rs#L297). Is passing the validation in scope for this PR or does it only focus on parsing?

JCapucho:
> I tried using this branch for [bevy's](https://github.com/bevyengine/bevy) glsl-to-spirv compilation, and while the parsing worked, using the `naga::valid::Validator` failed on the vertex attributes with `ValidationError::GlobalVariable { handle: [1], name: "Vertex_Position", error: InvalidUsage }` because [their storage type is `Function`](https://github.com/gfx-rs/naga/blob/b191c982ec146f97d3f495f7a8b7af648c189f0d/src/valid/interface.rs#L297). Is passing the validation in scope for this PR or does it only focus on parsing?

Good catch, while the main focus of this PR is to add the new parser and then follow up with any problem or missing that is encountered, small things like this will be fixed in this PR

JCapucho:
@jakobhellermann It should be fixed now

kvark (approved):
shipit!

PR #787, opened 2021-04-26
Add `MathFunction::Refract`

expenses:
This doesn't seem to be in wgsl (yet?) but it's in glsl, spv and metal so I think it's reasonable to add it into the IR.

Here's an glsl example:
```glsl
#version 450

void main() {
    vec3 incidence = vec3(0.0, 1.0, 0.0);
    vec3 normal = vec3(0.0, 1.0, 0.0);
    float ior = 1.0;
    colour = vec4(refract(incidence, normal, ior), 1.0);
}
```


kvark (approved):
Thank you!

PR #788, opened 2021-04-26
HLSL CI

Napokue:


Makefile
```diff
@@ -63,3 +63,6 @@ validate-wgsl: $(SNAPSHOTS_OUT)/*.wgsl
 		echo "Validating" $${file#"$(SNAPSHOTS_OUT)/"};	\
 		cargo run --features wgsl-in $${file}; \
 	done
+
+
+validate-hlsl: $(SNAPSHOTS_OUT)/*.hlsl
```
Napokue, line 68: I couldn't get any for loop variant for Powershell or Command line language (for Windows) working here. I also tried the for loop that is used for the other make functions, but none of them work on my Windows system.

When we solve this, I think the PR is 99% done.

tests/snapshots.rs
```diff
@@ -198,12 +200,13 @@ fn check_output_glsl(
 }
 
 #[cfg(feature = "hlsl-out")]
-fn check_output_hlsl(module: &naga::Module, destination: &PathBuf) {
+fn check_output_hlsl(module: &naga::Module, destination: &PathBuf, stage: naga::ShaderStage) {
```
kvark, line 203: why are we having the stage here? `hlsl::write_string` doesn't depend on it, so neither should the test output file.

kvark (commented):
I'll switch to windows and play with it later tonight

tests/snapshots.rs
Napokue: Ah yes right, well that is due to the naming for the snapshots. The dxc compiler has an argument (can't remember exactly what it was) that is required to specify the profile. So that will be shadermodel 6.0 (or higher) and the stage of the shader. The entrypoint also has to be specified, but for now I think we can assume that all test shaders their entrypoint names will be "name".

Napokue: I wasn't that far yet, but take a look at the arguments that [Shader Playground](https://github.com/tgjones/shader-playground/blob/master/src/ShaderPlayground.Core/Compilers/Dxc/DxcCompiler.cs#L142) passes to the dxc compiler

kvark: Interesting. We need one of the following:
  1. Find/request a way from DXC to check all the entry points - https://github.com/microsoft/DirectXShaderCompiler/issues/3714
  2. Generate a small file per test case with a list of entry points and their execution models, so that we can iterate the lines and feed them into DXC

Makefile
kvark: This is indeed tough, unexpectedly!
I found this to work:
```bash
validate-hlsl: $(SNAPSHOTS_OUT)/*.hlsl
	@FOR %%f IN ($^) DO dxc %%f -T ps_6_0
```

Napokue:
The CI works, even though it is failing 😃 

Reason it is failing is because the compute shader is not valid, as it requires "Thread Group declaration X, Y, and Z". So the validation works!

I propose that I will follow up with a fix for the shader itself.

kvark (approved):

kvark:
The problem with fixing it later is that people will see their stuff failing CI and assume something is wrong with their code.
How difficult would it be to fix it prior to landing this?

Napokue:
Sure, no problem, will look into that then in this PR. I wholeheartedly agree with what you are saying. I think it shouldn't be too hard.

Napokue:
PR is ready for last round of review, and if everything is correct, it can be landed.

kvark (approved):
Great!

PR #789, opened 2021-04-26
[spv-in] Properly parse empty names and set entry point function name

Gordon-F:
I'm not sure about these changes at all. Found these issues when developing `wgsl-out`.


src/front/spv/mod.rs
```diff
@@ -2371,7 +2371,8 @@ impl<I: Iterator<Item = u32>> Parser<I> {
         if left != 0 {
             return Err(Error::InvalidOperand);
         }
-        self.future_decor.entry(id).or_default().name = Some(name);
+        self.future_decor.entry(id).or_default().name =
+            if name.is_empty() { None } else { Some(name) };
```
kvark, line 2375: I don't think this is needed. We are parsing `OpName` here. If the incoming SPIRV has an empty name, so be it.

Gordon-F: It's better to fix this issue in backend side like we discuss in matrix? Otherwise we have variables with name `_`. `quad-vert.msl` in diff for example.

src/front/spv/function.rs
```diff
@@ -189,7 +189,7 @@ impl<I: Iterator<Item = u32>> super::Parser<I> {
         if let Some(ep) = self.lookup_entry_point.remove(&fun_id) {
             // create a wrapping function
             let mut function = crate::Function {
-                name: None,
+                name: Some(ep.name.clone()),
```
Gordon-F, line 192: @kvark How about this fix ?

src/front/spv/mod.rs
Gordon-F: I think it better to fix in `Namer`, not in spv frontend. Will do in next PR.

kvark: Does this actually *need* to be solved? If the backend ends up generating the name "_", so be it, given that the input annotated something with an empty string.

src/front/spv/function.rs
kvark: this seems fine to me, but we can probably do better, like `Some(format!("{}Wrap", ep.name))`. Otherwise, we'll have both the entry point and the function wrapper to have the same name.

src/front/spv/mod.rs
Gordon-F: > Does this actually need to be solved?

It's only about readability.

```metal
gl_PerVertex _ = {};
type10 {v_uv, _.gl_Position, _.gl_PointSize, {_.gl_ClipDistance[0]}};
```

```metal
gl_PerVertex global = {};
type10 {v_uv, global.gl_Position, global.gl_PointSize, {global.gl_ClipDistance[0]}};
```

kvark: Ok. I think we can assume that an empty name (that we want to handle) is only coming from SPIR-V, and only from the `gl_PerVertex` struct. Could we then have the code in place in spv-in specifically that is as narrow (for this use case) as possible? I don't think we want any sort of generic mechanism to eliminate empty names, that doesn't make sense. So a very specialized piece in the spv-in for this gl_PerVertex struct should be most appropriate.

Gordon-F: Ok, I decided to fix it on `namer` side. If it's ok, we can merge PR, or we or can leave it as is and close PR. It's not such an important issue.

src/proc/namer.rs
```diff
@@ -57,7 +57,14 @@ impl Namer {
 
     fn call_or(&mut self, label: &Option<String>, fallback: &str) -> String {
         self.call(match *label {
-            Some(ref name) => name,
+            Some(ref name) => {
+                // Name can be empty for some cases in SPIR-V
```
kvark, line 61: there is only one case where this happens in practice, and `spv` frontend should be handling it, not `Namer`

Gordon-F: One more try. Thank you for your patience :smile:

kvark (approved):
Awesome, thank you!

PR #793, opened 2021-04-27
[wgsl-out] Enable snapshot testing for spv-in quad shader

Gordon-F:
First attempt to generate shader from spv frontend. Draft only because I can't fix all validation errors.

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/proc/interpolator.rs:90:52
make: *** [validate-wgsl] Error 101
```

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -1,23 +1,19 @@
 // TODO: temp
 #![allow(dead_code)]
 use super::Error;
-use crate::FastHashMap;
 use crate::{
     back::{binary_operation_str, vector_size_str, wgsl::keywords::RESERVED},
-    proc::{EntryPointIndex, TypeResolution},
+    proc::{EntryPointIndex, NameKey, Namer, TypeResolution},
     valid::{FunctionInfo, ModuleInfo},
-    Arena, ArraySize, Binding, Constant, Expression, Function, GlobalVariable, Handle, ImageClass,
-    ImageDimension, Module, ScalarKind, ShaderStage, Statement, StorageFormat, StructLevel, Type,
-    TypeInner,
-};
-use crate::{
-    proc::{NameKey, Namer},
-    StructMember,
+    Arena, ArraySize, Binding, Constant, Expression, FastHashMap, Function, GlobalVariable, Handle,
+    ImageClass, ImageDimension, Interpolation, Module, Sampling, ScalarKind, ScalarValue,
+    ShaderStage, Statement, StorageFormat, StructLevel, StructMember, Type, TypeInner,
 };
 use bit_set::BitSet;
 use std::fmt::Write;
 
 const INDENT: &str = "    ";
+const COMPONENTS: &[char] = &['x', 'y', 'z', 'w'];
 const BAKE_PREFIX: &str = "_e";
 
 /// Shorthand result used internally by the backend
@@ -26,10 +22,12 @@ type BackendResult = Result<(), Error>;
 /// WGSL attribute
 /// https://gpuweb.github.io/gpuweb/wgsl/#attributes
 enum Attribute {
+    Access(crate::StorageAccess),
     Binding(u32),
     Block,
     BuiltIn(crate::BuiltIn),
     Group(u32),
+    Interpolate(Option<Interpolation>, Option<Sampling>),
     Location(u32),
     Stage(ShaderStage),
     Stride(u32),
@@ -75,23 +73,14 @@ impl<W: Write> Writer<W> {
         }
     }
 
-    pub fn write(&mut self, module: &Module, info: &ModuleInfo) -> BackendResult {
+    fn reset(&mut self, module: &Module) {
         self.names.clear();
         self.namer.reset(module, RESERVED, &[], &mut self.names);
+        self.named_expressions.clear();
+    }
 
-        // Write all constants
-        for (_, constant) in module.constants.iter() {
-            if constant.name.is_some() {
-                self.write_constant(&constant, true)?;
-            }
-        }
-
-        // Write all globals
-        for (_, global) in module.global_variables.iter() {
-            if global.name.is_some() {
-                self.write_global(&module, &global)?;
-            }
-        }
+    pub fn write(&mut self, module: &Module, info: &ModuleInfo) -> BackendResult {
+        self.reset(module);
 
         // Write all structs
         for (handle, ty) in module.types.iter() {
@@ -99,13 +88,46 @@ impl<W: Write> Writer<W> {
                 level, ref members, ..
             } = ty.inner
             {
-                let name = &self.names[&NameKey::Type(handle)].clone();
                 let block = level == StructLevel::Root;
-                self.write_struct(module, name, block, members)?;
+                self.write_struct(module, handle, block, members)?;
                 writeln!(self.out)?;
             }
         }
 
+        // Write all constants
+        for (handle, constant) in module.constants.iter() {
+            if constant.name.is_some() {
+                self.write_global_constant(&constant, handle)?;
+            }
+        }
+
+        // Write all globals
+        for (ty, global) in module.global_variables.iter() {
+            self.write_global(&module, &global, ty)?;
+        }
+
+        if !module.global_variables.is_empty() {
+            // Add extra newline for readability
+            writeln!(self.out)?;
+        }
+
+        // Write all regular functions
+        for (handle, function) in module.functions.iter() {
+            let fun_info = &info[handle];
+
+            let func_ctx = FunctionCtx {
+                ty: FunctionType::Function(handle),
+                info: fun_info,
+                expressions: &function.expressions,
+            };
+
+            // Write the function
+            self.write_function(&module, &function, &func_ctx)?;
+
+            writeln!(self.out)?;
+        }
+
+        // Write all entry points
         for (index, ep) in module.entry_points.iter().enumerate() {
             let attributes = match ep.stage {
                 ShaderStage::Vertex | ShaderStage::Fragment => vec![Attribute::Stage(ep.stage)],
@@ -115,7 +137,7 @@ impl<W: Write> Writer<W> {
                 ],
             };
 
-            self.write_attributes(&attributes)?;
+            self.write_attributes(&attributes, false)?;
             // Add a newline after attribute
             writeln!(self.out)?;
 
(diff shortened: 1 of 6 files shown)


kvark (approved):

Gordon-F:
@kvark Should we return an error from `interpolator` instead of panic?

kvark:
It should absolutely *not* assume that the module is correct, and not panic. cc @jimblandy 
At the same time, it doesn't need to return a proper error - this should be left up to the validator.

Still, I don't understand how this limitation affects your PR. Your PR doesn't add any new inputs for the snapshots, only outputs. The interpolation stuff runs before a module is generated, it doesn't depend on "wgsl-out".

kvark:
aaah, I see. Ok, yeah, we should not panic there. Also, the produced module is incorrect, so something in wgsl-out needs also to be fixed.

Gordon-F:
@kvark All fields from vertex output structs should have bindings?

```WGSL
struct type10 {
    [[location(0), interpolate(perspective)]] member: vec2<f32>;
    [[builtin(position)]] gl_Position1: vec4<f32>;
    gl_PointSize1: f32;
    gl_ClipDistance1: [[stride(4)]] array<f32,1>;
};

[[stage(vertex)]]
fn main1([[location(1)]] a_uv1: vec2<f32>, [[location(0)]] a_pos1: vec2<f32>) -> type10 {
...
```

How we should handle unknown builtins (`gl_PointSize`, `gl_ClipDistance`, etc.) in `wgsl-out`?

kvark:
> How we should handle unknown builtins (gl_PointSize, gl_ClipDistance, etc.) in wgsl-out?

As I wrote in the comment, we'll need the backend returning an error. Something like `Error::UnsupportedBuiltIn()`

Gordon-F:
> As I wrote in the comment, we'll need the backend returning an error. Something like `Error::UnsupportedBuiltIn()`

Does it mean that we can't generate even a simple shader from `glslang`? Maybe we can emulate unknown builtin somehow? 

```glsl
#version 450

void main() {
    gl_Position = vec4(1.0, 1.0, 1.0, 1.0);
}
```

Generated SPIR-V:
```
; SPIR-V
; Version: 1.3

 OpName %main "main"
 OpName %gl_PerVertex "gl_PerVertex"
 OpMemberName %gl_PerVertex 0 "gl_Position"
 OpMemberName %gl_PerVertex 1 "gl_PointSize"
 OpMemberName %gl_PerVertex 2 "gl_ClipDistance"
 OpMemberName %gl_PerVertex 3 "gl_CullDistance"
 OpName %_ ""
 OpMemberDecorate %gl_PerVertex 0 BuiltIn Position
 OpMemberDecorate %gl_PerVertex 1 BuiltIn PointSize
 OpMemberDecorate %gl_PerVertex 2 BuiltIn ClipDistance
 OpMemberDecorate %gl_PerVertex 3 BuiltIn CullDistance
```

kvark:
> Does it mean that we can't generate even a simple shader from glslang? Maybe we can emulate unknown builtin somehow?

I don't think we can emulate it. What we can is:
  1. error for these, and expect people to generate SPIRV that don't have clip/cull distances
  2. actually add support for clip/cull distances to WGSL
  3. try to ignore all the code writing to clip/cull outputs in WGSL-out, so effectively cutting it out

Gordon-F:
> 3. try to ignore all the code writing to clip/cull outputs in WGSL-out, so effectively cutting it out

It seems that [Tint](https://dawn.googlesource.com/tint/+/refs/heads/main/src/reader/spirv/parser_impl.cc#926) uses this variant. I will try to do the same for `wgsl-out`.

Gordon-F:
> It seems that Tint uses this variant. I will try to do the same for wgsl-out.

Finally fixed it! One of the test case that `wgsl-out` passed - running `cube` (wgpu-rs 0.8) with generated shaders from spv (wgpu-rs 0.6 branch).

src/valid/analyzer.rs
```diff
@@ -140,7 +140,7 @@ pub struct SamplingKey {
 pub struct ExpressionInfo {
     pub uniformity: Uniformity,
     pub ref_count: usize,
-    assignable_global: Option<Handle<crate::GlobalVariable>>,
+    pub assignable_global: Option<Handle<crate::GlobalVariable>>,
```
Gordon-F, line 143: @kvark It's ok to make it public?

src/back/wgsl/writer.rs
```diff
@@ -461,19 +568,49 @@ impl<W: Write> Writer<W> {
...
+                if let Some(expr) = result {
+                    self.start_baking_expr(module, expr, &func_ctx)?;
+                    self.named_expressions.insert(expr.index());
+                }
+                let func_name = &self.names[&NameKey::Function(function)];
+                write!(self.out, "{}(", func_name)?;
+                for (index, argument) in arguments.iter().enumerate() {
+                    self.write_expr(module, *argument, func_ctx)?;
+                    // Only write a comma if isn't the last element
+                    if index != arguments.len().saturating_sub(1) {
```
kvark, line 607: WGSL accepts trailing commas, so this can be simplified a but

src/back/wgsl/writer.rs
```diff
@@ -484,14 +621,18 @@ impl<W: Write> Writer<W> {
 
     fn start_baking_expr(
         &mut self,
+        module: &Module,
         handle: Handle<Expression>,
         context: &FunctionCtx,
     ) -> BackendResult {
         // Write variable name
-        write!(self.out, "let {}{}: ", BAKE_PREFIX, handle.index())?;
+        write!(self.out, "var {}{}: ", BAKE_PREFIX, handle.index())?;
```
kvark, line 629: why is this a `var`? I'd expect this being a `let`

src/back/wgsl/writer.rs
```diff
@@ -530,15 +671,79 @@ impl<W: Write> Writer<W> {
...
+                                        if let TypeInner::Struct { ref members, .. } =
+                                            module.types[base].inner
+                                        {
+                                            for (member_index, member) in members.iter().enumerate()
+                                            {
+                                                if let Some(Binding::BuiltIn(builtin)) =
+                                                    member.binding
+                                                {
+                                                    if !is_builtin_supported(builtin)
+                                                        && member_index == access_index as usize
```
kvark, line 717: if we know we are only looking for `member_index == access_index`, why not simply do `members[access_index as usize]` and work with it?

src/back/wgsl/writer.rs
```diff
@@ -802,6 +1071,25 @@ impl<W: Write> Writer<W> {
     }
 }
 
+fn is_builtin_supported(built_in: crate::BuiltIn) -> bool {
```
kvark, line 1074: why aren't we using `builtin_str(built_in).is_some()` instead of having this separate function?

src/back/wgsl/writer.rs
```diff
@@ -839,6 +1127,17 @@ fn scalar_kind_str(kind: ScalarKind) -> &'static str {
     }
 }
 
+fn scalar_value_str(value: ScalarValue) -> String {
```
kvark, line 1130: let's have this as a method on `self` instead, so that we can write directly to `self.out` instead of allocating this temporary string

src/back/wgsl/writer.rs
```diff
@@ -530,15 +671,79 @@ impl<W: Write> Writer<W> {
...
+                // We can, but otherwise, we can't generate WGSL shader from any glslang SPIR-V shaders.
+                // glslang generates gl_PerVertex struct with gl_CullDistance, gl_ClipDistance and gl_PointSize builtin inside by default.
+                // All of them are not supported by WGSL.
+
+                // We need to copy components to another vec because we don't know which of them we should write.
+                let mut components_to_write = Vec::with_capacity(components.len());
+                for component in components {
+                    let mut skip_component = false;
+                    if let Expression::Load { pointer } = func_ctx.expressions[*component] {
+                        let pointer_global = &func_ctx.info[pointer].assignable_global;
```
kvark, line 693: Any particular reason to use the `assignable_global` here?
Could we instead just match `function_ctx.expressions[pointer]` for `Expression::GlobalVariable` and get the handle from there?

src/valid/analyzer.rs
kvark: I don't think there is a strong case here for exposing it. You could just match the expression itself, instead.

src/back/wgsl/writer.rs
```diff
@@ -145,61 +167,100 @@ impl<W: Write> Writer<W> {
...
-                write!(self.out, ")")?;
+                FunctionType::EntryPoint(ep_index) => {
+                    self.names[&NameKey::EntryPointArgument(ep_index, index as u32)].clone()
+                }
+            };
+
+            write!(self.out, "{}: ", argument_name)?;
+            // Write argument type
+            self.write_type(module, arg.ty)?;
+            if index < func.arguments.len() - 1 {
```
kvark, line 198: we can add a trailing separator, no need for this condition

kvark (commented):
Amazing work!
A few things are noted to address, but nothing major really

src/back/wgsl/writer.rs
Gordon-F: All required expressions here are `Expression::AccessIndex` and we need to find a global associated with this expression.

src/back/wgsl/writer.rs
Gordon-F: TIL. Thank you! But here and in other places with the same logic, we also add spaces between arguments.

src/back/wgsl/writer.rs
kvark: do we expect these access chains to be more than 1 length?

kvark: This handling of `Compose` looks really strange to me. I think we'd need to rewrite it entirely.
It's currently looking specifically for a global variable that is a struct in Private storage class with this member...
Here is what we can do instead:
 
Check the output type. If it's a struct, check the binding on the member of the struct. If the binding is an unsupported builtin, skip the field.

This logic should be much simpler and more general at the same time.

src/back/wgsl/writer.rs
```diff
@@ -530,15 +688,74 @@ impl<W: Write> Writer<W> {
...
+                                    if let Some(Binding::BuiltIn(builtin)) =
+                                        members[access_index as usize].binding
+                                    {
+                                        if builtin_str(builtin).is_none() {
+                                            // glslang why you did this with us...
+                                            log::warn!(
+                                                "Skip component with unsupported builtin {:?}",
+                                                builtin
+                                            );
+                                            skip_component = true;
```
kvark, line 735: this can be just `continue`, right?

kvark (approved):
Thank you!
So we no longer need to use the `assignable_global` thing?

src/back/wgsl/writer.rs
Gordon-F: Unfortunately, no. We need to go through all this if chain to be sure that we find right struct and it member.

> So we no longer need to use the assignable_global thing?

Yes! I removed this changes from PR.

Issue #786, opened 2021-04-26
[spv-in] Handle `GLOp::Radians` and `GLOp::Degrees`


expenses:
I'm seeing these extended instructions in a quite a few shadertoy shaders, so it'd be nice to handle them. I don't have a preference as to whether we add a `MathFunction` to the IR or we emit multiplications instead.

kvark:
let's emit the multiplications (for now, at least)

expenses:
I guess we want to do something like this? I'm struggling with `Expression [13] can't be introduced - it's already in scope` . though.
```diff
diff --git a/src/front/spv/function.rs b/src/front/spv/function.rs
index a8e97ba..c3fae71 100644
--- a/src/front/spv/function.rs
+++ b/src/front/spv/function.rs
@@ -149,8 +149,8 @@ impl<I: Iterator<Item = u32>> super::Parser<I> {
                         fun_id,
                         &mut fun.expressions,
                         &mut fun.local_variables,
+                        &mut module.constants,
                         &module.types,
-                        &module.constants,
                         &module.global_variables,
                     )?;
 
diff --git a/src/front/spv/mod.rs b/src/front/spv/mod.rs
index ccea724..2a5d08a 100644
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -772,8 +772,8 @@ impl<I: Iterator<Item = u32>> Parser<I> {
         function_id: spirv::Word,
         expressions: &mut Arena<crate::Expression>,
         local_arena: &mut Arena<crate::LocalVariable>,
+        const_arena: &mut Arena<crate::Constant>,
         type_arena: &Arena<crate::Type>,
-        const_arena: &Arena<crate::Constant>,
         global_arena: &Arena<crate::GlobalVariable>,
     ) -> Result<ControlFlowNode, Error> {
         let mut block = Vec::new();
@@ -1695,84 +1695,132 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                     }
                     let inst_id = self.next()?;
                     let gl_op = Glo::from_u32(inst_id).ok_or(Error::UnsupportedExtInst(inst_id))?;
-                    let fun = match gl_op {
-                        Glo::Round => Mf::Round,
-                        Glo::Trunc => Mf::Trunc,
-                        Glo::FAbs | Glo::SAbs => Mf::Abs,
-                        Glo::FSign | Glo::SSign => Mf::Sign,
-                        Glo::Floor => Mf::Floor,
[...]

PR #797, opened 2021-04-28
[spv-in] Impl `GLOp::Radians` and `GLOp::Degrees`

expenses:
This adds a constant for `PI / 180` as suggested in https://github.com/gfx-rs/naga/issues/786#issuecomment-827092368.

As a test:
```glsl
#version 450

void main() {
    float deg = 15.0;
    float rad = radians(deg);
    float deg_again = degrees(rad);
    colour = vec4(vec3(deg_again), 1.0);
}
```
->
```metal
...
void main1(
    thread metal::float4& colour
) {
    float deg;
    float rad;
    float deg_again;
    deg = 15.0;
    rad = deg * 0.017453292519943295;
    deg_again = rad / 0.017453292519943295;
    metal::float3 _e16 = float3(deg_again);
    colour = metal::float4(_e16.x, _e16.y, _e16.z, 1.0);
    return;
}
...
```

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -1707,29 +1707,19 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                             self.lookup_expression.lookup(arg_id)?.handle
                         };
 
-                        let constant_expr = self.maths_constant_expressions[0];
-
-                        self.lookup_expression.insert(
-                            result_id,
-                            LookupExpression {
-                                handle: constant_expr,
-                                type_id: result_type_id,
-                            },
-                        );
-
-                        let op = match gl_op {
-                            Glo::Radians => crate::BinaryOperator::Multiply,
-                            Glo::Degrees => crate::BinaryOperator::Divide,
-                            _ => unreachable!(),
-                        };
+                        let pi_div_180 = self.maths_constant_expressions[0];
 
                         self.lookup_expression.insert(
                             result_id,
                             LookupExpression {
                                 handle: expressions.append(crate::Expression::Binary {
-                                    op,
+                                    op: match gl_op {
+                                        Glo::Radians => crate::BinaryOperator::Multiply,
+                                        Glo::Degrees => crate::BinaryOperator::Divide,
+                                        _ => unreachable!(),
+                                    },
                                     left: arg,
-                                    right: constant_expr,
+                                    right: pi_div_180,
                                 }),
                                 type_id: result_type_id,
                             },
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -2136,7 +2136,7 @@ impl<I: Iterator<Item = u32>> Parser<I> {
             self.index_constants.push(handle);
         }
 
-        // register radians -> degrees constants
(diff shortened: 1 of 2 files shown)


expenses:
There might be a _slight_ speedup in the resulting shader if we multiply by `180 / PI` instead of dividing by `PI / 180` for degrees.

src/front/spv/mod.rs
```diff
@@ -1695,84 +1699,111 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
-                        let arg_id = self.next()?;
-                        Some(self.lookup_expression.lookup(arg_id)?.handle)
+                    if gl_op == Glo::Radians || gl_op == Glo::Degrees {
+                        inst.expect(base_wc + 1)?;
+                        let arg = {
+                            let arg_id = self.next()?;
+                            self.lookup_expression.lookup(arg_id)?.handle
+                        };
+
+                        let pi_div_180 = self.maths_constant_expressions[0];
```
kvark, line 1710: oh, my apologies. I didn't realize that baking the constant to `self.math_constant_expressions` would not make this code simpler.

Since we'll have to have this complication anyway, we might as well create this constant on demand.
Would you want to try doing this? I imagine we'd do something like `module.constants.fetch_or_append(...)`.

expenses: Yeah so the problem I get with doing this is that I get this
```
        Expression [12] can't be introduced - it's already in scope
```
error and I'm not quite sure what to do about it.

expenses: Perhaps we need to insert the constant expression into `lookup_expression`? But I don't know what key to use for that.

expenses: Oh - I'm starting to see the problem. Because we're adding an extra expression before the multiply, `result_id` no longer matches the expression handle. The best way to fix this might be to go back to what we were doing before.

src/front/spv/mod.rs
```diff
@@ -1695,84 +1695,121 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+                        };
+
+                        let constant_handle = const_arena.fetch_or_append(crate::Constant {
+                            name: None,
+                            specialization: None,
+                            inner: crate::ConstantInner::Scalar {
+                                width: 4,
+                                value: crate::ScalarValue::Float(match gl_op {
+                                    Glo::Radians => std::f64::consts::PI / 180.0,
+                                    Glo::Degrees => 180.0 / std::f64::consts::PI / 180.0,
```
kvark, line 1713: something is not right here: 180 / x / 180

kvark (approved):

expenses:
Oh sorry, I should have clarified; this doesn't work at the moment on the spir-v file corresponding to the above glsl shader:
```
        Expression [12] can't be introduced - it's already in scope
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', bin/naga.rs:219:31
```

kvark:
I suspect it's referring to the `Expression::Constant`, which is *not* expected to be emitted (i.e. in any emit range of `Statement::Emit`).

PR #803, opened 2021-04-29
Add several additional ops to the spirv frontend.

lachlansneff:
This PR adds support for the `Int8`, `Int16`, and `Int64` capabilities to the spirv frontend, as well as support for `OpArrayLength`, `OpCopyMemory`, `OpInBoundsAccessChain`, `OpLogicalNotEqual`, and `OpLogicalEqual`.

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -1987,6 +1987,61 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                 Op::Fwidth | Op::FwidthFine | Op::FwidthCoarse => {
                     self.parse_expr_derivative(expressions, crate::DerivativeAxis::Width)?;
                 }
+                Op::ArrayLength => {
+                    inst.expect(5)?;
+                    let result_type_id = self.next()?;
+                    let result_id = self.next()?;
+                    let structure_id = self.next()?;
+                    let member_index = self.next()?;
+
+                    // Validate that all the types are correct.
+                    let result_type = self.lookup_type.lookup(result_type_id)?;
+                    match type_arena[result_type.handle].inner {
+                        crate::TypeInner::Scalar {
+                            kind: crate::ScalarKind::Uint,
+                            width: 4,
+                        } => {}
+                        _ => return Err(Error::InvalidParameter(Op::ArrayLength)),
+                    }
+                    let structure = self.lookup_expression.lookup(structure_id)?;
+                    let structure_type = self.lookup_type.lookup(structure.type_id)?;
+                    match type_arena[structure_type.handle].inner {
+                        crate::TypeInner::Pointer { base, .. } => {
+                            if let crate::TypeInner::Struct { ref members, .. } = type_arena[base].inner {
+                                if let Some(last) = members.last() {
+                                    match type_arena[last.ty].inner {
+                                        crate::TypeInner::Array { size: crate::ArraySize::Dynamic, .. } => if members.len() - 1 == member_index as _ {},
+                                        _ => return Err(Error::InvalidParameter(Op::ArrayLength)),
+                                    }
+                                } else {
+                                    return Err(Error::InvalidParameter(Op::ArrayLength))
+                                }
+                            } else {
+                                return Err(Error::InvalidParameter(Op::ArrayLength))
+                            }
+                        }
+                        _ => return Err(Error::InvalidParameter(Op::ArrayLength)),
+                    }
+
+                    let structure_expr = expressions.append(crate::Expression::Load {
+                        pointer: structure.handle,
+                    });
+
+                    let array = expressions.append(crate::Expression::AccessIndex {
+                        base: structure_expr,
+                        index: member_index,
+                    });
+
+                    let length = expressions.append(crate::Expression::ArrayLength(array));
+
+                    self.lookup_expression.insert(
+                        result_id,
+                        LookupExpression {
+                            handle: length,
+                            type_id: result_type_id,
+                        },
+                    );
+                }
                 _ => return Err(Error::UnsupportedInstruction(self.state, inst.op)),
             }
         };
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -1620,7 +1620,9 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                 | Op::ConvertSToF
                 | Op::ConvertUToF
                 | Op::ConvertFToU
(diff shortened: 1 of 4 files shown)


src/front/spv/mod.rs
```diff
@@ -1984,6 +1990,90 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+                        _ => return Err(Error::InvalidParameter(Op::ArrayLength)),
+                    }
+                    let structure = self.lookup_expression.lookup(structure_id)?;
+                    let structure_type = self.lookup_type.lookup(structure.type_id)?;
+                    match type_arena[structure_type.handle].inner {
+                        crate::TypeInner::Pointer { base, .. } => {
+                            if let crate::TypeInner::Struct { ref members, .. } = type_arena[base].inner {
+                                if let Some(last) = members.last() {
+                                    match type_arena[last.ty].inner {
+                                        crate::TypeInner::Array { size: crate::ArraySize::Dynamic, .. } => if members.len() - 1 == member_index as _ {},
```
kvark, line 2016: I don't think we need to be that detailed here. Let the validation catch what it can. So here, let's instead use the member index directly:
```rust
match type_arena[members[member_index as u32].ty].inner {
}
```

src/front/spv/mod.rs
```diff
@@ -1984,6 +1990,90 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+                                    return Err(Error::InvalidParameter(Op::ArrayLength))
+                                }
+                            } else {
+                                return Err(Error::InvalidParameter(Op::ArrayLength))
+                            }
+                        }
+                        _ => return Err(Error::InvalidParameter(Op::ArrayLength)),
+                    }
+
+                    let structure_expr = expressions.append(crate::Expression::Load {
```
kvark, line 2029: This is incorrect. We shouldn't load anything for determining the array length. It should be:
`ArrayLength -> Access -> GlobalVariable`

src/front/spv/mod.rs
```diff
@@ -1984,6 +1990,90 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+                    };
+
+                    block.extend(emitter.finish(expressions));
+
+                    // TODO: check if the source and target types are the same?
+                    let target = self.lookup_expression.lookup(target_id)?;
+                    let source = self.lookup_expression.lookup(source_id)?;
+                    
+                    // This operation is practically the same as loading and then storing, I think.
+                    let value_expr = expressions.append(crate::Expression::Load {
```
kvark, line 2067: this should be done before `emitter.finish(expressions)`, not after

src/front/spv/mod.rs
```diff
@@ -3105,25 +3181,55 @@ impl<I: Iterator<Item = u32>> Parser<I> {
             }
         }
 
-        let inner = match module.types[ty].inner {
-            crate::TypeInner::Scalar { kind, width } => make_scalar_inner(kind, width),
-            crate::TypeInner::Vector { size, kind, width } => {
+        let inner = match inner {
```
kvark, line 3184: let's match on `*inner` instead of taking references in the pattern match

kvark (commented):
This is great!
A few things to fix that I spotted.

src/front/spv/mod.rs
lachlansneff: This actually touches on something I'm wondering: should naga never panic, no matter what input it gets?

src/front/spv/mod.rs
lachlansneff: Does using `Expression::Access` on a pointer to a structure get a pointer to the field?

lachlansneff: Since `OpArrayLength` in spirv is given a pointer to the structure, maybe it makes sense to stick with that convention throughout naga?

kvark: > Does using Expression::Access on a pointer to a structure get a pointer to the field?

Yes.

> Since OpArrayLength in spirv is given a pointer to the structure, maybe it makes sense to stick with that convention throughout naga?

Thinking about this... I'm not seeing any issues with going this way. The concerns were 1) implementations complexity, 2) ability to have it on fixed-size arrays.

So for (1) we haven't implemented it on HLSL/MSL, but it seems like that code is not going to be any different. WGSL in/out would be a little bit more complex, SPIR-V in/out would be less complex. So all in all it's the same.

For (2), the frontend can just see the fixed-size array, if it needs to handle it, and spew `Expression::Constant` instead of `Expression::ArrayLength`.

So TL:DR; it seems fine to proceed with the SPIRV semantics here.

src/front/spv/mod.rs
kvark: Naga backends are generally free to panic on the IR that's not valid. However, even there we'd want to issue `Error::Validation` by the backend, because this condition may be a result of some miscommunication between the validator and the backend, so these conditions may still arise, and we still don't want to panic there.

Naga frontends, however, shouldn't panic, because we have no guarantee or expectation that there was any validation before us. That is to say, the frontends aren't expected to fully validate the source files, it's not our job.

src/front/spv/mod.rs
lachlansneff: I decided that I'd prefer if `Expression::ArrayLength` is passed a pointer to the array itself.

src/front/spv/mod.rs
```diff
@@ -1984,6 +1990,72 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
                     self.parse_expr_derivative(expressions, crate::DerivativeAxis::Width)?;
                 }
+                Op::ArrayLength => {
+                    inst.expect(5)?;
+                    let result_type_id = self.next()?;
+                    let result_id = self.next()?;
+                    let structure_id = self.next()?;
+                    let member_index = self.next()?;
+
+                    // Validate that all the types are correct.
```
kvark, line 2000: I don't think we need this, necessarily. If type expectations don't match, Validator should yell.
And if we want to check the result type ID, we'd probably do it in a more generic way for all expressions.

src/lib.rs
```diff
@@ -813,6 +813,10 @@ pub enum Expression {
     /// Result of calling another function.
     Call(Handle<Function>),
     /// Get the length of an array.
+    /// The expression must resolve to a pointer to an array with a dynamic size.
```
kvark, line 816: sweeet

kvark (commented):
Last few things!

PR #805, opened 2021-04-30
Add support for arrayLength to the wgsl frontend

lachlansneff:
This PR adds `arrayLength` support to the WGSL frontend.

PR #806, opened 2021-04-30
Add support for retrieving array length in msl backend

lachlansneff:
This PR adds support for `Expression::ArrayLength` to the msl backend.

The snapshot in this PR depends on #805.


bin/naga.rs
```diff
@@ -77,6 +77,10 @@ fn main() {
                         panic!("Unknown profile: {}", string)
                     };
                 }
+                #[cfg(feature = "msl-out")]
+                "sizes-buffer-slot" => {
```
kvark, line 81: I don't think this is useful right now. I.e. there aren't any command line arguments for the resource binding map.

src/back/msl/mod.rs
```diff
@@ -157,6 +157,10 @@ pub struct Options {
     pub spirv_cross_compatibility: bool,
     /// Don't panic on missing bindings, instead generate invalid MSL.
     pub fake_missing_bindings: bool,
+    /// The slot of a buffer that contains an array of `u32`,
+    /// one for the size of each bound buffer that contains a runtime array,
+    /// in order of declaration in the module.
```
kvark, line 162: let's say that it's the order of `GlobalVariable` declarations specifically

src/back/msl/writer.rs
```diff
@@ -1,10 +1,10 @@
 use super::{
     keywords::RESERVED, sampler as sm, Error, LocationMode, Options, PipelineOptions,
-    TranslationInfo,
+    ResolvedBinding, TranslationInfo,
 };
 use crate::{
     arena::{Arena, Handle},
-    back::vector_size_str,
+    back::{msl::BindTarget, vector_size_str},
```
kvark, line 7: The `BindTarget` should be included in the `use::super` section instead of here

src/back/msl/writer.rs
```diff
@@ -1061,25 +1063,37 @@ impl<W: Write> Writer<W> {
...
-                }
+                crate::TypeInner::Pointer { base, .. } => match context.module.types[base].inner {
+                    crate::TypeInner::Array { stride, .. } => {
+                        let handle = context.info[expr].assignable_global.unwrap();
+
+                        let global = &context.module.global_variables[handle];
+                        if let crate::TypeInner::Struct { ref members, .. } = context.module.types[global.ty].inner {
+                            if let Some(&crate::StructMember { offset, .. }) = members.last() {
+                                let buffer_idx = self.runtime_sized_buffers[&handle];
+                                write!(self.out, "((x_buffer_sizes.buffer_size{} - {}) / {})", buffer_idx, offset, stride)?;
```
kvark, line 1074: This may be a little bit more tricky. The stride can be larger than the size of the type. So we may end up in a situation where there is space for N+1 elements, but dividing by the stride will give you N.
You can get the span occupied by this type via `context.module.types[member.y].inner.span(&module.constants)`.

One of the ways to write the proper equation would be :
```
write!(self.out, "(1 + (x_buffer_sizes.buffer_size{} - {} - {}) / {})", buffer_idx, offset, type_span, stride)?;
```
(where the `type_span` is the span for `member.ty`).
Note: this array is guaranteed to have at least one element.

src/back/msl/writer.rs
```diff
@@ -1061,25 +1063,37 @@ impl<W: Write> Writer<W> {
...
+                            }
+                        } else {
+                            return Err(Error::Validation)
+                        }
+                    }
+                    _ => return Err(Error::Validation),
+                },
                 _ => return Err(Error::Validation),
             },
+            // crate::Expression::ArrayLength(expr) => match context.function.expressions[expr] {
```
kvark, line 1086: any reason to keep it here?

src/back/msl/writer.rs
```diff
@@ -1210,6 +1224,7 @@ impl<W: Write> Writer<W> {
         level: Level,
         statements: &[crate::Statement],
         context: &StatementContext,
+        has_buffer_sizes: bool,
```
kvark, line 1227: let's put this as a field in `StatementContext`

src/back/msl/writer.rs
```diff
@@ -1437,6 +1460,29 @@ impl<W: Write> Writer<W> {
...
+            for (handle, gv) in module.global_variables.iter() {
+                if let crate::TypeInner::Struct { ref members, .. } = module.types[gv.ty].inner {
+                    if let Some(member) = members.last() {
+                        if let crate::TypeInner::Array {
+                            size: crate::ArraySize::Dynamic,
+                            ..
+                        } = module.types[member.ty].inner
+                        {
+                            let idx = self.runtime_sized_buffers.len();
+                            writeln!(self.out, "{}metal::uint buffer_size{};", INDENT, idx)?;
```
kvark, line 1475: let's use the `NAMESPACE` constant here for `metal`

src/back/msl/writer.rs
```diff
@@ -1767,6 +1816,15 @@ impl<W: Write> Writer<W> {
                 tyvar.try_fmt(&mut self.out)?;
                 writeln!(self.out, "{}", separator)?;
             }
+
+            if options.sizes_buffer_binding.is_some() {
+                writeln!(
+                    self.out,
+                    "{}constant mslBufferSizes& x_buffer_sizes",
```
kvark, line 1823: why is the `x_` prefix?

tests/in/arraylength.wgsl
```diff
@@ -0,0 +1,12 @@
+[[block]]
```
kvark, line 1: I don't think we want to have a separate test file for each operation.
Would it be possible to make this a part of the `access.wgsl` instead?

kvark (commented):
Looks great! Got a few things we should touch before landing.

tests/snapshots.rs
```diff
@@ -154,7 +154,7 @@ fn check_output_msl(
     let options = &params.msl;
     #[cfg(not(feature = "deserialize"))]
     let options = if params.msl_custom {
-        println!("Skipping {}", destination);
+        println!("Skipping {}", destination.display());
```
kvark, line 157: why is this calling `display()`?

src/back/msl/writer.rs
```diff
@@ -1061,25 +1063,37 @@ impl<W: Write> Writer<W> {
...
-                    write!(self.out, "{}", coco)?;
-                }
-                crate::TypeInner::Array { .. } => {
-                    return Err(Error::FeatureNotImplemented(
-                        "dynamic array size".to_string(),
-                    ))
-                }
+                crate::TypeInner::Pointer { base, .. } => match context.module.types[base].inner {
+                    crate::TypeInner::Array { stride, .. } => {
+                        let handle = context.info[expr].assignable_global.unwrap();
```
kvark, line 1068: Not sure if exposing the assignable global (in public!) is justified for this case.

Perhaps, we could go another way here. Instead of asking about the type of `expr` and traversing the types from there, we could first match the expression itself, then reach to `Expression::GlobalVariable`, and that gives us both the type handle (to match for a struct) and the lookup key into `runtime_sized_buffers`.

How do you feel about trying this out? This would keep `assignable_global` private, and it seems conceptually more straightforward.

tests/snapshots.rs
lachlansneff: A path can't be printed without called `.display()`. I'm not sure how this worked originally.

src/back/msl/writer.rs
kvark: just start it with an underscore then. We use "_eXxx"  for expressions, for example

tests/snapshots.rs
kvark: But... `{}` *is* supposed to call `display`. What am I missing here?

lachlansneff: `{}` calls `<T as Display>::fmt`. Paths don't implement that directly because they can contain invalid unicode, so you have to call `.display()` to convert them into something that can be printed.

kvark: ok, so I'm still unsure how this worked :)
that is not a blocker for this to land

src/back/msl/writer.rs
```diff
@@ -1437,10 +1475,49 @@ impl<W: Write> Writer<W> {
         writeln!(self.out, "#include <simd/simd.h>")?;
         writeln!(self.out)?;
 
+        let uses_unsized_buffers = {
```
kvark, line 1478: Instead of having `uses_unsized_buffers`, we can just check `if !self.runtime_sized_buffers.is_empty()` in the other places.
In addition, we should be clearing `runtime_sized_buffers` here, for the case where the same writer is re-used for different modules.

tests/in/access.wgsl
```diff
@@ -1,7 +1,17 @@
...
+struct Bar {
+    data: [[stride(4)]] array<u32>;
+};
+
+[[group(0), binding(0)]]
+var<storage> bar: [[access(read_write)]] Bar;
+
 [[stage(vertex)]]
 fn foo([[builtin(vertex_index)]] vi: u32) -> [[builtin(position)]] vec4<f32> {
+	bar.data[arrayLength(&bar.data) - 1u] = 0u; // Comment this out!
```
kvark, line 13: comment needs to be removed?

kvark (commented):
Looks great! There is a few more things to polish here, and to make the tests pass.

kvark (approved):

PR #810, opened 2021-05-02
Error improvements in WGSL-in and SPV-in

kvark:
SPV-in's error didn't derive `Error` and wasn't even publicly exposed.
WGSL's error had a lifetime, which makes it unable to form error chains, see https://github.com/dtolnay/thiserror/issues/68
This PR is largely made to address https://github.com/gfx-rs/wgpu/issues/1359

### DIFF
--- a/bin/naga.rs
+++ b/bin/naga.rs
@@ -118,7 +118,7 @@ fn main() {
             match result {
                 Ok(v) => v,
                 Err(ref e) => {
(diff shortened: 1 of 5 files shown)


grovesNL (approved):
Passing `source` again in the public API feels a little strange because the spans we stored depend on the source remaining the same. e.g. technically the spans could be pointing to invalid offsets if a different `source` is provided

PR #816, opened 2021-05-03
Fix struct layouts for WGSL and SPIRV

### LINKED ISSUE #811: False positive: wrong array stride
Trying to compile the following compute shader works in 0.7 but fails in 0.8:

```glsl
#version 450
layout(local_size_x = 64) in;

struct Foo {
    uint lo;
    uint hi;
};

layout(set = 0, binding = 0) buffer FooBar {
    Foo[] foo;
};

void main() {
}
```

with:

```
wgpu error: Validation Error

Caused by:
      note: label = `foo.spv`
    Type [4] '' is invalid
    Array stride 8 is smaller than the base element size 16
```

kvark:
Fixes #811

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -191,6 +191,28 @@ impl crate::ImageDimension {
     }
 }
 
+//TODO: should this be shared with `crate::proc::wgsl::layout` logic?
+fn get_alignment(ty: Handle<crate::Type>, arena: &Arena<crate::Type>) -> crate::Alignment {
+    use crate::TypeInner as Ti;
+    match arena[ty].inner {
+        Ti::Scalar { width, .. } => crate::Alignment::new(width as u32).unwrap(),
+        Ti::Vector { size, width, .. }
+        | Ti::Matrix {
+            rows: size, width, ..
+        } => {
+            let count = if size >= crate::VectorSize::Tri { 4 } else { 2 };
+            crate::Alignment::new((count * width) as u32).unwrap()
+        }
+        Ti::Pointer { .. } | Ti::ValuePointer { .. } => crate::Alignment::new(1).unwrap(),
+        Ti::Array { stride, .. } => crate::Alignment::new(stride).unwrap(),
+        Ti::Struct { ref level, .. } => match *level {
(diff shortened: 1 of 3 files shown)


Napokue (approved):
LGTM, thanks!

src/front/spv/mod.rs
```diff
@@ -2843,17 +2875,7 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
-                None => match members.last() {
-                    Some(member) => {
-                        //TODO: is this needed?
-                        // If offsets are required, we can just put 0 here
-                        member.offset + module.types[member.ty].inner.span(&module.constants)
-                    }
-                    None => 0,
-                },
-            };
+            let offset = decor.offset.unwrap_or(0); //unexpected
```
jimblandy, line 2878: I think the SPIR-V spec language that covers this is:

> Composite objects in the StorageBuffer, PhysicalStorageBuffer, Uniform, and PushConstant Storage Classes must be explicitly laid out. The following apply to all the aggregate and matrix types describing such an object, recursively through their nested types:
> – Each structure-type member must have an Offset decoration.

It seems like this wouldn't apply to structs used for local variables, say. Does that matter here?

jimblandy (approved):
Looks good to me.

src/front/spv/mod.rs
```diff
@@ -191,6 +191,37 @@ impl crate::ImageDimension {
     }
 }
 
+//TODO: should this be shared with `crate::proc::wgsl::layout` logic?
```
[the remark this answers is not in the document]
kvark: Right, but this is SPV frontend, it technically doesn't need to obey to any WGSL rules.
So it's a deal between SPV-in and IR, which can be different... but it's not currently

src/front/spv/mod.rs
kvark: yeah, I'm changing this to an error now. Just was lazy and didn't want to panic there.

kvark: Actually, no, we still need to support the case where offsets are not present there: for interface structs (in the `Output` storage class, for example).

jimblandy (commented):
Looks good to me.

Issue #812, opened 2021-05-02
[glsl-out] Shader translation error: syntax_error, unexpected NEW_IDENTIFIER


agausmann:
Encountered while upgrading wgpu_glyph to wgpu version 0.8 (hecrj/wgpu_glyph#67), running examples on Ubuntu 20.10 with `gfx_backend_gl`.

## Logs

- `RUST_LOG=trace cargo run --example hello` (also remove the `features = ["cross"]` line from Cargo.toml) - [wgpu_glyph_naga.log](https://github.com/gfx-rs/naga/files/6410882/wgpu_glyph_naga.log) 
- `RUST_LOG=trace cargo run --example hello --features wgpu/cross` - [wgpu_glyph_spirv_cross.log](https://github.com/gfx-rs/naga/files/6410883/wgpu_glyph_spirv_cross.log)

<details>
<summary>Vertex shader generated by `spirv_cross`: (spoiler) </summary>

```
    #version 320 es

    layout(binding = 0, std140) uniform Transform
    {
        mat4 transform;
    } _90;

    layout(location = 0) in vec3 left_top;
    layout(location = 1) in vec2 right_bottom;
    layout(location = 0) out vec2 f_tex_pos;
    layout(location = 2) in vec2 tex_left_top;
    layout(location = 3) in vec2 tex_right_bottom;
    layout(location = 1) out vec4 f_color;
    layout(location = 4) in vec4 color;

    void main()
    {
        vec2 pos = vec2(0.0);
        float left = left_top.x;
        float right = right_bottom.x;
        float top = left_top.y;
        float bottom = right_bottom.y;
        switch (gl_VertexID)
        {
            case 0:
            {
                pos = vec2(left, top);
                f_tex_pos = tex_left_top;
                break;
            }
            case 1:
            {
                pos = vec2(right, top);
                f_tex_pos = vec2(tex_right_bottom.x, tex_left_top.y);
                break;
            }
            case 2:
            {
                pos = vec2(left, bottom);
                f_tex_pos = vec2(tex_left_top.x, tex_right_bottom.y);
                break;
            }
            case 3:
            {
                pos = vec2(right, bottom);
                f_tex_pos = tex_right_bottom;
                break;
            }
        }
        f_color = color;
        gl_Position = _90.transform * vec4(pos, left_top.z, 1.0);
    }
```
</details>

<details>
<summary>Vertex shader generated by `naga`: (spoiler)</summary>

```
    #version 320 es

    precision highp float;

[...]

PR #821, opened 2021-05-03
[glsl-out] Fix writing identifiers starting with gl_

### LINKED ISSUE #814: [glsl-out] Generated shader use identifiers starting with "gl_"
Found in https://github.com/gfx-rs/naga/issues/812

```bash
cat test.glsl.vert | glslangValidator --stdin -S vert
stdin
ERROR: 0:17: 'gl_VertexIndex' : identifiers starting with "gl_" are reserved 
```

Gordon-F:
Fix #814. One more fix for #812

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -363,7 +363,7 @@ impl<'a, W: Write> Writer<'a, W> {
 
         // Generate a map with names required to write the module
         let mut names = FastHashMap::default();
-        Namer::default().reset(module, keywords::RESERVED_KEYWORDS, &mut names);
+        Namer::default().reset(module, keywords::RESERVED_KEYWORDS, &["gl_"], &mut names);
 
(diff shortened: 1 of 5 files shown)


src/back/glsl/mod.rs
```diff
@@ -2122,6 +2117,17 @@ impl<'a, W: Write> Writer<'a, W> {
             uniforms,
         })
     }
+
+    /// Helper method used to produce GL specific unique name
+    /// GL identifiers starting with "gl_" are not allowed
+    fn get_name(&self, name_key: &NameKey) -> String {
```
kvark, line 2123: instead of handling it here, can we instead add support for "reserved prefixes" to `Namer` itself?

Gordon-F: Done!

kvark (approved):

PR #823, opened 2021-05-03
[spv-in] default output builtins

### LINKED ISSUE #820: Initialize output built-ins coming from SPIR-V
See downstream https://github.com/gfx-rs/wgpu/issues/1360

What happens here is that incoming SPIR-V has `gl_PerVertex` struct with stuff like `gl_ClipDistance` and `gl_Position` that doesn't get written ever. Thus, it's assumed to keep the defaults.
Our spv-in, however, copies stuff to/from private globals, where the defaults get erased.

kvark:
Fixes #820

### DIFF
--- a/src/front/spv/null.rs
+++ b/src/front/spv/null.rs
@@ -1,0 +1,149 @@
+use super::Error;
+use crate::arena::{Arena, Handle};
+
+fn make_scalar_inner(kind: crate::ScalarKind, width: crate::Bytes) -> crate::ConstantInner {
+    crate::ConstantInner::Scalar {
+        width,
+        value: match kind {
+            crate::ScalarKind::Uint => crate::ScalarValue::Uint(0),
+            crate::ScalarKind::Sint => crate::ScalarValue::Sint(0),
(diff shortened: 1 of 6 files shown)


src/front/spv/null.rs
```diff
@@ -0,0 +1,149 @@
...
+        },
+        Some(crate::BuiltIn::FragDepth) => crate::ConstantInner::Scalar {
+            value: crate::ScalarValue::Float(0.0),
+            width: 4,
+        },
+        Some(crate::BuiltIn::SampleMask) => crate::ConstantInner::Scalar {
+            value: crate::ScalarValue::Uint(!0),
+            width: 4,
+        },
+        //Some(crate::BuiltIn::ClipDistance) |
```
grovesNL, line 141: nit: leftover comment

grovesNL (approved):

src/front/spv/null.rs
kvark: this is actually by design, I'll add more wording

PR #825, opened 2021-05-03
Unify back::msl push constants map and sizes buffer

lachlansneff:
This allows the sizes buffer slots to be tightly packed.

### DIFF
--- a/src/back/msl/mod.rs
+++ b/src/back/msl/mod.rs
@@ -80,6 +80,32 @@ pub struct PushConstantsMap {
     pub cs_buffer: Option<Slot>,
 }
 
+#[derive(Clone, Debug, Default, Hash, Eq, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+pub struct PerStageResources {
+    #[cfg_attr(feature = "deserialize", serde(default))]
+    pub push_constant_buffer: Option<Slot>,
+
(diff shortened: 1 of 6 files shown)


src/back/msl/mod.rs
```diff
@@ -71,13 +71,27 @@ pub type BindingMap = std::collections::BTreeMap<BindSource, BindTarget>;
 #[derive(Clone, Debug, Default, Hash, Eq, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "serialize", derive(serde::Serialize))]
 #[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
-pub struct PushConstantsMap {
+pub struct PerStageResources {
```
kvark, line 74: excellent!

src/back/msl/mod.rs
```diff
@@ -282,6 +293,32 @@ impl Options {
             None => Err(EntryPointError::MissingPushConstants(stage)),
         }
     }
+
+    fn resolve_sizes_buffer(
+        &self,
+        stage: crate::ShaderStage,
+    ) -> Result<ResolvedBinding, EntryPointError> {
+        let slot = match stage {
+            crate::ShaderStage::Vertex => self.per_stage_map.vs.sizes_buffer,
```
kvark, line 302: if you feel fancy, you could implement `Index<crate::ShaderStage> for PerStageMap`

kvark (approved):
Wonderful!

src/back/msl/mod.rs
lachlansneff: That's a good idea, I'll do that.

PR #827, opened 2021-05-04
[spv-in] Fix radians/degrees parsing

expenses:
See https://github.com/gfx-rs/naga/pull/797#issuecomment-829284308.


src/front/spv/mod.rs
```diff
@@ -1731,29 +1738,23 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
-                                    _ => unreachable!(),
-                                }),
-                            },
-                        });
-
-                        let expr_handle =
-                            expressions.append(crate::Expression::Constant(constant_handle));
+                        let constant = match gl_op {
+                            Glo::Radians => {
+                                self.maths_constant_expressions[DEGREES_TO_RADIANS_CONST_INDEX]
```
kvark, line 1743: Let's freeze this until #817 lands (and then re-implement on top of it). After it, you'd just call `self.get_constant_scalar(..)` here, and you could remove `maths_constant_expressions` entirely

kvark (changes requested):

src/front/spv/mod.rs
kvark: this is merged now

expenses: I can only see a `get_constant_scalar` function in `back/spv`..

kvark: oh, sorry, I confused myself

src/front/spv/mod.rs
```diff
@@ -2222,6 +2228,29 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+            self.maths_constants
+                .push(module.constants.append(crate::Constant {
+                    name: None,
+                    specialization: None,
+                    inner: crate::ConstantInner::Scalar {
+                        width: 4,
+                        value: crate::ScalarValue::Float(std::f64::consts::PI / 180.0),
+                    },
+                }));
+            self.maths_constants
```
kvark, line 2243: given that this is a very narrow use case, can we get away with 1 constant instead of 2?

src/front/spv/mod.rs
```diff
@@ -424,6 +427,8 @@ pub struct Parser<I> {
     options: Options,
     index_constants: Vec<Handle<crate::Constant>>,
     index_constant_expressions: Vec<Handle<crate::Expression>>,
+    maths_constants: Vec<Handle<crate::Constant>>,
```
kvark, line 430: This is very similar to index constants, in spirit: in both cases we keep a hold of a constant handle, and in both cases there is an associated expression handle. We should probably merge them.

expenses:
Sorry for having this on-hold for so long! I'll try and get it done today.

src/front/spv/mod.rs
expenses: It's not a huge deal, but IMO the resulting shaders are going to be slightly nicer if we use 2 constants and multiply by both, rather than having 1 and choosing to multiply or divide.

src/front/spv/mod.rs
expenses: We could merge them, but this would make things a lot messier. The `index_constant_expressions` are currently indexed by the index of the vector component that is being extracted. Either we change that by an offset, or we change `DEGREES_TO_RADIANS_CONST_INDEX` and `RADIANS_TO_DEGREES_CONST_INDEX` to go after these 4 index constants. I'd prefer to keep things separate if possible.

kvark:
Needs a rebase

kvark:
Sorry, would you mind rebasing again?

expenses:
To be honestly, I'm not entirely sure why pull requests need to be rebased. As I'm not currently involved in naga or using wgpu, I'd rather you take over the PR and merge it.

kvark:
We can't merge it while there are merge conflicts, hence my request.

kvark:
I rewrote this to avoid any pre-work done, and added a test.

kvark (approved):

PR #828, opened 2021-05-04
[spv-in] Handle `Op::LogicalEqual`

expenses:
Looks like we just missed this one.
Needed for something like
```glsl
void main() {
    bool y = true;
    bool n = false;
    colour = vec4(vec3(y == n), 1.0);
}
```


kvark (approved):
Oh wow

PR #837, opened 2021-05-05
Add Capabilities bitset for validation

### LINKED ISSUE #835: Allow float64 by the validator
See https://github.com/gfx-rs/wgpu/issues/1362
Also related to #831

Basically, we need some sort of an `Options` struct for the validator. And it would control whether optional capabilities are allowed on the system (such as float64).

kvark:
Fixes #835


grovesNL (approved):
Nice!

PR #839, opened 2021-05-06
Support for width casting in the IR

kvark:
Follow-up to #837
Allows casting between f64 and f32, for example, which IR didn't support before.


jimblandy (commented):
Looks good to me.

PR #842, opened 2021-05-06
Barrier follow-up fixes

kvark:
Addressing Jim's review notes in #817


src/back/glsl/mod.rs
```diff
@@ -1393,12 +1393,11 @@ impl<'a, W: Write> Writer<'a, W> {
             // keyword which ceases all further processing in a fragment shader, it's called OpKill
             // in spir-v that's why it's called `Statement::Kill`
             Statement::Kill => writeln!(self.out, "{}discard;", INDENT.repeat(indent))?,
-            // Just issue a memory barrier, ignoring the flags.
+            // Issue am execution or a memory barrier.
```
jimblandy, line 1396: 'an'

jimblandy (approved):
Okay, this looks great.

Should we have a test case that exercises the control-only barrier case?

kvark:
>Should we have a test case that exercises the control-only barrier case?

WGSL doesn't appear to have that supported

PR #850, opened 2021-05-09
[msl] early out if the binding sizes are missing

kvark:
We can't just skip writing the rest of the entry point if we already started...
Follow-up to #825


kvark:
This is actually critical to run anything. gfx-24 tag isn't useful without this.

PR #857, opened 2021-05-12
[wgsl-out] More improvements

Gordon-F:
1. Implement all math functions (copy-paste from wgsl-in)
2. Rename entry point output structs if its name is none.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -61,6 +61,7 @@ pub struct Writer<W> {
     names: FastHashMap<NameKey, String>,
     namer: Namer,
     named_expressions: BitSet,
+    ep_results: Vec<(ShaderStage, Handle<Type>)>,
 }
 
 impl<W: Write> Writer<W> {
@@ -70,6 +71,7 @@ impl<W: Write> Writer<W> {
             names: FastHashMap::default(),
             namer: Namer::default(),
             named_expressions: BitSet::new(),
+            ep_results: vec![],
         }
     }
 
@@ -81,6 +83,13 @@ impl<W: Write> Writer<W> {
 
     pub fn write(&mut self, module: &Module, info: &ModuleInfo) -> BackendResult {
         self.reset(module);
+
+        // Save all ep result types
+        for (_, ep) in module.entry_points.iter().enumerate() {
+            if let Some(ref result) = ep.function.result {
+                self.ep_results.push((ep.stage, result.ty));
+            }
+        }
 
         // Write all structs
         for (handle, ty) in module.types.iter() {
@@ -173,6 +182,29 @@ impl<W: Write> Writer<W> {
         Ok(())
     }
 
+    /// Helper method used to write stuct name
+    ///
+    /// # Notes
+    /// Adds no trailing or leading whitespace
+    fn write_struct_name(&mut self, module: &Module, handle: Handle<Type>) -> BackendResult {
+        if module.types[handle].name.is_none() {
+            if let Some(&(stage, _)) = self.ep_results.iter().find(|&&(_, ty)| ty == handle) {
+                let name = match stage {
+                    ShaderStage::Compute => "ComputeOutput",
+                    ShaderStage::Fragment => "FragmentOutput",
+                    ShaderStage::Vertex => "VertexOutput",
+                };
+
+                write!(self.out, "{}", name)?;
+                return Ok(());
+            }
+        }
+
+        write!(self.out, "{}", self.names[&NameKey::Type(handle)])?;
+
+        Ok(())
+    }
+
     /// Helper method used to write structs
     /// https://gpuweb.github.io/gpuweb/wgsl/#functions
     ///
@@ -222,14 +254,11 @@ impl<W: Write> Writer<W> {
 
         // Write function return type
         if let Some(ref result) = func.result {
+            write!(self.out, " -> ")?;
             if let Some(ref binding) = result.binding {
-                write!(self.out, " -> ")?;
                 self.write_attributes(&map_binding_to_attribute(binding), true)?;
-                self.write_type(module, result.ty)?;
-            } else {
-                let struct_name = &self.names[&NameKey::Type(result.ty)].clone();
-                write!(self.out, " -> {}", struct_name)?;
             }
+            self.write_type(module, result.ty)?;
         }
 
         write!(self.out, " {{")?;
@@ -383,8 +412,10 @@ impl<W: Write> Writer<W> {
             self.write_attributes(&[Attribute::Block], false)?;
             writeln!(self.out)?;
         }
-        let name = &self.names[&NameKey::Type(handle)].clone();
-        write!(self.out, "struct {} {{", name)?;
+
+        write!(self.out, "struct ")?;
+        self.write_struct_name(module, handle)?;
+        write!(self.out, " {{")?;
         writeln!(self.out)?;
         for (index, member) in members.iter().enumerate() {
             // Skip struct member with unsupported built in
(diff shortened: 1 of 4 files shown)


src/back/wgsl/writer.rs
```diff
@@ -82,6 +84,13 @@ impl<W: Write> Writer<W> {
     pub fn write(&mut self, module: &Module, info: &ModuleInfo) -> BackendResult {
         self.reset(module);
 
+        // Save all ep result types
+        for (_, ep) in module.entry_points.iter().enumerate() {
+            if let Some(ref result) = ep.function.result {
+                self.ep_results.insert(result.ty, ep.stage);
```
kvark, line 90: what if two entry points of different stages happen to have the same types?

src/back/wgsl/writer.rs
```diff
@@ -185,7 +207,14 @@ impl<W: Write> Writer<W> {
         func_ctx: &FunctionCtx<'_>,
     ) -> BackendResult {
         let func_name = match func_ctx.ty {
-            FunctionType::EntryPoint(index) => self.names[&NameKey::EntryPoint(index)].clone(),
```
kvark, line 188: I don't understand, what was wrong with the original code?

src/back/wgsl/writer.rs
Gordon-F: Whoops. Thank you!

src/back/wgsl/writer.rs
Gordon-F: Sorry, I should had to describe why we need these changes first. It's all about readability (except math function).

Before:
```wgsl
fn main1([[location(1)]] a_uv1: vec2<f32>, [[location(0)]] a_pos1: vec2<f32>) -> type10 {
```

After:
```wgsl
fn vs_main([[location(1)]] a_uv1: vec2<f32>, [[location(0)]] a_pos1: vec2<f32>) -> VertexOutput {
```

Now it's easier to merge fragment and vertex shaders generated from spv into one wgsl file (otherwise all entry points functions named `main`).

grovesNL: Ideally I think we'd want to try to use the name of the entry point/types from SPIR-V instead of renaming it based on the stage.

There's also a risk if there are multiple entry points that use the same stage though I don't know if we currently allow that. e .g. multiple compute entry points in one file would collide

acowley: I use multiple compute entry points in single WGSL files quite a lot. I’d also like to retain the ability to go wgsl-in -> process -> wgsl-out.

Gordon-F: @grovesNL @acowley Thank you for your comments! I removed this part from PR.

src/back/wgsl/writer.rs
```diff
@@ -61,6 +61,7 @@ pub struct Writer<W> {
     names: FastHashMap<NameKey, String>,
     namer: Namer,
     named_expressions: BitSet,
+    ep_results: FastHashMap<ShaderStage, Handle<Type>>,
```
kvark, line 64: how is this supposed to work with multiple entry points? There may be different result types for multiple EPs inside the same module.

Gordon-F: Thank you! Fixed now.

src/back/wgsl/writer.rs
```diff
@@ -173,6 +182,31 @@ impl<W: Write> Writer<W> {
         Ok(())
     }
 
+    fn get_struct_name(&self, module: &Module, handle: Handle<Type>) -> String {
+        if module.types[handle].name.is_none() {
+            #[allow(clippy::pattern_type_mismatch)]
+            let shader_stage =
+                self.ep_results
+                    .iter()
+                    .find_map(|(stage, ty)| if *ty == handle { Some(stage) } else { None });
```
kvark, line 191: nit: this would be simpler with just `find()`

src/back/wgsl/writer.rs
```diff
@@ -173,6 +182,31 @@ impl<W: Write> Writer<W> {
         Ok(())
     }
 
+    fn get_struct_name(&self, module: &Module, handle: Handle<Type>) -> String {
```
kvark, line 185: can we return `&str` instead of `String`?

src/back/wgsl/writer.rs
```diff
@@ -173,6 +182,31 @@ impl<W: Write> Writer<W> {
...
+        if module.types[handle].name.is_none() {
+            #[allow(clippy::pattern_type_mismatch)]
+            let shader_stage =
+                self.ep_results
+                    .iter()
+                    .find_map(|(stage, ty)| if *ty == handle { Some(stage) } else { None });
+            if let Some(stage) = shader_stage {
+                match *stage {
```
kvark, line 193: we can move `return` out of the `match`

kvark (commented):
Almost there!

src/back/wgsl/writer.rs
Gordon-F: We have 2 self borrows here. I don't know how we can bit this error if we return `&str` here:

```bash
let name = self.get_struct_name(module, ty);
    |                            ---- immutable borrow occurs here
470 |                 write!(self.out, "{}", name)?;
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                 |
    |                 mutable borrow occurs here
    |                 immutable borrow later used by call
```

src/back/wgsl/writer.rs
Gordon-F: `find()` returns to us `(shader_stage, ty)` here, but we need only `shader_stage`.

kvark: right, so you'd do `find(...).0` for this

src/back/wgsl/writer.rs
kvark: is it trying to associate the `&Module` lifetime with the others? Let's make it not do this:
```rust
fn get_struct_name<'a>(&'a self, module: &Module, handle: Handle<Type>) -> &'a str {
```

kvark: Actually, I see what's going on there. Let's make this function insert the name into the output stream instead of returning it. I.e. `put_struct_name(&mut self, ..)`

Gordon-F: Thank you! Fixed.

src/back/wgsl/writer.rs
```diff
@@ -173,6 +182,30 @@ impl<W: Write> Writer<W> {
...
+    /// Helper method used to write stuct name
+    ///
+    /// # Notes
+    /// Adds no trailing or leading whitespace
+    fn write_struct_name(&mut self, module: &Module, handle: Handle<Type>) -> BackendResult {
+        if module.types[handle].name.is_none() {
+            #[allow(clippy::pattern_type_mismatch)]
+            let shader_stage = self.ep_results.iter().find(|(_, ty)| *ty == handle);
+            if let Some(stage) = shader_stage {
+                match stage.0 {
```
kvark, line 194: we should move things out of matches where possible, i.e. here it would be `let name = match { .. };` followed up by `write!(self.out, "{}", name)?;`

src/back/wgsl/writer.rs
```diff
@@ -173,6 +182,30 @@ impl<W: Write> Writer<W> {
         Ok(())
     }
 
+    /// Helper method used to write stuct name
+    ///
+    /// # Notes
+    /// Adds no trailing or leading whitespace
+    fn write_struct_name(&mut self, module: &Module, handle: Handle<Type>) -> BackendResult {
+        if module.types[handle].name.is_none() {
+            #[allow(clippy::pattern_type_mismatch)]
```
kvark, line 191: please no :)

src/back/wgsl/writer.rs
```diff
@@ -173,6 +182,30 @@ impl<W: Write> Writer<W> {
...
+            let shader_stage = self.ep_results.iter().find(|(_, ty)| *ty == handle);
```
kvark, line 192: let's do `let &(_, stage) = self.ep_results.`;

src/back/wgsl/writer.rs
Gordon-F: I just think it's a false positive clippy suggestion here. Sorry, my Rust-fu not so good :) Thanks a lot for your patience!

kvark (approved):
Nice, thank you!

PR #858, opened 2021-05-13
[msl] make sizes buffer local to each function

kvark:
Another (big) follow-up to #825
The original code was written in a way that doesn't consider entry points truly independent. We'd globally figure out if the sizes buffer is needed anywhere, and then blindly use it on all functions and entry points. This doesn't work if, say, the pipeline layout doesn't include some of the globals used by the program.
Instead, this PR determines, for each function, whether or not the sizes buffer is needed.

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -342,6 +342,21 @@ fn should_pack_struct_member(
     }
 }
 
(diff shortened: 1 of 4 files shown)


grovesNL (approved):
Looks great, thanks!

PR #859, opened 2021-05-13
TypeInner::span: Tolerate bad array length constants for now.

jimblandy:
If an array has a bogus length constant (a boolean, say), `TypeInner::span` will panic. However, the problem will be caught in validation, and this function is needed in front ends, which should never panic on invalid input.

This PR also moves `tests/errors.rs` to `tests/wgsl-errors.rs` and makes the entire file conditional on the `wgsl-in` feature, before adding some integration tests.


kvark (approved):
Good call!

PR #860, opened 2021-05-13
[spv-in] fix handling of break blocks (borrowed from Tint)

### LINKED ISSUE #856: InvalidEdgeClassification on collatz shader
https://github.com/linebender/piet-gpu/blob/master/piet-gpu-hal/examples/shader/collatz.spv

> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidEdgeClassification', bin/naga.rs:116:64

MatusT:
Fixes several things:
1. Construct's variables ```enclosing_*``` were not initialized. They are needed for proper classification.
2. Construct's depth was incorrectly initialized.
3. Construct's parent was incorrectly initialized.
2. Classification of break edges(Loop/Switch/If) was brought from Tint. Finalized from the previous pull request.
3. Add break only on LoopBreak.

Fixes #856

### DIFF
--- a/src/front/spv/flow.rs
+++ b/src/front/spv/flow.rs
@@ -222,14 +222,13 @@ impl FlowGraph {
                     begin_position,
                     end,
                     end_position,
+                    depth,
                     ..Default::default()
                 },
             );
 
-            let mut construct = constructs[node_index];
-
             let parent = {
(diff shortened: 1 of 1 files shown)


kvark (approved):
a bit of black magic here

PR #861, opened 2021-05-13
[wgsl-out] Enable snapshot testing for wgsl-in image test

Gordon-F:
Implement all statements and expressions required by image snapshot.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -6,8 +6,8 @@ use crate::{
     proc::{EntryPointIndex, NameKey, Namer, TypeResolution},
     valid::{FunctionInfo, ModuleInfo},
     Arena, ArraySize, Binding, Constant, Expression, FastHashMap, Function, GlobalVariable, Handle,
-    ImageClass, ImageDimension, Interpolation, Module, Sampling, ScalarKind, ScalarValue,
-    ShaderStage, Statement, StorageFormat, StructLevel, StructMember, Type, TypeInner,
+    ImageClass, ImageDimension, Interpolation, Module, SampleLevel, Sampling, ScalarKind,
+    ScalarValue, ShaderStage, Statement, StorageFormat, StructLevel, StructMember, Type, TypeInner,
 };
 use bit_set::BitSet;
 use std::fmt::Write;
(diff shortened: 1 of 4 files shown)


src/back/wgsl/writer.rs
```diff
@@ -843,32 +861,69 @@ impl<W: Write> Writer<W> {
...
-                expr,
-                kind,
-                convert: _, //TODO:
-            } => {
+            Expression::As { expr, kind, .. } => {
                 let inner = func_ctx.info[expr].ty.inner_with(&module.types);
                 let op = match *inner {
                     TypeInner::Matrix { columns, rows, .. } => {
-                        format!("mat{}x{}", vector_size_str(columns), vector_size_str(rows))
+                        format!(
```
kvark, line 909: let's write it out to the output instead of collecting into `String` first

kvark (commented):
A few fixes are needed, looks great overall!

src/back/wgsl/writer.rs
```diff
@@ -828,13 +847,12 @@ impl<W: Write> Writer<W> {
                 // TODO: comments
                 let fun_name = match level {
                     crate::SampleLevel::Auto => "textureSample",
-                    _ => {
-                        return Err(Error::Unimplemented(format!(
-                            "expression_imagesample_level {:?}",
-                            level
-                        )));
-                    }
+                    crate::SampleLevel::Zero => "textureSampleCompare",
```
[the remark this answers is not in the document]
Gordon-F: Thank you! Reworked this part. But I still not completely sure that I picked up right textures function here.

src/back/wgsl/writer.rs
```diff
@@ -819,56 +877,136 @@ impl<W: Write> Writer<W> {
...
-                // TODO: comments
-                let fun_name = match level {
-                    crate::SampleLevel::Auto => "textureSample",
-                    _ => {
-                        return Err(Error::Unimplemented(format!(
-                            "expression_imagesample_level {:?}",
-                            level
-                        )));
+                let texture_function = if let Some(depth_ref) = depth_ref {
+                    TextureFunction::SampleCompare(depth_ref)
```
kvark, line 886: is `TextureFunction` actually helping us here? I.e in this if/else/match we could just write the function name, and then later we could grab the expressions out of `level`

src/back/wgsl/writer.rs
```diff
@@ -819,56 +877,136 @@ impl<W: Write> Writer<W> {
...
                 self.write_expr(module, sampler, func_ctx)?;
                 write!(self.out, ", ")?;
                 self.write_expr(module, coordinate, func_ctx)?;
+
+                if let Some(array_index) = array_index {
+                    write!(self.out, ", ")?;
+                    self.write_expr(module, array_index, func_ctx)?;
+                }
+
+                if let TextureFunction::SampleBias(expr) = texture_function {
```
kvark, line 912: let's take this out of `level` instead: `if let crate::SampleLevel::Bias(expr) = level`.
Same for other cases

kvark (approved):
Great, thank you for patience!

PR #868, opened 2021-05-15
[wgsl-out] Implement all statements and expressions. Enable all wgsl-in tests

Gordon-F:
Draft because some snapshots don't pass validation.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -1,13 +1,12 @@
-// TODO: temp
-#![allow(dead_code)]
 use super::Error;
 use crate::{
     back::{binary_operation_str, vector_size_str, wgsl::keywords::RESERVED},
     proc::{EntryPointIndex, NameKey, Namer, TypeResolution},
     valid::{FunctionInfo, ModuleInfo},
-    Arena, ArraySize, Binding, Constant, Expression, FastHashMap, Function, GlobalVariable, Handle,
-    ImageClass, ImageDimension, Interpolation, Module, SampleLevel, Sampling, ScalarKind,
-    ScalarValue, ShaderStage, Statement, StorageFormat, StructMember, Type, TypeInner,
+    Arena, ArraySize, Binding, Constant, ConstantInner, Expression, FastHashMap, Function,
+    GlobalVariable, Handle, ImageClass, ImageDimension, Interpolation, Module, SampleLevel,
+    Sampling, ScalarKind, ScalarValue, ShaderStage, Statement, StorageClass, StorageFormat,
+    StructMember, Type, TypeInner,
 };
 use bit_set::BitSet;
 use std::fmt::Write;
@@ -107,7 +106,7 @@ impl<W: Write> Writer<W> {
         // Write all constants
         for (handle, constant) in module.constants.iter() {
             if constant.name.is_some() {
-                self.write_global_constant(&constant, handle)?;
+                self.write_global_constant(&module, &constant.inner, handle)?;
             }
         }
 
@@ -173,7 +172,7 @@ impl<W: Write> Writer<W> {
     fn write_scalar_value(&mut self, value: ScalarValue) -> BackendResult {
         match value {
             ScalarValue::Sint(value) => write!(self.out, "{}", value)?,
-            ScalarValue::Uint(value) => write!(self.out, "{}", value)?,
+            ScalarValue::Uint(value) => write!(self.out, "{}u", value)?,
             // Floats are written using `Debug` instead of `Display` because it always appends the
             // decimal part even it's zero
             ScalarValue::Float(value) => write!(self.out, "{:?}", value)?,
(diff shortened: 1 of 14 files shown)


src/back/wgsl/writer.rs
```diff
@@ -673,8 +672,96 @@ impl<W: Write> Writer<W> {
...
+            } => {
+                // Start the switch
+                write!(self.out, "{}", INDENT.repeat(indent))?;
+                write!(self.out, "switch(")?;
+                self.write_expr(module, selector, func_ctx)?;
+                writeln!(self.out, ") {{")?;
+
+                // Write all cases
+                let mut write_case = true;
+                if !cases.is_empty() {
```
kvark, line 698: this is strange, we shouldn't have this condition

src/back/wgsl/writer.rs
```diff
@@ -673,8 +672,96 @@ impl<W: Write> Writer<W> {
...
+                writeln!(self.out, "{}}}", INDENT.repeat(indent))?
+            }
+            Statement::Loop {
+                ref body,
+                ref continuing,
+            } => {
+                write!(self.out, "{}", INDENT.repeat(indent))?;
+                writeln!(self.out, "loop {{")?;
+
+                for sta in body.iter().chain(continuing.iter()) {
```
kvark, line 743: we are missing the `continuing:` stuff, we need to write it down

src/back/wgsl/writer.rs
```diff
@@ -1203,10 +1356,11 @@ impl<W: Write> Writer<W> {
     /// Ends in a newline
     fn write_global_constant(
         &mut self,
-        constant: &Constant,
+        module: &Module,
+        inner: &ConstantInner,
         handle: Handle<Constant>,
     ) -> BackendResult {
-        match constant.inner {
+        match *inner {
```
kvark, line 1363: We generally write scalar constants before the types, and composite constants after the types. This is needed because some array types may need to have the length of scalar constants, but the composite constants always need types to be defined.

src/back/wgsl/writer.rs
```diff
@@ -1343,6 +1512,16 @@ fn sampling_str(sampling: Sampling) -> &'static str {
     }
 }
 
+fn storage_class_str(storage_class: StorageClass) -> Option<&'static str> {
+    match storage_class {
+        StorageClass::Private => Some("private"),
+        StorageClass::Uniform => Some("uniform"),
+        StorageClass::Storage => Some("storage"),
+        StorageClass::PushConstant => Some("push_constant"),
+        _ => None,
```
kvark, line 1521: `WorkGroup` should also be `Some()`. That leaves only `Function` and `Handle` with `None`, so let's list them explictly

Gordon-F:
@kvark Last validation problem:
```wgsl
error: expected paren, found ']]'
   ┌─ wgsl:47:57
   │
47 │     vPos = particlesSrc.particles[global_invocation_id[0]].pos;
   │                                                         ^^ expected paren
```

Are WGSL allow indexing like this one?

kvark:
oooh, yeah. This is a WGSL parser bug: things like ">>" and "]]" need a bit of special treatment :(
You can work around by leaving a space between them: "] ]"

Gordon-F:
> You can work around by leaving a space between them: "] ]".

That's not a problem after we merged #871. All validation passed now 🎉

src/back/wgsl/writer.rs
Gordon-F: Here we write only global constants and writes types for scalar constants and composite.

src/back/wgsl/writer.rs
```diff
@@ -1128,8 +1222,70 @@ impl<W: Write> Writer<W> {
...
+                self.write_expr(module, reject, func_ctx)?;
+                write!(self.out, ", ")?;
+                self.write_expr(module, condition, func_ctx)?;
+                write!(self.out, ")")?
+            }
+            Expression::Derivative { axis, expr } => {
+                write!(
+                    self.out,
+                    "{}(",
+                    match axis {
```
kvark, line 1262: nit: let's do minimal work within macro invocations. Macros are always opaque to IDEs, so we should move the logic out. Here, we can match outside of the macro.

kvark (approved):
Looks great! I only got a few nits left

src/back/wgsl/writer.rs
Gordon-F: I remembered why we need such a strange condition because wgsl frontend allows writing switch without cases: 
```wgsl
fn main() {
    var pos: f32;
    switch (3) {
        default: { pos = 3.0; }
    }
}
```

src/back/wgsl/writer.rs
```diff
@@ -682,8 +681,102 @@ impl<W: Write> Writer<W> {
...
+                    }
+
+                    if case.fall_through {
+                        writeln!(self.out, "{}fallthrough;", INDENT.repeat(indent + 2))?;
+                    }
+                }
+
+                writeln!(self.out, "{}}}", INDENT.repeat(indent + 1))?;
+
+                if !default.is_empty() {
```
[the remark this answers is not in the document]
Gordon-F: Thank you for addressing this! Found another problem. Frontend allows writing switch like this:
```wgsl
fn main() {
    var pos: f32;
    switch (3) {
        case 0, 1, 2: { fallthrough; }
        default: { pos = 3.0; }
    }
}
```

PR #871, opened 2021-05-16
[wgsl-out] Create baking expression for texture function

Gordon-F:
Slightly improve readability of generated shaders


kvark (approved):
Very nice!

PR #873, opened 2021-05-17
Include span information in 'unknown identifier' errors.

jimblandy:


src/front/wgsl/mod.rs
```diff
@@ -175,6 +175,11 @@ impl<'a> Error<'a> {
                 labels: vec![(accessor_span.clone(), "invalid accessor".to_string())],
                 notes: vec![],
             },
+            Error::UnknownIdent(ref ident_span, ident) => ParseError {
+                message: format!("unknown identifier: '{}'", ident),
+                labels: vec![(ident_span.clone(), "unknown identifier".to_string())],
```
kvark, line 180: I wonder if the second element in this tuple should just be `&'static str`

kvark (approved):
Looks great!

src/front/wgsl/mod.rs
grovesNL: I think we want to be able to use `String` there too, e.g. we have some `format!`s above

kvark: there is quite a bit of repetition there between `message` and the second part of the label

PR #876, opened 2021-05-18
Builtins snapshot and a few fixes

### LINKED ISSUE #875: Problems with derived input variables in spv compute shaders
**Description**

When a compute shader in SPV format refers to gl_GlobalInvocationID, I get validation errors.

**Repro steps**

Apologies, I don't have this repo open-sourced yet, but hope to soon. In the meantime, this is a minimal compute shader that shows the issue. Try to create a shader module from this source:

```glsl
#version 450

layout(local_size_x = 16, local_size_y = 16) in;

layout(rgba8, set = 0, binding = 0) writeonly uniform image2D outputTex;

void main() {
    imageStore(outputTex, ivec2(gl_GlobalInvocationID.xy), vec4(1.0, 0.0, 0.0, 1.0));
    return;
}
```

I expect this to work, but I get this error:

```
wgpu error: Validation Error

Caused by:
    Function [1] 'main' is invalid
    Expression [15] is invalid
```

If I turn off ShaderFlags::VALIDATION, then I get this error instead:

```
wgpu error: Validation Error

Caused by:
    In Device::create_compute_pipeline
    unable to derive an implicit layout
    unable to reflect the shader COMPUTE interface
```

Note that if I remove  the reference to gl_GlobalInvocationID (storing to a hardcoded index), it runs fine.

**Extra materials**

Note: starting from wgsl source things work:

```wgsl
[[group(0), binding(0)]] var outputTex: [[access(write)]] texture_storage_2d<rgba8unorm>;

[[stage(compute), workgroup_size(16, 16)]]
fn main([[builtin(global_invocation_id)]] global_ix: vec3<u32>) {
    let rgba: vec4<f32> = vec4<f32>(1.0, 0.0, 0.0, 1.0);
    let write_ix = vec2<i32>(i32(global_ix.x), i32(global_ix.y));
    textureStore(outputTex, write_ix, rgba);
}
```

**Platform**
Information about your OS, version of `wgpu`, your tech stack, etc.
wgpu 0.8.1 (or 7486bdad, same errors), Windows 10, Vulkan

kvark:
Fixes #875
Related to #874

### DIFF
--- a/src/back/glsl/features.rs
+++ b/src/back/glsl/features.rs
@@ -97,7 +97,7 @@ impl FeaturesManager {
         // gl_ClipDistance is supported by core versions > 1.3 and aren't supported by an es versions without extensions
         check_feature!(CLIP_DISTANCE, 130, 300);
         check_feature!(CULL_DISTANCE, 450, 300);
-        check_feature!(SAMPLE_VARIABLES, 300, 300);
+        check_feature!(SAMPLE_VARIABLES, 400, 300);
 
         // Return an error if there are missing features
         if missing.is_empty() {
--- a/src/back/glsl/features.rs
+++ b/src/back/glsl/features.rs
@@ -30,6 +30,8 @@ bitflags::bitflags! {
         const SAMPLE_QUALIFIER = 1 << 12;
         const CLIP_DISTANCE = 1 << 13;
         const CULL_DISTANCE = 1 << 14;
+        // Sample ID
+        const SAMPLE_VARIABLES = 1 << 15;
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/back/glsl/features.rs
```diff
@@ -95,6 +97,7 @@ impl FeaturesManager {
         // gl_ClipDistance is supported by core versions > 1.3 and aren't supported by an es versions without extensions
         check_feature!(CLIP_DISTANCE, 130, 300);
         check_feature!(CULL_DISTANCE, 450, 300);
+        check_feature!(SAMPLE_VARIABLES, 300, 300);
```
Gordon-F, line 100: [gl_SampleID](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/gl_SampleID.xhtml) supported in GL 4+.

kvark:
Published in 0.4.2

PR #877, opened 2021-05-18
Use Cow more effectively.

jimblandy:
`Cow<'a, str>` implements `From<String>` and `From<&'a str>`, which makes creating them a bit nicer. And it seems like `Cow` is the right fit for the labels in `wgsl::ParseError`, which are sometimes fixed strings and sometimes formatted.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -25,6 +25,7 @@ use codespan_reporting::{
     },
 };
 use std::{
+    borrow::Cow,
     io::{self, Write},
     iter,
     num::NonZeroU32,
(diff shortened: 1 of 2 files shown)


jimblandy:
If you think it's valuable to readers to know that `Cow` is being used, then the `into` calls might be a bit Too Much Magic. My feeling is that using `Cow` for cases like this is common, and most Rust programmers will just recognize what's going on (or they will learn to soon enough).

src/front/wgsl/mod.rs
```diff
@@ -140,44 +141,44 @@ impl<'a> Error<'a> {
...
+                )],
                 notes: vec![],
             },
             Error::BadInteger(ref bad_span) => ParseError {
                 message: format!(
                     "expected integer literal, found `{}`",
                     &source[bad_span.clone()],
                 ),
-                labels: vec![(bad_span.clone(), "expected integer".to_string())],
+                labels: vec![(bad_span.clone(), "expected integer".to_string().into())],
```
kvark, line 155: why do we still need `to_string` here?

kvark (commented):
Looks good! Just one tiny thing

grovesNL (approved):
I think `Cow` would be great here, thanks!

src/front/wgsl/mod.rs
jimblandy: Oops - thanks!

PR #883, opened 2021-05-19
Flesh out the Naga module-level documentation.

jimblandy:


src/lib.rs
```diff
@@ -1,22 +1,101 @@
...
+by subsequent statements and their expressions.
+
+## `Expression` evaluation time and scope
+
+While the order of execution of [`Statement`]s is apparent from their structure,
+it is not so obvious exactly when a given [`Expression`] should be evaluated,
+since many `Statement`s and `Expression`s may refer to its value. But it is
+essential to clearly specify an expression's evaluation time, since that
+determines which statements' effects the expression should observe. It is also
+helpful to back ends to limit the visibility of an `Expression`'s value to a
```
kvark, line 55: same for "back ends"

src/lib.rs
```diff
@@ -1,22 +1,101 @@
...
+- The expression is an [`Expression::LocalVariable`] that is either the
+  `pointer` (destination) of a [`Statement::Store`], or initialized by some
+  previously executed `Statement::Store`.
+
+- The expression is the `result` of a [`Statement::Call`], representing the
+  call's return value. The call must be 'in scope' for the use (see below).
+
+- The expression is included in the range of some [`Statement::Emit`] that is
+  'in scope' for the use (see below). You can use the
+  [`Expression::needs_pre_emit`] method to determine whether an expression is in
```
kvark, line 73: it's the opposite: if something is "pre" emitted, it is not expected to show up with regular `Emit`

src/lib.rs
```diff
@@ -1,22 +1,101 @@
...
+helpful to back ends to limit the visibility of an `Expression`'s value to a
+portion of the statement tree.
+
+An `Expression` may only be used, whether by a `Statement` or another
+`Expression`, if one of the following is true:
+
+- The expression is an [`Expression::Constant`], [`Expression::FunctionArgument`], or
+  [`Expression::GlobalVariable`].
+
+- The expression is an [`Expression::LocalVariable`] that is either the
```
kvark, line 64: Maybe we should just say "the expression type is a pointer" here and nothing else?

src/lib.rs
```diff
@@ -1,22 +1,101 @@
...
+- The expression is included in the range of some [`Statement::Emit`] that is
+  'in scope' for the use (see below). You can use the
+  [`Expression::needs_pre_emit`] method to determine whether an expression is in
+  this category.
+
+When we say that a statement that establishes an expression's value, like a
+`Call` or `Emit`, is 'in scope' for some use of that expression, we mean that
+either:
+
+- the expression precedes the use in some `Emit` statement to which they both
```
kvark, line 80: I don't understand this bit. This says that an emitting statement is in scope for an expression if the expression precedes the `Emit`? Shouldn't this be the other way around?

src/lib.rs
```diff
@@ -1,22 +1,101 @@
...
+  belong, or
+
+- the statement that establishes the expression's value precedes some enclosing
+  statement of the use in a [`Statement::Block`] to which both statements
```
kvark, line 84: There should be nothing special about `Statement::Block`. Perhaps you meant just `Block`?

src/lib.rs
```diff
@@ -1,22 +1,101 @@
...
+
+- A [`Statement`] can have side effects and structured control flow.
+  `Statement`s do not produce a value, other than by storing one in some
+  designated place. `Statements` include blocks, conditionals, and loops, but also
+  operations that have side effects, like stores and function calls.
+
+`Statement`s form a tree, with pointers into the DAG of `Expression`s.
+
+Restricting side effects to statements simplifies analysis and code generation.
+A Naga back end can generate code to evaluate an `Expression` however and
```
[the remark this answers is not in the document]
jimblandy: I went with 'back end'.

src/lib.rs
jimblandy: Agreed, this is unclear; I'll spell things out a bit more.

But, the name of that method is not good. When `needs_pre_emit` is `true`, that means that the backend probably needs to do some preparation for it, but a `Statement::Emit` is *not* needed. So the word 'emit' in the name does not refer to the one thing in the IR named 'Emit'.

src/lib.rs
jimblandy: No, it's the right way around, it's just hard to parse. I'll rephrase.

src/lib.rs
jimblandy: At the moment, we don't document the types of any expressions, so the reader would have to go read the verifier to understand what "the expression type is a pointer" implies. But more broadly, the goal for the module-level documentation is to give the reader the orientation they need to make sense of the detailed documentation. So while it may be accurate to explain this in terms of what expressions generate pointers, using that approach here requires the reader to know details before they have an overview.

src/lib.rs
kvark: yep, renaming this to something clear would be good

src/lib.rs
jimblandy: Yes, good point. This will clean up some other bits as well.

src/lib.rs
```diff
@@ -1,22 +1,108 @@
...
+- The expression is the `result` of a [`Statement::Call`], representing the
+  call's return value. The call must be 'in scope' for the use (see below).
+
+- The expression is included in the range of some [`Statement::Emit`] that is
+  'in scope' for the use (see below). The [`Expression::needs_pre_emit`] method
+  returns `true` if the given expression does *not* need to be covered by an
+  `Emit` statement.
+
+The scope of an expression evaluated by an `Emit` statement covers the
+subsequent expressions in that `Emit`, any following statements in the `Block`
```
kvark, line 82: nicely written!

kvark (approved):

Issue #747, opened 2021-04-21
[meta] External data sets for testing


kvark:
It would be useful to have some place where we can store the bigger sets of shaders (SPIR-V, WGSL, GLSL, whatever). We'd then have a Github Action to fetch them and parse/validate. Since this would be a heavy action, we'd run it either manually, or on tag creation (seems most practical).

Here some info about SPIR-V corpus:
> vulkan CTS has 750K of them
> lots of them in SPIRV-Tools test suite accumulated over years
> When you build vulkan CTS, you can run external/vulkan/modules/vk-build-programs -v  to build all the shaders and run validation on them.
> You can hack that flow to dump;  I think there's a flow to dedup them and save them in a binary database of some kind but I never looked deeply at that.
> But they're not very diverse.  About 99% of them are generated from Glslang, so there's a monoculture problem.
The other 1% are generated from templated SPIR-V assembly text.
And recently there are a few hundred harder cases found through spirv-fuzz; using tech evolved from GraphicsFuzz folks
> All the .amber scripts in Vulkan CTS are under https://github.com/KhronosGroup/VK-GL-CTS/tree/master/external/vulkancts/data/vulkan/amber  with a subdir for graphicsfuzz

Gordon-F:
[Sascha Willems Vulkan examples](https://github.com/SaschaWillems/Vulkan/) have useful (and easier to integrate) [shaders](https://github.com/SaschaWillems/Vulkan/tree/master/data/shaders/glsl) test set. It should help to test `spv-in` and all naga backends. I don't think we should store any kind of snapshots in the naga repository with an external test set. Just be sure that the naga frontend and backend can handle them.

kvark:
Dota2 shader set was added in 47ada8182b83d43f47435df76d68f7dd278ab57a

kvark:
David's sample set - https://github.com/dneto0/spirv-samples

Gordon-F:
> David's sample set - https://github.com/dneto0/spirv-samples

@kvark Since we already have 2 external datasets, adding one more seems pretty straightforward task. We should just parse this external data set or produce valid `wgsl` code from it? Fail CI if something goes wrong or not?

kvark:
@Gordon-F I thought we just want to parse them into IR and validate. But generating WGSL is in line with our testing strategy, so it's also good.

Gordon-F:
`Tint` has a lot of tests - https://dawn.googlesource.com/tint/+/refs/heads/main/test/

[...]

PR #885, opened 2021-05-20
Add Vulkan tutorial shaders to lazy check

Gordon-F:
Part of #747.

Test run - https://github.com/Gordon-F/naga/runs/2629733528. Now we can validate 130 out of 300 shaders.

### DIFF
--- a/.github/workflows/lazy.yml
+++ b/.github/workflows/lazy.yml
@@ -44,3 +44,35 @@ jobs:
         run: cargo build --release --features spv-in,msl-out --bin naga
       - name: Convert shaders
         run: for file in data/*.spv ; do echo "Translating" ${file} && target/release/naga --validate 27 ${file} ${file}.metal; done
+  parse-vulkan-tutorial-shaders:
+    name: Parse Sascha Willems Vulkan tutorial shaders
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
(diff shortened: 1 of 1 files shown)


.github/workflows/lazy.yml
```diff
@@ -44,3 +44,35 @@ jobs:
...
+          set +e
+          touch counter
+          SUCCESS_RESULT_COUNT=0
+          FILE_COUNT=0
+          mkdir -p out
+          find "Vulkan/data/shaders/glsl/" -name '*.spv' | while read fname;
+          do
+              echo "Convert: $fname"
+              FILE_COUNT=$((FILE_COUNT+1))
+              target/release/naga --validate 27 $(realpath ${fname}) out/$(basename ${fname}).metal
```
Gordon-F, line 72: Should we test any backend here? Or just test `spv-in` frontend? If we decide to tests the backend here we should also validate output shaders.

kvark: We should probably enable the backends as they become operational

kvark (commented):
Good stuff! So what's the plan here? If we merge this, it's going to fail constantly, if I understand correctly. We need some sort of a mask to keep around.

Gordon-F:
> If we merge this, it's going to fail constantly, if I understand correctly

We can safety merge it as is. This code skip any errors because naga can't validate all of them now. All issues reported in #891.

kvark:
Ok, cool. The list looks quite fixable. Thank you for collecting all of it!

kvark (approved):

PR #886, opened 2021-05-20
[wgsl-out] Skip access to unsupported builtin inside Store statement

Gordon-F:
Maybe it would be better to add options to `wgsl-out` backend and let users decide what better to do with unsupported built-in.

But for now, since we already skip members of struct with unsupported built-in when writing struct, we should skip it everywhere.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -632,8 +632,15 @@ impl<W: Write> Writer<W> {
                 write!(self.out, "{}", INDENT.repeat(indent))?;
                 writeln!(self.out, "discard;")?
             }
-            // TODO: copy-paste from glsl-out
             Statement::Store { pointer, value } => {
+                // WGSL does not support all SPIR-V builtins and we should skip it in generated shaders.
+                // We already skip them when we generate struct type.
+                // Now we need to find expression that used struct with ignored builtins
+                if let Expression::AccessIndex { base, index } = func_ctx.expressions[pointer] {
+                    if access_to_unsupported_builtin(base, index, module, &func_ctx.info) {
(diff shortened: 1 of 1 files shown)


src/back/wgsl/writer.rs
```diff
@@ -632,8 +632,15 @@ impl<W: Write> Writer<W> {
                 write!(self.out, "{}", INDENT.repeat(indent))?;
                 writeln!(self.out, "discard;")?
             }
-            // TODO: copy-paste from glsl-out
             Statement::Store { pointer, value } => {
+                // WGSL does not support all SPIR-V builtins and we should skip it in generated shaders.
```
kvark, line 636: this is unfortunate... We shouldn't be in this situation, ideally.
If backends aren't expected to handle a particular builtin, the validator should catch this. And it would be responsibilty of spv-in to filter those out,

kvark (approved):

src/back/wgsl/writer.rs
Gordon-F: So, it means that `spv-in` should know about backend restrictions (capabilities?) and parse all that backend can handle and skip all other stuff?

What if the backend can emulate unsupported things somehow? Should I open a new issue for this?

kvark: No, spv-in wouldn't need to know about backends. It would need to know about what IR expects. IR shouldn't blindly pass everything under the sun, it recently got a bitset of capabilities itself, to validate. One of them is push constants, for example. So I imagine `spv-in` could have some options to omit particular things from the IR, and the user would provide them in tandem with specifying the validation capabilities for the IR.

PR #889, opened 2021-05-20
[glsl-in] Expression emitting

JCapucho:


src/front/glsl/ast.rs
```diff
@@ -1,12 +1,12 @@
 use super::{super::Typifier, constants::ConstantSolver, error::ErrorKind, SourceMetadata};
 use crate::{
-    proc::ResolveContext, Arena, BinaryOperator, Binding, Block, Constant, Expression, FastHashMap,
-    Function, FunctionArgument, GlobalVariable, Handle, Interpolation, LocalVariable, Module,
-    RelationalFunction, ResourceBinding, Sampling, ShaderStage, Statement, StorageClass, Type,
-    TypeInner, UnaryOperator,
+    front::Emitter, proc::ResolveContext, Arena, BinaryOperator, Binding, Block, Constant,
```
kvark, line 3: there is a `super::super` include in the previous line, let's use it here (i.e. that line is for all the frontend stuff)

src/front/glsl/ast.rs
```diff
@@ -141,99 +142,112 @@ pub enum Profile {
...
+    }
+
+    pub fn emit_flush(&mut self, body: &mut Block) {
+        body.extend(self.emitter.finish(&self.expressions))
+    }
+
+    pub fn add_expression(&mut self, expr: Expression, body: &mut Block) -> Handle<Expression> {
+        if expr.needs_pre_emit() {
+            self.emit_flush(body);
+            let expr = self.expressions.append(expr);
```
kvark, line 232: nit: let's not have `expr` name for bot the value itself and the handle to it

src/front/glsl/ast.rs
```diff
@@ -141,99 +142,112 @@ pub enum Profile {
...
-        name: &str,
-    ) -> Option<VariableReference> {
-        self.lookup_global_var_exps.get(name).cloned().or_else(|| {
-            let expr = match *program.lookup_global_variables.get(name)? {
+        for &(ref name, lookup) in program.global_variables.iter() {
+            let expr = match lookup {
                 GlobalLookup::Variable(v) => Expression::GlobalVariable(v),
                 GlobalLookup::BlockSelect(handle, index) => {
-                    let base = self.expressions.append(Expression::GlobalVariable(handle));
+                    let base = this.add_expression(Expression::GlobalVariable(handle), body);
```
kvark, line 190: Here we know that everything in this loop will be implicitly emitted (the global variables and accessing into them, so maybe we can just do `this.expressions.push()` instead of relying on the emitter?

src/front/glsl/ast.rs
```diff
@@ -141,99 +142,112 @@ pub enum Profile {
...
-    ) -> Option<VariableReference> {
-        self.lookup_constant_exps.get(name).cloned().or_else(|| {
-            let expr = self
-                .expressions
-                .append(Expression::Constant(*program.lookup_constants.get(name)?));
+            this.lookup_global_var_exps.insert(name.into(), var);
+        }
 
+        for &(ref name, handle) in program.constants.iter() {
+            let expr = this.add_expression(Expression::Constant(handle), body);
```
kvark, line 208: similarly, here we know that none of these expressions need to be explicitly emitted

src/front/glsl/functions.rs
```diff
@@ -135,91 +151,102 @@ impl Program<'_> {
...
-                                "exp2" => MathFunction::Exp2,
-                                "sign" => MathFunction::Sign,
-                                "transpose" => MathFunction::Transpose,
-                                "inverse" => MathFunction::Inverse,
-                                "normalize" => MathFunction::Normalize,
-                                _ => unreachable!(),
+                        Ok(ctx.add_expression(
+                            Expression::Math {
+                                fun: match name.as_str() {
+                                    "ceil" => MathFunction::Ceil,
```
kvark, line 157: would be good to move this match out into a `conv` module or something

src/front/glsl/parser.rs
```diff
@@ -499,15 +505,15 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
...
-                    .ok()
-            });
+            let maybe_constant = init
+                .clone()
+                .and_then(|(root, meta)| self.program.solve_constant(ctx.ctx, root, meta).ok());
 
             let pointer = ctx.add_var(self.program, ty, name, maybe_constant, meta)?;
 
             if let Some((value, _)) = init {
+                ctx.ctx.emit_flush(ctx.body);
```
kvark, line 515: this emit+start should probably be abstracted behind a helper, like `flush_expressions()`

src/front/mod.rs
```diff
@@ -14,8 +14,8 @@ use crate::{
 
 /// Helper class to emit expressions
 #[allow(dead_code)]
-#[derive(Default)]
-struct Emitter {
+#[derive(Default, Debug)]
+pub(crate) struct Emitter {
```
kvark, line 18: this doesn't need to be `pub(crate)`. You are using it from `front/glsl`, which is a child of this module

kvark (changes requested):
Just a few notes, looks good overall!

src/front/glsl/ast.rs
JCapucho: Access are implicitly emitted?

src/front/glsl/ast.rs
JCapucho: Yes but the previous iteration might have `Load` to be emitted but a flush before beginning the loop can fix this

kvark: sure, you can also just move this loop above the globals loop

src/front/glsl/ast.rs
kvark: accesses that result in pointers don't need to be emitted, since there is always going to be either a Store into them, or a Load (which is emitted)

kvark (approved):

PR #890, opened 2021-05-21
Validate built-ins to not repeat

### LINKED ISSUE #888: Using position from vertex output and frag coord causes panic on metal
Shader example:
```
struct VertexOutput {
};

[[stage(vertex)]]
fn vs_main() -> VertexOutput {
    var out: VertexOutput;
    out.position = vec4<f32>(0.0, 0.0, 0.0, 0.0);
    return out;
}

[[stage(fragment)]]
fn fs_main(
    input: VertexOutput,
    [[builtin(position)]] coord: vec4<f32>
) -> [[location(0)]] vec4<f32> {
    return vec4<f32>(0.0, 0.0, 0.0, 0.0);
}
```

The error:

```
wgpu error: Validation Error

Caused by:
      note: label = `iced_wgpu::quad pipeline`
    Internal error in VERTEX shader: Error compiling the shader "\"Compilation failed: \\n\\nprogram_source:106:17: error: declaration with attribute \\\'position\\\' already specified\\n, metal::float4 coord [[position]]\\n  ~~~~~~~~~~~~~~^~~~~\\nprogram_source:105:17: note: previous declaration with attribute \\\'position\\\' here\\n, metal::float4 position [[position]]\\n                ^\\n\""
```

 I do not get this on windows/vulkan. Logically it makes sense to me, but I don't know how this is normally handled in metal and can't find an example online. Is it allowed to have [[stage_in]] that defines position and then also a [[position]] frag coord as a parameter? If not, is there a way I can annotate this for it to understand better?

Note, I don't have a mac so I assume the top shader should produce the error, but it was from a report so I can't test it.

kvark:
Fixes #888

### DIFF
--- a/src/valid/interface.rs
+++ b/src/valid/interface.rs
@@ -41,14 +41,16 @@ pub enum VaryingError {
     InvalidInterpolation,
     #[error("Interpolation must be specified on vertex shader outputs and fragment shader inputs")]
     MissingInterpolation,
-    #[error("BuiltIn {0:?} is not available at this stage")]
+    #[error("Built-in {0:?} is not available at this stage")]
     InvalidBuiltInStage(crate::BuiltIn),
-    #[error("BuiltIn type for {0:?} is invalid")]
+    #[error("Built-in type for {0:?} is invalid")]
     InvalidBuiltInType(crate::BuiltIn),
     #[error("Struct member {0} is missing a binding")]
     MemberMissingBinding(u32),
     #[error("Multiple bindings at location {location} are present")]
     BindingCollision { location: u32 },
+    #[error("Built-in {0:?} is present more than once")]
+    DuplicateBuiltIn(crate::BuiltIn),
 }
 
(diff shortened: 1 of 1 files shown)


src/valid/mod.rs
```diff
@@ -94,6 +94,7 @@ pub struct Validator {
     types: Vec<r#type::TypeInfo>,
     layouter: Layouter,
     location_mask: BitSet,
+    built_ins: FastHashSet<crate::BuiltIn>,
```
jimblandy, line 97: Suggestion: we could probably just use a `BitSet` here, just as we do for `location_mask`, and cast `BuiltIn` to `usize` when we work with it. (The C programmer in me wants a `u32`, but that's not the way we do things any more...)

jimblandy (approved):
Looks good to me.

src/valid/mod.rs
kvark: oh indeed, I'll do this

Issue #807, opened 2021-04-30
[wgsl-in] panic on return without location


SNCPlay42:
This code:

```
[[stage(vertex)]]
fn main() -> f32 {
    return 0.0;
}
```

should error because the return has no location, but produces this panic, using naga 0.4.0:

```
  thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', \.cargo\registry\src\github.com-1ecc6299db9ec823\naga-0.4.0\src\proc\interpolator.rs:90:52  
  stack backtrace:
     0: std::panicking::begin_panic_handler
               at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0\/library\std\src\panicking.rs:493
     1: core::panicking::panic_fmt
               at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0\/library\core\src\panicking.rs:92
     2: core::panicking::panic
               at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0\/library\core\src\panicking.rs:50
     3: core::option::Option<mut naga::Binding*>::unwrap<mut naga::Binding*>
               at \.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\option.rs:386
     4: naga::proc::interpolator::{{impl}}::apply_common_default_interpolation::default_binding_or_struct
               at \.cargo\registry\src\github.com-1ecc6299db9ec823\naga-0.4.0\src\proc\interpolator.rs:90
     5: naga::Module::apply_common_default_interpolation
               at \.cargo\registry\src\github.com-1ecc6299db9ec823\naga-0.4.0\src\proc\interpolator.rs:135
     6: naga::front::wgsl::Parser::parse
               at \.cargo\registry\src\github.com-1ecc6299db9ec823\naga-0.4.0\src\front\wgsl\mod.rs:2745
     7: naga::front::wgsl::parse_str
               at \.cargo\registry\src\github.com-1ecc6299db9ec823\naga-0.4.0\src\front\wgsl\mod.rs:2754
     8: build_script_build::main
               at .\build.rs:24
     9: core::ops::function::FnOnce::call_once<fn(),tuple<>>
               at \.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ops\function.rs:227
```

If the return value is given a location attribute, parsing succeeds as expected:

```
// no panic
[[stage(vertex)]]
[...]

PR #894, opened 2021-05-21
Improve error messages for missing I/O bindings.

jimblandy:
The `apply_common_default_interpolation` helper function would panic if bindings
were missing, but missing bindings should be something that front ends can count
on validation to detect, so the helper should just return silently.

The validator returned `InvalidType` errors for missing bindings, apparently
because variables without bindings must be structs that do have bindings. But
this is unhelpful when you've just forgotten to label an argument. So this patch
adds a new, more specific, `VaryingError` variant.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -729,7 +729,7 @@ impl Writer {
                 }
                 Instruction::type_int(id, bits, signedness)
             }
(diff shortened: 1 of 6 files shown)


kvark:
Supposedly fixes #807?

kvark (approved):
Looks fancy and reasonable :)
Thank you!

Issue #887, opened 2021-05-20
[glsl-in] Builtin global have empty storage access 


Gordon-F:
```glsl
#version 450

void main() {
    gl_Position = vec4(1.0, 1.0, 1,0, 1.0);
}
```

```bash
Global variable [1] 'gl_Position' is invalid:
        Storage access LOAD | STORE exceeds the allowed (empty)
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', bin/naga.rs:286:70
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

JCapucho:
@Gordon-F could you please test #898

Gordon-F:
@JCapucho fix worked, thank you! But one more thing, glsl frontend can't validate vec constructor. For example:
```glsl
gl_Position = vec4(1.0, 1.0, 1.0, 1.0); // Ok (4 float compose components)
gl_Position = vec4(1.0, 1.0,    1,   0,     1.0); // Not Ok (5 compose components)
```

I can open another issue for this problem.

JCapucho:
@Gordon-F This is expected, the glsl frontend can generate bad IR for bad code, it's the job of the validator to catch it

PR #898, opened 2021-05-22
[glsl-in] Validation fixes

### LINKED ISSUE #884: GLSL parsing doesn't pass validation
The glsl parser currently produces code which does not pass the validation.
For example:

```glsl
#version 450

void main() {}
```
Running `cargo run --features glsl-in test.vert` results in 
```
Entry point main at Vertex is invalid:
        Expression [1] is invalid
```

<details>
<summary>Module</summary>

```rust
Module {
    types: {
        [1]: Type {
            name: None,
            inner: Struct {
                top_level: true,
                members: [],
                span: 0,
            },
        },
    },
    constants: {},
    global_variables: {},
    functions: {
        [1]: Function {
            name: Some(
                "main",
            ),
            arguments: [],
            result: None,
            local_variables: {},
            expressions: {},
            body: [
                Return {
                    value: None,
                },
            ],
        },
    },
    entry_points: [
        EntryPoint {
            name: "main",
            stage: Vertex,
            early_depth_test: None,
            workgroup_size: [
                0,
                0,
                0,
            ],
            function: Function {
                name: None,
                arguments: [],
                result: Some(
                    FunctionResult {
                        ty: [1],
                        binding: None,
                    },
                ),
                local_variables: {},
                expressions: {
                    [1]: Compose {
                        ty: [1],
                        components: [],
                    },
                },
                body: [
                    Call {
                        function: [1],
                        arguments: [],
                        result: None,
                    },
                    Return {
                        value: Some(
                            [1],
                        ),
                    },
                ],
            },
        },
    ],
}

```
</details>

Another example:

```glsl
#version 450

layout(location = 0) in vec3 Vertex_Position;
layout(location = 1) in vec3 Vertex_Normal;

layout(location = 0) out vec3 v_WorldPosition;
layout(location = 1) out vec3 v_WorldNormal;

void main() {
    mat4 Model = mat4(0.0);
    mat4 ViewProj = mat4(0.0);

    vec4 world_position = Model * vec4(Vertex_Position, 1.0);
    v_WorldNormal = mat3(Model) * Vertex_Normal;
    gl_Position = ViewProj * world_position;
}

```
results in 
```
        Expression [24] is invalid
        Type resolution failed
        Incompatible operands: Matrix { columns: Quad, rows: Quad, width: 4 } as Float
```

JCapucho:
Fixes #884 and #887

Fixes included
- The output struct no longer produces a `VaryingError`
- Builtins are checked if they can be emitted in the stage now
- Globals now have proper storage access
- Single argument vector and matrix constructors now support dimension casting
- The lod value in TextureLod is now casted
- Parameter qualifiers now set the type of the argument
- Default interpolation is now provided


JCapucho:
For whoever reviews this the commit message for the commits fixing the vector/matrix constructors and the one fixing the parameter qualifier have some more information in them

src/front/glsl/functions.rs
```diff
@@ -524,6 +525,26 @@ impl Program<'_> {
     }
 }
 
+fn should_read(built_in: BuiltIn, stage: ShaderStage) -> bool {
```
kvark, line 561: This looks like a list that the validator uses. Let the validator take care of this. Here, if you need to exclude specific things, then make it an exclusion list specifically, instead of listing all the possibilities.

kvark (changes requested):

src/front/glsl/ast.rs
```diff
@@ -203,12 +203,15 @@ impl<'function> Context<'function> {
 
         for &(ref name, lookup) in program.global_variables.iter() {
             this.emit_flush(body);
-            let expr = match lookup {
-                GlobalLookup::Variable(v) => Expression::GlobalVariable(v),
+            let (expr, load) = match lookup {
+                GlobalLookup::Variable(v) => (
+                    Expression::GlobalVariable(v),
+                    StorageClass::Handle != program.module.global_variables[v].class,
```
kvark, line 209: nit: doing the "constant != variable" comparison is super weird. Can we please use the normal order of things?

src/front/glsl/parser.rs
```diff
@@ -776,7 +776,7 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
             ty,
             init: None,
             // TODO
```
kvark, line 789: is it still a TODO?

src/front/glsl/variables.rs
```diff
@@ -331,9 +331,23 @@ impl Program<'_> {
...
         }
 
-        let class = match storage {
-            StorageQualifier::StorageClass(class) => class,
-            _ => StorageClass::Private,
+        let (class, storage_access) = match self.module.types[ty].inner {
+            TypeInner::Image { class, .. } => (
+                StorageClass::Handle,
+                if let ImageClass::Storage(_) = class {
+                    StorageAccess::all()
```
kvark, line 339: this should depend on `readonly`/`writeonly`/`readwrite` qualifiers

src/front/glsl/variables.rs
```diff
@@ -343,7 +357,7 @@ impl Program<'_> {
             ty,
             init,
             // TODO
```
kvark, line 360: is it still a TODO?

src/front/glsl/variables.rs
```diff
@@ -310,7 +311,7 @@ impl Program<'_> {
             self.entry_args.push((
                 Binding::Location {
                     location,
-                    interpolation,
+                    interpolation: Some(interpolation.unwrap_or(Interpolation::Perspective)),
```
kvark, line 314: careful with the type here: it can only be Perspective for floating point stuff, has to be Flat for integers

src/front/glsl/functions.rs
JCapucho: The problem here is that the validator complains if a builtin is added as a function argument in a stage where it shouldn't

src/front/glsl/ast.rs
JCapucho: And what would be the normal order? An equality comparison?

src/front/glsl/variables.rs
JCapucho: I know, I just haven't gotten around to implementing them, I'll add a TODO

src/front/glsl/variables.rs
JCapucho: Read on the glsl spec that the default was perspective but missed that it was only for floats

src/front/glsl/functions.rs
kvark: I don't quite understand. If you have an entry point in GLSL that's using some built-in, then just add it to the arguments. If it's not using this built-in, and you aren't adding it to the arguments, then the validator have nothing to complain about.

src/front/glsl/ast.rs
kvark: "variable != something" or "variable == something"
I think the "constant == variable" comes from some very old times where people were afraid to accidentally write assignment expression instead of equality (in C/C++).

kvark:
Thank you for addressing the concerns! I think only "should_read" stuff is left to discuss.

src/front/glsl/ast.rs
JCapucho: Sorry I'm really not understanding what you want here, do you want me to change the order of the comparsion i.e.
```rust
program.module.global_variables[v].class != StorageClass::Handle
```

src/front/glsl/functions.rs
JCapucho: You have the case where the built in is being used but shouldn't be written for example (vertex stage)
```glsl
void main() {
  gl_Postition = vec4(1);
}
```
without this fix the backend would produce this pseudocode
```glsl
vec4 position;

void wrapper(
  [builtin=position] vec4 x
) -> [builtin=position] vec4 {
  position = x;
  main();
  return position;
}

void main() {
  position = vec4(1);
}
```
this is invalid because position can't be read in the vertex stage, with this change the generated pseudocode would look like this

```glsl
vec4 position;

void wrapper() -> [builtin=position] vec4 {
  main();
  return position;
}

void main() {
  position = vec4(1);
}
```

src/front/glsl/ast.rs
JCapucho: Glad we managed to get on the same page

src/front/glsl/functions.rs
kvark: Thank you for writing this down! I think there is a problem in the first step. Why does the pseudocode have "position = x" at all? When GLSL is parsed, we should see that `gl_Position` (not specifically it, anything goes) is on the left side of an assignment, so there is no reading going on.

JCapucho: The frontend doesn't keep global use information even though it should because with multiple stages we can't blindly emit globals arguments, I think we should add this for now and I'll make another PR which collects the global use of the functions and the functions it calls recursively and only emit the loads and the stores for the return based on it.

JCapucho:
@kvark it should be ready for merging now

kvark (approved):

PR #903, opened 2021-05-24
Require dynamically sized arrays to appear in 'storage' memory.

jimblandy:
WGSL requires that runtime-sized arrays appear only as the last member of a
structure in in the `storage` storage class. It seems to me that Naga should
enforce this restriction on its own IR as well.


kvark (approved):
Thank you!
So the `Storage` class was already validated properly, but this PR also fixes the validation of other classes.

PR #904, opened 2021-05-25
[wgsl-in] Improve wgsl lexer error messages

HalfVoxel:
Improves error messages by adding spans for most wgsl lexer errors and adding more information to many messages to remove ambiguity.

For example:

**before**
```rust
error: expected paren, found '('
   ┌─ wgsl:49:13
   │
49 │         vec4(c)
   │             ^ expected paren
```
**after**
```rust
error: expected '<', found '('
   ┌─ wgsl:49:13
   │
49 │         vec4(c)
   │             ^ expected '<'
```

or
**before**
```rust
error: invalid type cast to `vec4`
```
**after**
```rust
error: cannot cast a struct to a vec4<f32>
   ┌─ wgsl:50:18
   │
50 │         vec4<f32>(in)
   │                  ^^^^ cannot cast a struct to a vec4<f32>
```

This also fixes a bug in which an integer or floating point constant with some invalid widths could be accepted. For example, the constant `3.0f33` was treated as an `f32` constant.

Some files are getting reallly large, maybe they should be split into several smaller ones?

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -28,7 +28,7 @@ use std::{
     borrow::Cow,
     io::{self, Write},
     iter,
-    num::NonZeroU32,
+    num::{NonZeroU32, ParseFloatError, ParseIntError},
     ops,
 };
 use thiserror::Error;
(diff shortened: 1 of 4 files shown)


src/front/wgsl/mod.rs
```diff
@@ -12,7 +12,7 @@ use crate::{
     proc::{
         ensure_block_returns, Alignment, Layouter, ResolveContext, ResolveError, TypeResolution,
     },
-    FastHashMap,
+    FastHashMap, ScalarKind,
```
kvark, line 15: Let's not include IR stuff at module scope here, except for the container wrappers (hashmap, arena, handle). This would be inconsistent with the rest of the codebase.

src/front/wgsl/mod.rs
```diff
@@ -135,38 +169,85 @@ pub enum Error<'a> {
...
                     &source[unexpected_span.clone()],
                 ),
                 labels: vec![(
                     unexpected_span.clone(),
-                    format!("expected {}", expected).into(),
+                    format!("expected {}", expected_str).into(),
                 )],
                 notes: vec![],
+            }
+        },
```
kvark, line 222: something strange happened here with indentation

src/front/wgsl/mod.rs
```diff
@@ -190,6 +340,49 @@ impl<'a> Error<'a> {
     }
 }
 
+fn type_to_name(tp: &crate::TypeInner) -> String {
```
kvark, line 343: We can also do `impl crate::TypeInner { fn to_name(&self) {...} }` or something like this.

kvark (commented):
Very nice work here, thank you!
Just a couple of notes.

src/front/wgsl/mod.rs
```diff
@@ -181,6 +262,75 @@ impl<'a> Error<'a> {
                 labels: vec![(ident_span.clone(), "unknown identifier".into())],
                 notes: vec![],
             },
+            Error::UnknownScalarType(ref bad_span) => ParseError {
+                message: format!("unknown scalar type: '{}'", &source[bad_span.clone()]),
+                labels: vec![(bad_span.clone(), "unknown scalar type".into())],
+                notes: vec!["Valid scalar types are f16, f32, f64, i8, i16, i32, i64, u8, u16, u32, u64, bool".into()],
```
grovesNL, line 268: Are these supposed to be included here yet? There's listed under [Reserved Keywords](https://gpuweb.github.io/gpuweb/wgsl/#reserved-keywords) and IIRC why the width validation message only listed 32

src/lib.rs
```diff
@@ -454,7 +454,7 @@ pub struct Type {
 }
 
 /// Enum with additional information, depending on the kind of type.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
```
grovesNL, line 457: Is there anywhere the clone is necessary vs. a reference?

src/front/wgsl/mod.rs
kvark: I suppose this is forward looking, and the use of unavailable types is caught by the validation anyway.
So we can go either way.

src/lib.rs
HalfVoxel: This is used to provide better error messages for Error::BadTypeCast. I can unfortunately not use a reference because the struct may even be temporary. Fortunately it is only used in the error path, in which performance is less of an issue.
I did worry a bit about using it in BadTypeCast though because the TypeInner struct is quite large. Possibly they should be boxed to avoid copying costs?

image

src/front/wgsl/mod.rs
HalfVoxel: I added that note because those scalar types are exactly what the lexer accepts right now:
image

and as you say it will get caught in validation if the type isn't actually supported.

src/front/wgsl/mod.rs
HalfVoxel: But `TypeInner` is not a wgsl specific type, but this name is the name used in wgsl. It didn't feel quite right to put wgsl-specific formatting code outside the wgsl directory.

HalfVoxel:
Rebased on main and fixed some review comments.

src/front/wgsl/mod.rs
kvark: It's totally cool because the method would be private, only visible to WGSL. We do this in other frontends/backends.

src/lib.rs
kvark: Let's try to find another way. For example, these could be strings, output from `type_to_name`.
We don't want to have `TypeInner` cloneable in general, because it invites for inefficient code paths to appear.

HalfVoxel: Ok, I changed those to strings.

src/front/wgsl/mod.rs
HalfVoxel: Ok. I added an `to_wgsl` function which is quite nice.
Eventually we might want to re-use some printing code from wgsl-out. But I'm not sure how easy that is to do right now.

HalfVoxel:
Rebased on main and fixed some review comments.

I also added validation to make sure texture sample types are correct (see last commit).

kvark (approved):
Do you mind if we squash this?

HalfVoxel:
No problems. Squashed.

PR #907, opened 2021-05-25
Include the type arena in `ResolveContext`, since it's always needed.

jimblandy:
This removes a bunch of parameter passing, and I think there's only one type arena that is meaningful in these cases anyway.


kvark (approved):
Looks reasonable! I was thinking that there are places where the type arena is not convenient to provide, and the type resolution doesn't require it, strictly speaking. It appears that there are no such places, since the CI is happy here :)

PR #911, opened 2021-05-27
Don't treat pointers as data.

jimblandy:
All commits should be CI-clean.

- Don't allow pointers in return values.

- [wgsl-in]: Use a `matches!` expression for brevity.

- Don't permit `Expression::Select` to operate on pointers. At the moment, Naga
  doesn't support anything like SPIR-V's 'variable pointers'.

- Naga `TypeInner::Pointer` types are not `TypeFlags::DATA`.

  Change `Validator::validate_type` not to mark pointers with the
  `TypeFlags::Data` flag. The `TypeFlags::DATA` flag is supposed to correspond to
  WGSL's concept of a 'plain type': a scalar, atomic, or composite type, excluding
  textures, samplers, and pointers.

  With this change, there is no type where `TypeFlags::DATA` and
  `TypeFlags::INTERFACE` differ, but I'm not removing anything in this PR, because
  I think we'll want to re-introduce a different distinction between them soon.

### DIFF
--- a/src/valid/function.rs
+++ b/src/valid/function.rs
@@ -86,6 +86,8 @@ pub enum FunctionError {
     },
     #[error("Argument '{name}' at index {index} has a type that can't be passed into functions.")]
     InvalidArgumentType { index: usize, name: String },
+    #[error("The function's given return type cannot be returned from functions")]
+    NonConstructibleReturnType,
     #[error("Argument '{name}' at index {index} is a pointer of space {space:?}, which can't be passed into functions.")]
     InvalidArgumentPointerSpace {
         index: usize,
@@ -891,6 +893,17 @@ impl super::Validator {
                     name: argument.name.clone().unwrap_or_default(),
                 }
                 .with_span_handle(argument.ty, &module.types));
+            }
+        }
+
+        #[cfg(feature = "validate")]
+        if let Some(ref result) = fun.result {
+            if !self.types[result.ty.index()]
+                .flags
+                .contains(super::TypeFlags::CONSTRUCTIBLE)
+            {
+                return Err(FunctionError::NonConstructibleReturnType
+                    .with_span_handle(result.ty, &module.types));
             }
         }
 
--- a/src/valid/type.rs
+++ b/src/valid/type.rs
@@ -52,6 +52,15 @@ bitflags::bitflags! {
 
         /// This type can be passed as a function argument.
         const ARGUMENT = 0x40;
+
+        /// A WGSL [constructible] type.
+        ///
+        /// The constructible types are scalars, vectors, matrices, fixed-size
+        /// arrays of constructible types, and structs whose members are all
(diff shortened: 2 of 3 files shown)


src/front/wgsl/mod.rs
```diff
@@ -1232,12 +1232,11 @@ impl Parser {
         mut handle: Handle<crate::Expression>,
         allow_deref: bool,
     ) -> Result<Handle<crate::Expression>, Error<'a>> {
-        let mut needs_deref = match ctx.expressions[handle] {
-            crate::Expression::LocalVariable(_) | crate::Expression::GlobalVariable(_) => {
-                allow_deref
-            }
-            _ => false,
-        };
+        let mut needs_deref = allow_deref
```
kvark, line 1235: Is this really better than the old code?
I think we should only use macros where their benefit outweight the cost of having an extra layer of indirection. Macros slow down compilation, confuse IDEs, and generally shouldn't be used so sparingly.

src/valid/expression.rs
```diff
@@ -755,6 +755,9 @@ impl super::Validator {
                 if !condition_good || accept_inner != reject_inner {
                     return Err(ExpressionError::InvalidSelectTypes);
                 }
+                if let Ti::Pointer { .. } = *accept_inner {
```
kvark, line 758: I think this is a wrong level to check this. We are checking for pointers, but what if it's a texture, for example?
Instead, we need to check the derived type flags on this and ensure this is `DATA | SIZED`

src/front/wgsl/mod.rs
jimblandy: Okay, this is the case where I think `matches!` actually works. Putting `allow_deref` in the first arm of the `match` is a really obscure way of saying `&&`. And if you do use `&&`, then the `match` just has arms with `=> true` and `=> false`, which is classic `matches!`. Since this is a `std` macro, I think IDEs are going to pick up on it pretty quickly; `rust-analyzer`, at least, hasn't been having any problems with it.

But if you don't find that persuasive, it's not something I feel strongly about.

kvark: I think I agree. It's not striking me as a case that would need to be an exception though if we try to avoid matches in general (I checked briefly and didn't find a clippy for this).

teoxoy:
@jimblandy what is the status of this PR?

jimblandy:
Probably stale. It needs to be reevaluated in light of the current state of the language spec and the code, which has evolved.

jimblandy:
The easiest thing to do would be to demote it (re-file it) as an issue instead of a PR.

jimblandy:
Converted to draft - just realized there were other commits on the branch I hadn't noticed, and whose changes need to be incorporated.

teoxoy (approved):
PR description is a bit outdated but the changes look good

PR #912, opened 2021-05-27
The `bool` type is neither a host-shareable nor an interface type.

jimblandy:
However, `bool` is permitted for built-ins.

### DIFF
--- a/src/valid/type.rs
+++ b/src/valid/type.rs
@@ -211,13 +211,17 @@ impl super::Validator {
                 if !self.check_width(kind, width) {
                     return Err(TypeError::InvalidWidth(kind, width));
                 }
+                let shareable = if kind.is_numeric() {
+                    TypeFlags::IO_SHAREABLE | TypeFlags::HOST_SHAREABLE
+                } else {
+                    TypeFlags::empty()
+                };
                 TypeInfo::new(
                     TypeFlags::DATA
                         | TypeFlags::SIZED
                         | TypeFlags::COPY
-                        | TypeFlags::IO_SHAREABLE
-                        | TypeFlags::HOST_SHAREABLE
-                        | TypeFlags::ARGUMENT,
+                        | TypeFlags::ARGUMENT
+                        | shareable,
                     width as u32,
                 )
             }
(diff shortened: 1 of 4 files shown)


jimblandy:
Actually, it might be best to combine this with a fix for `TypeFlags::INTERFACE` handling.

src/valid/type.rs
```diff
@@ -332,7 +335,12 @@ impl super::Validator {
                         return Err(TypeError::InvalidData(member.ty));
                     }
                     if top_level && !base_info.flags.contains(TypeFlags::INTERFACE) {
-                        return Err(TypeError::InvalidBlockType(member.ty));
+                        // Certain non-`INTERFACE` types like arrays and `bool` are
+                        // permitted for built-ins, whose types we verify in
+                        // `VaryingContext::validate`.
+                        if !matches!(member.binding, Some(crate::Binding::BuiltIn(_))) {
```
kvark, line 341: let's not use matches please

jimblandy: Here, `matches!` doesn't buy much, but there's another spot I ran into (in another patch series) where the `match` is worse. I'll point it out when we get there, curious what you'll think.

kvark (approved):

teoxoy:
Can we merge this?

jimblandy:
It's probably worth re-reviewing, since it's so old.

teoxoy:
I can re-review it but I think it needs to be rebased first.

jimblandy:
Okay - this version is basically a rewrite for the current sources. It depends on #1875 (and thus is still a draft), and conflicts with #1874. If we can get those landed, I'll re-base and move this out of "draft" state.

jimblandy:
This has some renames and some refactoring, so it's probably easiest to review commit-by-commit.

Each commit should pass CI individually, so it's not necessary to squash before merging.

teoxoy (approved):
Looks good but the title indicates that the PR will also remove `bool`s from being `HOST_SHAREABLE`.

teoxoy (approved):
Looks great!

Issue #392, opened 2021-01-30
Cached/named expressions


kvark:
Here is a fun case we have in `shadow.wgsl` test:
```rust
        const light: Light = s_lights.data[i];
        const shadow: f32 = fetch_shadow(i, light.proj * in_position_fs);
        const light_dir: vec3<f32> = normalize(light.pos.xyz - in_position_fs.xyz);
        const diffuse: f32 = max(0.0, dot(normal, light_dir));
        color = color + shadow * diffuse * light.color.xyz;
```
Which translates to the following abomination in MSL:
```
color1 = (color1 + ((fetch_shadow(i, (s_lights.data[i].proj * input.in_position_fs)) * metal::max(0.0, metal::dot(metal::normalize(input.in_normal_fs), metal::normalize((metal::float3(s_lights.data[i].pos.x, s_lights.data[i].pos.y, s_lights.data[i].pos.z) - metal::float3(input.in_position_fs.x, input.in_position_fs.y, input.in_position_fs.z)))))) * metal::float3(s_lights.data[i].color.x, s_lights.data[i].color.y, s_lights.data[i].color.z)));
```

The problem is - our IR doesn't know how to keep the values of expressions around, so it has to nest the full expression tree in the spot where a `const` is used. This is contrary to the user expectations.

One choice that we could make is - just make the WGSL frontend generate a local variable for every const. This would automatically support naming. It will result, however, in each access to this value being an `OpLoad` in SPIR-V. It might be optimized away by the drivers, but it seems weird that we have that `const` mechanism at all, and we end up doing the same thing as for `var`...

Another choice is introducing a new `CachedExpression` kind of thing:
```rust
struct CachedExpression {
  name: Option<String>,
  expr: Expression,
}
struct Block {
  cached_expressions: Arena<CachedExpression>,
  statements: Vec<Statement>,
}
```
The concern here is that this new construct is a complication designed specifically for SPIRV output, so far. It might be useful for AIR and DXIL in the near future though, so maybe we should bite that bullet now.

kvark:
Another possible way to get that is adding `mutable: bool` flag to the `struct LocalVariable`. This might be the simplest we can do. I'm just a bit worried about what this flag = `false` would mean. I.e. what if there are multiple assignments, but they are in unrelated branches? I guess we could say that this is very strict: immutable variables can only have 1 assignment ever (unlike, say, Rust).

kvark:
Let's consider this closed now, thanks to @Gordon-F efforts!

PR #914, opened 2021-05-27
Set bake_ref_count to 1 for Expression::Load

### LINKED ISSUE #910: Statements are incorrectly ordered from `mtl` and `glsl` backends.
I've created a gist with steps to reproduce [here](https://gist.github.com/paulgb/a8f705d9e80bbfd0ca42410086a5c04f).

The input WGSL code is:

    [...snip; full code in linked repo]
    var a: f32 = 0.0;
    let b: f32 = a;
    a = 1.0;
    v_data.data = b;
    [snip...]

The correct resulting value of `v_data.data` is `0.0`.

The code returned by `naga` for `metal` is:

    [snip]
    float a = 0.0;
    a = 1.0;
    v_data.data = a;
    return;
    [snip...]

The resulting value of `v_data.data` is `1.0`. The value `b` is written out entirely, but instead of taking on the value at the time of its assignment, it takes on the value of `a` at the end.

The `.spv` file itself appears to be ordered correctly, so the issue appears to occur between loading the `.spv` file and generating the output in the backend. To confirm this, I used `spirv-cross` to generate metal and glsl output from the same input `.spv` returned by `naga`; these files are correctly ordered.

I'm planning to keep digging into this, but I wanted to raise it as an issue in case it's come up elsewhere.

In case anyone needs a workaround, changing `let` expressions to `var` expressions seems to fix it, possibly with a small performance penalty.

paulgb:
This fixes #910 by forcing `Load` expressions to be baked at the time they are emitted, so that they take on the value at that time rather than at the time when they are used.

I would be happy to add some tests, but I haven't been able to figure out how the `tests/{in,out}` test cases are actually run.

### DIFF
--- a/tests/out/quad-vert.Vertex.glsl
+++ b/tests/out/quad-vert.Vertex.glsl
@@ -28,7 +28,8 @@ layout(location = 0) in vec2 _p2vs_location0;
 smooth layout(location = 0) out vec2 _vs2fs_location0;
 
 void main1() {
-    v_uv = a_uv;
+    vec2 _expr12 = a_uv;
+    v_uv = _expr12;
     vec2 _expr13 = a_pos;
     perVertexStruct.gen_gl_Position = vec4(_expr13.x, _expr13.y, 0.0, 1.0);
     return;
@@ -40,7 +41,12 @@ void main() {
     a_uv = a_uv1;
     a_pos = a_pos1;
     main1();
-    type10 _tmp_return = type10(v_uv, perVertexStruct.gen_gl_Position, perVertexStruct.gen_gl_PointSize, perVertexStruct.gen_gl_ClipDistance, perVertexStruct.gen_gl_CullDistance);
+    vec2 _expr10 = v_uv;
+    vec4 _expr11 = perVertexStruct.gen_gl_Position;
+    float _expr12 = perVertexStruct.gen_gl_PointSize;
+    [1] _expr13 = perVertexStruct.gen_gl_ClipDistance;
+    [1] _expr14 = perVertexStruct.gen_gl_CullDistance;
+    type10 _tmp_return = type10(_expr10, _expr11, _expr12, _expr13, _expr14);
     _vs2fs_location0 = _tmp_return.member;
     gl_Position = _tmp_return.gen_gl_Position1;
     return;
--- a/src/back/mod.rs
+++ b/src/back/mod.rs
@@ -28,6 +28,8 @@ impl crate::Expression {
             crate::Expression::ImageSample { .. } | crate::Expression::ImageLoad { .. } => 1,
             // derivatives use the control flow
             crate::Expression::Derivative { .. } => 1,
+            // load can depend on state that changes during program flow
+            crate::Expression::Load { .. } => 1,
             // cache expressions that are referenced multiple times
             _ => 2,
         }
--- a/tests/out/access.msl
+++ b/tests/out/access.msl
@@ -24,5 +24,7 @@ vertex fooOutput foo(
 , device Bar& bar [[buffer(0)]]
 , constant _mslBufferSizes& _buffer_sizes [[buffer(24)]]
 ) {
-    return fooOutput { static_cast<float4>(metal::int4(type5 {bar.data[(1 + (_buffer_sizes.size0 - 64 - 4) / 4) - 1u], static_cast<int>(bar.matrix[3u].x), 3, 4, 5}.inner[vi])) };
+    metal::float4 _e5 = bar.matrix[3u];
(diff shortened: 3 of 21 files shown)


kvark (approved):
Makes sense, thank you!

kvark:
They are run by `tests/snapshot.rs`. So if you do `cargo test`, you'll see the outputs updated automatically, and they need to be included into this PR.
Speaking of which, this PR has an unfortunate side effect: textual backends that use variables now generate more expressions than needed. We can follow-up with this separately. They'll probably need some backend-specific logic to "reverse" the effect of this PR.

kvark:
Let's not worry about this right now, and just get the CI happy to land this fix, as it is important for correctness.

tests/out/quad-vert.Vertex.glsl
```diff
@@ -40,7 +41,12 @@ void main() {
     a_uv = a_uv1;
     a_pos = a_pos1;
     main1();
-    type10 _tmp_return = type10(v_uv, perVertexStruct.gen_gl_Position, perVertexStruct.gen_gl_PointSize, perVertexStruct.gen_gl_ClipDistance, perVertexStruct.gen_gl_CullDistance);
+    vec2 _expr10 = v_uv;
+    vec4 _expr11 = perVertexStruct.gen_gl_Position;
+    float _expr12 = perVertexStruct.gen_gl_PointSize;
+    [1] _expr13 = perVertexStruct.gen_gl_ClipDistance;
```
Gordon-F, line 47: That's looks strange. We needed to fix load expression logic for glsl-out if we want to merge these changes.

kvark: @Gordon-F would you be able to help with this! I don't want to leave #914 hanging in the air

Gordon-F: We could make this logic less aggressive by checking the pointer inside `Expression::Load { pointer }`. If it's pointing to `Expression::Constant` or `Expression::LocalVariable` we should return 1 bake count, otherwise 2 (default).

@paulgb Would you like to fix this in such a way? Also, it would be awesome to add your case to naga snapshot testing.

paulgb: I'm happy to add this, but `bake_ref_count` only has access to the `Handle` reference, not the arena needed to look it up.

I wonder if a cleaner approach would be to add a boolean field to `ExpressionInfo` called `may_vary`, set to  `true` iff the expression is a `Load` that depends on a global or local variable. The backend would then bake an expression if `may_vary || min_ref_count < expression_info.ref_count`.

paulgb: I made an attempt at the latter approach. It does fix the GLSL validation problem. It's a separate branch, but I could bring it into this PR if it's a desirable approach.

https://github.com/paulgb/naga/commit/3a11e22933b1c13efdc0ba92ec9563a9430e6551

Gordon-F: @kvark @paulgb Since we will have to change the public api, maybe it's easier to add `Arena<Expression>` reference to bake_ref_count ? It is used only in backends. 

Anyway, none of these solutions looks like a correct fix for this situation. We should return to this later (maybe with https://github.com/gfx-rs/naga/issues/392).

kvark: Is this still a problem when rebased on top of master? ae367f6b74f34f3f7acbf7c8db0e03dcd37f74e4 could be affecting this

Gordon-F: It's still a problem on master. We can't handle arrays typed builtin (https://github.com/gfx-rs/naga/issues/851) in `glsl-out`.

Before:
```glsl
[1] _expr13 = perVertexStruct.gen_gl_ClipDistance;
```

After master rebase:
```
float _expr13 = perVertexStruct.gen_gl_ClipDistance;
```

Error:
```
Validating quad-vert.Vertex.glsl
stdin
ERROR: 0:43: '=' :  cannot convert from ' global 1-element array of highp float' to ' temp highp float'
ERROR: 0:43: '' : compilation terminated 
ERROR: 2 compilation errors.  No code generated.
```

Open a new PR #925 with another fix of this problem without validation errors.

kvark: Can we just fix the GLSL variable declarations to support arrays better?
Would this work if it looked like `float _expr13[1] = perVertexStruct.gen_gl_ClipDistance;`?
In the worst case, we'd do what MSL backend does, and generate something like:
```
float _expr13[1] = { perVertexStruct.gen_gl_ClipDistance[0] };
```

Gordon-F:
@kvark @paulgb This PR can be closed. Fixed in #925.

PR #916, opened 2021-05-28
Include array bounds checks in SPIR-V output

jimblandy:
This is a recreation of #847, which became orphaned for technical GitHub reasons. I've copied over the unresolved comments here. Still a draft; remaining items:

- [x] Investigate possibly dropped `x` access for `bar.matrix[index].x` in `access.wgsl`.

- [x] Handle structs passed by pointer.

- [ ] Avoid generating many identical `OpArrayLength - 1` expressions.

- [ ] Handle nested structs.

- [ ] Handle arrays sized by pipeline-overrideable constants

- [x] @kvark 's request copied from #847: We need some sort of policy parameter like this:

          enum OutOfBoundsPolicy {
            Restrict, // may be faster than `Avoid`, but we'll need to bench thios
            Avoid,  // portable - matches D3D
            Ignore, // unsafe but low overhead, useful for WebGPU on native (behind a non-portable unsafe feature)
          }

    - [x] Generate proper WGSL zero values for all types.
    - [x] Make stores properly conditional.


jimblandy:
>     * [x]  Investigate possibly dropped `x` access for `bar.matrix[index].x` in `access.wgsl`.

For `matrix[index].x` we generate three SPIR-V instructions:

    %34 = OpAccessChain  %32  %18 %33 %3
    %35 = OpLoad  %13  %34
    %36 = OpCompositeExtract  %14  %35 0

instead of a single `OpAccessChain`, so I was worried. But the code is correct, if not ideal.

src/proc/mod.rs
```diff
@@ -108,6 +118,37 @@ impl super::TypeInner {
...
+            Self::Pointer { base, .. } => {
+                // When assigning types to expressions, ResolveContext::Resolve
+                // does a separate sub-match here instead of a full recursion,
+                // so we'll do the same.
+                let base_inner = &module.types[base].inner;
+                match *base_inner {
+                    Self::Vector { size, .. } => Some(size as _),
+                    Self::Matrix { columns, .. } => Some(columns as _),
+                    Self::Array { size, .. } => size.to_known_length(module),
+                    _ => unreachable!("pointed-to type does not have a length: {:?}", base_inner),
```
Gordon-F, line 146: How about return `None` here and add warn log with this message?

jimblandy: Hmm, `None` means a dynamically sized array; we take advantage of that in `back/spv/writer.rs`. Maybe this could return `Result<Option<u32>>`?

jimblandy:
Unfortunately, my code for removing redundant length computations was naive, so that had to come out.

But this version implements all three proposed bounds check policies.

jimblandy:
I think the remaining unchecked items are optional, so this is ready for review.

jimblandy:
Support for specialization-constant-sized arrays is still broken in this patch. It won't just fail with an error; it'll try to treat it like a runtime-sized array, and SPIR-V validation will fail. So that is a legit remaining unchecked item.

But in any case, it's ready for review on the rest.

src/proc/mod.rs
jimblandy: @kvark What approach would you recommend here?

src/back/mod.rs
```diff
@@ -1,5 +1,10 @@
 //! Functions which export shader modules into binary and text formats.
 
+#[cfg(feature = "deserialize")]
```
kvark, line 3: nit: if it's just one structure, much easier to write it as `serde::Deserialize` there direction instead of adding these lines

src/back/mod.rs
```diff
@@ -13,6 +18,39 @@ pub mod spv;
...
+    /// When presented with an out-of-bounds index check, reads return zero, and
+    /// writes have no effect.
+    ReadZeroSkipWrite,
+
+    /// Out-of-bounds indexes are undefined behavior. Generate the fastest code
+    /// possible.
+    UndefinedBehavior,
+}
+
+impl Default for IndexBoundsCheckPolicy {
```
kvark, line 48: This is a tough one to default... and I guess it has to be defaulted, since it's a part of `Options`, where a default is expected.
I think, instead of following the WebGPU requirements here, the default should follow the basic idea of explicit opt-in into more work or features. Since restricting or zeroing is extra code, it's a feature. So the default should be `UndefinedBehavior`.

src/back/spv/mod.rs
```diff
@@ -73,6 +76,7 @@ impl Default for Options {
             lang_version: (1, 0),
             flags,
             capabilities: None,
+            index_bounds_check_policy: super::IndexBoundsCheckPolicy::Restrict,
```
kvark, line 79: similarly here, we should default to `UndefinedBehavior`

src/back/spv/writer.rs
```diff
@@ -270,8 +297,6 @@ struct GlobalVariable {
     /// prelude block (and reset before every function) as `OpLoad` of the variable.
     /// It is then used for all the global ops, such as `OpImageSample`.
     handle_id: Word,
-    /// SPIR-V storage class.
-    class: spirv::StorageClass,
```
kvark, line 274: this goes away after rebase, I assume?

src/back/spv/writer.rs
```diff
@@ -1320,6 +1355,234 @@ impl Writer {
         })
     }
 
+    /// Emit code to compute the length of a run-time array.
+    ///
+    /// Given `array`, an expression referring to the final member of a struct,
+    /// where the member in question is a runtime-sized array, return the insn
```
kvark, line 1361: what does "insn" stand for?

src/back/spv/writer.rs
```diff
@@ -1320,6 +1355,234 @@ impl Writer {
...
+        array: Handle<crate::Expression>,
+        ir_module: &crate::Module,
+        ir_function: &crate::Function,
+        function: &Function,
+        block: &mut Block,
+    ) -> Result<Word, Error> {
+        let length_id =
+            self.write_runtime_array_length(array, ir_module, ir_function, function, block)?;
+
+        let limit_type = LocalType::Value {
```
kvark, line 1438: we could return this from `write_runtime_array_length` (as a part of the tuple), since it's already getting this type

src/back/spv/writer.rs
```diff
@@ -1320,6 +1355,234 @@ impl Writer {
...
+        ir_function: &crate::Function,
+        fun_info: &FunctionInfo,
+        function: &Function,
+        block: &mut Block,
+    ) -> Result<Word, Error> {
+        let array_ty = fun_info[array].ty.inner_with(&ir_module.types);
+        let index_id = self.cached[index];
+        let index_type_id = self.get_expression_type_id(&fun_info[index].ty)?;
+
+        let limit_id = if let Some(known_length) = array_ty.to_known_length(ir_module) {
```
kvark, line 1488: nit: doing `if let ... = ... { ... } else {..}` is a weird substitute for `match`

src/back/spv/writer.rs
```diff
@@ -1320,6 +1355,234 @@ impl Writer {
...
+            kind: crate::ScalarKind::Bool,
+            width: 1,
+            pointer_class: None,
+        };
+        let bool_type_id = self.get_type_id(bool_type.into())?;
+
+        // Compare the index against the length.
+        let condition_id = self.id_gen.next();
+        block.body.push(Instruction::binary(
+            spirv::Op::ULessThan,
```
kvark, line 1566: A bit of a problem here: the `write_runtime_array_length` result is actually "length(xx) - 1". So comparing with `<` is incorrect with it.

src/back/spv/writer.rs
```diff
@@ -1320,6 +1355,234 @@ impl Writer {
...
+            condition_id,
+            index_id,
+            length_id,
+        ));
+
+        // Indicate that we did generate the check.
+        Ok(Some(condition_id))
+    }
+
+    /// Decide whether to put off emitting insns for `expr_handle`.
```
kvark, line 1577: oh, "insns" stands for "instructions"? :)

src/back/spv/writer.rs
```diff
@@ -1357,40 +1623,133 @@ impl Writer {
...
+                        if let Some(comparison_id) = self.write_index_comparison(
+                            base,
+                            index,
+                            ir_module,
+                            ir_function,
+                            fun_info,
+                            function,
+                            block,
+                        )? {
+                            // Branch based on whether the index was in bounds.
```
kvark, line 1679: this is quite a bit of logic - it would be great to move it out into a helper

src/back/spv/writer.rs
```diff
@@ -1357,40 +1623,133 @@ impl Writer {
...
+                            //
+                            // As it turns out, our out-of-bounds branch block would
+                            // contain no instructions: it just produces a constant
+                            // zero, whose instruction is in the module's declarations
+                            // section at the front. In this case, SPIR-V lets us omit
+                            // the empty 'else' block, and branch directly to the
```
kvark, line 1685: a small win here!

src/back/spv/writer.rs
```diff
@@ -1768,19 +2127,77 @@ impl Writer {
...
+                )? {
+                    ExpressionPointer::Ready { pointer_id } => {
+                        let id = self.id_gen.next();
+                        block
+                            .body
+                            .push(Instruction::load(result_type_id, id, pointer_id, None));
+                        id
+                    }
+                    ExpressionPointer::Conditional { condition, access } => {
+                        let header_block = block.label_id;
```
kvark, line 2146: similarly, this code should be a candidate for moving out into a helper

src/proc/mod.rs
```diff
@@ -108,6 +118,37 @@ impl super::TypeInner {
...
         }
     }
+
+    /// Return the length of a subscriptable type, if known at compile time.
+    ///
+    /// The value returned is appropriate for bounds checks on subscripting. If
+    /// the length is dynamic (say, a WGSL runtime-sized array), then return
+    /// `None`.
+    ///
+    /// `self` must be a vector, matrix, or array.
```
kvark, line 128: or a pointer to any of these

src/proc/mod.rs
```diff
@@ -108,6 +118,37 @@ impl super::TypeInner {
...
+    pub fn to_known_length(&self, module: &super::Module) -> Option<u32> {
```
kvark, line 129: probably needs a rename to something like `array_access_length` to indicate that we are talking about array-like types here, specifically. Otherwise, it's not clear (judging by the name) why structures are excluded, for example

src/proc/mod.rs
kvark: Is this supposed to be called by the both the backends and the frontends? If it's just the backends, moving it to `back/mod.rs` and leaving the unreachable is fine.
If it's both, let's return a result

src/valid/function.rs
```diff
@@ -565,6 +565,22 @@ impl super::Validator {
                     name: argument.name.clone().unwrap_or_default(),
                 });
             }
+
+            // Pointers to unsized values are permitted, but only if they're structs
+            // with an unsized final member, not runtime-sized arrays.
+            let argument_inner = &module.types[argument.ty].inner;
+            if let crate::TypeInner::Pointer { base, class: _ } = *argument_inner {
+                let base_info = &self.types[base.index()];
+                if !base_info.flags.contains(TypeFlags::SIZED) {
```
kvark, line 574: would it be easier (and still correct) to simply ensure that `TypeFlags::DATA` is there for the base type?
if it's not a structure, the general type validation should see that and complain

tests/in/pointer-access.param.ron
```diff
@@ -0,0 +1,5 @@
+(
```
kvark, line 1: any particular reason we are adding a SPIR-V test here (on the input)?
I'd think that bounds checking code is 99% in the backends (as the way we are trying to get it here), so the format of the input test shouldn't matter, in which case we should strongly prefer WGSL on the input.

Can this logic be just merged into the existing `access.wgsl`?

tests/out/collatz.spvasm
```diff
@@ -1,7 +1,7 @@
 ; SPIR-V
 ; Version: 1.0
 ; Generator: rspirv
-; Bound: 59
+; Bound: 65
```
kvark, line 4: We don't want all outputs to show the bounds checks. let's enable it selectively for just a few, if anything, on top of the `access.wgsl`

kvark (changes requested):
Amazing work here! Excited to see this happening 🚀

src/proc/mod.rs
jimblandy: It's used in expression validation to do some checks at shader translation time. I'll make it return a `Result`.

src/back/mod.rs
jimblandy: I'd argue that undefined behavior is something people should have to opt into. It seems to me that the "damn the torpedos, full speed ahead" approach of `UndefinedBehavior` is the feature, and that predictable behavior, following the spec, is the boring status quo, that you get if you're not really paying attention or a newbie.

jimblandy: While I do want to push a bit for a safe default here, it's your call. If you think UB by default is a necessity, then I'm happy to make the change.

src/back/spv/writer.rs
jimblandy: It should, yes.

src/back/spv/writer.rs
jimblandy: "Instruction". I think it's a GCC-ism. I'll write it out.

src/proc/mod.rs
jimblandy: Makes sense. Renamed to `array_known_length`. I think `known` is an important component of the name.

src/back/mod.rs
kvark: Yes, I'm convinced Naga should be principled in opting into extra work, here, not safey.
To add more justification, Naga by itself can't generally provide safety, anyway. It relies on the host API, for example to pass the buffer sizes into internal data (like we do in Metal for the `ArrayLength`). The only layer that *can* guarantee safety is `wgpu`, and it's defaulting to safe behaviors. This defaulting includes configuring Naga in a way it needs.

jimblandy: Okay, sounds good.

src/valid/function.rs
jimblandy: You're right, type validation should forbid `ptr<_, array<f32>>`. (It doesn't, but we can fix that.) In that light, the existing check that argument types are sized should handle everything.

jimblandy: Actually, we must permit pointers to dynamically sized arrays, since otherwise we have no type for the intermediate `AccessIndex` expression that refers to a dynamically sized array appearing as the last member of a struct.

But making such pointers non-`DATA` (and actually requiring local variables to be `SIZED` `DATA`) prevents the cases we care about preventing.

tests/in/pointer-access.param.ron
jimblandy: This is SPIR-V only because our WGSL front end doesn't implement the `*` dereferencing operator yet. Once we implement that, it can be replaced with WGSL.

src/back/spv/writer.rs
jimblandy: I added a helper function for producing the unsigned int type.

src/back/spv/writer.rs
jimblandy: The nice thing about `if let else` is that it saves you a level of indentation, compared to match. But I can make this a match if you like.

src/back/spv/writer.rs
jimblandy: I don't think that's correct - `write_runtime_array_length` is the length of the array; it's the `_limit` functions that return one less than the length. I'll change the name of the `_limit` functions to `_max_index`, to make this clearer.

src/back/spv/writer.rs
jimblandy: I've removed this abbreviation.

src/back/spv/writer.rs
jimblandy: I took a shot at this, with, I think, mixed results. Let me know what you think of the current version.

cli/src/main.rs
```diff
@@ -62,6 +63,25 @@ fn main() {
...
+                        }
+                        other => {
+                            panic!(
+                                "Unrecognized '--index-bounds-check-policy' value: {:?}",
+                                other
+                            );
+                        }
+                    };
+                }
+                #[cfg(feature = "spv-in")]
```
Gordon-F, line 84: Accidentally added feature gate?

jimblandy: Oh dear. I'd better look over the rest of this PR more thoroughly. Thank you!

jimblandy:
Okay, fixed the garbage @Gordon-F noticed, and some debugging junk. This, I actually believe is worth reviewing.

src/back/spv/writer.rs
```diff
@@ -1318,6 +1400,494 @@ impl Writer {
...
+            function,
+            block,
+        )? {
+            KnownOrComputed::Known(known_length) => {
+                // When the length and index are known at compile time, we do the bounds check
+                // during validation, so there's no need to check anything here.
+                // Mimic the logic in `Validator::validate_expression`.
+                if let crate::Expression::Constant(index_k) = ir_function.expressions[index] {
+                    if let Some(known_index) = ir_module.constants[index_k].to_array_length() {
+                        return Ok(KnownOrComputed::Known(known_index));
```
kvark, line 1664: I'm confused about this. What is the expected type returned here? I'd think it's a boolean, i.e. the function returns something that is a boolean index comparison. But this branch returns an index. How so?

src/back/spv/writer.rs
```diff
@@ -1318,6 +1400,494 @@ impl Writer {
...
+
+        result_id
+    }
+
+    /// Emit code for bounds checks, per self.index_bounds_check_policy.
+    ///
+    /// Return a `BoundsCheckResult` indicating how the index should be
+    /// consumed. See that type's documentation for details.
+    #[allow(clippy::too_many_arguments)]
+    fn write_bounds_check(
```
kvark, line 1754: best function evar

src/back/spv/writer.rs
```diff
@@ -1318,6 +1400,494 @@ impl Writer {
...
+        ir_function: &crate::Function,
+        fun_info: &FunctionInfo,
+        function: &mut Function,
+        base: Handle<crate::Expression>,
+        index: Handle<crate::Expression>,
+        block: &mut Block,
+    ) -> Result<BoundsCheckResult, Error> {
+        Ok(match self.index_bounds_check_policy {
+            IndexBoundsCheckPolicy::Restrict => {
+                match self.write_restricted_index(
```
kvark, line 1766: why not return `BoundsCheckResult` from here?

src/back/spv/writer.rs
```diff
@@ -1318,6 +1400,494 @@ impl Writer {
...
+                    KnownOrComputed::Known(restricted_index) => {
+                        BoundsCheckResult::KnownInBounds(restricted_index)
+                    }
+                    KnownOrComputed::Computed(restricted_index_id) => {
+                        BoundsCheckResult::Computed(restricted_index_id)
+                    }
+                }
+            }
+            IndexBoundsCheckPolicy::ReadZeroSkipWrite => {
+                match self.write_index_comparison(
```
kvark, line 1784: similarly, here? I think returning `BoundsCheckResult` would address the other concern I raised, if it's feasible

src/proc/mod.rs
```diff
@@ -110,6 +118,45 @@ impl super::TypeInner {
     }
 }
 
+impl crate::TypeInner {
```
kvark, line 121: can we move all the indexable logic into `src/proc/index.rs` and re-export here?

tests/out/pointer-access.info.ron
```diff
@@ -0,0 +1,355 @@
+(
+    functions: [
```
kvark, line 2: do we need these RON snapshots for pointer-access? They are fairly verbose.
I think we should omit them unless there is a strong reason to include them

kvark (commented):
Great! The code is shaping up just beautifully ❤️ 
Just a couple organizational comments. Overall looking quite landable.

src/back/spv/writer.rs
```diff
@@ -1318,6 +1400,494 @@ impl Writer {
...
+    /// Compute the maximum valid index of a subscriptable value.
+    ///
+    /// Given `sequence`, an expression referring to some indexable type, return
+    /// its maximum valid index - one less than its length. The result may
+    /// either be computed, or known at shader translation time.
+    ///
+    /// `sequence` may be a `Vector`, `Matrix`, or `Array`, a `Pointer` to any
+    /// of those, or a `ValuePointer`. An array may be fixed-size, dynamically
+    /// sized, or use a specializable constant as its length.
+    fn write_sequence_max_index(
```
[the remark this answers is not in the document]
jimblandy: To avoid leaking a bunch of `writer`-local types, this needed to be a sub-module of `writer`: `pub(in ...)` can only restrict visibility to parents of the module containing the definition. So it's `src/back/spv/writer/index.rs`, which is a bit deep, but not the end of the world.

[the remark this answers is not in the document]
jimblandy: If making `index` a child module of `writer` is too much nesting, then there's the alternative of adding a bunch of non-`pub` `use` declarations to `src/back/spv/mod.rs`. I didn't explore that possibility.

src/proc/mod.rs
jimblandy: Sure.

tests/out/pointer-access.info.ron
jimblandy: Okay, sounds good.

jimblandy (commented):
Okay, I think I've addressed all the comments. New push in a sec.

jimblandy:
Okay, hmm, I wasn't expecting GitHub to treat my replies as a separate review. I'll be sure not to do that again.

Upshot was, all comments addressed as expected.

jimblandy:
Lovely. I can't reproduce those clippy complaints locally. Need to walk dogs, will fix once I get back.

src/back/spv/writer.rs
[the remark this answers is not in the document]
kvark: yeah, I don't think we need this level of nesting in either parts of Naga.
Writer isn't meant to be this huge sub-module of SPV-out. It's just accidental that it has all the logic today. Coincidentally, "writer-local" is pretty much "spv-in" namespace, since "writer" dominates it. So `pub(super)` would work ok here.

src/back/spv/writer.rs
```diff
@@ -1,12 +1,16 @@
+mod index;
```
kvark, line 1: can we make it just `back/spv/index.rs` instead of `back/spv/writer/index.rs`?

jimblandy:
(just trying to placate clippy and accommodate changes to error behavior on master; some review comments not yet addressed)

src/back/spv/writer.rs
jimblandy: I've implemented "approach C", and it was fine.

jimblandy:
Okay - that's all green, and with the indexing code in `src/back/spv/index.rs`. I think this is ready.

jimblandy:
For ease of review, I separated out moving `Writer` from `writer.rs` to `mod.rs` into its own commit, which (hopefully) includes no significant code changes. So if you look at the individual commits, you'll get a more meaningful view of the actual work.

jimblandy:
(rebase onto latest master)

src/back/spv/index.rs
```diff
@@ -0,0 +1,510 @@
+//! Bounds-checking for SPIR-V output.
+
+use super::{Block, Function, IdGenerator};
+use super::{Error, Instruction, Word, Writer};
```
kvark, line 4: nit: we generally use one `use` for each thing (like `super`)

kvark (approved):

kvark:
Waiting for #990 and the release happen first

PR #917, opened 2021-05-28
[spv-out]: Don't pass in the type arena where it's not needed.

jimblandy:
It turns out that the module's `Arena<Type>` was being passed around in a lot of places and never actually used. I guess Rust's unused variable detection was fooled by mutual recursion.

If we discover in the future that we do need the type arena somewhere deep in the depths of the call tree, this change would be a pain to undo. But I think it would be a lot more readable simply to let `Writer` hold a reference to the `Module` it's currently working on. There are ways to accomplish that while still using the same `Writer` for multiple `Module`s and retaining its allocations.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -364,33 +364,25 @@ impl Writer {
         Err(Error::MissingCapabilities(capabilities.to_vec()))
     }
 
-    fn get_type_id(
-        &mut self,
-        arena: &Arena<crate::Type>,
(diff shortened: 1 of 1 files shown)


kvark (approved):
Whoa, so `write_type_declaration_local` and `get_type_id` called each other, so Rust wasn't able to see that an argument is unused.
Great that you found it!

kvark:
published in 0.4.2

PR #923, opened 2021-05-29
[glsl-in ] Implicit conversion for expressions

JCapucho:
One of the major pieces missing from the frontend, this adds only local implicit conversion and not across functions(will be done in another PR)


kvark (approved):
Looks fancy and reasonable :)

PR #925, opened 2021-05-30
Fix `Expression::Load` generating

### LINKED ISSUE #910: Statements are incorrectly ordered from `mtl` and `glsl` backends.

Gordon-F:
One more attempt to fix #910. Also added this case as a snapshot.

Previous discussion in #914.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2130,7 +2130,18 @@ impl<'a, W: Write> Writer<'a, W> {
                 self.write_value_type(inner)?;
             }
         }
-        write!(self.out, " {} = ", name)?;
+
+        let base_ty_res = &ctx.info[handle].ty;
+        let resolved = base_ty_res.inner_with(&self.module.types);
+
+        // If rhs is a array type, we should write temp variable as a dynamic array
+        let array_str = if let TypeInner::Array {..} = *resolved {
+            "[]"
+        } else {
+            ""
+        };
+
+        write!(self.out, " {}{} = ", name, array_str)?;
         self.write_expr(handle, ctx)?;
         writeln!(self.out, ";")?;
         self.named_expressions.insert(handle, name);
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -521,7 +521,11 @@ impl<'a, W: Write> Writer<'a, W> {
                 // glsl has no concept of samplers so we just ignore it
                 TypeInner::Sampler { .. } => continue,
                 // All other globals are written by `write_global`
-                _ => self.write_global(handle, global)?,
+                _ => {
+                    self.write_global(handle, global)?;
+                    // Add a newline (only for readability)
+                    writeln!(self.out)?;
+                }
             }
         }
 
@@ -2135,7 +2139,7 @@ impl<'a, W: Write> Writer<'a, W> {
         let resolved = base_ty_res.inner_with(&self.module.types);
 
         // If rhs is a array type, we should write temp variable as a dynamic array
-        let array_str = if let TypeInner::Array {..} = *resolved {
+        let array_str = if let TypeInner::Array { .. } = *resolved {
             "[]"
         } else {
             ""
--- a/src/back/mod.rs
+++ b/src/back/mod.rs
@@ -28,6 +28,10 @@ impl crate::Expression {
             crate::Expression::ImageSample { .. } | crate::Expression::ImageLoad { .. } => 1,
             // derivatives use the control flow
             crate::Expression::Derivative { .. } => 1,
+            // TODO: We need a better fix for named `Load` expressions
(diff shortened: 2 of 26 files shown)


src/back/mod.rs
```diff
@@ -20,16 +20,24 @@ impl crate::Expression {
...
+    fn bake_ref_count(&self, expressions: &crate::Arena<crate::Expression>) -> usize {
+        const OTHER_EXPR_BAKE_COUNT: usize = 2;
         match *self {
             // accesses are never cached, only loads are
             crate::Expression::Access { .. } | crate::Expression::AccessIndex { .. } => !0,
             // sampling may use the control flow, and image ops look better by themselves
             crate::Expression::ImageSample { .. } | crate::Expression::ImageLoad { .. } => 1,
             // derivatives use the control flow
             crate::Expression::Derivative { .. } => 1,
+            crate::Expression::Load { pointer } => match expressions[pointer] {
```
kvark, line 32: This is not a complete fix, right? The expression can also be an access into something that is a variable, or it can be a global variable. In all the cases we want to load early.
So technically, returning `1` here is the only good solution. Why are we not doing this?

kvark (changes requested):

src/back/mod.rs
Gordon-F: It's a temporary solution. We can safely add Global Variable here without any validation error. I will try to investigate how we can fix glsl array and named expression on wgsl frontend side.

kvark: What about `Access` and `AccessIndex`?

I'm not sure how you are seeing this being addressed on WGSL frontend. This seems like a backend issue.

Gordon-F:
Ok, I guess I should fix #851 first.

Gordon-F:
@kvark CI is happy now.

src/back/glsl/mod.rs
```diff
@@ -521,7 +521,11 @@ impl<'a, W: Write> Writer<'a, W> {
                 // glsl has no concept of samplers so we just ignore it
                 TypeInner::Sampler { .. } => continue,
                 // All other globals are written by `write_global`
-                _ => self.write_global(handle, global)?,
+                _ => {
+                    let end_with_newline =
+                        index != self.module.global_variables.len().saturating_sub(1);
```
kvark, line 526: This isn't going to work, since we are skipping the unused globals anyway. So your item may end up being the last in the list.

src/back/glsl/mod.rs
```diff
@@ -769,6 +774,10 @@ impl<'a, W: Write> Writer<'a, W> {
         };
         writeln!(self.out, ";")?;
 
+        if add_a_newline {
```
kvark, line 777: seems a bit weird to add this extra parameter and bit of logic, given that it can be done outside of this function. Is this to re-use it better?

src/back/mod.rs
kvark: Wouldn't it be simpler to just return 1 here without checking the actual expressions? I don't understand what this gives us.

tests/in/baked-load.wgsl
```diff
@@ -0,0 +1,16 @@
+[[block]]
+struct IOData {
```
kvark, line 2: would it be possible to merge this into `access.wgsl`?

JCapucho:
If i may ask why do you want to add a newline between globals in `glsl-out` and why do you check if it's the last global to write a newline just writing always a newline would in the worst case add two newlines and even then I don't think it will because you need to write the varyings

Gordon-F:
@kvark, @JCapucho Thank you! All suggestions addressed. I hope we will find a way how we can avoid unnecessary code generation.

PR #933, opened 2021-06-01
[glsl-in] track entry points arguments use

### LINKED ISSUE #931: [glsl-in] expression already in scope with constant above function
```glsl
#version 450

const int constant = 10;

float function() {
    return 0.0;
}
```
fails to validate with the following error:
```
Function [1] 'function' is invalid:
	Expression [1] can't be introduced - it's already in scope
```
Changing the `function` to return `void` or moving the constant below the function "fixes" the error.

This is the module it generates:
```rust
Module {
    types: {
        [1]: Type {
            name: None,
            inner: Scalar {
                kind: Sint,
                width: 4,
            },
        },
        [2]: Type {
            name: None,
            inner: Scalar {
                kind: Float,
                width: 4,
            },
        },
    },
    constants: {
        [1]: Constant {
            name: None,
            specialization: None,
            inner: Scalar {
                width: 4,
                value: Sint(
                    10,
                ),
            },
        },
        [2]: Constant {
            name: None,
            specialization: None,
            inner: Scalar {
                width: 4,
                value: Float(
                    0.0,
                ),
            },
        },
    },
    global_variables: {},
    functions: {
        [1]: Function {
            name: Some(
                "function",
            ),
            arguments: [],
            result: Some(
                FunctionResult {
                    ty: [2],
                    binding: None,
                },
            ),
            local_variables: {},
            expressions: {
                [1]: Constant(
                    [1],
                ),
                [2]: Constant(
                    [2],
                ),
            },
            body: [
                Emit(
                    [1..1],
                ),
                Return {
                    value: Some(
                        [2],
                    ),
                },
            ],
        },
    },
    entry_points: [],
}
```

JCapucho:
- Fixes problems with multiple entry points
- Fixes the emitting of constants (fixes #931)
- Reenables the snapshots tests :)

### DIFF
--- a/src/front/glsl/functions.rs
+++ b/src/front/glsl/functions.rs
@@ -1,8 +1,7 @@
 use crate::{
-    proc::ensure_block_returns, Arena, BinaryOperator, Binding, Block, BuiltIn, EntryPoint,
-    Expression, Function, FunctionArgument, FunctionResult, Handle, MathFunction,
-    RelationalFunction, SampleLevel, ScalarKind, ShaderStage, Statement, StructMember,
-    SwizzleComponent, Type, TypeInner,
+    proc::ensure_block_returns, Arena, BinaryOperator, Block, EntryPoint, Expression, Function,
+    FunctionArgument, FunctionResult, Handle, MathFunction, RelationalFunction, SampleLevel,
+    ScalarKind, Statement, StructMember, SwizzleComponent, Type, TypeInner,
 };
 
 use super::{ast::*, error::ErrorKind, SourceMetadata};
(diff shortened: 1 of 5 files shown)


src/front/glsl/functions.rs
```diff
@@ -462,17 +465,86 @@ impl Program<'_> {
...
+                ref continuing,
+            } => {
+                for stmt in body.iter().chain(continuing.iter()) {
+                    self.check_call_global(caller, function_arg_use, stmt)
+                }
+            }
+            Statement::Call { function, .. } => {
+                let callee_len = function_arg_use[function.index()].len();
+                let caller_len = function_arg_use[caller.index()].len();
+                function_arg_use[caller.index()].extend(
```
kvark, line 513: Since the function is mutating, it shouldn't be innocently called `check_call_global`. Instead, perhaps something like `patch_calls`?

kvark (approved):

src/front/glsl/functions.rs
kvark: we can leave it for the follow up, since no other concerns are here to block this

PR #934, opened 2021-06-01
Glsl error message improvements

jakobhellermann:
Before:
image
After:
image

### DIFF
--- a/src/front/glsl/error.rs
+++ b/src/front/glsl/error.rs
@@ -69,8 +69,8 @@ pub enum ErrorKind {
     #[error("Unknown layout qualifier: {1}")]
     UnknownLayoutQualifier(SourceMetadata, String),
     #[cfg(feature = "glsl-validate")]
-    #[error("Variable already declared: {0}")]
-    VariableAlreadyDeclared(String),
+    #[error("Variable already declared: {1}")]
+    VariableAlreadyDeclared(SourceMetadata, String),
     #[error("{1}")]
     SemanticError(SourceMetadata, Cow<'static, str>),
 }
@@ -85,6 +85,8 @@ impl ErrorKind {
             | ErrorKind::UnknownLayoutQualifier(metadata, _)
             | ErrorKind::SemanticError(metadata, _)
             | ErrorKind::UnknownField(metadata, _) => Some(metadata),
+            #[cfg(feature = "glsl-validate")]
+            ErrorKind::VariableAlreadyDeclared(metadata, _) => Some(metadata),
             ErrorKind::InvalidToken(ref token, _) => Some(token.meta),
             _ => None,
         }
--- a/src/front/glsl/error.rs
+++ b/src/front/glsl/error.rs
@@ -1,9 +1,51 @@
 use super::{
     constants::ConstantSolvingError,
-    token::{SourceMetadata, Token},
+    token::{SourceMetadata, Token, TokenValue},
 };
 use std::borrow::Cow;
 use thiserror::Error;
+
+fn join_with_comma(list: &[ExpectedToken]) -> String {
+    let mut string = "".to_string();
+    for (i, val) in list.iter().enumerate() {
+        string.push_str(&val.to_string());
+        match i {
+            i if i == list.len() - 1 => {}
+            i if i == list.len() - 2 => string.push_str(" or "),
+            _ => string.push_str(", "),
+        }
+    }
+    string
+}
+
+#[derive(Debug, PartialEq)]
+pub enum ExpectedToken {
+    Token(TokenValue),
+    TypeName,
+    Identifier,
+    IntLiteral,
+    FloatLiteral,
+    BoolLiteral,
+    EOF,
+}
+impl From<TokenValue> for ExpectedToken {
+    fn from(token: TokenValue) -> Self {
+        ExpectedToken::Token(token)
+    }
+}
+impl std::fmt::Display for ExpectedToken {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match *self {
+            ExpectedToken::Token(ref token) => write!(f, "{:?}", token),
+            ExpectedToken::TypeName => write!(f, "a type"),
+            ExpectedToken::Identifier => write!(f, "identifier"),
+            ExpectedToken::IntLiteral => write!(f, "integer literal"),
+            ExpectedToken::FloatLiteral => write!(f, "float literal"),
+            ExpectedToken::BoolLiteral => write!(f, "bool literal"),
+            ExpectedToken::EOF => write!(f, "end of file"),
+        }
+    }
+}
 
 #[derive(Debug, Error)]
 #[cfg_attr(test, derive(PartialEq))]
@@ -14,8 +56,8 @@ pub enum ErrorKind {
     InvalidProfile(SourceMetadata, String),
     #[error("Invalid version: {1}")]
     InvalidVersion(SourceMetadata, u64),
-    #[error("Unexpected token: {0}")]
-    InvalidToken(Token),
+    #[error("Expected {}, found {0}", join_with_comma(.1))]
+    InvalidToken(Token, Vec<ExpectedToken>),
     #[error("Not implemented {0}")]
     NotImplemented(&'static str),
     #[error("Unknown variable: {1}")]
@@ -43,7 +85,7 @@ impl ErrorKind {
             | ErrorKind::UnknownLayoutQualifier(metadata, _)
             | ErrorKind::SemanticError(metadata, _)
             | ErrorKind::UnknownField(metadata, _) => Some(metadata),
-            ErrorKind::InvalidToken(ref token) => Some(token.meta),
+            ErrorKind::InvalidToken(ref token, _) => Some(token.meta),
             _ => None,
         }
     }
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -11,9 +11,9 @@ use super::{
     Program,
 };
 use crate::{
-    arena::Handle, Arena, ArraySize, BinaryOperator, Block, Constant, ConstantInner, Expression,
-    Function, FunctionResult, ResourceBinding, ScalarValue, Statement, StorageClass, StructMember,
-    SwitchCase, Type, TypeInner, UnaryOperator,
+    arena::Handle, front::glsl::error::ExpectedToken, Arena, ArraySize, BinaryOperator, Block,
+    Constant, ConstantInner, Expression, Function, FunctionResult, ResourceBinding, ScalarValue,
+    Statement, StorageClass, StructMember, SwitchCase, Type, TypeInner, UnaryOperator,
 };
 use core::convert::TryFrom;
 use std::{iter::Peekable, mem};
@@ -38,7 +38,10 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
 
         match token.value {
             TokenValue::Identifier(name) => Ok((name, token.meta)),
-            _ => Err(ErrorKind::InvalidToken(token)),
+            _ => Err(ErrorKind::InvalidToken(
+                token,
+                vec![ExpectedToken::Identifier],
+            )),
         }
     }
 
@@ -46,7 +49,7 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
         let token = self.bump()?;
 
         if token.value != value {
-            Err(ErrorKind::InvalidToken(token))
+            Err(ErrorKind::InvalidToken(token, vec![value.into()]))
         } else {
             Ok(token)
         }
@@ -90,7 +93,12 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
                 440 | 450 | 460 => self.program.version = i.value as u16,
                 _ => return Err(ErrorKind::InvalidVersion(version.meta, i.value)),
             },
-            _ => return Err(ErrorKind::InvalidToken(version)),
+            _ => {
+                return Err(ErrorKind::InvalidToken(
+                    version,
+                    vec![ExpectedToken::IntLiteral],
+                ))
+            }
         }
 
         let profile = self.lexer.peek();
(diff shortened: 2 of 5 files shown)


src/front/glsl/error.rs
```diff
@@ -14,8 +22,8 @@ pub enum ErrorKind {
     InvalidProfile(SourceMetadata, String),
     #[error("Invalid version: {1}")]
     InvalidVersion(SourceMetadata, u64),
-    #[error("Unexpected token: {0}")]
-    InvalidToken(Token),
+    #[error("Expected {1}, got token '{0}'")]
+    InvalidToken(Token, Cow<'static, str>),
```
JCapucho, line 26: We should store a `Vec` of the possible tokens here instead of a string and the message should be `Expected {0}, found {1}` since it's more compact and doesn't involve the user with implementation details like tokens

src/front/glsl/error.rs
```diff
@@ -73,6 +82,27 @@ pub struct ParseError {
     pub kind: ErrorKind,
 }
 
+impl ParseError {
+    fn diagnostic(&self) -> Diagnostic<()> {
+        match self.kind.metadata() {
+            Some(metadata) => Diagnostic::error()
+                .with_message("failed to parse glsl")
```
JCapucho, line 89: The message should be the error message itself

src/front/glsl/error.rs
```diff
@@ -73,6 +82,27 @@ pub struct ParseError {
...
+                .with_labels(vec![Label::primary((), metadata.start..metadata.end)
+                    .with_message(self.kind.to_string())]),
```
JCapucho, line 91: Since the message would be in the message there should be no message here

src/front/glsl/error.rs
```diff
@@ -73,6 +82,27 @@ pub struct ParseError {
...
+            None => Diagnostic::error().with_message(self.kind.to_string()),
+        }
+    }
+
+    /// Emits a summary of the error to standard error stream.
+    pub fn emit_to_stderr(&self, source: &str) {
+        let files = SimpleFile::new("glsl", source);
```
JCapucho, line 98: Here the file name should be supplied by the user

src/front/glsl/error.rs
```diff
@@ -73,6 +82,27 @@ pub struct ParseError {
...
+        let config = codespan_reporting::term::Config::default();
+        let writer = StandardStream::stderr(ColorChoice::Always);
```
JCapucho, line 100: This will mess up log files and piping please use `Auto`

JCapucho (changes requested):
Besides my personal hate for error reporting in libraries there are some problems especially with the expected case in the `InvalidToken` error (which should probably be renamed to `UnexpectedToken`), great work nonetheless :)

Also I think that in the long run we should rename `ParseError` to `Error` and add a `meta` field to it so that all errors are forced to have metadata information, but that's for another PR.

jakobhellermann:
> I personally don't like the library being responsible for error reporting

I'm fine with removing the codespan reporting from this PR and having it just provide more context to the errors, I agree that it isn't really the responsibility of the library.

Currently the `wgsl` frontend also has `emit_to_stderr` and `emit_to_string` functions, do you think they should be moved to `bin/naga.rs` instead?

JCapucho:
> > I personally don't like the library being responsible for error reporting
> 
> I'm fine with removing the codespan reporting from this PR and having it just provide more context to the errors, I agree that it isn't really the responsibility of the library.
> 
> Currently the `wgsl` frontend also has `emit_to_stderr` and `emit_to_string` functions, do you think they should be moved to `bin/naga.rs` instead?

No need to remove it, it's just my personal opinion and the other maintainers already showed support for it being included in the library and it's already included in wgsl as you said but I think we should at least make it optional since some end users may only want to compile their already known working shaders and bringing in another dependency would only increase their compile times with no real benefit

Cargo.toml
```diff
@@ -30,7 +30,7 @@ pp-rs = { git = "https://github.com/Kangz/glslpp-rs.git", optional = true, rev =
 [features]
 default = []
 dot-out = []
-glsl-in = ["pp-rs"]
+glsl-in = ["pp-rs", "codespan-reporting"]
```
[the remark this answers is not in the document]
kvark: For context, WGSL links to it because WebGPU needs to report good errors.
Contrary, Naga's GLSL frontent is likely to be used in glsl-to-wgsl or glsl-to-spirv converters. These should prioritize code size (for WASM), so for GLSL frontend rich error reporting should be optional.

bin/naga.rs
```diff
@@ -139,7 +146,10 @@ fn main() {
                     defines: Default::default(),
                 },
             )
-            .unwrap_pretty()
+            .unwrap_or_else(|err| {
```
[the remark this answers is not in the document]
jakobhellermann: I'm not sure I understand. With `unwrap_pretty`, the `print_err` function would be used instead of the fancy error reporting, no?

Cargo.toml
```diff
@@ -41,10 +41,12 @@ spv-out = ["spirv"]
 wgsl-in = ["codespan-reporting"]
 wgsl-out = []
 hlsl-out = []
+error-reporting = ["codespan-reporting"]
```
jakobhellermann, line 44: I made the reporting optional but I'm not sure it is the best way forward.
Firstly, it requires `required-features` to be added to `bin/naga.rs` so every invocation must use `cargo run --features error-reporting,...`, which is a bit annoying. Also, I agree with @JCapucho that it isn't the responsibility of the library but rather of the user to handle error reporting (maybe someone already has their own infrastructure for that or want's more flexibilty).

Maybe moving the error reporting to `bin/naga.rs` and having no feature to enable it in `naga` itself would be the better move. What do you think?

kvark: If we can move it fully into bin/naga, that's absolutely the best, yes.

Cargo.toml
```diff
@@ -41,10 +41,12 @@ spv-out = ["spirv"]
...
 
 [[bin]]
 name = "naga"
 path = "bin/naga.rs"
+required-features = ["error-reporting"]
```
Gordon-F, line 49: Is this needed? If we keep this, cargo will require us to pass this feature all the time.

For example:
```
cargo run --features wgsl-in,wgsl-out test.wgsl test_gen.wgsl module.txt

warning: invalid feature `error-reporting` in required-features of target `naga`: `error-reporting` is not present in [features] section
error: target `naga` in package `naga` requires the features: `error-reporting`
Consider enabling them by passing, e.g., `--features="error-reporting"`
```

Cargo.toml
jakobhellermann: I moved it to naga.rs but there is one annoying issue, there seems to be no way to have dependencies (codespan-reporting) for `bin`s: https://stackoverflow.com/questions/35711044/how-can-i-specify-binary-only-dependencies.
The solutions listed on stack overflow are 
1. move the executable to an example
2. use required-features and a feature just for the bin
3. make the binary it's own package

kvark: I see. We've been fighting with it quite a bit.
Sounds like we still need the `glsl-error-reporting` feature (note the glsl in the name) then.

jakobhellermann: For what its worth, for option 3. with a separate `naga-cli` crate, `cargo run` would still work in root directory if we specify 
```toml
[workspace]
members = [".", "naga-cli"]
default-members = ["naga-cli"]
```

https://doc.rust-lang.org/cargo/reference/workspaces.html#package-selection

kvark: oh interesting, I didn't know a "default-members" is a thing!
The other problem with moving it into a crate is that essentially we'd have to duplicate all the features.
Overall, it's a whole different discussion. Let's have it, but maybe separately from this PR.

jakobhellermann: Opened a PR for discussion: #938

kvark:
Thank you! Changes look reasonable. Let's get @JCapucho having another look.

JCapucho:
I would like to fix the way the expected tokens are handled in invalid token before merging

src/front/glsl/error.rs
jakobhellermann: Done.

src/front/glsl/error.rs
```diff
@@ -14,8 +27,8 @@ pub enum ErrorKind {
     InvalidProfile(SourceMetadata, String),
     #[error("Invalid version: {1}")]
     InvalidVersion(SourceMetadata, u64),
-    #[error("Unexpected token: {0}")]
-    InvalidToken(Token),
+    #[error("Expected {}, found '{0}'", join_with_comma(.1))]
+    InvalidToken(Token, Vec<Cow<'static, str>>),
```
JCapucho, line 31: What I meant here was a `Vec<Token>`, this makes it so that if any user wants to handle the error in their code they can deal with the enum instead of a stringified representation and since in the future i intend to implement `Display` in `Token` you get the same output. And it also makes it easier to handle in the `expect` method in parser

src/front/glsl/parser.rs
```diff
@@ -342,7 +350,16 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
...
             // TODO: handle Shared?
-            _ => Err(ErrorKind::InvalidToken(token)),
+            _ => Err(ErrorKind::InvalidToken(
+                token,
+                vec![
+                    "identifier".into(),
+                    "push_constant".into(),
+                    "std140".into(),
+                    "std430".into(),
+                    "early_fragment_tests".into(),
```
JCapucho, line 360: We should only use identifier here since unknown layout directives is another error

JCapucho (changes requested):
I'm sorry I didn't express my intent clearly but when I meant a list of tokens I was talking about a literal `Vec<Token>`

src/front/glsl/error.rs
jakobhellermann: How would `InvalidToken(_, vec!["identifier".into()])` work in that case? Use a `TokenValue::Identifier("".to_string())` and special case identifiers to not print their inner string in the error?

JCapucho: Exactly, we could maybe add a check to see if the string isn't empty but may original idea was to just print the enum kind and not it's content this would also be true for integers

jakobhellermann: One alternative to dummy values for `Identifier`, `TypeName` etc. is what wgsl does:
Have an
```rust
enum ExpectedToken {
  Token(TokenValue),
  Identifier,
  TypeName,
  ...
}
```
and use that instead.

I implemented this in https://github.com/gfx-rs/naga/pull/934/commits/bccd6a8e0d1d26b510e2b43f002d309968012626, what do you think?

jakobhellermann: The wgsl frontend doesn't print the enum variant names `Identifier`, `Float`, and `Paren` but `identifier`, `floating point literal` and `(` instead. Show I do that for the glsl frontend as well?

JCapucho: I think the names should a little more user friendly (like the ones in the wgsl frontend as you mentioned) and instead of `expected_str` you should implement `Display`. For tokens if you want to add pretty printing please do it in a separate PR since it's quite a big change

jakobhellermann: [Done](https://github.com/gfx-rs/naga/pull/934/commits/1dee90ece62ceb6bd5c6bb2d0c623c306f87a238)

JCapucho:
Looks good to me, let's just have @kvark take a look, I'm going to squash this afterward unless you want to rebase the broken commits with the ci fixes

jakobhellermann:
I rebased the commits.

src/front/glsl/parser.rs
```diff
@@ -405,7 +425,10 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
             let token = self.bump()?;
             match token.value {
                 TokenValue::Semicolon if self.program.version == 460 => Ok(()),
-                _ => Err(ErrorKind::InvalidToken(token)),
+                _ => Err(ErrorKind::InvalidToken(
+                    token,
+                    vec![TokenValue::Semicolon.into()],
```
JCapucho, line 430: This is version dependent

src/front/glsl/parser.rs
```diff
@@ -697,7 +735,14 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
 
                                     Ok(true)
                                 }
-                                _ => Err(ErrorKind::InvalidToken(token)),
+                                _ if external => Err(ErrorKind::InvalidToken(
+                                    token,
+                                    vec![TokenValue::LeftBrace.into()],
```
JCapucho, line 740: prototypes are also allowed here so you must add `Semicolon`

src/front/glsl/variables.rs
```diff
@@ -515,15 +515,17 @@ impl Program<'_> {
...
     ) -> Result<Handle<Expression>, ErrorKind> {
         #[cfg(feature = "glsl-validate")]
         if let Some(ref name) = name {
             if ctx.lookup_local_var_current_scope(name).is_some() {
-                return Err(ErrorKind::VariableAlreadyDeclared(name.clone()));
+                return Err(ErrorKind::VariableAlreadyDeclared(meta, name.clone()));
             }
         }
+        #[cfg(not(feature = "glsl-validate"))]
+        let _ = meta;
```
JCapucho, line 528: nit: maybe a allow unused is better here

JCapucho (changes requested):
I've double checked the PR and found some small things but we can merge it anyways and fix it later

JCapucho:
Please fix CI and if possible squash the last commit and I'll merge this

src/front/glsl/variables.rs
JCapucho: Please apply it only to the variable in question and not the whole struct

src/front/glsl/parser.rs
JCapucho: Since you added eof please make it always part of the expected list

src/front/glsl/variables.rs
jakobhellermann: I tried, but the warning still exists when only applied to the one variable.

JCapucho: Oh ok then let's leave it as his

JCapucho (approved):

JCapucho:
Thank you for your contribution

PR #936, opened 2021-06-02
Insert attribute glsl location qualifier for opengl ES 3.0

Frizi:
GLSL ES 3.0 allows location to be specified on attributes. This is not allowed for varyings, hence the extra guard against interpolations.

This change fixes wgpu's Cube example in WebGL2 (and possibly others, but I haven't tested that yet), particularily on MacOS, where the driver likes to assign locations out of order.

### DIFF
--- a/tests/out/quad-vert.Vertex.glsl
+++ b/tests/out/quad-vert.Vertex.glsl
@@ -21,7 +21,7 @@ struct gen_gl_PerVertex_block_0 {
 vec2 a_pos = vec2(0, 0);
 layout(location = 1) in vec2 _p2vs_location1;
 layout(location = 0) in vec2 _p2vs_location0;
-smooth layout(location = 0) out vec2 _vs2fs_location0;
+smooth out vec2 _vs2fs_location0;
 
 void main1() {
     v_uv = a_uv;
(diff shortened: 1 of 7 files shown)


kvark:
@Frizi please run `make` (or `cargo test --all-features`) to update the reftest output, so that it's included in the PR.

Frizi:
Reftests updated.

tests/out/quad-vert.Vertex.glsl
```diff
@@ -21,7 +21,7 @@ struct gen_gl_PerVertex_block_0 {
 vec2 a_pos = vec2(0, 0);
 layout(location = 1) in vec2 _p2vs_location1;
 layout(location = 0) in vec2 _p2vs_location0;
-smooth layout(location = 0) out vec2 _vs2fs_location0;
+smooth out vec2 _vs2fs_location0;
```
kvark, line 24: I'm confused here. Why are we removing the output locations from these varyings now?

kvark (changes requested):

tests/out/quad-vert.Vertex.glsl
Frizi: This is not supported under webgl. According to this wiki, it requires at least version 4.1 and `ARB_separate_shader_objects`. I'm not sure if we care about that right now. Varying locations are not required when you always link all the stages together.

kvark (approved):
We'd still want at least one of those snapshots to show how varying locations are assigned (by selecting the output GLSL version in the xxx.param.ron), but otherwise the reasoning is all good

PR #937, opened 2021-06-02
Add ability to generate named expressions

Gordon-F:
These changes allowed to generated expressions with correct names provided by a frontend. If a frontend can parse variable name (for example `let` in `wgsl`) it could add a map of names and expressions inside the `Function` structure.

This PR implements such functionality for `wgsl-in` and `wgsl-out`, `glsl-out`, but it can be implemented for all fronts/backs.

Part of #910, #392 (not fixed, because we should implement this for all fronts/backs).

Example of generated shader:
```wgsl
[[stage(fragment)]]
fn main([[location(0)]] uv : vec2<f32>) -> [[location(0)]] vec4<f32> {
  let color = textureSample(u_texture, u_sampler, uv);
  if (color.a == 0.0) {
    discard;
  }
  // forcing the expression here to be emitted in order to check the
  // uniformity of the control flow a bit more strongly.
  let premultiplied = color.a * color;
  return premultiplied;
}
```

Before:
```glsl
void main() {
    vec2 uv2 = _vs2fs_location0;
    vec4 _expr4 = texture(_group_0_binding_0, vec2(uv2));
    if((_expr4.w == 0.0)) {
        discard;
    }
    _fs2p_location0 = (_expr4.w * _expr4);
    return;
}
```

After:
```glsl
void main() {
    vec2 uv2 = _vs2fs_location0;
    vec4 color = texture(_group_0_binding_0, vec2(uv2));
    if((color.w == 0.0)) {
        discard;
    }
    vec4 premultiplied = (color.w * color);
    _fs2p_location0 = premultiplied;
    return;
}
```


src/front/wgsl/mod.rs
```diff
@@ -1004,13 +1004,16 @@ impl std::error::Error for ParseError {
 pub struct Parser {
     scopes: Vec<Scope>,
     lookup_type: FastHashMap<String, Handle<crate::Type>>,
+    /// Map of expressions that have associated variable names
+    named_expressions: FastHashMap<Handle<crate::Expression>, String>,
```
kvark, line 1008: Can we avoid this? It's a per-function thing, and keeping it at "global" state here is a bit undesired. Instead, when parsing a function, a hashmap needs to be created on the stack, then passed to the `StatementContext`. After the function is done parsing, it would just be moved into the produced function.

tests/snapshots.rs
```diff
@@ -297,6 +297,10 @@ fn convert_wgsl() {
             "globals",
             Targets::SPIRV | Targets::METAL | Targets::GLSL | Targets::WGSL,
         ),
+        (
+            "baked-load",
```
kvark, line 301: is a new test really necessary here? I'd think that this code already affects all the snapshot tests. I don't think it needs to be tested exclusively.

kvark (changes requested):
Very nice work!
The tough part is figuring out how exactly to represent the errors in IR.
It could be `struct Expression { name: Option<String>, inner: ExpressionInner }`.
It could be an array of `Option<String>`, one for each expression.
Or it could be a hashmap. I think the hashmap is our best shot right now, and this is what PR does.

Gordon-F:
> It could be struct Expression { name: Option<String>, inner: ExpressionInner }.
> It could be an array of Option<String>, one for each expression.
> Or it could be a hashmap

I tried all of these solutions a little bit. Hashmap inside the `Function` structure required fewer works and can be implemented easier step by step in all frontend/backend.

tests/snapshots.rs
Gordon-F: This test case is based on #910. If we didn't catch this error previously, we need this case, to be sure we don't break it again.

kvark: I see the change here in this PR as pure formatting. There is no semantical changes expected to be here.
So adding a test case for #910 is not something that should be a part of this.

src/back/glsl/mod.rs
```diff
@@ -1505,7 +1495,12 @@ impl<'a, W: Write> Writer<'a, W> {
     /// # Notes
     /// Doesn't add any newlines or leading/trailing spaces
     fn write_expr(&mut self, expr: Handle<Expression>, ctx: &FunctionCtx<'_>) -> BackendResult {
-        if let Some(name) = self.cached_expressions.get(&expr) {
+        if let Some(name) = self.named_expressions.get(&expr) {
+            write!(self.out, "{}", name)?;
+            return Ok(());
+        }
+
+        if let Some(name) = self.named_expressions.get(&expr) {
```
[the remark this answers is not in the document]
Gordon-F: Where is Clippy? He's the hero we deserve. 😄

tests/snapshots.rs
Gordon-F: Ok, let's look at the wgsl code:
```wgsl
var a: f32 = 0.0;
// We should check that backed doesn't skip this expression
let b: f32 = a;
a = 1.0;
v_data.data = b;
```

Previously we are skipped `let b: f32 = a;` because it's has not enough ref count to be a baked expression. But now can track every `let` expressions and generate correct code without any additional unnecessary code. (like we tried in #910 and #925).

kvark: Ok, thank you for explaining. I think there is a bit of confusion about what this PR does.

I want it to be in line with all the other names we have in IR: all of them are non-semantical and optional. This matches SPIR-V. In this sense, the PR itself is non-semantical.

I.e. adding a name to an expression doesn't change the meaning of the program (in terms of IR semantics). This PR isn't here to fix the issue. The issue needs to be fixed by adjusting the bake ref count, instead - #925.

Gordon-F: Agree with you, thank you! I will remove this part from PR and will update #925. But we definitely need to figure out how to reduce the amount unnecessary code in such case.

kvark:
Actually, there might be a bit of a problem here. So names are non-semantical. Technically, they can be invalid symbols for the target encoding. They can also repeat:
```
let a1 = 1;
let a2 = 2;
let c = a1;
```
If the IR has `a1` and `a2` expressions with the same name, that's still a valid IR. But when we output WGSL/MSL/GLSL, we'd end up with `let c = a2;`, accidentally.

So all names have to go through the `Namer` in the backend (if it's not done already).

Gordon-F:
We are skipped all this let in IR (if they not used at all):
```wgsl
let a1 = 1.0;
let a2 = 2.0;
let c = a1 + a2;
return vec4<f32>(a1, 1.0, 1.0, 1.0);
```

Generates to:
```
let c: f32 = (1.0 + 2.0);
return vec4<f32>(1.0, 1.0, 1.0, 1.0);
```

Should we have something like `Expression::Assignment` in IR ?

Anyway, we should use `Namer`, because we can have names that are banned on the backend side (`gl_*`, for example).

kvark:
Definitely NO assignment expressions, please!

Gordon-F:
@kvark I think all suggestions are addressed.

src/back/glsl/mod.rs
```diff
@@ -1187,30 +1192,18 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                            TypeResolution::Value(ref inner) => {
-                                self.write_value_type(inner)?;
-                            }
-                        }
                         let name = format!("_expr{}", handle.index());
-                        write!(self.out, " {} = ", name)?;
-                        self.write_expr(handle, ctx)?;
-                        writeln!(self.out, ";")?;
-                        self.cached_expressions.insert(handle, name);
+                        self.write_named_expr(handle, name.clone(), ctx)?;
```
kvark, line 1206: doesn't need a `clone()` here

src/back/glsl/mod.rs
```diff
@@ -1187,30 +1192,18 @@ impl<'a, W: Write> Writer<'a, W> {
             // This is where we can generate intermediate constants for some expression types.
             Statement::Emit(ref range) => {
                 for handle in range.clone() {
+                    if let Some(name) = ctx.named_expressions.get(&handle) {
+                        write!(self.out, "{}", INDENT.repeat(indent))?;
+                        let sanitized_name = self.namer.sanitize(name);
```
kvark, line 1197: this turned out to be pretty nice! Now there isn't a concern that the local map partially duplicates `named_expressions`, since in fact they may be different

src/back/wgsl/writer.rs
```diff
@@ -53,13 +52,25 @@ struct FunctionCtx<'a> {
     info: &'a FunctionInfo,
     /// The expression arena of the current function being written
     expressions: &'a Arena<Expression>,
+    /// Map of expressions that have associated variable names
+    named_expressions: &'a FastHashMap<Handle<Expression>, String>,
+}
+
+impl<'a> FunctionCtx<'a> {
```
kvark, line 59: nit: can do `FunctionCtx<'_>`

src/back/wgsl/writer.rs
```diff
@@ -53,13 +52,25 @@ struct FunctionCtx<'a> {
     info: &'a FunctionInfo,
     /// The expression arena of the current function being written
     expressions: &'a Arena<Expression>,
+    /// Map of expressions that have associated variable names
+    named_expressions: &'a FastHashMap<Handle<Expression>, String>,
```
kvark, line 56: looks like this type could be a typedef somewhere in lib.rs? it's already used a bunch

src/back/wgsl/writer.rs
```diff
@@ -570,6 +580,20 @@ impl<W: Write> Writer<W> {
...
+                        write!(self.out, "{}", INDENT.repeat(indent))?;
+                        let sanitized_name = self.namer.sanitize(name);
+                        self.start_named_expr(module, handle, &func_ctx, &sanitized_name)?;
+                        self.write_expr(module, handle, &func_ctx)?;
+                        writeln!(self.out, ";")?;
+
+                        // Front end provides names for all variables at the start of writing.
+                        // But we write them to step by step. We need to recache them.
+                        // Otherwise, we could accidentally write variable name instead of full expression.
+                        self.named_expressions.insert(handle, name.clone());
```
kvark, line 593: why are we adding `name.clone()` here and not `sanitized_name`?

kvark (changes requested):
Thank you! A few more things to fix.

Gordon-F:
@kvark Does the `WGSL` spec support variable shadowing ? How we should deal with code like this: 
```wgsl
[[stage(fragment)]]
fn main1([[location(0), interpolate(perspective)]] uv2: vec2<f32>) -> [[location(0)]] vec4<f32> {
    let a1 = 1.0;
    let a1 = 1.0;
    let a2 = 2.0;
    let c = a1 + a2;
    let c = a2 + a1;
    return vec4<f32>(a1, c, c, 1.0);
}
```

And how backend should deal with it if it should?

src/back/glsl/mod.rs
```diff
@@ -295,9 +297,6 @@ pub enum Error {
     /// A scalar with an unsupported width was requested
     #[error("A scalar with an unsupported width was requested: {0:?} {1:?}")]
     UnsupportedScalar(ScalarKind, Bytes),
-    /// [`Interpolation::Patch`](crate::Interpolation::Patch) isn't supported
-    #[error("Patch interpolation isn't supported")]
-    PatchInterpolationNotSupported,
```
Gordon-F, line 300: Unused error, found when I tried to generate docs.

kvark:
WGSL currently allows having local names shadowing the global names. But I think this is largely irrelevant to the issue at hands: IR names can also come from GLSL, SPIR-V, or other sources.
In the code you mentioned:
```
let a1 = 1.0;
let a1 = 1.0;
```
These two lines would manifest in two separate expressions, but assigned the same name.

src/back/glsl/mod.rs
```diff
@@ -194,9 +194,11 @@ struct FunctionCtx<'a> {
     info: &'a FunctionInfo,
     /// The expression arena of the current function being written
     expressions: &'a Arena<Expression>,
+    /// Map of expressions that have associated variable names
+    named_expressions: &'a crate::NamedExpressions,
 }
 
-impl<'a> FunctionCtx<'a> {
+impl<'a> FunctionCtx<'_> {
```
kvark, line 201: do we still use the `<'a>` thing? if not, we should just write `impl FunctionCtx<'_>`

kvark (approved):

kvark:
Let's merge and follow-up with more fronteds/backends

PR #938, opened 2021-06-02
move bin/naga.rs to a separate crate in workspace

jakobhellermann:
Splitting this off #934 for a place to discuss this change.

Because of the `default-members`, running `cargo run --features spv-out,wgsl-in input.wgsl output.spv` still works.
A disadvantage is that all the cargo features have to be duplicated.


Gordon-F:
Maybe we can enable all features for bin target? And remove all feature gates from the bin code. Are there any disadvantages, except compile time?

Cargo.toml
```diff
@@ -8,6 +8,7 @@ homepage = "https://github.com/gfx-rs/naga"
 repository = "https://github.com/gfx-rs/naga"
 keywords = ["shader", "SPIR-V", "GLSL", "MSL"]
 license = "MIT OR Apache-2.0"
+resolver = "2"
```
Gordon-F, line 11: A new feature resolver has been introduced in [rust 1.51](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html). Naga MSRV is 1.43 (based on info from CI task, but this definitely should be in README).

Is it required here?

jakobhellermann: It is required if the naga cli uses features, because otherwise `cargo run -p naga-cli --features ...` does not work: https://github.com/rust-lang/cargo/issues/5364.

Is the MSRV just for _users_ of `naga` or also for users of the `naga` binary and naga developers? The `resolver = 2` only affects the latter, older cargo version can still use naga as a dependency.

jakobhellermann:
> Maybe we can enable all features for bin target? And remove all feature gates from the bin code. Are there any disadvantages, except compile time?

I don't see any, except maybe binary size.

Here's the (approximate) difference in compile speed:

|                              | Compile time (debug) | Incremental compile time (debug) | Compile time (release) |
|------------------------------|--------------|----------------------------------|------------------------|
| No features                  | 24s          | 1s                               | 32s                    |
| All shader language features | 50s          | 2.3s                             | 110s                   |

kvark:
I'm getting warmed up to this path, especially since `cargo run` would just work (that was the main blocker).
I happen to run the converter a lot, and lately I just stopped bothering with feature selection and switched to `--all-features`. So following @Gordon-F suggestion to just unconditionally have that, at least for now (we can reconsider later), sounds like a win.

Cargo.toml
Gordon-F: Thank you for explain! MSRV for `naga` lib users. Everything is ok, since CI is happy.

naga-cli/Cargo.toml
```diff
@@ -9,22 +9,11 @@ repository = "https://github.com/gfx-rs/naga"
...
-wgsl-out = ["naga/wgsl-out"]
-hlsl-out = ["naga/hlsl-out"]
-
 [[bin]]
 name = "naga"
 path = "src/main.rs"
 
 [dependencies]
-naga = { path = "../" }
+naga = { path = "../", features = ["wgsl-in", "wgsl-out", "glsl-in", "glsl-out", "spv-in", "spv-out", "msl-out", "hlsl-out", "dot-out"] }
```
kvark, line 17: I wish there was "all_features = true"

naga-cli/src/main.rs
```diff
@@ -288,10 +269,7 @@ fn main() {
             }
             other => {
                 let _ = params;
```
kvark, line 271: no longer needed (here and in other places)

kvark (commented):
We should strongly add `env_logger` dependency now and uncomment the first line in main()

kvark:
Also `path = "src/main.rs"` isn't needed in the toml, I think

jakobhellermann:
> 
> 
> Also `path = "src/main.rs"` isn't needed in the toml, I think

I added the `[[bin]]` to keep the binary name as `naga` instead of `naga-cli`, and removing the `path` makes cargo error.

naga-cli/Cargo.toml
jakobhellermann: The `serialize` and `deserialize` features aren't needed anyway. `glsl-validate` should probably be enabled as well?

kvark: we can keep the serialization off now, indeed, but the validation for GLSL should be ON

Cargo.toml
```diff
@@ -43,13 +44,13 @@ wgsl-in = ["codespan-reporting"]
...
-
 [dev-dependencies]
 diff = "0.1"
 ron = "0.6"
 serde = { version = "1.0", features = ["derive"] }
 spirv = { package = "spirv_headers", version = "1.5", features = ["deserialize"] }
 rspirv = "0.7"
+
+[workspace]
+members = [".", "naga-cli"]
```
kvark, line 55: Can this live in `cli` folder? We have `naga-fuzz` today sitting there under just `fuzz`, so this would be consistent.
I don't recall why the fuzzer needs its own workspace tho

naga-cli/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[[bin]]
+name = "naga"
+path = "src/main.rs"
+
+[dependencies]
+naga = { path = "../", features = ["wgsl-in", "wgsl-out", "glsl-in", "glsl-out", "spv-in", "spv-out", "msl-out", "hlsl-out", "dot-out", "glsl-validate"] }
+log = "0.4"
+codespan-reporting = "0.11"
+env_logger = "0.8"
```
kvark, line 20: nit: add a newline

kvark (approved):
thank you!

PR #941, opened 2021-06-02
Add note about MSRV in README

Gordon-F:


README.md
```diff
@@ -8,6 +8,8 @@
 
 The shader translation library for the needs of [wgpu](https://github.com/gfx-rs/wgpu) and [gfx-rs](https://github.com/gfx-rs/gfx) projects.
 
+**Minimum Supported Rust Version:** `1.43.0`
```
kvark, line 11: do you think we could have a real github badge for this instead of a note? I think I saw some in other projects

Gordon-F: I saw only generated with `shields.io`. Like this one:

kvark: looks neat!

Gordon-F: Done. Rendered - https://github.com/Gordon-F/naga/blob/patch-2/README.md.

kvark (approved):

PR #942, opened 2021-06-03
[glsl-in] Read global in prologue only if allowed

JCapucho:
Adds a check to all entry points arguments if they should be read.

Fixes validation for the following snippet:
```glsl
// Vertex
void main() {
       gl_Position = vec4(1);
	fragColor = gl_Position;
}
```

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -46,6 +46,31 @@ bitflags::bitflags! {
     }
 }
 
+bitflags::bitflags! {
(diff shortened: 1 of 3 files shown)


kvark (approved):
Thanks!

Gordon-F:
Should we add this snippet as a test case for `glsl-in`?

JCapucho:
> Should we add this snippet as a test case for `glsl-in`?

@Gordon-F this wouldn't have been caught by the parser tests it would need to be a snapshot

PR #943, opened 2021-06-03
[glsl-in] Add integration tests to CI

JCapucho:
Adds a new workflow which tests some glsl shaders from https://github.com/JCapucho/naga-glsl-in-tests.

The workflow is ran on pull request merge or with manual activation.


kvark:
Interesting to see the manual activation thing.
Any reason this isn't just added to the `lazy` pipeline we have?

Gordon-F:
Maybe it should be a part of `naga` repository?

JCapucho:
> Any reason this isn't just added to the `lazy` pipeline we have?

I didn't add it to `lazy` firstly because it didn't have manual activation and I didn't know if it was desirable to add it and also if a PR only touches glsl-in a manual activation won't need to run the spirv and metal tests.

I can combine it into the `lazy` pipeline if that's desirable.

kvark:
I think this is very in line with tests with have in `lazy`. The decision to either switch it on manually or automatically should probably be done for the whole thing. Unless you are thinking that we'd be running the slow tests selectively?..
Maybe, but it seems to me that just running all of them on merges is simplest and gives us what we want here.

> Maybe it should be a part of naga repository?

Unsure. There is going to be quite a few corpuses. If they account to a resonable size in total, and they aren't changed often, then it makes sense to keep them in. Dota2 shaders weight less than a megabyte, so it seems promising.

Gordon-F:
> if a PR only touches glsl-in a manual activation won't need to run the spirv and metal tests.

You can start CI based on changed file path.

https://github.com/gfx-rs/naga/blob/90c2cf6aa6ea2b595098a27b148e3ad2f91317cb/.github/workflows/validation-macos.yml#L3-L5

JCapucho:
> > if a PR only touches glsl-in a manual activation won't need to run the spirv and metal tests.
> 
> @JCapucho You can start CI based on changed file path.
> 
> https://github.com/gfx-rs/naga/blob/90c2cf6aa6ea2b595098a27b148e3ad2f91317cb/.github/workflows/validation-macos.yml#L3-L5

@Gordon-F that starts up the whole workflow so it would get the same result

kvark:
Ok, it sounds like we need to split the `lazy` one into different workflows per area, so that we trigger them less often?

JCapucho:
> Unsure. There is going to be quite a few corpuses. If they account to
> a resonable size in total, and they aren&rsquo;t changed often, then it
> makes sense to keep them in. Dota2 shaders weight less than a
> megabyte, so it seems promising.

I can add them to naga itself, I only added them to a separate repo because of our discussion in the matrix room where you said you didn&rsquo;t want to fill up the repo with many small test cases. I intend to add some more tests and add them as issues are fixed and I&rsquo;m also looking at adding some glslang tests.

> Ok, it sounds like we need to split the lazy one into different workflows per area, so that we trigger them less often?

Since currently `lazy` is only ran in the master branch I don&rsquo;t think it makes that much of a difference but if we want to add manual activation to all `lazy` workflows it might make sense since it allows for quicker feedback

kvark:
> Since currently lazy is only ran in the master branch I don’t think it makes that much of a difference but if we want to add manual activation to all lazy workflows it might make sense since it allows for quicker feedback

Sorry, I should clarify. I mean, still running them automatically, but at least detecting more carefully when we need to run them. I.e. a PR changing README doesn't need to re-run the SPV-in tests.

> I can add them to naga itself, I only added them to a separate repo because of our discussion in the matrix room where you said you didn’t want to fill up the repo with many small test cases. I intend to add some more tests and add them as issues are fixed and I’m also looking at adding some glslang tests.

Would you be open to including them as a ZIP file? We'd have a folder with all the big tests, compressed.

JCapucho:
> Sorry, I should clarify. I mean, still running them automatically, but
> at least detecting more carefully when we need to run them. I.e. a PR
> changing README doesn&rsquo;t need to re-run the SPV-in tests.

We can add a condition to check if any files were modified in the src or tests directory then

> Would you be open to including them as a ZIP file? We&rsquo;d have a folder
> with all the big tests, compressed.

For me it&rsquo;s okay but where would the zip file be stored?

kvark:
> For me it’s okay but where would the zip file be stored?

Just here, `data/name_of_the_corpus.zip`

JCapucho:
Okay then my only concern is that reviewing new tests will be harder

Gordon-F:
> reviewing new tests will be harder

Agreed with @JCapucho.

Is a zip file required here? glsl frontend tests have equality file size with a wgsl frontend. Because it's just a text file.

kvark:
This scheme is not for tests that are gradually added. It's more for the existing corpuses that are mostly immutable (or can be reviewed externally).
So for the tests @JCapucho wants to add, let's first add them as regular snapshots (GLSL -> WGSL). If this set happens to grow large and stabilize in the future, we may reconsider moving it out.

JCapucho:
> This scheme is not for tests that are gradually added. It's more for the existing corpuses that are mostly immutable (or can be reviewed externally).
> So for the tests @JCapucho wants to add, let's first add them as regular snapshots (GLSL -> WGSL). If this set happens to grow large and stabilize in the future, we may reconsider moving it out.

I misunderstood what you meant I thought it was something to do right now.

So let me see if I understand, we add now the tests directly to the snapshots and then later if needed we zip them all.

kvark:
Right, let's add them to snapshots

Gordon-F:
One note about zip files. Maybe it's a little be paranoid, but if we stored something with a zip archive inside the repo, we should be careful about all PR that changes these files because of security reasons.

JCapucho:
@kvark just one small question, the current corpus has 10 tests do I add them directly to the snapshots `in` folder or do I add it in a special folder?

kvark:
Good question! Maybe we can have a subfolder? I.e. `tests/in/glsl/...`

JCapucho:
@kvark I've added the tests to the snapshot folder and added a runner that reads all the shaders figures out the stage and validates and if wgsl-out is enabled writes to wgsl.

I've also included a fix about the less and greater being swapped which I noticed while writing out the snapshots (They are already doing their job :) )

JCapucho:
hmm CI is complaining but the shader looks fine can someone who knows wgsl look at the shader and see if something's wrong in the generated ouput?

Gordon-F:
@JCapucho Could you disable `wgsl-out` validation for a while? I'll take a look at what's wrong with `wgsl-out`.

JCapucho:
I've disabled the wgsl generation as @Gordon-F asked and removed the shader and some others related to empty struct generation I'll fix it in another PR

kvark (approved):

PR #949, opened 2021-06-04
Permit dynamic indexing of arrays and matrices only behind a pointer.

jimblandy:
This makes Naga IR validation impose the restrictions added to WGSL in
gpuweb/gpuweb#1801.

Remove code in the SPIR-V writer to spill arrays to temporary variables in order
to index them dynamically. If such IR is encountered, treat it as a failure of
validation.

(This simplifies #916 a little bit, so I took it on first.)

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -72,7 +72,6 @@ struct Function {
     signature: Option<Instruction>,
     parameters: Vec<Instruction>,
     variables: crate::FastHashMap<Handle<crate::LocalVariable>, LocalVariable>,
-    internal_variables: Vec<LocalVariable>,
     blocks: Vec<Block>,
     entry_point_context: Option<EntryPointContext>,
 }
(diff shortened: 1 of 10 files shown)


src/back/spv/writer.rs
```diff
@@ -1324,62 +1320,6 @@ impl Writer {
         })
     }
 
-    #[allow(clippy::too_many_arguments)]
-    fn promote_access_expression_to_variable(
```
kvark, line 1328: hoorrray!

src/valid/expression.rs
```diff
@@ -168,6 +167,9 @@ impl super::Validator {
                         return Err(ExpressionError::InvalidIndexType(index));
                     }
                 }
+                if !is_pointer && function.expressions[index].is_dynamic_index(module) {
```
kvark, line 170: what if it's a vector? can still be dynamically indexed

kvark (changes requested):
Great stuff! Just one note here

src/valid/expression.rs
jimblandy: Arg, thanks!

jimblandy:
This version puts the docs in the right place, as well, and adds tests for the cases that should have worked.

kvark (approved):

PR #952, opened 2021-06-05
Add snapshot output files for quad-glsl.glsl.

jimblandy:
These began appearing in 344acc31 (João Capucho, 2021-06-01), but were
apparently left out of the PR.

### DIFF
--- a/tests/out/quad-glsl.wgsl
+++ b/tests/out/quad-glsl.wgsl
@@ -1,0 +1,40 @@
+struct VertexOutput {
+    [[location(0), interpolate(perspective)]] member: vec2<f32>;
+    [[builtin(position)]] member1: vec4<f32>;
+};
+
+struct FragmentOutput {
+    [[location(0), interpolate(perspective)]] member2: vec4<f32>;
+};
+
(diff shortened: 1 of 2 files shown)


kvark:
@jimblandy thank you for noticing this!
How do you feel about fixing the test configuration instead?
In the latest take on testing - #752 - I expressed the idea that we should have WGSL -> everything and everything -> WGSL. If we do everything -> everything, then it would unproportionally put the load on the backends.

So, in the spirit of this approach, the GLSL-in tests should purely go GLSL -> WGSL and not have any other outputs.

JCapucho:
@jimblandy thank you for noticing this, I also noticed this while working on another thing and have fixed it in the branch I was working on, also as @kvark said I changed spirv to wgsl so I think it's better for us to close this one. But thank you non the less.

jimblandy:
Testing is important, and the plans in #752 sound good, but at the moment, I want to finish the other work I've started. For this PR I just wanted git to stop complaining.

JCapucho (approved):
Lgtm

Issue #945, opened 2021-06-03
[wgsl-in] Indexing into matrices must be done beyond pointers


Gordon-F:
Found in #943

Original code:
```glsl
// AUTHOR: enfipy
// ISSUE: #210
// FIX: #898
#version 450

layout(location = 0) in vec3 Vertex_Position;
layout(location = 1) in vec3 Vertex_Normal;
layout(location = 2) in vec2 Vertex_Uv;

layout(location = 0) out vec3 v_Position;
layout(location = 1) out vec3 v_Normal;
layout(location = 2) out vec2 v_Uv;

layout(set = 0, binding = 0) uniform Camera {
    mat4 ViewProj;
};

layout(set = 2, binding = 0) uniform Transform {
    mat4 Model;
};

void main() {
    v_Normal = (Model * vec4(Vertex_Normal, 1.0)).xyz;
    v_Normal = mat3(Model) * Vertex_Normal;
    v_Position = (Model * vec4(Vertex_Position, 1.0)).xyz;
    v_Uv = Vertex_Uv;
    gl_Position = ViewProj * vec4(v_Position, 1.0);
}
```

Consumed wgsl:
```wgsl
[[block]]
struct Camera {
    ViewProj: mat4x4<f32>;
};

[[block]]
struct Transform {
    Model: mat4x4<f32>;
};

struct VertexOutput {
    [[location(0), interpolate(perspective)]] member: vec3<f32>;
    [[location(1), interpolate(perspective)]] member1: vec3<f32>;
    [[location(2), interpolate(perspective)]] member2: vec2<f32>;
    [[builtin(position)]] member3: vec4<f32>;
};

var<private> Vertex_Position: vec3<f32>;
var<private> Vertex_Normal: vec3<f32>;
var<private> Vertex_Uv: vec2<f32>;
var<private> v_Position: vec3<f32>;
var<private> v_Normal: vec3<f32>;
var<private> v_Uv: vec2<f32>;
[[group(0), binding(0)]]
var<uniform> global: Camera;
[[group(2), binding(0)]]
var<uniform> global1: Transform;
var<private> gl_Position: vec4<f32>;

fn main() {
    let _e3: vec3<f32> = Vertex_Normal;
    let _e17: mat4x4<f32> = global1.Model;
    let _e20: vec4<f32> = (_e17 * vec4<f32>(_e3, 1.0));
    v_Normal = vec3<f32>(_e20.x, _e20.y, _e20.z);
[...]

PR #955, opened 2021-06-06
[glsl-in] Compute shader fixes and additions

### LINKED ISSUE #246: [glsl-in] Unsupported compute stage layout qualifiers
Running on https://github.com/gfx-rs/gfx/blob/17249afe4c3b219200abb3a5c979984bdcb4a640/examples/compute/shader/collatz.comp
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseError { kind: NotImplemented("unsupported layout qualifier(s)") }', examples/convert.rs:91:14

### LINKED ISSUE #902: [glsl-in] buffer isn't treated as a keyword
The following code will produce `Unexpected token: Identifier("buffer")`

```
#version 450 core
buffer test {};
```

The problem is that `buffer` isn't treated as a keyword, this is a good first issue if anyone wants guidance ping me here or on the matrix room.

JCapucho:
Contains a bunch of bug fixes and implements more parts of the parser to
finally be able to parse the collatz compute shader and other compute
shaders.

When I started working on this I thought it would be a simple thing but
it proved more complex, here&rsquo;s a list of the things included in this PR

-   Add parsing for compute shader workgroup modifiers (fixes #246)
-   Add support for the buffer storage class (fixes #902)
-   Add mod function and use implicit conversions for it&rsquo;s arguments
-   Add implicit conversions for locals initializers
-   Add glGlobalInvocationID builtin
-   Fixes a panic when calculating global use
-   Refractor the way global and blocks are handled to reduce code
    duplication
-   Fixes unsized types being loaded
-   Fixes assignments for function arguments in validation
-   Generates the wgsl for the quad shader
-   Add the collatz shader to testing

Due to #945 the glsl snapshot folder isn&rsquo;t currently generating wgsl

Also due to load expressions being cached the produced wgsl isn&rsquo;t
correct but everything else looks okay, here&rsquo;s the wgsl for the collatz
shader

    [[block]]
    struct PrimeIndices {
        indices: [[stride(4)]] array<u32>;
    };

    [[group(0), binding(0)]]
    var<storage> global: [[access(read_write)]] PrimeIndices;
    var<private> gl_GlobalInvocationID: vec3<u32>;

    fn collatz_iterations(n: u32) -> u32 {
        var n1: u32;
        var i: u32;
        var local: u32;

        n1 = n;
        let _e4: u32 = n1;
        i = u32(0);
        loop {
            if (!((_e4 != u32(1)))) {
                break;
            }
            {
                if (((_e4 % u32(2)) == u32(0))) {
                    {
                        n1 = (_e4 / u32(2));
                    }
                } else {
                    {
                        n1 = ((u32(3) * _e4) + u32(1));
                    }
                }
                let _e28: u32 = i;
                local = _e28;
                i = (_e28 + 1u);
            }
        }
        return i;
    }

    fn main() {
        var index: u32;

        let _e6: u32 = index;
        index = gl_GlobalInvocationID.x;
        let _e12: u32 = collatz_iterations(global.indices[_e6]);
        global.indices[_e6] = _e12;
        return;
    }

    [[stage(compute), workgroup_size(1, 1, 1)]]
    fn main1([[builtin(global_invocation_id)]] param: vec3<u32>) {
        gl_GlobalInvocationID = param;
        main();
        return;
    }

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -260,7 +260,24 @@ impl<'function> Context<'function> {
                         .expressions
                         .append(Expression::AccessIndex { base, index });
 
-                    (expr, true)
+                    (expr, {
+                        let ty = program.module.global_variables[handle].ty;
+
+                        match program.module.types[ty].inner {
+                            TypeInner::Struct { ref members, .. } => {
+                                if let TypeInner::Array {
+                                    size: crate::ArraySize::Dynamic,
+                                    ..
+                                } = program.module.types[members[index as usize].ty].inner
+                                {
+                                    false
+                                } else {
+                                    true
+                                }
+                            }
+                            _ => true,
+                        }
+                    })
                 }
             };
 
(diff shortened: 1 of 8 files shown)


kvark (approved):
Looks sensible, I didn't inspect all the details.
Is this meant to be landed as is? Or squashed?

JCapucho:
I made the commits so that they would all pass validation and contain a specific fix each so I think it's better to merge it has is without squashing

PR #957, opened 2021-06-07
[msl-out] Implement named expression generation

Gordon-F:


Gordon-F:
CI errors looks unrelated to PR changes.

src/back/msl/writer.rs
```diff
@@ -1306,14 +1307,38 @@ impl<W: Write> Writer<W> {
...
+                            self.start_baking_expression(
+                                handle,
+                                &context.expression,
+                                &sanitized_name,
+                            )?;
+                            self.put_expression(handle, &context.expression, true)?;
+                            self.named_expressions.insert(handle, sanitized_name);
+
+                            writeln!(self.out, ";")?;
+                            continue;
```
kvark, line 1329: Can we avoid control flow here, please?
Another thing, can we merge this with the other bit of code here?
I.e. something like
```
let name = match context.expression.function.named_expressions.get(&handle) {
  Some(name) => Some(self.namer.call_unique(name)),
  None => {
    let min_ref_count = ...;
    if min_ref_count <= ... {
      Some(format!(...))
    } else {
      None
    }
  };
  if let Some(name) = name {
    // spew out the expression
  }
}
```

kvark (commented):
Please update the block size expectations, the CI failure is related to this change.
I.e. it adds more logic to `put_block`, which affects its stack size. So the higher bound needs to be raised to 16000

src/back/msl/writer.rs
Gordon-F: Great idea, thank you! Fixed.

kvark (approved):
Great!

PR #958, opened 2021-06-07
[spv-out]: Don't have write_expression_pointer return a storage class.

jimblandy:
Nobody uses that value. This lets us simplify `GlobalVariable` and
`write_global_variable` as well.


kvark (approved):
Thank you!
It was used at some point :)

PR #970, opened 2021-06-13
Fix testing infrastructure

Gordon-F:
One thing that we missed in #938. Currently, `cargo run --all-features` running 0 tests.


kvark (approved):
Oh wow

PR #971, opened 2021-06-14
Makefile: Make validate-wgsl target use new 'cli' sub-package.

jimblandy:
Without this, `make validate-wgsl` fails, complaining that `cli` doesn't have a `wgsl-in` feature.


Makefile
```diff
@@ -61,7 +61,7 @@ validate-dot: $(SNAPSHOTS_OUT)/*.dot
 validate-wgsl: $(SNAPSHOTS_OUT)/*.wgsl
 	@set -e && for file in $^ ; do \
 		echo "Validating" $${file#"$(SNAPSHOTS_OUT)/"};	\
-		cargo run --features wgsl-in $${file}; \
+		cargo run --manifest-path cli/Cargo.toml $${file}; \
```
kvark, line 64: I don't think the manifest path is needed. We wanted `cargo run` to run the CLI by default ,anyway

kvark (approved):

PR #972, opened 2021-06-14
[hlsl-out] More improvements. Enable quad snapshot testing

Gordon-F:


Makefile
```diff
@@ -65,7 +65,15 @@ validate-wgsl: $(SNAPSHOTS_OUT)/*.wgsl
 	done
 
 validate-hlsl: $(SNAPSHOTS_OUT)/*.hlsl
-	@set -e && for file in $(SNAPSHOTS_OUT)/*.Compute.hlsl ; do \
+	@set -e && for file in $(SNAPSHOTS_OUT)/*.Vertex.hlsl ; do \
```
kvark, line 68: we discussed this briefly, and we should be able to generate a single HLSL for the module

README.md
```diff
@@ -40,7 +40,8 @@ Naga includes a default binary target, which allows to test the conversion of di
 cargo run --features wgsl-in -- my_shader.wgsl # validate only
 cargo run --features spv-in -- my_shader.spv my_shader.txt # dump the IR module into a file
 cargo run --features spv-in,msl-out -- my_shader.spv my_shader.metal --flow-dir flow-dir # convert the SPV to Metal, also dump the SPIR-V flow graph to `flow-dir`
-cargo run --features wgsl-in,glsl-out -- my_shader.wgsl my_shader.vert --profile es310 # convert the WGSL to GLSL vertex stage under ES 3.20 profile
+cargo run --features wgsl-in,glsl-out -- my_shader.wgsl my_shader.glsl.vert --profile es310 # convert the WGSL to GLSL vertex stage under ES 3.20 profile
+cargo run --features wgsl-in,hlsl-out -- my_shader.wgsl my_shader.hlsl.vert --shader-model 50 # convert the WGSL to HLSL vertex stage with shader model 5.0
```
kvark, line 44: the features are no longer needed, the CLI binary now has them unconditionally

src/back/hlsl/writer.rs
```diff
@@ -1,16 +1,22 @@
 //TODO: temp
 #![allow(dead_code)]
+use super::Options;
 use super::{Error, ShaderModel};
 use crate::back::hlsl::keywords::RESERVED;
+use crate::back::vector_size_str;
```
kvark, line 6: Ideally, the code style here would follow other backends such as MSL and SPIR-V specifically.
But at the very minimum, let's combine the use cases so that there is one per crate. So `use crate::` is there only once.

src/back/hlsl/writer.rs
```diff
@@ -72,42 +92,267 @@ impl<W: Write> Writer<W> {
...
-        // Write all entry points
-        for (index, ep) in module.entry_points.iter().enumerate() {
-            let func_ctx = FunctionCtx {
-                ty: FunctionType::EntryPoint(index as u16),
-                info: info.get_entry_point(index),
-                expressions: &ep.function.expressions,
-                named_expressions: &ep.function.named_expressions,
+        // Write all constants
+        for (handle, constant) in module.constants.iter() {
+            if constant.name.is_some() {
```
kvark, line 116: The name here should be totally cosmetic. It shouldn't be used as a condition to emit the global constant.

src/back/hlsl/writer.rs
```diff
@@ -72,42 +92,267 @@ impl<W: Write> Writer<W> {
...
+
+    fn write_ep_input_struct(
+        &mut self,
+        module: &Module,
+        func: &Function,
+        stage: ShaderStage,
+    ) -> BackendResult {
+        if !func.arguments.is_empty() {
+            let struct_name = match stage {
+                ShaderStage::Vertex => self.namer.call_unique("VertexInput"),
```
kvark, line 208: nit: we could move out the logic from `match`. Only the inner `str` is different

src/back/hlsl/writer.rs
```diff
@@ -72,42 +92,267 @@ impl<W: Write> Writer<W> {
...
+            let mut members = Vec::with_capacity(func.arguments.len());
+
+            write!(self.out, "struct {}", &struct_name)?;
+            writeln!(self.out, " {{")?;
+
+            for arg in func.arguments.iter() {
+                let arg_name = if let Some(ref name) = arg.name {
+                    name
+                } else {
+                    "member"
```
kvark, line 222: Maybe "argument"? I'd expect to see "member" for the structure members but maybe not in the arguments

src/back/hlsl/writer.rs
```diff
@@ -72,42 +92,267 @@ impl<W: Write> Writer<W> {
...
+    /// # Notes
+    /// Always adds a newline
+    fn write_global(&mut self, module: &Module, handle: Handle<GlobalVariable>) -> BackendResult {
+        let global = &module.global_variables[handle];
+        let inner = &module.types[global.ty].inner;
+
+        let register_ty = match *inner {
+            TypeInner::Image { .. } => "t",
+            TypeInner::Sampler { .. } => "s",
+            _ => "",
```
kvark, line 267: should be expanded (possibly in a follow-up) to fill https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-variable-register

src/back/hlsl/writer.rs
```diff
@@ -72,42 +92,267 @@ impl<W: Write> Writer<W> {
...
+        };
+
+        let register = if let Some(ref binding) = global.binding {
+            format!("register({}{})", register_ty, binding.binding)
+        } else {
+            String::from("")
```
kvark, line 273: this will show up as `xxx: ;`, which will not fly :)

src/back/hlsl/writer.rs
```diff
@@ -72,42 +92,267 @@ impl<W: Write> Writer<W> {
...
+        handle: Handle<Constant>,
+    ) -> BackendResult {
+        match *inner {
+            ConstantInner::Scalar {
+                width: _,
+                ref value,
+            } => {
+                let name = &self.names[&NameKey::Constant(handle)];
+                let (ty, value) = match *value {
+                    crate::ScalarValue::Sint(value) => ("int", format!("{}", value)),
```
kvark, line 300: It would be good to try to avoid heap allocations from the start.
On one hand, this is unlikely going to manifest too much in performance profiles, at least here specifically.
On the other hand, we have parties constantly looking to pre-compile HLSL and friends, which is a bit problematic. We don't need to give them more reason to do so (by making Naga slower).

src/back/hlsl/writer.rs
```diff
@@ -131,6 +376,52 @@ impl<W: Write> Writer<W> {
...
+                    scalar_kind_str(kind, width)?,
+                    vector_size_str(size)
+                )?;
+            }
+            TypeInner::Matrix {
+                columns,
+                rows,
+                width,
+            } => {
+                //TODO: int matrix ?
```
kvark, line 392: no need for a TODO - the IR only has float matrices anyway

src/back/hlsl/writer.rs
```diff
@@ -311,15 +823,79 @@ impl<W: Write> Writer<W> {
...
+
+    /// Helper method used to write [`ScalarValue`](ScalarValue)
+    ///
+    /// # Notes
+    /// Adds no trailing or leading whitespace
+    fn write_scalar_value(&mut self, value: ScalarValue) -> BackendResult {
+        match value {
+            ScalarValue::Sint(value) => write!(self.out, "{}", value)?,
+            ScalarValue::Uint(value) => write!(self.out, "{}u", value)?,
+            // Floats are written using `Debug` instead of `Display` because it always appends the
```
kvark, line 852: pretty cool!

kvark (commented):
Amazing work here! Excited to see the less trivial HLSLs compiling 🎉

README.md
Gordon-F: I will remove all features requirements from README.

src/back/hlsl/writer.rs
Gordon-F: `rust-analyzer` add one line per "Import item ...", and don't automatically merge it. I always forget to merge it manually.

src/back/hlsl/writer.rs
Gordon-F: This mostly for readability. For example:
```wgsl
// vertex
let c_scale: f32 = 1.2;
return VertexOutput(uv, vec4<f32>(c_scale * pos, 0.0, 1.0));
``` 

If I remove this part of the code, `c_scale` will be inlined.

src/back/hlsl/writer.rs
Gordon-F: Almost every PR you suggest me to avoid logic inside `match`, and I forget about it every time. 😄 

Thank you!

src/back/hlsl/writer.rs
Gordon-F: That's not a good variable name. It should be `member_name` since all entry point arguments are wrapped into a struct. All function arguments for an entry point function become `wrappeed_struct.member`.

src/back/hlsl/writer.rs
Gordon-F: It's a copy-paste from `glsl-out` :) There are a lot of copy-paste between at least 3 backends: glsl, wgsl, hlsl 🤔

src/back/hlsl/writer.rs
Gordon-F: TIL, Thank you!

src/back/hlsl/writer.rs
Gordon-F: Would it be better if we make 2 times match? We can avoid heap allocation with something like this:

```rust
let ty = match *value {
    crate::ScalarValue::Sint(value) => "int",
...

match *value {
    crate::ScalarValue::Sint(value) => write!(..),
    crate::ScalarValue::Uint(value) => write!(..),
    crate::ScalarValue::Float(value) => write!(..),
```

kvark: that sounds fine, yes

src/back/hlsl/writer.rs
kvark: I see. I suppose we can roll with this :)

Makefile
```diff
@@ -65,7 +65,15 @@ validate-wgsl: $(SNAPSHOTS_OUT)/*.wgsl
...
-		echo "Validating" $${file#"$(SNAPSHOTS_OUT)/"};\
-		dxc $${file} -T cs_5_0;\
+	@set -e && for file in $^ ; do \
+		echo "Validating" $${file#"$(SNAPSHOTS_OUT)/"}; \
+		config="$$(dirname $${file})/$$(basename $${file}).config"; \
+		vertex=""\
+		fragment="" \
+		compute="" \
+		. $${config}; \
+		[ ! -z "$${vertex}" ] && echo "Vertex Stage:" && dxc $${file} -T $${vertex} -E $${vertex_name} -Wno-parentheses-equality  -Zi -Qembed_debug > /dev/null; \
```
kvark, line 75: the "-Wno-parentheses-equality  -Zi -Qembed_debug" part can be moved outside, e.g.
```bash
DXC_PARAMS=-Wno-parentheses-equality  -Zi -Qembed_debug
```

kvark (approved):

PR #978, opened 2021-06-16
Add `.` to the workspace's default members. Test with no features.

jimblandy:
Naga is now a workspace with `naga` and `cli` as its two members. The default
package for cargo commands is `cli`, so that `cargo run` will just run the CLI.

However, this has a few unexpected consequences:

- Now `cargo test` will just try to run `cli`'s tests, of which there are none.
  Adding `"."` to the `default-members` list in the workspace's `Cargo.toml`
  seems to fix this, without breaking `cargo run`.

- Even with `"."` added to `default-members`, `cargo test` will build `naga` by
  default with the features requested for it in `cli/Cargo.toml`: all the front
  and back ends, but no `serialize` or `deserialize`. This means that our CI job
  meant to verify no-feature builds isn't doing that job any more. We need to
  pass `--package naga` to `cargo test` to make it test naga directly.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -19,7 +19,13 @@ jobs:
       - uses: actions-rs/cargo@v1
         name: Default test
         with:
+          # Our intention here is to test `naga` with no features enabled. But
+          # since `cli` is the default package, a plain `cargo test` will build
+          # `naga` with the features requested in `cli/Cargo.toml`. Passing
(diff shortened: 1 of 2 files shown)


jimblandy:
I clearly don't understand how `pipeline.yml` works...

jimblandy:
Okay, this does seem to run the no-features tests, and then the all-features tests. Ready for review.

kvark (approved):

PR #981, opened 2021-06-17
Update `wgsl-out` status in README

Gordon-F:
Since `wgsl-out` can handle all snapshot test set, we can update its status.


kvark (approved):
horray!

PR #982, opened 2021-06-17
[wgsl-in] use AccessIndex for accesses using constant index

Frizi:
This change allows constant index expressions parsed by wgsl-in to be consumed by spv-out. Otherwise, the access is always treated as dynamic. Those accesses are often generated by `glsl-in -> wgsl-out` translation for matrix cast expression, which previously prevented those shaders to be further translated into spirv.

The msl-out change is also required, as this change surfaced a bug when generating those const accesses, emitting incorrect code from snapshot tests.


kvark (approved):
Looks reasonable!
It will be fairly easy to revert this if the WGSL spec moves a different way

PR #984, opened 2021-06-17
[glsl-in] parse all math functions

Frizi:
Added support for all missing math functions from glsl. Also fixes `mod` to not incorrectly parse as `modf`, which is a separate function.


src/front/glsl/functions.rs
```diff
@@ -236,38 +251,42 @@ impl Program<'_> {
                             body,
                         )))
                     }
+                    "atan" => match args.len() {
+                        1 => Ok(Some(ctx.add_expression(
```
kvark, line 255: nit: the match could return `Expression`, which we'd then add outside of the match

tests/in/glsl/math-functions.vert
```diff
@@ -0,0 +1,55 @@
+#version 450
```
kvark, line 1: any reason this doesn't have a corresponding WGSL output snapshot?

Frizi: generation of wgsl output is right now disabled in snapshots, though i'd like to enable it. I can do that in this PR if that's ok.

kvark: fine by me :)

kvark:
> thread 'convert_glsl_folder' panicked at 'called `Result::unwrap()` on an `Err` value: Function { handle: [1], name: "collatz_iterations", error: Expression { handle: [15], error: InvalidBinaryOperandTypes(Modulo, [5], [14]) } }', tests/snapshots.rs:427:10

JCapucho:
can you please rebase on master

Frizi:
Rebased, I see snapshots just got enabled as well, so removed that part.

JCapucho (approved):
Looks good to me, pinging @kvark since he reviewed originally this PR, do you have anything else you would like fixed or can we merge this?

PR #988, opened 2021-06-18
reorder namer assignment order to preserve names for reflection

Frizi:
Previously, namer was processing the module in such order, that names of entry points and bindings were pretty much guaranteed to be non-unique, resulting with additional numeral postfix. This resulted in bugs when specifying wgpu render pipeline, as the entry point name didn't match. Additionaly, binding names are often inspected through shader reflection to determine which buffers to bind, so it's good idea to preserve those as well.

This change modifies the order of operations in the namer, such that important names are processed first, and collisions are resolved by modifying names of private symbols.

Additionaly, because struct members essentially have their own namespace, those are no longer deduplicated globally, and are uniquely named only in scope of given struct. That both aids reflection and helps with debugging translated shaders.

### DIFF
--- a/src/proc/namer.rs
+++ b/src/proc/namer.rs
@@ -86,6 +86,18 @@ impl Namer {
         })
     }
 
+    fn namespace(&mut self, reserved_keywords: &[&str], f: impl FnOnce(&mut Self)) {
+        let parent_unique = std::mem::take(&mut self.unique);
+        self.unique.extend(
+            reserved_keywords
+                .iter()
+                .map(|string| (string.to_string(), 0)),
+        );
+
+        f(self);
+        self.unique = parent_unique;
+    }
+
     pub fn reset(
         &mut self,
         module: &crate::Module,
@@ -110,10 +122,42 @@ impl Namer {
             output.insert(NameKey::Type(ty_handle), ty_name);
 
             if let crate::TypeInner::Struct { ref members, .. } = ty.inner {
-                for (index, member) in members.iter().enumerate() {
-                    let name = self.call_or(&member.name, "member");
-                    output.insert(NameKey::StructMember(ty_handle, index as u32), name);
-                }
+                // struct members have their own namespace, because access is always prefixed
+                self.namespace(reserved_keywords, |namer| {
+                    for (index, member) in members.iter().enumerate() {
+                        let name = namer.call_or(&member.name, "member");
+                        output.insert(NameKey::StructMember(ty_handle, index as u32), name);
+                    }
+                })
+            }
+        }
+
+        for (ep_index, ep) in module.entry_points.iter().enumerate() {
+            let ep_name = self.call(&ep.name);
+            output.insert(NameKey::EntryPoint(ep_index as _), ep_name);
+            for (index, arg) in ep.function.arguments.iter().enumerate() {
+                let name = self.call_or(&arg.name, "param");
+                output.insert(
+                    NameKey::EntryPointArgument(ep_index as _, index as u32),
+                    name,
+                );
+            }
+            for (handle, var) in ep.function.local_variables.iter() {
+                let name = self.call_or(&var.name, "local");
+                output.insert(NameKey::EntryPointLocal(ep_index as _, handle), name);
+            }
+        }
+
+        for (fun_handle, fun) in module.functions.iter() {
+            let fun_name = self.call_or(&fun.name, "function");
+            output.insert(NameKey::Function(fun_handle), fun_name);
+            for (index, arg) in fun.arguments.iter().enumerate() {
+                let name = self.call_or(&arg.name, "param");
+                output.insert(NameKey::FunctionArgument(fun_handle, index as u32), name);
+            }
+            for (handle, var) in fun.local_variables.iter() {
+                let name = self.call_or(&var.name, "local");
+                output.insert(NameKey::FunctionLocal(fun_handle, handle), name);
             }
         }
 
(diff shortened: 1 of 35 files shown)


kvark:
> This resulted in bugs when specifying wgpu render pipeline, as the entry point name didn't match.

Fixing the namer is not the right way to solve this. The namer has a simple task: generate non-conflicting names. It doesn't know about your pipeline interface (that is broken here). Backends have their own limitations on what entry point names are valid and what not. Therefore:
  - if the backend can accept any entry point name that IR reasonably allows, then it should just pass through the EP names without touching the namer for it
  - if the backend has additional restrictions, it needs to adjust the names and return some sort of a naming map to the user (of the backend). MSL backend does this.

Gordon-F:
@Frizi Nice feature, thank you! One more option, how we can deal with unique EP, pass to backend your own entry point name, like we did with `hlsl-out`:
https://github.com/gfx-rs/naga/blob/e3827f8e0899377c0177faa1f106ec2f3f8a9e3e/src/back/hlsl/mod.rs#L27-L47

Frizi:
> if the backend can accept any entry point name that IR reasonably allows, then it should just pass through the EP names without touching the namer for it

This is not right, as other non-entry-point functions can also conflict with that name. This is currently the case for glsl frontend, as it generates an extra function named exactly the same as entry point. So that conflict has to be resolved, thus namer must be involved. It just have to be involved in the right order.

> if the backend has additional restrictions, it needs to adjust the names and return some sort of a naming map to the user (of the backend). MSL backend does this.

Is that `binding_map` generated? All I see is that MSL backend accepts it as input through options, but the user has to provide it manually.

> how we can deal with unique EP, pass to backend your own entry point name, like we did with hlsl-out

Entry points are already provided to glsl frontend through it's options. Wgsl backend in particular allows multiple entry points to be generated on the output, so it shouldn't need any extra options. Just generate what's in the module. All I want to see is that entry point names, as specified in the source shader, survive the translation.

Gordon-F:
> This is currently the case for glsl frontend, as it generates an extra function named exactly the same as entry point

Why glsl frontend generate extra function with the same name? It look like a frontend problem.

Frizi:
> Why glsl frontend generate extra function with the same name? It look like a frontend problem.

I don't know the reason. It was like that since I started looking at it. My assumption is that this generated function maps closer to the true contents of glsl entry point, which interfaces with bindings through globals. So the true entry point function is responsible purely for populating those globals and reading them back for output. Unfortunately, both are named the same, thus the collision.

kvark:
@Frizi 
> This is not right, as other non-entry-point functions can also conflict with that name. This is currently the case for glsl frontend, as it generates an extra function named exactly the same as entry point. So that conflict has to be resolved, thus namer must be involved. It just have to be involved in the right order.

Yes, you are right. Namer should be involved, but ideally not just via changing the order of calling into it. Instead, it should be notified that these names are occupied. I suppose what your PR does will result in very similar behavior, so it's still a good improvement.

> Is that binding_map generated? All I see is that MSL backend accepts it as input through options, but the user has to provide it manually.

It's the [TranslationInfo](https://github.com/gfx-rs/naga/blob/e3827f8e0899377c0177faa1f106ec2f3f8a9e3e/src/back/msl/mod.rs#L442) returned by the backend.

src/proc/namer.rs
```diff
@@ -86,6 +86,18 @@ impl Namer {
         })
     }
 
+    fn namespace(&mut self, reserved_keywords: &[&str], f: impl FnOnce(&mut Self)) {
+        let parent_unique = std::mem::take(&mut self.unique);
+        self.unique.extend(
```
kvark, line 91: This is a bit unfortunate, it introduces a whole lot of heap allocations on every switch.
We need a more lightweight way to switch the namespaces

kvark (commented):
This is a solid improvement!
Let's just make it less heavy on the heap allocator, please :)

src/proc/namer.rs
```diff
@@ -86,16 +102,12 @@ impl Namer {
...
-                .iter()
-                .map(|string| (string.to_string(), 0)),
-        );
-
+    fn namespace(&mut self, f: impl FnOnce(&mut Self)) {
+        self.namespace_index += 1;
         f(self);
-        self.unique = parent_unique;
+        let current_ns = self.namespace_index;
+        self.unique.retain(|&(_, ns), _| ns != current_ns);
```
kvark, line 109: how could anything be in here? we just bumped the namespace index, which was never seen before

kvark: why are we clearing this up at all?

src/proc/namer.rs
```diff
@@ -52,29 +55,42 @@ impl Namer {
...
                 *e.get_mut() += 1;
-                format!("{}{}", e.key(), e.get())
+                format!("{}{}", e.key().0, e.get())
             }
             Entry::Vacant(e) => {
-                let name = e.key().to_string();
-                e.insert(0);
-                name
+                let name = &e.key().0;
+                if self.keywords.contains(&e.key().0) {
```
kvark, line 85: Hmm, so with this, we are going to scan through the keyword list on each and every token added? There can be a lot of keywords. So this may turn out to be somewhat inefficient.

kvark (commented):
I think we should carefully re-think the way Namer will function in multiple namespaces.

src/proc/namer.rs
Frizi: This is after the closure call, that is about to add stuff to that hashmap by using `call` method. After all this is the reason tu use `namespace` method in the first place.

The idea is that once you are done with given namespace, you clear it out, so next time the index is bumped, the list is empty.

In old implementation, every time you enter the namespace, a new unique map was created, and it was discarded once you exit the namespace. Here, it's being cleared instead.

src/proc/namer.rs
Frizi: Yes, once per unique keyword. The list is not checked in `Occupied` match. I don't think how to avoid that double-check. If your unique map is no longer global, you have to check it somehow.

src/proc/namer.rs
kvark: I don't quite understand why we are able to clear everything. Once you write down a structure with its members, Namer needs to generate their names. But then, anywhere through the backend code, it may need to ask the Namer: "what was the name of this field?" and get that back. If we are clearing the subset, Namer wouldn't be able to answer this.

src/proc/namer.rs
Frizi: Actually important destinction: This is not a list, but a hash set. That should make the lookup significantly more tolerable 😉

src/proc/namer.rs
Frizi: The unique hashmap is only used for collision detection. The result of renames is written down to the `output` map, which is a separate entity inside `reset` method. It has nothing to do with `unique`. Thus, the result of naming is still properly persisted.

kvark: ah, ok, that's good :)

kvark:
Ok, I think we should still carefully think about it and improve.
But also it has a strong user benefit right now, so we should proceed with it even if some concerns are reserved :)
Thank you for driving this @Frizi !

kvark (approved):

PR #992, opened 2021-06-18
Reorganize test snapshot folder

### LINKED ISSUE #976: Reorganize test snapshot folder
Since the number of tests is increasing, so I suggest moving test snapshots to subfolders for each front/back.

As is:
```
./test/in
./test/out
```
To be:
```
./test/in/wgsl
./test/in/spv
...

./test/out/wgsl
./test/out/glsl
...
```

Gordon-F:
Fix #976

- All files in `tests/cases/*` removed, because they are unused.
- `wgsl` and `params.ron` files are unchanged (lives in root folder).
- `spv-in` snapshots moved into `tests/in/spv`
- `multiple_entry_points.glsl` added as a replacement of `quad-glsl`.
- `convert_glsl` removed since it unused
- all backends targets moved into its folder in `test/out`.


JCapucho:
The split of quad-glsl concerns me since now we aren't testing multiple entry points on one shader.

kvark (commented):
Looks great, thank you!
@JCapucho is right though, we need to keep that single GLSL for quad_glsl. Is it causing issues?

Gordon-F:
@JCapucho `multiple_entry_points.glsl` added for testing multiple entry points.

kvark (approved):

Issue #940, opened 2021-06-02
spv-in miscompilation of inlined if statement with return


khyperia:
This is a follow-up to https://github.com/gfx-rs/naga/issues/939, when I looked at the output of naga's GLSL backend on that code.

The SPIR-V input (and the GLSL that created it) can be found in this [shader-playground link](http://shader-playground.timjones.io/75e4625445b11eaacbf214d519ae166d). Check the input GLSL program vs. the result GLSL program: if `thingy > 5`, then I would expect `101010` to be returned. But, `0` is returned unconditionally!

Input GLSL:

```glsl
#version 460

layout (location = 1) in float thingy;
layout (location = 0) out float fragColor;

float f() {
    if (thingy > 5) {
        return 101010;
    }
    return 0;
}

void main()
{
	fragColor = f();
}
```

<details>
  <summary>Disassembled SPIR-V, click to expand</summary>
  
```spirv
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 10
; Bound: 49
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main" %thingy %fragColor
               OpExecutionMode %main OriginUpperLeft
               OpSource GLSL 460
               OpName %main "main"
               OpName %thingy "thingy"
               OpName %fragColor "fragColor"
               OpDecorate %thingy Location 1
               OpDecorate %fragColor Location 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
%_ptr_Input_float = OpTypePointer Input %float
     %thingy = OpVariable %_ptr_Input_float Input
    %float_5 = OpConstant %float 5
       %bool = OpTypeBool
%float_101010 = OpConstant %float 101010
    %float_0 = OpConstant %float 0
%_ptr_Output_float = OpTypePointer Output %float
  %fragColor = OpVariable %_ptr_Output_float Output
       %uint = OpTypeInt 32 0
     %uint_0 = OpConstant %uint 0
       %main = OpFunction %void None %3
          %5 = OpLabel
               OpSelectionMerge %46 None
               OpSwitch %uint_0 %41
         %41 = OpLabel
         %42 = OpLoad %float %thingy
         %43 = OpFOrdGreaterThan %bool %42 %float_5
               OpSelectionMerge %45 None
[...]

PR #993, opened 2021-06-19
Refactor control flow validation, allow break in switches

kvark:
Helps #940

### DIFF
--- a/src/valid/function.rs
+++ b/src/valid/function.rs
@@ -65,8 +65,10 @@ pub enum FunctionError {
     InvalidArgumentType { index: usize, name: String },
     #[error("There are instructions after `return`/`break`/`continue`")]
     InstructionsAfterReturn,
-    #[error("The `break`/`continue` is used outside of a loop context")]
-    BreakContinueOutsideOfLoop,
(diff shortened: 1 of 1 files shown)


src/valid/function.rs
```diff
@@ -325,24 +333,37 @@ impl super::Validator {
...
+                    if !context.abilities.contains(ControlFlowAbility::CONTINUE) {
+                        return Err(FunctionError::ContinueOutsideOfLoop);
                     }
                     finished = true;
                 }
                 S::Return { value } => {
-                    if !context.flags.contains(Flags::CAN_JUMP) {
+                    if !context.abilities.contains(ControlFlowAbility::RETURN) {
                         return Err(FunctionError::InvalidReturnSpot);
                     }
```
jimblandy, line 368: Are `Kill` statements permitted in `Loop` `continuing` blocks?

jimblandy (approved):
Looks good to me.

src/valid/function.rs
kvark: I assume it is

kvark:
Thank you for the review!

PR #994, opened 2021-06-19
[glsl-in] Implicit casts for function calls

JCapucho:
Took me some time to figure it out but here's one of the largest missing parts for the glsl frontend


kvark (approved):
UUh, that's quite a bit of logic.
I suppose it's there to support GLSL specifically, right?
I wonder to what extent this is requested. Maybe there is a way to accept some GLSL ES version instead of GLSL desktop, so that implicit casts aren't a thing?

PR #995, opened 2021-06-19
Support u8 and i8 constants

DJMcNab:
If we want to keep this as a targeted change, this PR could be considered complete, however it might be more correct to add int8 and int16 capabilities.

This is part of fixing some issues caused using the wgpu naga validation with [rust-gpu](https://github.com/EmbarkStudios/rust-gpu), specifically. However, this does not block any work there, since we can just not validate.

This still doesn't get the rust-gpu compute example passing, because we now get:
```
Function [1] 'compute_shader::collatz' is invalid: 
        Expression [29] is invalid
```
which I haven't debugged (some way to link these error expressions with their spir-v id in error message would be extremely welcome, so that you can at least hope to debug naga itself).

### DIFF
--- a/src/valid/type.rs
+++ b/src/valid/type.rs
@@ -137,7 +137,11 @@ impl super::Validator {
             crate::ScalarKind::Float => {
                 width == 4 || (width == 8 && self.capabilities.contains(Capabilities::FLOAT64))
             }
-            crate::ScalarKind::Sint | crate::ScalarKind::Uint => width == 4,
+            crate::ScalarKind::Sint | crate::ScalarKind::Uint => {
+                width == 4
+                    || (width == 1/* TODO?: && self.capabilities.contains(Capabilities::INT8)? */)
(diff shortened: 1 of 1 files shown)


src/valid/type.rs
```diff
@@ -137,7 +137,11 @@ impl super::Validator {
             crate::ScalarKind::Float => {
                 width == 4 || (width == 8 && self.capabilities.contains(Capabilities::FLOAT64))
             }
-            crate::ScalarKind::Sint | crate::ScalarKind::Uint => width == 4,
+            crate::ScalarKind::Sint | crate::ScalarKind::Uint => {
+                width == 4
+                    || (width == 1/* TODO?: && self.capabilities.contains(Capabilities::INT8)? */)
```
kvark, line 142: I think having these capabilities would make sense, but nothing would be able to set them to true until the proper support is added to wgpu-hal backends

PR #998, opened 2021-06-20
[glsl-in] more fixes

JCapucho:
A bundle of some more glsl bug fixes and new function implementations.

All commits are self contained and implement a specific feature so they can be reviewed individually


kvark (approved):
Very nice!

PR #1004, opened 2021-06-21
Clear Validator::valid_expression_list before validating each function.

jimblandy:
If validation fails, `Validator` may be left with entries in
`valid_expression_list` and `valid_expression_set`. The validator does clear the
set before examining a function's body, but not the list, which means a
long-lived `Validator` could accumulate an unbounded amount of garbage in the
list.

(Minor cleanup: use structural update syntax where appropriate.)

### DIFF
--- a/src/valid/function.rs
+++ b/src/valid/function.rs
@@ -147,16 +147,7 @@ impl<'a> BlockContext<'a> {
     }
 
     fn with_flags(&self, flags: Flags) -> Self {
-        BlockContext {
-            flags,
-            info: self.info,
-            expressions: self.expressions,
-            types: self.types,
-            global_vars: self.global_vars,
-            functions: self.functions,
-            prev_infos: self.prev_infos,
-            return_type: self.return_type,
-        }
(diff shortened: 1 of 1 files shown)


jimblandy:
This will need to be rebased on #993, so best to put this off until that's merged.

src/valid/function.rs
```diff
@@ -147,16 +147,7 @@ impl<'a> BlockContext<'a> {
...
-            flags,
-            info: self.info,
-            expressions: self.expressions,
-            types: self.types,
-            global_vars: self.global_vars,
-            functions: self.functions,
-            prev_infos: self.prev_infos,
-            return_type: self.return_type,
-        }
+        BlockContext { flags, ..*self }
```
kvark, line 150: niiice

kvark (approved):
The list is actually guaranteed to be clear, since we always drain from it. But it holds *only* if the validation was successful... So you are right, if the validation failed, we can't re-use the validator without this fix.

PR #1007, opened 2021-06-21
[hlsl-out] Implement Access and Unary expressions

Gordon-F:
Required for running gfx-rs quad example with naga path.


kvark (approved):
Thank you!

PR #1008, opened 2021-06-22
Move `FunctionCtx` into back module

Gordon-F:
`FunctionCtx` is used by `wgsl`, `glsl` and `hlsl` backends.


kvark (approved):
Oh interesting, I didn't expect it to be shared, since it may need backend-specific info.
But it seems to be working here.

PR #1013, opened 2021-06-22
[glsl-in] Deduplicate constants so array types match

fintelia:
This fixes GLSL shaders like:
```glsl
#version 450 core

void foo(int i[3]) {}

void main() {
    int array[3];
    foo(array);
}
```

Previous error:
<pre><font color="#F66151"><b>error</b></font><b>: Unknown function &apos;foo&apos;</b>
  <font color="#12488B">┌─</font> array.vert:7:5
  <font color="#12488B">│</font>
<font color="#12488B">7</font> <font color="#12488B">│</font>     <font color="#C01C28">foo(array)</font>;
  <font color="#12488B">│</font>     <font color="#C01C28">^^^^^^^^^^</font>
</pre>

### DIFF
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -1032,7 +1032,7 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
             }
         };
 
-        let handle = self.program.module.constants.append(Constant {
+        let handle = self.program.module.constants.fetch_or_append(Constant {
             name: None,
             specialization: None,
             inner: ConstantInner::Scalar { width, value },


JCapucho (approved):
Looks good to me, my only concern is with larger shaders for every literal we will have to walk the entire constant arena in the worst case.
Maybe a better solution is to have the type comparison check constants for equality.

kvark:
I don't understand. How is constant deduplication fixing this? If we have 2 separate constants with value "3", they should be interchangeable.

fintelia:
The problem is that two handles to the constant "3" don't evaluate to equal:
https://github.com/gfx-rs/naga/blob/a5874eb0cfab4592323e6071d76eab81bceda556/src/front/glsl/functions.rs#L522

kvark:
Why are they not equal?

JCapucho:
Because the size is of type `Handle<Constant>` so it only checks for the handle index and not it's content on the arena

JCapucho:
We need to have a helper function here to perform the comparison, if the type is a an array we need to see if the contents of size match and not the handles

kvark:
Interesting, hmm. So this is a problem with other frontends as well, potentially. I.e. the validation for "a = b" will yell if these are the same-sized arrays just with different constant handles.
We could require the frontends to do at least scalar constant de-duplication, I guess?

JCapucho:
My only concern is that deduplication is `O(n)` over the constant arena and for larger shaders this can start to be a performance problem, not to mention that it destroys the caches.

kvark:
Thinking about this some more. Even if we require this, we wouldn't be able to fully compare `TypeInner` by value anyway, since there are references to other types in there, not just constants.
So full deduplication is not helping much :/ Let's proceed with this PR and keep thinking about the problem.

kvark (approved):

PR #1014, opened 2021-06-22
glsl: Replace todo!() with a NotImplemented error

magcius:
So we can get a source location rather than a panic.

### DIFF
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -797,7 +797,10 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
                             //TODO: declaration
                             // type_qualifier IDENTIFIER SEMICOLON
                             // type_qualifier IDENTIFIER identifier_list SEMICOLON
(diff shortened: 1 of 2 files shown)


kvark (approved):
Nice!

kvark:
Thank you for the first contribution!
Clippy isn't happy

src/front/glsl/parser.rs
```diff
@@ -797,7 +797,10 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
                             //TODO: declaration
                             // type_qualifier IDENTIFIER SEMICOLON
                             // type_qualifier IDENTIFIER identifier_list SEMICOLON
-                            todo!()
+                            Err(ErrorKind::NotImplemented(
+                                token.meta,
+                                "type qualifier",
```
JCapucho, line 802: This message is quite ambiguous can't we name something like `variable modifiers`?

JCapucho (approved):
LGTM

PR #1016, opened 2021-06-23
Reduce global import usage

Gordon-F:
Following up suggestion in #1008. Reduce global import usage for `glsl`, `hlsl`, `msl` and `wgsl` backends.


kvark (approved):
lovely, thank you!

PR #1022, opened 2021-06-24
Handle texture/sampler function arguments

JCapucho:
WIP.

spv-in accepts shaders with sampling using a function argument and the validator partially supports it, the `sampling_set` of `FunctionInfo` only supports globals still so it can't validate shaders with this yet.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -890,7 +890,14 @@ impl<'a, W: Write> Writer<'a, W> {
             back::FunctionType::EntryPoint(_) => &[][..],
             back::FunctionType::Function(_) => &func.arguments,
         };
-        self.write_slice(arguments, |this, i, arg| {
+        let arguments: Vec<_> = arguments
+            .iter()
+            .filter(|arg| match self.module.types[arg.ty].inner {
+                TypeInner::Sampler { .. } => false,
+                _ => true,
+            })
+            .collect();
+        self.write_slice(&arguments, |this, i, arg| {
             // Write the argument type
             match this.module.types[arg.ty].inner {
                 // We treat images separately because they might require
@@ -915,8 +922,6 @@ impl<'a, W: Write> Writer<'a, W> {
                     // any spaces at the beginning or end
                     this.write_image_type(dim, arrayed, class)?;
                 }
-                // glsl has no concept of samplers so we just ignore it
-                TypeInner::Sampler { .. } => return Ok(false),
                 // All other types are written by `write_type`
                 _ => {
                     this.write_type(arg.ty)?;
@@ -927,7 +932,7 @@ impl<'a, W: Write> Writer<'a, W> {
             // The leading space is important
             write!(this.out, " {}", &this.names[&ctx.argument_key(i)])?;
 
-            Ok(true)
+            Ok(())
         })?;
 
         // Close the parentheses and open braces to start the function body
@@ -1023,16 +1028,14 @@ impl<'a, W: Write> Writer<'a, W> {
     /// # Notes
     /// - Adds no newlines or leading/trailing whitespace
     /// - The last element won't have a trailing `,`
-    fn write_slice<T, F: FnMut(&mut Self, u32, &T) -> BackendResult<bool>>(
+    fn write_slice<T, F: FnMut(&mut Self, u32, &T) -> BackendResult>(
         &mut self,
         data: &[T],
         mut f: F,
     ) -> BackendResult {
         // Loop trough `data` invoking `f` for each element
         for (i, item) in data.iter().enumerate() {
-            if !f(self, i as u32, item)? {
-                continue;
-            }
+            f(self, i as u32, item)?;
 
             // Only write a comma if isn't the last element
             if i != data.len().saturating_sub(1) {
@@ -1077,8 +1080,7 @@ impl<'a, W: Write> Writer<'a, W> {
 
                 // Write the comma separated constants
                 self.write_slice(components, |this, _, arg| {
-                    this.write_constant(&this.module.constants[*arg])?;
-                    Ok(true)
+                    this.write_constant(&this.module.constants[*arg])
                 })?;
 
                 write!(self.out, ")")?
@@ -1498,16 +1500,18 @@ impl<'a, W: Write> Writer<'a, W> {
                     self.named_expressions.insert(expr, name);
                 }
                 write!(self.out, "{}(", &self.names[&NameKey::Function(function)])?;
-                self.write_slice(arguments, |this, i, arg| {
-                    let arg_ty = this.module.functions[function].arguments[i as usize].ty;
-                    if let TypeInner::Sampler { .. } = this.module.types[arg_ty].inner {
-                        return Ok(false);
-                    }
-
-                    this.write_expr(*arg, ctx)?;
-
-                    Ok(true)
-                })?;
+                let arguments: Vec<_> = arguments
+                    .iter()
+                    .enumerate()
+                    .filter_map(|(i, arg)| {
+                        let arg_ty = self.module.functions[function].arguments[i].ty;
+                        match self.module.types[arg_ty].inner {
+                            TypeInner::Sampler { .. } => None,
+                            _ => Some(*arg),
+                        }
+                    })
+                    .collect();
+                self.write_slice(&arguments, |this, _, arg| this.write_expr(*arg, ctx))?;
                 writeln!(self.out, ");")?
             }
         }
@@ -1607,10 +1611,7 @@ impl<'a, W: Write> Writer<'a, W> {
                 self.write_type(ty)?;
 
                 write!(self.out, "(")?;
-                self.write_slice(components, |this, _, arg| {
-                    this.write_expr(*arg, ctx)?;
-                    Ok(true)
-                })?;
+                self.write_slice(components, |this, _, arg| this.write_expr(*arg, ctx))?;
                 write!(self.out, ")")?
             }
             // Function arguments are written as the argument name
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1846,8 +1846,6 @@ impl<'a, W: Write> Writer<'a, W> {
                         write!(self.out, ")",)?;
                     }
                 }
-                // TODO: Why is this not implemented ???
-                // return Err(Error::Custom("ImageQuery not implemented".to_string()));
             }
             // `Unary` is pretty straightforward
             // "-" - for `Negate`
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -218,7 +218,7 @@ impl fmt::Display for VaryingName<'_> {
 }
 
 /// Shorthand result used internally by the backend
-type BackendResult = Result<(), Error>;
+type BackendResult<T = ()> = Result<T, Error>;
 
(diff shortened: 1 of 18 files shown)


src/front/spv/image.rs
```diff
@@ -432,26 +441,42 @@ impl<I: Iterator<Item = u32>> super::Parser<I> {
...
+        let image_ty = match expressions[si_lexp.image] {
+            crate::Expression::GlobalVariable(handle) => {
+                if let Some(flags) = self.handle_sampling.get_mut(&handle) {
+                    *flags |= sampling_bit;
+                }
+
+                global_arena[handle].ty
+            }
+            crate::Expression::FunctionArgument(i) => {
+                let flags = function_info.parameters_sampling[i as usize]
```
kvark, line 459: isn't the parameters sampling required to have the size equal to the number of arguments already?

src/front/spv/image.rs
```diff
@@ -432,26 +441,42 @@ impl<I: Iterator<Item = u32>> super::Parser<I> {
...
+                arguments[i as usize].ty
+            }
+            ref other => return Err(Error::InvalidGlobalVar(other.clone())),
+        };
+        match expressions[si_lexp.sampler] {
+            crate::Expression::GlobalVariable(handle) => {
+                *self.handle_sampling.get_mut(&handle).unwrap() |= sampling_bit
+            }
+            crate::Expression::FunctionArgument(i) => {
+                let flags = function_info.parameters_sampling[i as usize]
```
kvark, line 472: same here

src/valid/analyzer.rs
```diff
@@ -181,6 +199,10 @@ pub struct FunctionInfo {
     ///
     /// Each item corresponds to an expression in the function.
     expressions: Box<[ExpressionInfo]>,
+    /// Vector with information of wether or not a functin argument is used for sampling
```
kvark, line 202: "Vector with information of whether or not a function argument is used for sampling"

src/valid/analyzer.rs
```diff
@@ -159,6 +159,24 @@ impl ExpressionInfo {
...
+enum GlobalOrArgument {
+    Global(Handle<crate::GlobalVariable>),
+    Argument(u32),
+}
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+struct ArgumentSampling {
+    /// Whether this argument is used as an image or a sampler
```
kvark, line 174: why do we need to track this? Since we know the type of the argument, it's always unambiguous if it's used as a sampler or as an image

src/valid/analyzer.rs
```diff
@@ -181,6 +199,10 @@ pub struct FunctionInfo {
     ///
     /// Each item corresponds to an expression in the function.
     expressions: Box<[ExpressionInfo]>,
+    /// Vector with information of wether or not a functin argument is used for sampling
+    ///
+    /// Each item corresponds to a function argument
+    argument_sampling: Box<[Option<ArgumentSampling>]>,
```
kvark, line 205: why it be easier to track this as something like:
```rust
struct Sampling {
  image: GlobalOrArgument,
  sampler: Option<GlobalOrArgument>,
}

sampling: FastHashSet<Sampling>,
```

src/valid/analyzer.rs
```diff
@@ -273,21 +295,101 @@ impl FunctionInfo {
...
+        expression_arena: &Arena<crate::Expression>,
+    ) -> Result<FunctionUniformity, FunctionError> {
         for key in info.sampling_set.iter() {
             self.sampling_set.insert(key.clone());
         }
+        for (i, arg_sampling) in info
+            .argument_sampling
+            .iter()
+            .enumerate()
+            .filter_map(|(i, s)| Some((i, s.as_ref()?)))
```
kvark, line 311: nit: `|(i, s)| s.as_ref().map(|s| (i, s))`

src/valid/analyzer.rs
```diff
@@ -397,16 +499,45 @@ impl FunctionInfo {
...
-                        crate::Expression::GlobalVariable(var) => var,
-                        _ => return Err(ExpressionError::ExpectedGlobalVariable),
-                    },
-                    sampler: match expression_arena[sampler] {
-                        crate::Expression::GlobalVariable(var) => var,
-                        _ => return Err(ExpressionError::ExpectedGlobalVariable),
-                    },
-                });
+                let image_storage = match expression_arena[image] {
+                    crate::Expression::GlobalVariable(var) => GlobalOrArgument::Global(var),
```
kvark, line 503: we could make this nicer by something like this:
```rust
impl crate::Expression {
  fn to_global_or_argument(&self) -> Option<GlobalOrArgument> {...}
}
```

src/valid/type.rs
```diff
@@ -239,9 +245,9 @@ impl super::Validator {
                 // `DATA`.
                 let base_info = &self.types[base.index()];
                 let data_flag = if base_info.flags.contains(TypeFlags::SIZED) {
-                    TypeFlags::DATA
+                    TypeFlags::DATA | TypeFlags::ARGUMENT
```
kvark, line 248: shouldn't the `ARGUMENT` always be there for pointers? currently there is a branch of the code where it doesn't get set

src/valid/type.rs
```diff
@@ -350,7 +356,7 @@ impl super::Validator {
                             return Err(TypeError::NonPositiveArrayLength(const_handle));
                         }
 
-                        TypeFlags::SIZED
+                        TypeFlags::SIZED | TypeFlags::ARGUMENT
```
kvark, line 359: is this true that anything `SIZED` can be passed as an argument?

kvark (commented):
Nice work!

src/back/glsl/mod.rs
```diff
@@ -892,14 +892,42 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                    arrayed,
+                    class,
+                } => {
+                    // Write the storage format if needed
+                    if let TypeInner::Image {
+                        class: crate::ImageClass::Storage(format),
+                        ..
+                    } = this.module.types[arg.ty].inner
+                    {
+                        write!(this.out, "layout({}) ", glsl_storage_format(format))?;
```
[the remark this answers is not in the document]
JCapucho: This is a function argument so we shouldn't need to write binding information unless I'm mistaken

src/back/glsl/mod.rs
```diff
@@ -1803,7 +1846,8 @@ impl<'a, W: Write> Writer<'a, W> {
                         write!(self.out, ")",)?;
                     }
                 }
-                return Err(Error::Custom("ImageQuery not implemented".to_string()));
+                // TODO: Why is this not implemented ???
```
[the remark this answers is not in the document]
JCapucho: I didn't remove it because I wasn't sure why it was there in the first place. From what I can see all the necessary logic is there already.

src/front/spv/image.rs
```diff
@@ -19,12 +22,15 @@ bitflags::bitflags! {
...
         handle: Handle<crate::Expression>,
-    ) -> Result<Handle<crate::GlobalVariable>, Error> {
+        global_vars: &Arena<crate::GlobalVariable>,
+        arguments: &[FunctionArgument],
+    ) -> Result<Handle<crate::Type>, Error> {
         match self[handle] {
-            crate::Expression::GlobalVariable(handle) => Ok(handle),
+            crate::Expression::GlobalVariable(handle) => Ok(global_vars[handle].ty),
+            crate::Expression::FunctionArgument(i) => Ok(arguments[i as usize].ty),
             ref other => Err(Error::InvalidGlobalVar(other.clone())),
```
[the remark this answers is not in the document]
JCapucho: Yes, this was just me being lazy

src/front/spv/image.rs
JCapucho: Yes, I think you are confused about the `get_or_insert` right? The type of `parameters_sampling` is `Vec<Option<image::SamplingFlags>>` so the `get_or_insert` refers to the `Option` so we are getting a mutable reference whilst making it `Some` if it wasn't already.

src/valid/type.rs
JCapucho: Now that I mention it. I'm not sure but I think the arrays with constant size that can't be passed aren't themselves valid types, but I'm not sure I'll have to look into it.

src/valid/type.rs
JCapucho: upon closer inspection this isn't true since the following wgsl is invalid (according to our testing suite)
```rust
fn unacceptable_unsized(arg: ptr<storage, array<f32>>) { }
```

JCapucho:
Backend status:

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<tbody>
<tr>
<td class="org-left">Backend</td>
<td class="org-left">Status</td>
</tr>


<tr>
<td class="org-left">Spirv</td>
<td class="org-left">Broken</td>
</tr>


<tr>
<td class="org-left">GLSL</td>
<td class="org-left">Working</td>
</tr>


<tr>
<td class="org-left">WGSL</td>
<td class="org-left">Not tested</td>
</tr>


<tr>
<td class="org-left">Metal</td>
<td class="org-left">Not tested</td>
</tr>


<tr>
<td class="org-left">HLSL</td>
<td class="org-left">Working</td>
</tr>
</tbody>
</table>

Whilst the HLSL backend theoritically works it can&rsquo;t parse the shader in the issue because it has pointers.

The WGSL and Metal backends emit without errors but I don&rsquo;t know if the code is valid.

Frontend status:

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<tbody>
<tr>
<td class="org-left">Frontend</td>
<td class="org-left">Status</td>
</tr>


<tr>
<td class="org-left">Spirv</td>
<td class="org-left">Working</td>
</tr>


<tr>
<td class="org-left">GLSL</td>
<td class="org-left">Working</td>
</tr>


<tr>
<td class="org-left">WGSL</td>
<td class="org-left">Broken</td>
</tr>
</tbody>
</table>

kvark:
This is great, thank you for taking care of this complex feature!

> The WGSL and Metal backends emit without errors but I don’t know if the code is valid.

If we add a snapshot test, the CI will check if the Metal produced code is valid. WGSL you should be able to check locally as well.

> Now that you mention it. I'm not sure but I think the arrays with constant size that can't be passed aren't themselves valid types, but I'm not sure I'll have to look into it.

Still seems to me that we could get away with just SIZED here.

JCapucho:
> If we add a snapshot test, the CI will check if the Metal produced code is valid. WGSL you should be able to check locally as well.

Yes CI can check metal but we can't check WGSL because the frontend is broken.

> Still seems to me that we could get away with just SIZED here.

Sure I'll revert we didn't accept them as arguments so this isn't breaking anything.

src/valid/type.rs
JCapucho: Actually the old code accepts any sized array as an argument

kvark (commented):
A few more small things

kvark (approved):
great!

PR #1026, opened 2021-06-25
[spv-out] Properly reset `Writer`s between module writes.

jimblandy:
A single `Writer` can be reused to convert any number of Naga IR modules to
SPIR-V. This lets us reuse the heap allocations, but makes us responsible for
resetting the state of the `Writer` fully between modules. The old code
neglected to clear several different fields: logical_layout (all fields),
lookup_type, and cached_constants.

This commit uses Rust struct literal expressions to construct the re-initialized
`Writer`, which makes the compiler check that each field has been handled.

It also introduces a policy limiting how much storage will be retained, so that
processing the occasional whale doesn't leave us with leviathan buffers
permanently.

### DIFF
--- a/src/back/spv/recyclable.rs
+++ b/src/back/spv/recyclable.rs
@@ -1,0 +1,74 @@
+//! Reusing collections' previous allocations.
+
+/// A value that can be reset to its initial state, retaining its current allocations.
+///
+/// Naga attempts to lower the cost of SPIR-V generation by allowing clients to
+/// reuse the same `Writer` for multiple Module translations. Reusing a `Writer`
+/// means that the `Vec`s, `HashMap`s, and other heap-allocated structures the
+/// `Writer` uses internally begin the translation with heap-allocated buffers
+/// ready to use.
+///
+/// But this approach introduces the risk of `Writer` state leaking from one
+/// module to the next. When a developer adds fields to `Writer` or its internal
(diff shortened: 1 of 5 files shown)


src/back/spv/recyclable.rs
```diff
@@ -0,0 +1,74 @@
...
+
+// Stock values for various collections.
+
+/// Maximum extra capacity that a recycled vector is allowed to have. If the
+/// actual capacity is larger, we re-allocate the vector storage with lower
+/// capacity.
+const MAX_EXTRA_CAPACITY_PERCENT: usize = 200;
+
+/// Minimum extra capacity to keep when re-allocating the vector storage.
+const MIN_EXTRA_CAPACITY_PERCENT: usize = 20;
```
kvark, line 38: Do we *actually* need to have this capacity shrinking? We talked about this on the matrix, and I tried to say that we are likely OK without ever shrinking.

kvark (approved):

src/back/spv/recyclable.rs
jimblandy: Sorry, I wasn't trying to be intransigent - I think I just misunderstood the conclusion we'd reached in chat. I'll try to confirm decisions more fully in the future.

jimblandy: Filed #1054 to remove capacity shrinking.

PR #1032, opened 2021-06-26
[spv-out] Track block termination statically.

jimblandy:
Rather than giving `Block` an optional `terminator` field in addition to its
`body`, track block termination statically, with two types:

- `Block` is a block without a termination instruction. This is what most code
  generation functions operate on.

- `TerminatedBlock` is a block with a termination instruction. This is what
  `Function::blocks` holds.

The `Function::consume` method takes a `Block` by value, together with a
termination instruction, and turns it into a `TerminatedBlock`.

This lets us remove some unwraps and awkward conditions.

As part of this change, `Writer::write_block` no longer hits an `unimplemented!`
for Naga statements following `Break`, `Return`, and so on. Instead, it simply
doesn't emit code for them, which is a correct translation. If we want to forbid
these, we should handle that in validation instead.

### DIFF
--- a/src/back/spv/mod.rs
+++ b/src/back/spv/mod.rs
@@ -72,10 +72,26 @@ impl IdGenerator {
     }
 }
 
+/// A SPIR-V block to which we are still adding instructions.
+///
+/// A `Block` represents a SPIR-V block that does not yet have a termination
+/// instruction like `OpBranch` or `OpReturn`.
+///
+/// The `OpLabel` that starts the block is implicit. It will be emitted based on
+/// `label_id` when we write the block to a `LogicalLayout`.
+///
+/// To terminate a `Block`, pass the block and the termination instruction to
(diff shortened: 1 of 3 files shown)


jimblandy:
I believe all commits are CI-clean, so this does not need to be squashed.

src/back/spv/mod.rs
```diff
@@ -72,18 +72,33 @@ impl IdGenerator {
...
+/// `Function::consume`. This takes ownership of the `Block` and transforms it
+/// into a `TerminatedBlock`.
 struct Block {
     label_id: Word,
     body: Vec<Instruction>,
-    termination: Option<Instruction>,
+}
+
+/// A SPIR-V block that ends with a termination instruction.
+struct TerminatedBlock {
```
kvark, line 92: could also just have it as a generic argument on `Block`

kvark (approved):
Looks like a solid improvement, thank you!

PR #1033, opened 2021-06-26
[spv-out] Introduce a `BlockContext` type to simplify writer methods

jimblandy:
This PR introduces a `BlockContext` type to carry common arguments for SPIR-V generation, reducing long argument lists in `src/back/spv`; all Clippy `#[allow]` attrs for long argument lists are removed.

This PR is dependent on #1032, textually, but not technically. I wanted to avoid a rebase if both are accepted, but if you don't like #1032, this can be adapted to work without it.

This PR is probably best reviewed by looking at individual commits. It unavoidably moves a bunch of functions around, from the `Writer` impl block to a new `BlockContext` impl block. But it also needs to make lots of minor changes to their content. To make review tractable, the code motion is separated out into its own commit that splits the `impl Writer` block into two separate `impl Writer` blocks. Then, the commit that introduces `BlockContext` simply changes one of them to an `impl BlockContext` block, so that the accompanying changes to the function bodies can be seen clearly.


jimblandy:
I believe all commits in this PR are CI-clean, so it does not need to be squashed before merging.

kvark (approved):
Love those extensive comments, thank you!
The only thing that we are possibly missing here is taking the opportunity to move things out of the writer module as we are heavily refactoring it.

kvark:
(needs a rebase)

jimblandy:
rebased, will try moving `BlockContext` into its own file

jimblandy:
Okay, this version adds a final commit that moves `BlockContext` into `back/spv/block.rs`, and makes a few associated cleanups.

kvark:
Crazy diff of 2K LOC!

PR #1038, opened 2021-06-26
[glsl-in] Fix function lookup

magcius:
First look for exact matches before trying the implicit
lookup. Otherwise, finding two wrong matches first will
result in an ambiguous function error.

This logic could use with a bit of a cleanup, keeping
track of a closer score and doing more exact matching
on types, but this might be good enough for now.

### DIFF
--- a/src/front/glsl/functions.rs
+++ b/src/front/glsl/functions.rs
@@ -507,6 +507,7 @@ impl Program<'_> {
                         })?;
 
                         let mut maybe_decl = None;
+                        let mut ambiguous = false;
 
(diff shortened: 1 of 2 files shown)


JCapucho (changes requested):
Whilst this solution work you're duplicating code and adding another loop which can get expensive real quick, there's a simpler solution add a boolean called `ambiguous` and only throw the error if `ambiguous` is `true` and `exact` is `false`

magcius:
Done.

JCapucho (approved):
LGTM

PR #1040, opened 2021-06-27
[hlsl-out] Implement loop statement and composite constants

Gordon-F:


src/back/hlsl/writer.rs
```diff
@@ -323,15 +324,13 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                writeln!(self.out, ";")?;
             }
-            crate::ConstantInner::Composite { .. } => {
-                return Err(Error::Unimplemented(format!(
-                    "write_global_constant Composite {:?}",
-                    inner
-                )))
+            crate::ConstantInner::Composite { ty, ref components } => {
+                self.write_type(module, ty)?;
+                self.write_composite_constant(module, ty, components)?;
```
kvark, line 330: we need to separate the passes in 2. Instead of going through all of the constants, we need to go through the scalar constants first, then write down the types (since the arrays may depend on the scalar constants), and afterwards go through the composite constants (since they depend on types).

kvark (approved):
Nice!

PR #1041, opened 2021-06-27
[spv-in] First cut at reveal-dominators algorithm for #904

jimblandy:
Here's a very poor implementation of the switch-only algorithm I proposed here:
https://github.com/gfx-rs/naga/issues/940#issuecomment-869207997

Problems:
- quadratic break disposition test
- too many splices and drains


jimblandy:
Hmm. Pretty sure the impl Disposition for Block is wrong.

jimblandy:
I think we've decided that we can't fully address #940 with structural manipulations of the Naga IR.

PR #1043, opened 2021-06-28
glsl-out: adjust coordinate space

kvark:
Similar to SPIR-V, by default we should assume the "native" coordinate space of the language. In case of OpenGL, this means depth from -1 to 1.
@JCapucho we'll need to do something opposite for glsl-in, unless we are already doing something :)

Draft because I'm still figuring out how to make `shadow` example working

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -126,6 +126,15 @@ impl fmt::Display for Version {
     }
 }
 
+bitflags::bitflags! {
+    #[cfg_attr(feature = "serialize", derive(serde::Serialize))]
+    #[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
+    pub struct WriterFlags: u32 {
+        /// Extend output Z from (0,1) to (-1,1).
+        const ADJUST_COORDINATE_SPACE = 0x1;
+    }
+}
+
 /// Structure that contains the configuration used in the [`Writer`](Writer)
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "serialize", derive(serde::Serialize))]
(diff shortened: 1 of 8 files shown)


JCapucho:
Thanks for warning me, indeed the glsl frontend needs this. At first I thought that the newer versions followed the vulkan coordinate spec but no they are still tied to OpenGL.

src/back/glsl/mod.rs
```diff
@@ -1455,6 +1470,22 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                            if self.module.entry_points[ep_index as usize].stage
+                                == crate::ShaderStage::Vertex
+                                && self
+                                    .options
+                                    .writer_flags
+                                    .contains(WriterFlags::ADJUST_COORDINATE_SPACE)
+                            {
+                                writeln!(
+                                    self.out,
+                                    "gl_Position.z = gl_Position.z * 2.0 - gl_Position.w;",
```
JCapucho, line 1484: I don't know much about the coordinate systems but from what I know and read about whilst searching for reviewing this, first shouldn't we flip the Y axis? And second from what I read here https://matthewwellings.com/blog/the-new-vulkan-coordinate-system/ it says that the code for doing this should be
```
gl_Position.z = (gl_Position.z + gl_Position.w) / 2.0;
```
Once again I'll say that I don't really know much about this, so feel free to ignore this comment

kvark: I don't think the Y needs to be flipped. Otherwise, we'd see this right away in the GL backend.
The transformation you posted is for the *opposite*: going from (-1,1) to (0,1).

kvark:
Alright, I'm struggling to debug this on GL/Linux. I think this PR is a solid improvement anyway and needs to go in.

PR #1047, opened 2021-06-29
[wgsl-out] Convert shift expressions to u32

### LINKED ISSUE #1037: [wgsl-out] Signed shifts are not supported
This GLSL code does not export to WGSL correctly:

```glsl
// Convert from 0...1 UNORM range to SNORM range
vec3 ConvertToSignedInt(vec3 t_Input) {
    ivec3 t_Num = ivec3(t_Input * 255.0);
    // Sign extend
    t_Num = t_Num << 24 >> 24;
    return vec3(t_Num) / 127.0;
}
```

Tint says:

```
1 error(s) generated while compiling the shader:
256:20 error: Binary expression operand types are invalid for this operation: vec3<i32> shift_left i32
    t_Num = ((_e26 << 24) >> 24);
```

We must emit 24u for both shift sizes.

Gordon-F:
Fix #1037

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -908,10 +908,56 @@ impl<W: Write> Writer<W> {
                 write!(self.out, "{}", name)?;
             }
             Expression::Binary { op, left, right } => {
+                let (convert_to_uint_left, convert_to_uint_right) = match op {
+                    crate::BinaryOperator::ShiftLeft | crate::BinaryOperator::ShiftRight => {
+                        let left_inner = func_ctx.info[left].ty.inner_with(&module.types);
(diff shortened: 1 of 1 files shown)


kvark (changes requested):
Thank you for taking a stab at this!
I believe this is the wrong place to fix it. Instead, we should enforce the IR to have unsigned shifts. So the work would be done in the frontends and the validator.

Gordon-F:
@kvark Ok, this PR can be closed. 

Only GLSL accept shift operations with signed scalar/vector?

kvark:
Iirc, negative shifts are UB in C, and some of the shadong languages inherit this UB.
By default, our IR tries to follow WGSL, unless there is a good reason not to.

PR #1048, opened 2021-06-29
[CI] Add dneto0 spirv-samples to lazy task

Gordon-F:
Test run - https://github.com/Gordon-F/naga/runs/2943811724?check_suite_focus=true

Part of #747

### DIFF
--- a/.github/workflows/lazy.yml
+++ b/.github/workflows/lazy.yml
@@ -76,3 +76,76 @@ jobs:
               echo "Result: $(expr $FILE_COUNT - $SUCCESS_RESULT_COUNT) / $FILE_COUNT" > counter
           done
           cat counter
+  dneto0_spirv-samples:
+    name: Parse dneto0 spirv-samples
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+      - name: Download shaders
+        run: git clone https://github.com/dneto0/spirv-samples.git
+      - name: Build Naga
+        run: cargo build --release --bin naga
+      - name: Install spirv-tools
+        run: |
+           wget -q https://storage.googleapis.com/spirv-tools/artifacts/prod/graphics_shader_compiler/spirv-tools/linux-clang-release/continuous/1489/20210629-121459/install.tgz
+           tar zxf install.tgz
+           ./install/bin/spirv-as --version
+      - name: Compile spv from spvasm
+        run: |
+           cd spirv-samples
+           mkdir -p spv
+
+           find "./spvasm" -name '*.spvasm' | while read fname;
+           do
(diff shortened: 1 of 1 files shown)


.github/workflows/lazy.yml
```diff
@@ -76,3 +76,77 @@ jobs:
...
+          profile: minimal
+          toolchain: stable
+      - name: Download shaders
+        run: git clone https://github.com/dneto0/spirv-samples.git
+      - name: Build Naga
+        run: cargo build --release --bin naga
+      - name: Install vulkan-sdk
+        run: |
+           wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo apt-key add -
+           sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.2.176-focal.list https://packages.lunarg.com/vulkan/1.2.176/lunarg-vulkan-1.2.176-focal.list
```
kvark, line 95: is this necessary? can we just install `vulkan-tools` from the standard package repositories?

.github/workflows/lazy.yml
```diff
@@ -76,3 +76,77 @@ jobs:
...
+           sudo apt install vulkan-sdk
+      - name: Compile spv from spvasm
+        run: |
+           cd spirv-samples
+           mkdir -p spv
+
+           find "./spvasm" -name '*.spvasm' | while read fname;
+           do
+               echo "Convert to spv with glslc: $fname"
+               glslc --target-spv=spv1.3 $(realpath ${fname}) -o ./spv/$(basename ${fname}).spv
```
kvark, line 106: we should use `spirv-as` instead of `glslc` for this

kvark (commented):
Very nice, thank you!
Just a few notes

.github/workflows/lazy.yml
Gordon-F: I don't know how I can fix this error (Everything is ok on my machine 😄):
```
Convert to spv with glslc: ./spvasm/SpvParserImportTest_Import_NonSemantic_IgnoredExtInsts.spvasm
error: 26: 27: Couldn't translate unknown extended instruction name 'ArgumentInfo' to unsigned integer.
```

Looks like `spirv-tools` is outdated on build agent:
```
SPIRV-Tools v2020.1 unknown hash, 2020-03-23T08:58:00
Target: SPIR-V 1.5
```

https://github.com/Gordon-F/naga/runs/2953885438?check_suite_focus=true

kvark: does it matter that spirv-tools is from 2020? it should still work. It's not like we are using any fancy features in SPIR-V anyway

Gordon-F: It works with latest prebuild [spirv-tools](https://github.com/KhronosGroup/SPIRV-Tools).

https://github.com/Gordon-F/naga/runs/2954042540?check_suite_focus=true

kvark (approved):
Awesome work, thank you!

PR #1051, opened 2021-06-30
[glsl-in] Add support for texture function with LOD bias

magcius:


src/front/glsl/functions.rs
```diff
@@ -152,9 +152,12 @@ impl Program<'_> {
                         Ok(Some(args[0].0))
                     }
                     "texture" => {
-                        if args.len() != 2 {
```
JCapucho, line 155: You can write this as
```rust
if !(2..=3).contains(&args.len()) {
  return Err(ErrorKind::wrong_function_args(name, 2, args.len(), meta));
}
```

src/front/glsl/functions.rs
```diff
@@ -152,9 +152,12 @@ impl Program<'_> {
...
-                            return Err(ErrorKind::wrong_function_args(name, 2, args.len(), meta));
-                        }
+                        let mut level = SampleLevel::Auto;
+                        match args.len() {
+                            3 => { level = SampleLevel::Bias(args[2].0); },
+                            2 => { },
+                            _ => { return Err(ErrorKind::wrong_function_args(name, 2, args.len(), meta)); }
+                        };
```
JCapucho, line 160: You can write this as
```rust
args.get(2).map(SampleLevel::Bias).unwrap_or(SampleLevel::Auto);
```

JCapucho (changes requested):
We can write this in more idiomatic rust

magcius:
Thank you! Still getting used to idiomatic Rust

JCapucho (approved):
Looks good now, thank you for your contribution.

PR #1059, opened 2021-07-02
[glsl-in] Convert bool -> scalar cast to Select

magcius:
Code is quite terrible since I'm not a Rust expert yet, but it seems to work in my (admittedly limited) testing. Would love help making it better.

### DIFF
--- a/src/front/glsl/functions.rs
+++ b/src/front/glsl/functions.rs
@@ -1,13 +1,35 @@
 use crate::{
-    proc::ensure_block_returns, Arena, BinaryOperator, Block, EntryPoint, Expression, Function,
+    proc::ensure_block_returns, Arena, BinaryOperator, Block, Constant, ConstantInner, EntryPoint, Expression, Function,
     FunctionArgument, FunctionResult, Handle, ImageQuery, LocalVariable, MathFunction,
-    RelationalFunction, SampleLevel, ScalarKind, Statement, StructMember, SwizzleComponent, Type,
+    RelationalFunction, SampleLevel, ScalarKind, ScalarValue, Statement, StructMember, SwizzleComponent, Type,
     TypeInner, VectorSize,
 };
 
 use super::{ast::*, error::ErrorKind, SourceMetadata};
 
 impl Program<'_> {
+    fn add_constant_value(
+        &mut self,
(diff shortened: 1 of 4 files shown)


src/front/glsl/functions.rs
```diff
@@ -24,13 +46,41 @@ impl Program<'_> {
...
+                    let vector_size = match *expr_type {
+                        TypeInner::Vector{ size, .. } => Some(size),
+                        _ => None,
                     };
 
+                    // Special case: if casting from a bool, we need to use Select and not As.
+                    let result_type_inner = &self.module.types[ty].inner;
+                    let result_scalar_kind = result_type_inner.scalar_kind();
+                    if expr_type.scalar_kind() == Some(ScalarKind::Bool) && result_scalar_kind.is_some() && result_scalar_kind != Some(ScalarKind::Bool) {
+                        let c0 = self.add_constant_value(result_scalar_kind.unwrap(), 0u64);
```
kvark, line 60: A dirty way to simplify this would be `let result_scalar_kind = result_type_inner.scalar_kind().unwrap_or(ScalarKind::Bool);`

A pure/clean way would be to express this as a `match`:
```rust
match result_type_inner.scalar_kind() {
  Some(result_scalar_kind) if expr_type.scalar_kind() == Some(ScalarKind::Bool) => {...}
  _ => {}
}
```

magcius: Done.

src/front/glsl/functions.rs
```diff
@@ -1,13 +1,35 @@
...
+        value: u64,
+    ) -> Handle<Constant> {
+        let value = match scalar_kind {
+            ScalarKind::Uint => ScalarValue::Uint(value),
+            ScalarKind::Sint => ScalarValue::Sint(value as i64),
+            ScalarKind::Float => ScalarValue::Float(value as f64),
+            _ => unreachable!(),
+        };
+
+        self.module.constants.append(Constant {
```
kvark, line 23: given that these are likely going to repeat, let's use `fetch_or_append` here

kvark (commented):
Sorry, one more small thing!

kvark (approved):

PR #1063, opened 2021-07-02
[spv-out] accept texture trough function argument

JCapucho:
Follow up to #1022

Still a draft because it seems like vulkan doesn't quite like it

### DIFF
--- a/src/back/mod.rs
+++ b/src/back/mod.rs
@@ -175,3 +175,13 @@ fn vector_size_str(size: crate::VectorSize) -> &'static str {
         crate::VectorSize::Quad => "4",
     }
 }
+
+impl crate::TypeInner {
+    #[allow(unused)]
+    fn is_handle(&self) -> bool {
+        match *self {
+            crate::TypeInner::Image { .. } | crate::TypeInner::Sampler { .. } => true,
+            _ => false,
+        }
+    }
+}
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -703,7 +703,7 @@ impl<'w> BlockContext<'w> {
                 array_index,
                 index,
             } => {
-                let image_id = self.get_image_id(block, image)?;
+                let image_id = self.get_image_id(image);
                 let coordinate_id =
                     self.write_texture_coordinates(coordinate, array_index, block)?;
 
@@ -773,7 +773,7 @@ impl<'w> BlockContext<'w> {
             } => {
                 use super::instructions::SampleLod;
                 // image
-                let image_id = self.get_image_id(block, image)?;
+                let image_id = self.get_image_id(image);
                 let image_type = self.fun_info[image].ty.handle().unwrap();
                 // Vulkan doesn't know about our `Depth` class, and it returns `vec4<f32>`,
                 // so we need to grab the first component out of it.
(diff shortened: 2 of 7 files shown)


JCapucho:
Spirv validation no longer complains

src/back/spv/writer.rs
```diff
@@ -269,7 +269,18 @@ impl Writer {
         let mut parameter_type_ids = Vec::with_capacity(ir_function.arguments.len());
         for argument in ir_function.arguments.iter() {
             let class = spirv::StorageClass::Input;
-            let argument_type_id = self.get_type_id(LookupType::Handle(argument.ty))?;
+            let handle_ty = match ir_module.types[argument.ty].inner {
```
kvark, line 272: we should have something like this in `src/back/mod.rs`
```rust
impl crate::TypeInner {
  fn is_handle(&self) -> bool {.. }
}
```

src/back/spv/block.rs
```diff
@@ -1233,17 +1233,32 @@ impl<'w> BlockContext<'w> {
...
             }
+            crate::Expression::FunctionArgument(i) => {
+                let result_type_id = self.get_type_id(LookupType::Handle(
+                    self.ir_function.arguments[i as usize].ty,
+                ))?;
+                let pointer_id = self.function.parameter_id(i);
+                let id = self.gen_id();
+                block
+                    .body
+                    .push(Instruction::load(result_type_id, id, pointer_id, None));
```
kvark, line 1257: this seems strange, a bit. Is `global_variables` storing the loaded value in the other branch? There is no reason why they should diverge. If the other doesn't load, this one shouldn't either.

kvark (changes requested):

src/back/spv/block.rs
JCapucho: The globals are implicitly loaded in `writer.rs:407`, I could add a new struct similar to `GlobalVariable` (or reuse it).

kvark: I think my general point still stands - this code needs to be more straightforward. As such, we can either emit the `OpLoad` expressions on all of the arguments of the function that are handles at the beginning of the function. Or we can stop trying to do this for globals as well ahead of time, and instead only `OpLoad` at the time of need, e.g. right before the sampling instruction.

JCapucho: I've pushed some changes that make the global be loaded as needed let me know what you think about it

src/back/mod.rs
```diff
@@ -175,3 +175,12 @@ fn vector_size_str(size: crate::VectorSize) -> &'static str {
         crate::VectorSize::Quad => "4",
     }
 }
+
+impl crate::TypeInner {
+    fn is_handle(&self) -> bool {
```
kvark, line 180: let's put `#[allow(unused)]` on this

tests/out/spv/image.spvasm
```diff
@@ -1,19 +1,19 @@
 ; SPIR-V
 ; Version: 1.1
 ; Generator: rspirv
-; Bound: 174
+; Bound: 195
```
kvark, line 4: ok, unfortunately this is sort of a blocker. We are increasing the ID count by 10% here without changing any functionality. We need to find a way to avoid this.
If I understand correctly, this PR is currently doing `OpLoad` on demand for globals and parameters.
What if we do it at the function start instead?

kvark (changes requested):

tests/out/spv/image.spvasm
JCapucho: Done, now the everything is loaded ahead of time

kvark (approved):
Great, thank you!

PR #1065, opened 2021-07-02
[glsl-in] Add support for gl_FragCoord / gl_FragDepth

magcius:
And remove the ability to read gl_Position in fragment.

### DIFF
--- a/src/front/glsl/variables.rs
+++ b/src/front/glsl/variables.rs
@@ -100,7 +100,26 @@ impl Program<'_> {
                 },
                 BuiltIn::Position,
                 true,
+                PrologueStage::empty(),
+            ),
+            "gl_FragCoord" => add_builtin(
+                TypeInner::Vector {
+                    size: VectorSize::Quad,
+                    kind: ScalarKind::Float,
+                    width: 4,
+                },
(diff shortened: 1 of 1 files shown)


src/front/glsl/variables.rs
```diff
@@ -100,6 +100,25 @@ impl Program<'_> {
...
+                PrologueStage::FRAGMENT,
+            ),
+            "gl_FragDepth" => add_builtin(
+                TypeInner::Scalar {
+                    kind: ScalarKind::Float,
+                    width: 4,
+                },
+                BuiltIn::FragDepth,
+                true,
                 PrologueStage::FRAGMENT,
```
JCapucho, line 122: I don't think that it can be read in the fragment stage has its only defined as `out`

JCapucho (changes requested):
Overall looks good, but I have some doubts regarding the `gl_FragDepth`

src/front/glsl/variables.rs
magcius: Right, I forgot that PrologueStage was about reading. I wrote this before we chatted about that.

JCapucho (approved):
Looks good now

PR #1066, opened 2021-07-04
wgsl-in: add more error tests

grovesNL:
Added some more error snapshots, and removed a few error variants that aren't used anymore.


kvark (approved):
Great, thank you!

PR #1068, opened 2021-07-05
wgsl-in: add more error tests

grovesNL:
Most errors still need improvements (using debug output for type handles, better descriptions, etc.), but just trying to capture most of the existing behavior for now, and switch to spans wherever it's easily possible.

We cover most of `wgsl::Error` with snapshots now, but still missing a few places (e.g. `ResolveError` variants or all of the possible cases for `Unexpected` tokens).

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -88,84 +88,43 @@ pub enum ExpectedToken<'a> {
     StructAttribute,
 }
 
-#[derive(Clone, Debug, Error)]
+#[derive(Clone, Debug)]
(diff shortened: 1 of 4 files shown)


src/front/wgsl/mod.rs
```diff
@@ -88,84 +88,43 @@ pub enum ExpectedToken<'a> {
     StructAttribute,
 }
 
-#[derive(Clone, Debug, Error)]
+#[derive(Clone, Debug)]
```
kvark, line 91: neat!

kvark (approved):

PR #1069, opened 2021-07-05
[spv-out] Support object-wise select

JCapucho:
The IR accepts select expressions which take a bool condition and select
between two vectors, whilst spirv only allows for component-wise
selects.

Now if the condition is a scalar and the object a vector we implicitly
splat the condition.

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -933,9 +933,42 @@ impl<'w> BlockContext<'w> {
                 reject,
             } => {
                 let id = self.gen_id();
-                let condition_id = self.cached(condition);
+                let mut condition_id = self.cached(condition);
                 let accept_id = self.cached(accept);
                 let reject_id = self.cached(reject);
+
+                let condition_ty = self.fun_info[condition]
+                    .ty
+                    .inner_with(&self.ir_module.types);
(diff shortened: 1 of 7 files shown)


src/back/spv/block.rs
```diff
@@ -933,10 +933,40 @@ impl<'w> BlockContext<'w> {
...
                 let accept_id = self.cached(accept);
                 let reject_id = self.cached(reject);
 
+                let condition_ty = self.fun_info[condition]
+                    .ty
+                    .inner_with(&self.ir_module.types);
+                let object_ty = self.fun_info[accept].ty.inner_with(&self.ir_module.types);
+
+                if let (
+                    &crate::TypeInner::Scalar { kind, width },
```
kvark, line 946: This should always check for booleans in the kind: `kind: crate::ScalarKind::Bool`

tests/in/object-select.wgsl
```diff
@@ -0,0 +1,7 @@
+[[stage(fragment)]]
```
kvark, line 1: I don't think we should have an individual test file for every feature of the IR. Let's put this into `operations.wgsl`, can we?

kvark (changes requested):
Looks good!
Just a couple small things

kvark (approved):
Thank you!

PR #1071, opened 2021-07-05
[hlsl-out] More improvements 

Gordon-F:
1. Implement `Select` expression.
2. Enable `interface` snapshot
3. Return only real names for entry point in ReflectionInfo, instead of tuple (stage, name).
4. Properly write locations for fragment output (`SV_TargetN`, instead of `LOCN`)


kvark (approved):
Nice, thank you!

PR #1072, opened 2021-07-05
Type inference for local variables in WGSL

kvark:
Looks like the WGSL samples now require this.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -124,7 +124,8 @@ pub enum Error<'a> {
     ZeroSizeOrAlign(Span),
     InconsistentBinding(Span),
     UnknownLocalFunction(Span),
-    LetTypeMismatch(Span, Handle<crate::Type>),
+    InitializationTypeMismatch(Span, Handle<crate::Type>),
+    MissingType(Span),
     Other,
 }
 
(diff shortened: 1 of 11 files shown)


src/front/wgsl/mod.rs
```diff
@@ -326,11 +327,16 @@ impl<'a> Error<'a> {
                 labels: vec![(span.clone(), "unknown local function".into())],
                 notes: vec![],
             },
-            Error::LetTypeMismatch(ref name_span, ref expected_ty) => ParseError {
+            Error::InitializationTypeMismatch(ref name_span, ref expected_ty) => ParseError {
                 message: format!("the type of `{}` is expected to be {:?}", &source[name_span.clone()], expected_ty),
                 labels: vec![(name_span.clone(), format!("definition of `{}`", &source[name_span.clone()]).into())],
                 notes: vec![],
             },
+            Error::MissingType(ref name_span) => ParseError {
```
grovesNL, line 335: nit: would be useful to have a snapshot test for this too

grovesNL (approved):

src/front/wgsl/mod.rs
kvark: good point, adding now!

PR #1074, opened 2021-07-06
[glsl-in] Improvements and fixes

JCapucho:
yet another one of these types of PR, this time we have
- Support for parsing all texture types and for shadow samplers
- Implicit splatting of arguments for the `max` builtin
- A fix for implicit conversions using bitcast instead of a regular cast
- Support for parsing array initializers of the form `type[maybe_size](components)`, glsl has this quirk that initializers for arrays without a size aren't actually dynamic but the size comes from the argument count, so things like `float[](1,2,3,4)` is of type `float[4]`


JCapucho:
The commits as always incorporate a single feature/fix so reviewing them individually might prove more useful

kvark (approved):
Wow, amazing work!

PR #1075, opened 2021-07-06
[hlsl-out] Zero initialize variables. Write interpolation and sampling qualifier

Gordon-F:


src/back/hlsl/writer.rs
```diff
@@ -1388,3 +1410,97 @@ fn scalar_kind_str(kind: crate::ScalarKind, width: crate::Bytes) -> Result<&'sta
...
+        Some("ByteAddressBuffer")
+    } else if storage_access.is_all() {
+        Some("RWByteAddressBuffer")
+    } else {
+        None
+    }
+}
+
+/// Helper function that return string with default zero initialization for supported types
+fn zero_init_value_str(inner: &TypeInner) -> Option<String> {
```
kvark, line 1425: Can we just use the C-style initialization:
```cpp
Something foo = {}; // initializes to zeroes automatically
```

tests/snapshots.rs
```diff
@@ -362,7 +362,7 @@ fn convert_wgsl() {
         ),
         (
             "interpolate",
-            Targets::SPIRV | Targets::METAL | Targets::GLSL | Targets::WGSL,
+            Targets::SPIRV | Targets::METAL | Targets::GLSL | Targets::HLSL | Targets::WGSL,
```
kvark, line 365: very nice!

kvark (changes requested):
Looks great! Just one note here

src/back/hlsl/writer.rs
Gordon-F: You mean for `struct`? Currently this code initialize only scalar, vector and matrix.

kvark: I assume `= {}` works for all types in HLSL just like it does in C.

Gordon-F: TIL, Thank you! Fixed.

kvark: Why do we still need this function then?

Gordon-F: This not working for hlsl:
```
./shader.hlsl:7:14: error: too few elements in vector initialization (expected 2 elements, have 0)
  float2 v = {};
```

Based on your suggestion I reworked from:
```hlsl
float2 v = float2(0.0, 0.0);
```
To:
```hlsl
float2 v = { 0.0, 0.0 };
```

kvark: Apologies, I confused you with this advice! HLSL has zero initialization semantics but with a different syntax from C:
```
Foo foo = (Foo)0;
```
This should work on all types.

Gordon-F: Thank you! I think it's finally fixed :smile:

Reworked this function and now it's write directly to buffer to avoid heap allocation.

src/back/hlsl/writer.rs
```diff
@@ -1331,6 +1353,62 @@ impl<'a, W: Write> Writer<'a, W> {
 
         Ok(())
     }
+
+    /// Helper function that write default zero initialization for supported unitialized types
+    fn write_default_init(&mut self, module: &Module, ty: Handle<crate::Type>) -> BackendResult {
```
kvark, line 1358: sorry, I still don't understand why we have this function.
Zero-initialization should be as simple as:
```cpp
write!(self.out, "(")?;
self.write_type(ty)?;
write!(self.out, ")0")?;
```

src/back/hlsl/writer.rs
```diff
@@ -43,13 +43,13 @@ pub struct Writer<'a, W> {
...
 impl<'a, W: Write> Writer<'a, W> {
-    pub fn new(out: W, options: &'a Options) -> Self {
+    pub fn new(out: W, options: &'a Options, module: &Module) -> Self {
         Self {
             out,
             names: crate::FastHashMap::default(),
             namer: proc::Namer::default(),
             options,
-            ep_inputs: Vec::with_capacity(3),
+            ep_inputs: Vec::with_capacity(module.entry_points.len()),
```
kvark, line 52: let's not pass the `module` just for this. Instead, just use `Vec::new` and expect this to be re-used

src/back/hlsl/writer.rs
Gordon-F: Sorry, I missed the fact, that it can that simple. Thank you!

kvark (commented):
Great! One small nit :)

kvark (approved):

PR #1077, opened 2021-07-06
[spv-out] Move cached expression table to BlockContext.

jimblandy:
Since the table of cached expressions is only meaningful within a single
function, it's really something that should only be accessed from
`BlockContext`.

However, to save heap allocations, it makes sense to retain it in the `Writer`
between functions. But the `Writer` field should have a different name, to
ensure people don't use it by accident.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -82,7 +82,7 @@ impl Writer {
             constant_ids: Vec::new(),
             cached_constants: crate::FastHashMap::default(),
             global_variables: Vec::new(),
-            cached: CachedExpressions::default(),
+            saved_cached: CachedExpressions::default(),
             gl450_ext_inst_id,
             temp_list: Vec::new(),
         })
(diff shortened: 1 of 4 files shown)


src/back/spv/writer.rs
```diff
@@ -131,7 +131,7 @@ impl Writer {
             constant_ids: take(&mut self.constant_ids).recycle(),
             cached_constants: take(&mut self.cached_constants).recycle(),
             global_variables: take(&mut self.global_variables).recycle(),
-            cached: take(&mut self.cached).recycle(),
+            saved_cached: take(&mut self.saved_cached).recycle(),
```
kvark, line 134: why doesn't `recycle()` do the `take` internally?

kvark (approved):

src/back/spv/writer.rs
jimblandy: The documentation for the `Recycle` trait explains the idea here: if the method takes a `&mut`, then that makes it easy for implementations to just update fields of interest, instead of constructing a fresh value so that we'll get compile-time errors if a field is forgotten. Passing by value encourages implementations to do the right thing.

Of course, none of this guarantees anything. An implementation could perfectly well just assign to a few fields of `self` and then return `self`. It's just a prompt. I couldn't figure out any way to enforce the rule that every field should be listed.

If you think it'd be okay for `Recyclable::recycle` to just take `&mut self` and remind people to do the `take` in the documentation, then we can do that.

kvark: That's good reasoning, let's keep it

PR #1083, opened 2021-07-13
Improve WGSL interpolation attributes

pyrotechnick:
This PR improves Naga's WGSL output by omitting parts of interpolation attributes which match default values.

 It also enables the parsing of [`center` interpolation sampling](https://gpuweb.github.io/gpuweb/wgsl/#interpolation).

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -324,30 +324,25 @@ impl<W: Write> Writer<W> {
                 Attribute::Binding(id) => format!("binding({})", id),
                 Attribute::Group(id) => format!("group({})", id),
                 Attribute::Interpolate(interpolation, sampling) => {
-                    let sampling_part = match sampling {
-                        Some(sampling) if sampling != crate::Sampling::default() => {
-                            Some(sampling_str(sampling))
-                        }
-                        _ => None,
-                    };
-                    let parts: Vec<_> = vec![
-                        if sampling_part.is_some()
-                            || (interpolation.is_some()
-                                && interpolation != Some(crate::Interpolation::default()))
-                        {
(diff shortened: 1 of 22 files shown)


src/back/wgsl/writer.rs
```diff
@@ -324,25 +324,30 @@ impl<W: Write> Writer<W> {
...
-                            } else {
-                                format!(",{}", sampling_str(sampling))
-                            }
+                    let sampling_part = match sampling {
+                        Some(sampling) if sampling != crate::Sampling::default() => {
+                            Some(sampling_str(sampling))
+                        }
+                        _ => None,
+                    };
+                    let parts: Vec<_> = vec![
```
kvark, line 333: can we refactor this to avoid heap allocation? I understand that this particular case may not be too problematic, but generally across Naga codebase we try to avoid heap allocation in frontends and backends, unless necessary

kvark (commented):
Thank you, looks good!
Just one note

pyrotechnick:
236341e removes the use of `Vec`; and by extension heap allocation.

I have also removed the `Default` trait on both `Interpolation` and `Sampling`.

Both `Interpolation` and `Sampling` are not WGSL backend specific structs and the defaults specified by WGSL are not necessarily the defaults of other targets/backends.

kvark:
> I have also removed the Default trait on both Interpolation and Sampling.
> Both Interpolation and Sampling are not WGSL backend specific structs and the defaults specified by WGSL are not necessarily the defaults of other targets/backends.

Pretty sure these are defaults in all targetss/backends we work with :) But it's ok, we can do it separately.

kvark (approved):

PR #1086, opened 2021-07-13
[hlsl-out] Texture functions implementation

Gordon-F:


src/back/hlsl/writer.rs
```diff
@@ -1023,18 +1032,74 @@ impl<'a, W: Write> Writer<'a, W> {
...
+
+                if let Some(depth_ref) = depth_ref {
+                    write!(self.out, ", ")?;
+                    self.write_expr(module, depth_ref, func_ctx)?;
+                }
+
+                match level {
+                    Sl::Auto => {}
+                    Sl::Zero => {
+                        // Level 0 is implied for depth comparison
```
kvark, line 1193: since `SampleCmpLevelZero` is used, why do we need to add `0.0` even for non-depth?

src/back/hlsl/writer.rs
```diff
@@ -1,5 +1,27 @@
...
+// See `WrappedImageQuery` struct that represents a unique function and will be generated before writing all statements and expressions.
+// This allowed to works with `Expression::ImageQuery` as expression and write wrapped function.
+//
+// For example:
+// ```wgsl
+// let dim_1d = textureDimensions(image_1d);
+// ```
+//
+// ```hlsl
+// int NagaGetDimensions1DDimensions(Texture1D<float4>)
```
kvark, line 14: maybe we can make a nicer name here :)

src/back/hlsl/writer.rs
```diff
@@ -30,16 +50,84 @@ struct EpStructMember {
...
+    NumLevels,
+    NumLayers,
+    NumSamples,
+}
+
+impl From<crate::ImageQuery> for ImageQuery {
+    fn from(q: crate::ImageQuery) -> Self {
+        use crate::ImageQuery as Iq;
+        match q {
+            Iq::Size { level } => {
```
kvark, line 105: let's use patterns for this, i.e. `Iq::Size { level:Some(_) } => ...`

src/back/hlsl/writer.rs
```diff
@@ -30,16 +50,84 @@ struct EpStructMember {
...
+    }
+}
+
 pub struct Writer<'a, W> {
     out: W,
     names: crate::FastHashMap<NameKey, String>,
     namer: proc::Namer,
     /// HLSL backend options
+    #[allow(dead_code)]
     options: &'a Options,
```
kvark, line 125: nit: in cases like this I find renaming this to `_options` to be better. Otherwise, you may start using the options later and forget that there is a decoration put on it.

src/back/hlsl/writer.rs
```diff
@@ -154,6 +244,9 @@ impl<'a, W: Write> Writer<'a, W> {
             };
             let name = self.names[&NameKey::Function(handle)].clone();
 
+            // Write wrapped struction for `Expression::ImageQuery` before writing all statements and expressions
```
kvark, line 247: "struction"?

src/back/hlsl/writer.rs
```diff
@@ -182,6 +275,9 @@ impl<'a, W: Write> Writer<'a, W> {
 
             let name = self.names[&NameKey::EntryPoint(index as u16)].clone();
 
+            // Write wrapped struction for `Expression::ImageQuery` before writing all statements and expressions
```
kvark, line 278: same here

src/back/hlsl/writer.rs
```diff
@@ -1441,6 +1566,180 @@ impl<'a, W: Write> Writer<'a, W> {
...
+            self.out,
+            "NagaGetDimensions{}{}{}{}",
+            image_dimension_str(query.dim),
+            match query.class {
+                crate::ImageClass::Sampled { multi: true, .. } => "MS",
+                crate::ImageClass::Depth => "Depth",
+                _ => "",
+            },
+            if query.arrayed { "Array" } else { "" },
+            match query.query {
```
kvark, line 1812: Let's not put any logic inside the `write!` macro invocation. For any tools the macros are harder to work with, so we should take advantage of plain code where possible.

src/back/hlsl/writer.rs
```diff
@@ -1441,6 +1566,180 @@ impl<'a, W: Write> Writer<'a, W> {
...
+    /// Helper function that write wrapped function for `Expression::ImageQuery`
+    /// https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-getdimensions
+    fn write_wrapped_image_query_functions(
+        &mut self,
+        module: &Module,
+        func_body: &[crate::Statement],
+        func_ctx: &back::FunctionCtx<'_>,
+    ) -> BackendResult {
+        for sta in func_body.iter() {
+            if let crate::Statement::Emit(ref range) = *sta {
```
kvark, line 1669: instead of iterating the emitted ranges, let's just iterate the expressions directly

src/back/hlsl/writer.rs
```diff
@@ -1441,6 +1566,180 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                for handle in range.clone() {
+                    if let crate::Expression::ImageQuery { image, query } =
+                        func_ctx.expressions[handle]
+                    {
+                        let image_ty = func_ctx.info[image].ty.inner_with(&module.types);
+                        if let crate::TypeInner::Image {
```
kvark, line 1675: in order to reduce indendation, let's bail out with `continue` if this isn't hit

src/back/hlsl/writer.rs
```diff
@@ -1441,6 +1566,180 @@ impl<'a, W: Write> Writer<'a, W> {
...
+    fn write_wrapped_image_query_functions(
+        &mut self,
+        module: &Module,
+        func_body: &[crate::Statement],
+        func_ctx: &back::FunctionCtx<'_>,
+    ) -> BackendResult {
+        for sta in func_body.iter() {
+            if let crate::Statement::Emit(ref range) = *sta {
+                for handle in range.clone() {
+                    if let crate::Expression::ImageQuery { image, query } =
```
kvark, line 1671: same here, let's bail out instead of adding indentation, it's already quite deep

src/back/hlsl/writer.rs
```diff
@@ -1441,6 +1566,180 @@ impl<'a, W: Write> Writer<'a, W> {
 
         Ok(())
     }
+
+    /// Helper function that write wrapped function for `Expression::ImageQuery`
+    /// https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-getdimensions
+    fn write_wrapped_image_query_functions(
```
kvark, line 1662: We should start thinking about how to split the HLSL backend into sub-modules. This isn't needed for this PR, but technically we could move all the image-related processing into a module in the future.

src/back/hlsl/writer.rs
```diff
@@ -1441,6 +1566,180 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                                writeln!(self.out, "{{")?;
+                                const RETURN_VARIABLE_NAME: &str = "ret";
+
+                                use crate::ImageDimension as IDim;
+                                use crate::ImageQuery as Iq;
+                                // GetDimensions Overloaded Methods
+                                // https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-getdimensions#overloaded-methods
+                                let (ret_swizzle, number_of_params) = match query {
+                                    Iq::Size { .. } => match dim {
+                                        IDim::D1 => ("x", if arrayed { 2 } else { 1 }),
```
kvark, line 1717: nit: could just add `+1` if arrayed after the match

src/back/hlsl/writer.rs
```diff
@@ -1441,6 +1566,180 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                                    RETURN_VARIABLE_NAME
+                                )?;
+                                write!(self.out, "{}", back::INDENT)?;
+                                self.write_expr(module, image, func_ctx)?;
+                                write!(self.out, ".GetDimensions(")?;
+                                match query {
+                                    Iq::Size { level: Some(_) } => {
+                                        write!(self.out, "{}, ", MIP_LEVEL_PARAM)?;
+                                    }
+                                    _ => match (dim, class) {
```
kvark, line 1745: this is a weird match! Effectively we are only checking each of those once. Is the idea to extend this later?
If we aren't going to touch this, let's opt into separate if lets or matches

src/back/hlsl/writer.rs
```diff
@@ -1441,6 +1566,180 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                                        (_, crate::ImageClass::Sampled { multi: true, .. }) => {}
+                                        (IDim::D2 | IDim::D3 | IDim::Cube, _) => {
+                                            write!(self.out, "0, ")?;
+                                        }
+                                        _ => {}
+                                    },
+                                }
+                                for i in 0..number_of_params {
+                                    if i == (number_of_params - 1) {
```
kvark, line 1754: can we instead have the loop iterating like `for component in back::COMPONENTS[..number_of_params-1].iter()`?
Then we can follow with this code that handles the last value

src/back/hlsl/writer.rs
```diff
@@ -1538,3 +1829,23 @@ fn sampling_str(sampling: crate::Sampling) -> Option<&'static str> {
...
+        | Sf::R32Float
+        | Sf::Rg32Float
+        | Sf::Rgba32Float => Some("float4"),
+        Sf::Rgba8Uint | Sf::Rgba16Uint | Sf::R32Uint | Sf::Rg32Uint | Sf::Rgba32Uint => {
+            Some("uint4")
+        }
+        Sf::Rgba8Sint | Sf::Rgba16Sint | Sf::R32Sint | Sf::Rg32Sint | Sf::Rgba32Sint => {
+            Some("int4")
+        }
+        _ => None,
```
kvark, line 1939: is this unfinished on purpose? are there any blockers or concerns here?

src/back/hlsl/writer.rs
```diff
@@ -1227,6 +1244,79 @@ impl<'a, W: Write> Writer<'a, W> {
...
             }
+            Expression::ImageLoad {
+                image,
+                coordinate,
+                array_index: _, // TODO:
+                index: _,       // TODO:
+            } => {
+                // https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-load
+                let image_ty = func_ctx.info[image].ty.inner_with(&module.types);
+                if let crate::TypeInner::Image {
```
kvark, line 1255: similarly to other places, we should just bail out (with `unreachable` or `continue`) if the match fails instead of indenting this much code here

src/back/hlsl/writer.rs
```diff
@@ -1227,6 +1244,79 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                    let ms = if let crate::ImageClass::Sampled { multi: true, .. } = class {
+                        true
+                    } else {
+                        false
+                    };
+
+                    // Input location type based on texture type
+                    let (coordinate_param_ty, numb_of_components) = match dim {
+                        IDim::D1 => {
+                            if arrayed {
```
kvark, line 1275: similarly, it looks like we can just add `+1` for `arrayed` after the fact

src/back/hlsl/writer.rs
```diff
@@ -1849,3 +1939,11 @@ fn storage_format_to_texture_type(format: crate::StorageFormat) -> Option<&'stat
         _ => None,
     }
 }
+
+fn number_of_components(vector_size: crate::VectorSize) -> usize {
```
kvark, line 1943: no need for this one, can just do `vector_size as usize`

src/back/hlsl/writer.rs
Gordon-F: Waiting for `Tint` to finish this :laughing:

https://dawn.googlesource.com/tint/+/refs/heads/main/src/writer/hlsl/generator_impl.cc#69

src/back/hlsl/writer.rs
Gordon-F: Now It's better a little bit 😄

src/back/hlsl/writer.rs
Gordon-F: I guess no, because we have `number_of_params = 1` here for 1D texture.

src/back/hlsl/writer.rs
Gordon-F: Now we returns error here.

src/back/hlsl/writer.rs
```diff
@@ -1456,3 +1675,38 @@ fn sampling_str(sampling: crate::Sampling) -> Option<&'static str> {
...
+        Sf::Rg16Float => "float2",
+        Sf::Rg8Unorm => "unorm float2",
+        Sf::Rg8Snorm => "snorm float2",
+
+        Sf::Rg8Sint | Sf::Rg16Sint => "int2",
+        Sf::Rg8Uint | Sf::Rg16Uint => "uint2",
+
+        Sf::Rg11b10Float => "float3",
+
+        Sf::Rgba8Unorm
```
kvark, line 1698: should this be "unorm float4" instead?

src/back/hlsl/writer.rs
```diff
@@ -1456,3 +1675,38 @@ fn sampling_str(sampling: crate::Sampling) -> Option<&'static str> {
...
+        | Sf::Rgba16Float
+        | Sf::R32Float
+        | Sf::Rg32Float
+        | Sf::Rgba32Float => "float4",
+        Sf::Rgba8Uint
+        | Sf::Rgba16Uint
+        | Sf::R32Uint
+        | Sf::Rg32Uint
+        | Sf::Rgba32Uint
+        | Sf::Rgb10a2Unorm => "uint4",
```
kvark, line 1709: should this one be "unorm float4" instead?

src/back/hlsl/writer.rs
```diff
@@ -1456,3 +1675,38 @@ fn sampling_str(sampling: crate::Sampling) -> Option<&'static str> {
...
+        Sf::Rg8Unorm => "unorm float2",
+        Sf::Rg8Snorm => "snorm float2",
+
+        Sf::Rg8Sint | Sf::Rg16Sint => "int2",
+        Sf::Rg8Uint | Sf::Rg16Uint => "uint2",
+
+        Sf::Rg11b10Float => "float3",
+
+        Sf::Rgba8Unorm
+        | Sf::Rgba8Snorm
```
kvark, line 1699: perhaps "snorm float4" instead?

src/back/mod.rs
```diff
@@ -24,7 +24,7 @@ const BAKE_PREFIX: &str = "_e";
 ///
 /// Also stores data needed to identify it (handle for a regular function or index for an entry point)
 #[allow(dead_code)]
-enum FunctionType {
+pub(crate) enum FunctionType {
```
kvark, line 27: why is this needed? Private types are fully visible to the sub-modules

src/back/mod.rs
```diff
@@ -33,7 +33,7 @@ enum FunctionType {
 
 /// Helper structure that stores data needed when writing the function
 #[allow(dead_code)]
-struct FunctionCtx<'a> {
+pub(crate) struct FunctionCtx<'a> {
```
kvark, line 36: similarly, this shouldn't need `pub(crate)`

kvark (commented):
Last few fixes are needed. Great work overall!

kvark (approved):

PR #1087, opened 2021-07-13
Document `ImageLoad`, `ImageStore`, and a few other random things.

jimblandy:
I've inferred most of this documentation from reading the code, so read it with skepticism. Let me know what I need to fix.


kvark (approved):
Very nice!

PR #1091, opened 2021-07-14
[wgsl] Implicitly readable global storage buffers

pyrotechnick:
This PR improves Naga's WGSL support by always allowing reads of global storage buffers; that is, globals of type `Array`/`Struct` that are declared in global scope via `var<storage>`.

These changes encompass the first set of changes that must be made for Naga's behaviour to reflect that of the WGSL specification as of 12 July 2021.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -3269,13 +3269,22 @@ impl Parser {
                         _ => crate::StorageClass::Private,
                     },
                 };
+                let storage_access = match class {
+                    crate::StorageClass::Storage => match module.types[pvar.ty].inner {
+                        crate::TypeInner::Struct { .. } | crate::TypeInner::Array { .. } => {
+                            pvar.access | crate::StorageAccess::LOAD
+                        }
+                        _ => pvar.access,
+                    },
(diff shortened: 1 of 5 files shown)


tests/in/access.wgsl
```diff
@@ -9,15 +9,19 @@ struct Bar {
 [[group(0), binding(0)]]
 var<storage> bar: [[access(read_write)]] Bar;
 
+[[group(0), binding(1)]]
+var<storage> qux: Bar;
```
kvark, line 13: any reason to have both the new and the old syntax (for `bar`) in the same test? If we are updating the syntax, let's not leave any trace behind

src/front/wgsl/mod.rs
```diff
@@ -3269,13 +3269,22 @@ impl Parser {
                         _ => crate::StorageClass::Private,
                     },
                 };
+                let storage_access = match class {
```
kvark, line 3272: this doesn't really parse any new access qualifier syntax yet

kvark (changes requested):
Thank you!
I think we should get it into a shape where the new syntax is supported, and not merge half-way.

PR #1092, opened 2021-07-14
Proof of concept for "span" feature, with WGSL & GLSL parsing augmented.

ElectronicRU:
Proof of concept for #358 .

Currently not used in any kind of back error reporting, just wanted to see what it takes to generate span information in a parser. Turns out it's pretty okay!

### DIFF
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -269,27 +269,6 @@ impl<T> Arena<T> {
         return self.span_info.get(handle.index()).unwrap_or(&Span::Unknown);
         &Span::Unknown
     }
-
-    #[cfg_attr(not(feature = "span"), allow(unused_variables))]
-    #[deprecated]
-    pub fn set_span(&mut self, handle: Handle<T>, span: Span) {
-        #[cfg(feature = "span")]
-        {
-            self.span_info[handle.index()].clone_from(&span);
-        }
-    }
-
-    #[cfg_attr(not(feature = "span"), allow(unused_variables))]
-    #[deprecated]
-    pub fn set_span_if_unknown(&mut self, handle: Handle<T>, span: Span) {
-        #[cfg(feature = "span")]
-        {
-            let ref mut existing = self.span_info[handle.index()];
-            if let Span::Unknown = existing {
-                existing.clone_from(&span);
-            }
-        }
-    }
 }
 
 #[cfg(feature = "deserialize")]
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -218,7 +218,12 @@ impl<T> Arena<T> {
     /// returns a handle pointing to
     /// an existing element if the check succeeds, or adds a new
     /// element otherwise.
-    pub fn fetch_if_or_append<F: Fn(&T, &T) -> bool>(&mut self, value: T, span: Span, fun: F) -> Handle<T> {
+    pub fn fetch_if_or_append<F: Fn(&T, &T) -> bool>(
+        &mut self,
+        value: T,
+        span: Span,
+        fun: F,
+    ) -> Handle<T> {
         if let Some(index) = self.data.iter().position(|d| fun(d, &value)) {
             let index = unsafe { Index::new_unchecked((index + 1) as u32) };
             Handle::new(index)
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -195,8 +195,9 @@ impl<T> Arena<T> {
     }
 
     /// Adds a new value to the arena, returning a typed handle.
-    #[cfg_attr(not(feature = "span"), allow(unused_variables))]
     pub fn append(&mut self, value: T, span: Span) -> Handle<T> {
+        #[cfg(not(feature = "span"))]
+        let _ = span;
         let position = self.data.len() + 1;
         let index =
             Index::new(position as u32).expect("Failed to append to Arena. Handle overflows");
@@ -262,12 +263,16 @@ impl<T> Arena<T> {
         self.data.clear()
     }
 
-    #[cfg_attr(feature = "span", allow(unreachable_code))]
-    #[cfg_attr(not(feature = "span"), allow(unused_variables))]
     pub fn get_span(&self, handle: Handle<T>) -> &Span {
         #[cfg(feature = "span")]
-        return self.span_info.get(handle.index()).unwrap_or(&Span::Unknown);
-        &Span::Unknown
+        {
+            return self.span_info.get(handle.index()).unwrap_or(&Span::Unknown);
+        }
+        #[cfg(not(feature = "span"))]
+        {
+            let _ = handle;
+            &Span::Unknown
+        }
     }
 }
 
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -277,7 +277,10 @@ impl<T> Arena<T> {
 }
 
 #[cfg(feature = "deserialize")]
-impl<'de, T> serde::Deserialize<'de> for Arena<T> {
+impl<'de, T> serde::Deserialize<'de> for Arena<T>
+where
+    T: serde::Deserialize<'de>,
+{
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: serde::Deserializer<'de>,
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -262,16 +262,15 @@ impl<T> Arena<T> {
         self.data.clear()
     }
 
-    #[allow(unused_variables)]
+    #[cfg_attr(feature = "span", allow(unreachable_code))]
+    #[cfg_attr(not(feature = "span"), allow(unused_variables))]
     pub fn get_span(&self, handle: Handle<T>) -> &Span {
         #[cfg(feature = "span")]
-        {
-            return self.span_info.get(handle.index()).unwrap_or(&Span::Unknown);
-        }
+        return self.span_info.get(handle.index()).unwrap_or(&Span::Unknown);
         &Span::Unknown
     }
 
-    #[allow(unused_variables)]
+    #[cfg_attr(not(feature = "span"), allow(unused_variables))]
     #[deprecated]
     pub fn set_span(&mut self, handle: Handle<T>, span: Span) {
         #[cfg(feature = "span")]
@@ -280,7 +279,7 @@ impl<T> Arena<T> {
         }
     }
 
-    #[allow(unused_variables)]
+    #[cfg_attr(not(feature = "span"), allow(unused_variables))]
     #[deprecated]
     pub fn set_span_if_unknown(&mut self, handle: Handle<T>, span: Span) {
         #[cfg(feature = "span")]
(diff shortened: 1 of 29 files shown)


ElectronicRU:
@kvark wyt?

src/arena.rs
```diff
@@ -240,6 +247,21 @@ impl<T> Arena<T> {
             marker: PhantomData,
         }
     }
+
+    #[allow(unused_variables)]
+    pub fn get_span(&self, handle: Handle<T>) -> Span {
+        #[cfg(feature = "span")] {
+            return self.span_info.get(handle.index()).map_or(Span::Unknown, |x| x.clone())
```
kvark, line 254: Generally arena's access out of bounds panics (unless you use the "try_" method version). Any reason to not just panic here?

src/front/wgsl/mod.rs
```diff
@@ -1585,7 +1596,9 @@ impl Parser {
             })
         };
 
-        self.scopes.pop();
+        let span = self.pop_scope(lexer);
+        const_arena.add_span(handle, ArenaSpan::ByteRange(span));
```
kvark, line 1600: how would this work in a case where we called `fetch_or_append` and took an existing handle?

kvark (commented):
Very nice work!

src/front/wgsl/mod.rs
ElectronicRU: Not very well, unfortunately, it does indeed stretch the span across :D Thanks, that's a nice catch - I suppose this is a usecase for `set_span_if_unknown` method.

ElectronicRU:
Handled fetch_or_append cases for constants and types better.
Unfortunate side effect is that because of uniqueness/deduplication, only the span of one (currently first) occurrence is saved for types. It's probably fine if backend uses type spans only to emit errors for unsupported type declarations.

src/span.rs
```diff
@@ -0,0 +1,24 @@
+use std::ops::Range;
+
+// A source code span, used for error reporting.
+#[derive(Clone)]
+pub enum Span {
```
kvark, line 5: do we need an entirely new module for this? maybe it can live in `lib.rs`

src/span.rs
```diff
@@ -0,0 +1,24 @@
...
+    // Span is unknown - no source information.
+    Unknown,
+    // Byte range.
+    ByteRange(Range<usize>),
+}
+
+impl Span {
```
kvark, line 12: should this stuff be behind the `span` feature?

ElectronicRU: This would creep the feature flags outside of lib/arena and into all of the lexers. Dead code elimination should take care of it regardless, since without span feature, setting a span does nothing.

src/span.rs
ElectronicRU: Originally I planned to put span derivation impls in here, but so far they've been unnecessary. I propose we merge this into lib.rs if by the end of prototype this turns out to be sufficiently smol.

ElectronicRU:
Hey - as a heads up I'm gonna continue work on this, I've been in quarantine but Imma out and fine :)

kvark:
@ElectronicRU looking forward to see this landing! Naga's error messages are number 1 problem with wgpu today, I'd say.

ElectronicRU:
@kvark I went with "split spans into a separate Vec, mask stuff with Derefs" approach for Blocks. It's not super pretty, but it should get the job done.

src/arena.rs
```diff
@@ -240,6 +250,52 @@ impl<T> Arena<T> {
...
+            return self.span_info.get(handle.index()).unwrap_or(&Span::Unknown);
+        }
+        &Span::Unknown
+    }
+
+    #[allow(unused_variables)]
+    pub fn set_span(&mut self, handle: Handle<T>, span: Span) {
+        #[cfg(feature = "span")]
+        {
+            self.span_info[handle.index()].clone_from(&span);
```
kvark, line 267: nit: if this requires a reference anyway, let's have the argument as a reference

src/arena.rs
```diff
@@ -240,6 +250,52 @@ impl<T> Arena<T> {
...
+        }
+    }
+
+    #[allow(unused_variables)]
+    pub fn set_span_if_unknown(&mut self, handle: Handle<T>, span: Span) {
+        #[cfg(feature = "span")]
+        {
+            let ref mut existing = self.span_info[handle.index()];
+            if let Span::Unknown = existing {
+                existing.clone_from(&span);
```
kvark, line 277: same here

src/lib.rs
```diff
@@ -1147,8 +1150,128 @@ pub enum Expression {
...
+        #[cfg(feature = "span")]
+        let span_info = std::iter::repeat(Span::Unknown).take(body.len()).collect();
+        Self {
+            body,
+            #[cfg(feature = "span")]
+            span_info,
+        }
+    }
+
+    pub fn body(&self) -> &Vec<Statement> {
```
kvark, line 1186: nit: any reason this doesn't return `&[Statement]`?

src/lib.rs
```diff
@@ -1147,8 +1150,128 @@ pub enum Expression {
...
+        &self.body
+    }
+
+    #[allow(unused_variables)]
+    pub fn push(&mut self, end: Statement, span: Span) {
+        self.body.push(end);
+        #[cfg(feature = "span")]
+        self.span_info.push(span);
```
kvark, line 1194: one of the risks here, I suppose, is that the number of elements in `body` diverges from the number of elements in `span_info`. Even though the fields are private, they are visible to all the submodules.
I can think of 2 approaches:
  - the lazy one - just do the check in validation
  - the strict one - move the block definition and implementation into a `block` submodule, and re-export it. This would mean the fields are really concealed, and one can at least look at the logic to ensure that both vectors are modified at the same time.

src/lib.rs
```diff
@@ -1147,8 +1150,128 @@ pub enum Expression {
...
+        self.span_info
+            .splice(range.clone(), other.span_info.into_iter());
+        self.body.splice(range, other.body.into_iter());
+    }
+
+    pub fn iter(&self) -> std::slice::Iter<Statement> {
+        self.body.iter()
+    }
+
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = (&mut Statement, Option<&mut Span>)> {
```
kvark, line 1226: this is a bit weird. There is no symmetry between `iter()` and `iter_mut()` in terms of returned type?

kvark (commented):
I think I see the trade-offs here. And it's along the lines we discussed, so this is acceptable.
In terms of span collection, in addition to the compile feature, we'll probably need each backend to have a flag for enabling/disabling this. We can follow up, no need to do it here.

Do you consider this ready for merging overall? Looks like it's getting close :)

src/lib.rs
ElectronicRU: Renamed iter_mut to a different beast, and removed iter altogether since it's provided by Deref.

src/lib.rs
ElectronicRU: Moved to a separate module, there's a decent amount of code now anyway.

src/lib.rs
ElectronicRU: Oh that's just bad anyway.

ElectronicRU:
@kvark 
re: some nits and general design,
as I'm doing more arena stuff (especially for types), I tend to see `fetch_or_append` together with `set_span_if_unknown`, and append together with `set_span`. My inner designer tells me to just add Span arguments to `*append`.

`append` functions of `Arena` are used either in the frontend (that's where spans live), or, sparingly, in some tests. It won't be a huge deal for the code-base to make this change (and you can totally plug `Default::default()` instead of spans). However, this is public API, therefore I'm kind of cagey. Wyt?

Re: ready status, I'd consider this ready when we actually do something with spans :) IE validator or backend errors actually incorporate spans, and CLI is able to handle them. Ideally SPIR-V also gets OpLine handling. But it's getting there for sure!

kvark:
It would certainly be great if the common way to add elements to arenas was exposed with spans. Maybe `append` should always have spans, and something like `append_nospan` to have the current semantics.

ElectronicRU:
Okay, I've finished refactoring Arena interface as well.

Todo list:

- [ ] Add a Span that can reference files (probably needs an Rc or Arc of str, sadly; the only saving grace is there are usually not that many files, but that means nontrivial Drop...)
- [ ] Make use of it in the SPIR-V backend; most of the groundwork is there, we basically only generate spans for single opcodes there
- [ ] Actually emit some kind of useful data based on spans in validators/backends.

IMO this PR has already swollen well beyond reasonable :( Shotgun API changes did it in. I propose we declare the above beyond the scope of current PR and merge it before I have to rebase this another time.

Alternative is a complete do-over that first introduces all the necessary API changes, and then wires the logic up. Obviously the history wasn't as pretty. If that would be desirable, I might well spend some hours doing that, but I'd really rather not.

src/arena.rs
ElectronicRU: Nuked from orbit.

src/arena.rs
ElectronicRU: Ditto.

src/span.rs
ElectronicRU: Looks like we're at end. IMO a separate module is fine, since Span tries to be a semi-opaque type (even in frontend, at a certain level you just get, clone and merge spans). Wyt?

src/arena.rs
```diff
@@ -195,13 +195,14 @@ impl<T> Arena<T> {
     }
 
     /// Adds a new value to the arena, returning a typed handle.
-    pub fn append(&mut self, value: T) -> Handle<T> {
+    #[cfg_attr(not(feature = "span"), allow(unused_variables))]
```
kvark, line 198: let's handle this inside the function, not outside, i.e.
```rust
#[cfg(not(feature = "span"))]
let _ = span;
```

src/arena.rs
```diff
@@ -261,6 +267,7 @@ impl<T> Arena<T> {
     }
 
     #[allow(unused_variables)]
+    #[deprecated]
```
kvark: can we just remove these instead of marking deprecated?

ElectronicRU: I did ;)

src/arena.rs
ElectronicRU: Donezo.

kvark:
We'd be happy to merge what you got, let's just rebase please.

ElectronicRU:
@kvark please merge - I won't survive another cross-cutting refactor D:

kvark:
Clippy looks unhappy.

kvark:
@ElectronicRU Clippy and the last rebase are needed. Sorry about breaking your stuff (again)!
I'm really excited to see this landing. 🚀

ElectronicRU:
Should be done!

kvark (approved):

ElectronicRU:
@kvark I am thinking squashed. While the work was piecewise, I temporarily broke some tests and features for some of the commits.

ElectronicRU:
Oh and I accidentally commited in wrong test results - should be good now.

ElectronicRU:
🎉

PR #1097, opened 2021-07-16
Support `builtin(primitive_index)`

initial-algebra:
Quick and dirty validation and code generation for the SPIR-V builtin `PrimitiveID`.  Currently only attempts to support the SPIR-V backend.


src/back/hlsl/writer.rs
```diff
@@ -1388,6 +1388,7 @@ fn builtin_str(built_in: crate::BuiltIn) -> &'static str {
         // fragment
         Bi::FragDepth => "SV_Depth",
         Bi::FrontFacing => "SV_IsFrontFace",
+        // Bi::PrimitiveId => "SV_PrimitiveID",
```
kvark, line 1391: why is this commented out?

src/lib.rs
```diff
@@ -269,6 +269,7 @@ pub enum BuiltIn {
     // fragment
     FragDepth,
     FrontFacing,
+    PrimitiveId,
```
kvark, line 272: the current naming convention for builtins is such that "XxxID" are vectors, and all the scalars are "XxxIndex".
So let's call this `PrimitiveIndex`.

src/valid/mod.rs
```diff
@@ -58,6 +58,7 @@ bitflags::bitflags! {
         const PUSH_CONSTANT = 0x1;
         /// Float values with width = 8.
         const FLOAT64 = 0x2;
+        const PRIMITIVE_ID = 0x4;
```
kvark, line 61: similarly, let's call it `PRIMITIVE_INDEX`

kvark (commented):
Great start!

src/back/hlsl/writer.rs
initial-algebra: I could certainly uncomment it, but I haven't tested the HLSL path at all.  I will do that.  Speaking of backends, I don't have an Apple device or any knowledge of Metal, so I won't be able to work on that platform.  I can certainly throw together a Linux VM to work on the OpenGL backend at some point.

src/valid/mod.rs
initial-algebra: Should we maybe call it `GEOMETRY` to reflect the generic shader capability?

kvark: No, we aren't going to support geometry shaders any time soon. From the point of view of Naga, this should be a capability unrelated to geometry shaders. It's just the Naga client's responsibility to check on Vulkan that geometry is allowed.

src/back/hlsl/writer.rs
kvark: HLSL is emerging now, and we have all the things checked on CI which we care not to break. So don't afraid to be bold on HLSL and MSL. If CI is happy, we are likely happy.
Speaking of which, we'll need to add the use of primitive ID to one of the test. I think "extra.wgsl" should be this, given that it's extra functionality.

initial-algebra:
Very untested, since I'm currently performing SPIR-V passthrough after updating wgpu to `master`.

kvark (approved):
Shipit!

kvark:
Could you try using it in `extra.wgsl` test and seeing what comes up?

initial-algebra:
Seems to work!

kvark:
@initial-algebra the tests are failing because you didn't add some of the snapshot outputs, it seems

kvark (approved):

initial-algebra:
I have no idea where to begin with validating the MSL version.

kvark:
Interesting approach! I'll follow up with a way to communicate the language version to the validator, directly.

initial-algebra:
Thank you @kvark for being so patient!

PR #1100, opened 2021-07-17
glsl-out: fix emission of location/binding attributes

kvark:
Should help https://github.com/gfx-rs/wgpu/issues/1617
Here is the breakdown:
  - "location" is supported unconditionally for VS inputs and FS outputs
  - "location" requires GLSL ES 310 for varyings (VS outputs and FS inputs)
  - "binding" requires GLSL ES 310

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -101,9 +101,14 @@ impl Version {
         }
     }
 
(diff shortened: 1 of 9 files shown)


kvark:
Reviewed by @magcius

PR #1106, opened 2021-07-18
[hlsl-out] Validate all entry points in Makefile

Gordon-F:
All entry points validations passed 🚀.

Still uses config files, but with a more readable `Makefile`. My attempt to implement validation with comments inside hlsl shaders makes `Makefile` unreadable at all and full of dark magic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,17 +65,27 @@ validate-wgsl: $(SNAPSHOTS_BASE_OUT)/wgsl/*.wgsl
 		cargo run $${file}; \
 	done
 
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -65,17 +65,27 @@ validate-wgsl: $(SNAPSHOTS_BASE_OUT)/wgsl/*.wgsl
 		cargo run $${file}; \
 	done
 
+validate-hlsl: SHELL:=/bin/bash # required because config files uses arrays
```
kvark, line 68: are we defining `validate-hlsl` twice now?

kvark (approved):

PR #1108, opened 2021-07-19
Update GLSL snapshots for quad, after e76824a.

jimblandy:
Unless I'm doing something wrong, we should find out why CI didn't warn us about these being out of date.


jimblandy:
Error between chair and keyboard. I wasn't using `--all-features`.

PR #1109, opened 2021-07-19
Add tests for sampling mipmapped and multisampled images.

jimblandy:
Prior to this change, nothing in the test suite exercised the SPIR-V backend
code for generating LOD or Sample arguments to OpImageFetch instructions.


kvark (approved):
Nice!

PR #1111, opened 2021-07-19
Add Explicit Int Precision for GLSL ES Backend

zicklag:
I ran into an [error](https://matrix.to/#/!FZyQrssSlHEZqrYcOb:matrix.org/$XhLgbZcu0uCYJjUm07uVTNUjwCxjOXv8etwaOt7_y8E?via=matrix.org&via=mozilla.org&via=kyju.org) while targeting WebGL where the shaders failed to link because of differing precision between vertex and fragment shaders in a uniform with an integer field. This explicitly sets the integer precision to `highp` so that they match across vertex and fragment shaders.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -409,11 +409,12 @@ impl<'a, W: Write> Writer<'a, W> {
             writeln!(self.out, "#extension GL_EXT_texture_shadow_lod : require")?;
         }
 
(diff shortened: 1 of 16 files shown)


kvark (approved):

kvark:
If you run `make` locally, it should modify the snapshot tests. These outputs have to be included in this PR.

zicklag:
Updated with snapshot tests.

PR #1116, opened 2021-07-20
Improve CLI Argument Doc and Error Handling

### LINKED ISSUE #1107: Improved CLI Argument Parsing
Would you be open to a PR for improving the current Naga CLI argument parsing?

Nowadays I tend to prefer [structopt](https://docs.rs/structopt/0.3.22/structopt/), but if you were concerned about code size or anything I've round [argh](https://docs.rs/argh/0.1.5/argh/) a good minimal alternative.

zicklag:
- Use argh for parsing commandline arguments and generating help message
- Remove almost all panics and handle errors by printing them and
exiting 1

Resolves #1107 

---

This adds the `argh` dependency for argument parsing which adds a measurable amount of compile time:

**Before:**

Targets: | naga-cli 0.4.0 (bin "naga")
-- | --
Total time: | 23.6s

**After:**

Targets: | naga-cli 0.4.0 (bin "naga")
-- | --
Total time: | 25.3s

That's a 1.7s ( 7.2% ) increase in build time on a clean release build.

If this is undesired I can switch it back to using the old parser, update the help message for the old parser to include available options, and keep the improved error handling introduced in this PR.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,7 +1,111 @@
 #![allow(clippy::manual_strip)]
 #[allow(unused_imports)]
 use std::fs;
-use std::{env, error::Error, path::Path};
+use std::{error::Error, fmt, path::Path, str::FromStr};
+
+/// Translate shaders to different formats
+#[derive(argh::FromArgs, Debug, Clone)]
+struct Args {
+    /// validate the shader during translation
+    #[argh(switch)]
+    validate: bool,
+
+    /// what policy to use for index bounds checking.
+    ///
+    /// May be `Restrict`, `ReadZeroSkipWrite`, or `UndefinedBehavior`
+    #[argh(option)]
+    index_bounds_check_policy: Option<IndexBoundsCheckPolicyArg>,
+
+    /// directory to dump the SPIR-V flow dump to
+    #[argh(option)]
+    flow_dir: Option<String>,
+
+    /// the shader entrypoint to use when compiling to GLSL
+    #[argh(option)]
+    entry_point: Option<String>,
+
+    /// the shader profile to use, for example `es`, `core`, `es330`, if translating to GLSL
+    #[argh(option)]
+    profile: Option<GlslProfileArg>,
+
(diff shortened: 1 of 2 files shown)


cli/src/main.rs
```diff
@@ -43,122 +144,87 @@ impl<T, E: Error> PrettyResult for Result<T, E> {
...
+        print_err(e.as_ref());
+        std::process::exit(1);
+    }
+}
+
+/// Error type for the CLI
+#[derive(Debug, Clone)]
+struct CliError(String);
+impl std::fmt::Display for CliError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
kvark, line 165: nit: we should add `std::fmt` to includes

cli/src/main.rs
```diff
@@ -43,122 +144,87 @@ impl<T, E: Error> PrettyResult for Result<T, E> {
...
 fn main() {
+    if let Err(e) = run() {
+        print_err(e.as_ref());
+        std::process::exit(1);
+    }
+}
+
+/// Error type for the CLI
+#[derive(Debug, Clone)]
+struct CliError(String);
```
kvark, line 163: Would `type CliError = Cow<'static, str>` allow us to write less boilerplate?
Actually, maybe even `type CliError = &'static str`?

cli/src/main.rs
```diff
@@ -347,6 +446,7 @@ use codespan_reporting::{
         termcolor::{ColorChoice, StandardStream},
     },
 };
+use naga::back::{glsl::Version, hlsl::ShaderModel, IndexBoundsCheckPolicy};
```
kvark, line 449: If this is needed for the code above, let's have it at the top.
Actually, do we need to include `glsl::Version` and `hlsl::ShaderModel`? We generally avoid including backend-specific things.

zicklag:
Updated!

kvark (approved):

PR #1121, opened 2021-07-21
Create Processor for Detecting Unused Module Items

zicklag:
Creates a module processor that returns the list of items in the module
that are not used for a particular entrypoint. Also integrated with the
GLSL backend to prune unused items from generated shaders.

---

Not super tested yet, but I think it's working! I'll mark it not a draft after more testing.

### DIFF
--- a/src/proc/unused.rs
+++ b/src/proc/unused.rs
@@ -1,0 +1,415 @@
+use std::collections::HashSet;
+
+use crate::Handle;
+
+/// Collection of handles for unused items in a module
+#[derive(Default, Clone, Debug)]
+pub struct UnusedItems {
+    pub types: HashSet<Handle<crate::Type>>,
+    pub global_variables: HashSet<Handle<crate::GlobalVariable>>,
+    pub constants: HashSet<Handle<crate::Constant>>,
+    pub functions: HashSet<Handle<crate::Function>>,
+}
+
+/// Gets the list of unused items for a given module
+///
+/// You may optionally specify an entry point and this will detect all items not used by that
+/// entrypoint
+pub fn get_unused_items<'a>(
+    module: &'a crate::Module,
+    entry_point: Option<&'a crate::EntryPoint>,
+) -> UnusedItems {
+    // Lists of used items
+    let mut used_global_variables = HashSet::with_capacity(module.global_variables.len());
+    let mut used_types = HashSet::with_capacity(module.types.len());
+    let mut used_constants = HashSet::with_capacity(module.constants.len());
+    let mut used_functions = HashSet::with_capacity(module.functions.len());
+
+    // Lists of unused items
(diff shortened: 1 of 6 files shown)


zicklag:
OK, I fixed a bug revealed from the snapshot examples and I think this is ready for review. Not tested in WGPU yet, but I'll do that tomorrow.

kvark:
Let's not rush into generalizing things. Instead, let's first make things work, and then see what needs to be generalized.
For the GLSL backend, we only care about skipping the types that are not used by the global variables. Writing this code is a couple of lines. Instead, you are suggesting this general (as in - platform independent processor) PR with 400+ lines.

src/proc/unused.rs
```diff
@@ -0,0 +1,386 @@
...
+fn crawl_type(
+    ty: Handle<crate::Type>,
+    types: &crate::Arena<crate::Type>,
+    used_types: &mut HashSet<Handle<crate::Type>>,
+) {
+    used_types.insert(ty);
+
+    // Crawl any sub-types
+    match types[ty].inner {
+        crate::TypeInner::Scalar { kind: _, width: _ } => (),
```
Gordon-F, line 125: You can write `crate::TypeInner::Scalar { .. } => ...` if you not used any type fields.

zicklag:
@kvark If there was a way to do it simply I'm all for that, but I couldn't figure out how to prune the unused structs, without crawling the function bodies to figure out which structs are used.

For the exact case we face with WebGL, the struct we need to prune is only used in a global variable, but what if you wrote a struct that you just used in one of the shader functions? That struct would get pruned because it wasn't used in a global, and then the function would be invalid.

I might be missing something, though. Does that not make sense?

src/proc/unused.rs
zicklag: I was thinking maybe it would be good to do it like that so that if fields were added to the variant later it would cause an error. Do you think that's not really worth worrying about?

Gordon-F: I don't think we should worry about it here. But if we will, it would be great to add an extra comment with an explanation :)

zicklag: OK, I'll just switch to `..` then. :+1: It's probably over-paranoid to do it the other way. :)

kvark:
@zicklag that's a fair concern! Indeed if we just prune the types based on global usage, even if we only prune top-level structs, users could technically use them internally without any bindings. However, our actual problem is narrower - we are really only concerned about top-level structs with unsized arrays. In this sense, pruning is perfectly fine, since you can't use such structs internally in shaders (since they are unsized). So I think we should still go for the easy 5-line solution here in GLSL before considering anything more general, like a processor in the PR.

zicklag:
Closing in favor of #1122.

PR #1122, opened 2021-07-21
Prune Unused, Unsized Types From GLSL Output

zicklag:
Prunes all unused, unsized types that are not referenced by globals used by the current entrypoint for the GLSL backend.

### DIFF
--- a/src/back/glsl/features.rs
+++ b/src/back/glsl/features.rs
@@ -205,6 +205,8 @@ impl<'a, W> Writer<'a, W> {
     /// If the version doesn't support any of the needed [`Features`](Features) a
     /// [`Error::MissingFeatures`](super::Error::MissingFeatures) will be returned
     pub(super) fn collect_required_features(&mut self) -> BackendResult {
+        let ep_info = self.info.get_entry_point(self.entry_point_idx as usize);
+
         if let Some(depth_test) = self.entry_point.early_depth_test {
             self.features.request(Features::IMAGE_LOAD_STORE);
 
@@ -224,7 +226,7 @@ impl<'a, W> Writer<'a, W> {
             self.features.request(Features::COMPUTE_SHADER)
         }
 
-        for (_, ty) in self.module.types.iter() {
+        for (ty_handle, ty) in self.module.types.iter() {
             match ty.inner {
                 TypeInner::Scalar { kind, width } => self.scalar_required_features(kind, width),
                 TypeInner::Vector { kind, width, .. } => self.scalar_required_features(kind, width),
@@ -236,8 +238,42 @@ impl<'a, W> Writer<'a, W> {
                         self.features.request(Features::ARRAY_OF_ARRAYS)
                     }
 
+                    // If the array is dynamically sized
                     if size == crate::ArraySize::Dynamic {
-                        self.features.request(Features::DYNAMIC_ARRAY_SIZE);
+                        let mut is_used = false;
+
+                        // Check if this type is used in a global that is needed by the current entrypoint
+                        for (global_handle, global) in self.module.global_variables.iter() {
+                            // Skip unused globals
+                            if ep_info[global_handle].is_empty() {
+                                continue;
+                            }
+
+                            // If this array is the type of a global, then this array is used
+                            if global.ty == ty_handle {
+                                is_used = true;
+                                break;
+                            }
+
+                            // If the type of this global is a struct
+                            if let crate::TypeInner::Struct { ref members, .. } =
+                                self.module.types[global.ty].inner
+                            {
+                                // Check the last element of the struct to see if it's type uses
+                                // this array
+                                if let Some(last) = members.last() {
+                                    if last.ty == ty_handle {
+                                        is_used = true;
+                                        break;
+                                    }
+                                }
+                            }
+                        }
+
+                        // If this dynamically size array is used, we need dynamic array size support
+                        if is_used {
+                            self.features.request(Features::DYNAMIC_ARRAY_SIZE);
+                        }
                     }
                 }
                 TypeInner::Image {
@@ -289,7 +325,10 @@ impl<'a, W> Writer<'a, W> {
             }
         }
 
-        for (_, global) in self.module.global_variables.iter() {
+        for (handle, global) in self.module.global_variables.iter() {
+            if ep_info[handle].is_empty() {
+                continue;
+            }
             match global.class {
                 StorageClass::WorkGroup => self.features.request(Features::COMPUTE_SHADER),
                 StorageClass::Storage => self.features.request(Features::BUFFER_STORAGE),
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -545,9 +545,11 @@ impl<'a, W: Write> Writer<'a, W> {
                 TypeInner::Sampler { .. } => continue,
                 // All other globals are written by `write_global`
                 _ => {
-                    self.write_global(handle, global)?;
-                    // Add a newline (only for readability)
-                    writeln!(self.out)?;
+                    if !ep_info[handle].is_empty() {
+                        self.write_global(handle, global)?;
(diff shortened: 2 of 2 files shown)


src/back/glsl/mod.rs
```diff
@@ -777,6 +818,10 @@ impl<'a, W: Write> Writer<'a, W> {
         handle: Handle<crate::GlobalVariable>,
         global: &crate::GlobalVariable,
     ) -> BackendResult {
+        if self.info.get_entry_point(self.entry_point_idx as usize)[handle].is_empty() {
```
kvark, line 821: it would be easier to do this check at the caller side, since they already have `ep_info` by hand

src/back/glsl/mod.rs
```diff
@@ -453,6 +490,10 @@ impl<'a, W: Write> Writer<'a, W> {
         // This are always ordered because of the IR is structured in a way that you can't make a
         // struct without adding all of it's members first
         for (handle, ty) in self.module.types.iter() {
+            if self.unused_unsized_types.contains(&handle) {
+                continue;
+            }
+
             if let TypeInner::Struct { ref members, .. } = ty.inner {
                 // No needed to write a struct that also should be written as a global variable
                 let is_global_struct = self
```
kvark, line 499: oh, look! here we are already doing the scan of globals, finding the variable corresponding to this type.
This code ensures that we aren't writing down any structs that have globals using them.
I think this code needs to be fixed in a number of ways:
  - support the case where there is a global using the struct, and also there are local users. Currently, the logic allows them exclusive.
  - skip globals that aren't used by the entry point

But we don't have to do this in your PR, we can follow-up.
The important takeaway here is that this code would already skip any structs that are used by globals, so you don't need to add any extra checks here.

src/back/glsl/mod.rs
```diff
@@ -368,6 +404,7 @@ impl<'a, W: Write> Writer<'a, W> {
 
             block_id: IdGenerator::default(),
             named_expressions: crate::NamedExpressions::default(),
+            unused_unsized_types: unused_unsized_structs,
```
kvark, line 407: in fact, even this array isn't needed.
The current logic of the backend is such that the struct is only written down together with the global variable itself. This logic needs to be extended, but for your fix, the *only* extra logic is needed is checking that `ep_info` before calling `write_global`.

kvark (changes requested):

src/back/glsl/mod.rs
zicklag: Oh, interesting! I'll take that check out.

src/back/glsl/mod.rs
zicklag: Updated locally.

src/back/glsl/mod.rs
zicklag: I also need the list of unused structs in the `collect_required_features()`, though.

src/back/glsl/mod.rs
```diff
@@ -317,6 +317,13 @@ pub struct Writer<'a, W> {
     block_id: IdGenerator,
     /// Set of expressions that have associated temporary variables
     named_expressions: crate::NamedExpressions,
+    /// List of unused structs that have dynamic-sized arrays in them
+    ///
+    /// This is used to prune out the unsized structs that are not used by globals accessed by the
+    /// current entrypoint. This is done to make sure that, if an entrypoint doesn't need SSBOs,
+    /// that it doesn't require them and cause shader compilation failure on platforms where they
+    /// are not supported.
+    unused_unsized_types: Vec<Handle<crate::Type>>,
```
Gordon-F, line 326: If we are looking only for globals, maybe it would be easier to use `Vec<Handle<GlobalVariable>` ?

zicklag: The only time we use that struct now is actually [here](https://github.com/gfx-rs/naga/pull/1122/files#diff-66bc1c8d2690a79f6feff4967c60eb32e58447d0bf01fb83d9ee45de9720026bR227) where the required feature check is looping through `Handle<crate::Type>`'s, so I using handles to types is right.

src/back/glsl/mod.rs
zicklag: We could actually remove the field and then make it an argument of `collect_required_features()` though if you thought that was better.

src/back/glsl/mod.rs
```diff
@@ -353,6 +360,35 @@ impl<'a, W: Write> Writer<'a, W> {
...
 
+        // Collect the list of unused, unsized structs for this entrypoint
+        let mut unused_unsized_types = Vec::default();
+        for (handle, global) in module.global_variables.iter() {
+            if info.get_entry_point(ep_idx)[handle].is_empty() {
+                match module.types[global.ty].inner {
+                    // If the global type is an unsized array
+                    crate::TypeInner::Array { size, .. } => {
+                        if crate::ArraySize::Dynamic == size {
+                            unused_unsized_types.push(global.ty);
```
kvark, line 371: I don't think we can do this. According to this logic, if one global is used, and another one is unused, and they have the same type, this type would still go into the `unused_unsized_types`, even though it's used by the other global

src/back/glsl/mod.rs
```diff
@@ -353,6 +360,35 @@ impl<'a, W: Write> Writer<'a, W> {
         let mut namer = proc::Namer::default();
         namer.reset(module, keywords::RESERVED_KEYWORDS, &["gl_"], &mut names);
 
+        // Collect the list of unused, unsized structs for this entrypoint
+        let mut unused_unsized_types = Vec::default();
+        for (handle, global) in module.global_variables.iter() {
+            if info.get_entry_point(ep_idx)[handle].is_empty() {
+                match module.types[global.ty].inner {
+                    // If the global type is an unsized array
+                    crate::TypeInner::Array { size, .. } => {
```
kvark, line 369: nit: can just pattern match here: `size: crate::ArraySize::Dynamic`

src/back/glsl/mod.rs
```diff
@@ -353,6 +360,35 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                    }
+                    // Or if it is a struct with an unsized array as the last element
+                    crate::TypeInner::Struct { ref members, .. } => {
+                        if let Some(last) = members.last() {
+                            if let crate::TypeInner::Array { size, .. } =
+                                module.types[last.ty].inner
+                            {
+                                if crate::ArraySize::Dynamic == size {
+                                    unused_unsized_types.push(global.ty);
+                                    unused_unsized_types.push(last.ty);
```
kvark, line 382: we shouldn't add the contained type to the list, we only care about the global types here.

kvark (changes requested):

src/back/glsl/mod.rs
zicklag: Yeah, you're right. I think I keep overcomplicating this. I've pushed another try.

This one should make more sense. :crossed_fingers:

zicklag (commented):
I pushed another attempt that should work better and be less convoluted hopefully.

src/back/glsl/features.rs
```diff
@@ -236,8 +238,46 @@ impl<'a, W> Writer<'a, W> {
...
+                                break;
+                            }
+
+                            // If the type of this global is a struct
+                            if let crate::TypeInner::Struct { ref members, .. } =
+                                self.module.types[global.ty].inner
+                            {
+                                // Check the last element of the struct to see if it's type uses
+                                // this array
+                                if let Some(last) = members.last() {
```
kvark, line 264: it seems to me that we should just compare `last.ty == ty_handle` here instead of matching the type

kvark (commented):
Almost there!

src/back/glsl/features.rs
zicklag: Ah, good point, because we already checked.

Pushed an update.

kvark (approved):

PR #1123, opened 2021-07-21
Skip Early Depth Test When Unsupported in GLSL

zicklag:


src/back/glsl/features.rs
```diff
@@ -206,7 +206,12 @@ impl<'a, W> Writer<'a, W> {
     /// [`Error::MissingFeatures`](super::Error::MissingFeatures) will be returned
     pub(super) fn collect_required_features(&mut self) -> BackendResult {
         if let Some(depth_test) = self.entry_point.early_depth_test {
-            self.features.request(Features::IMAGE_LOAD_STORE);
+            // If IMAGE_LOAD_STORE is supported for this version of GLSL
```
kvark, line 209: let's just remove this check

src/back/glsl/mod.rs
```diff
@@ -430,22 +430,27 @@ impl<'a, W: Write> Writer<'a, W> {
 
         // Enable early depth tests if needed
         if let Some(depth_test) = self.entry_point.early_depth_test {
-            writeln!(self.out, "layout(early_fragment_tests) in;")?;
+            // If early depth test is supported for this version of GLSL
+            if !(self.options.version < Version::Desktop(130)
```
kvark, line 434: let's do `log::warn!` in the `else` branch of this

src/back/glsl/features.rs
zicklag: On platforms where it is supported, we do check for the flag when writing out the extensions needed by the shader, does that mean we would still need the that?

src/back/glsl/mod.rs
```diff
@@ -430,22 +430,27 @@ impl<'a, W: Write> Writer<'a, W> {
 
         // Enable early depth tests if needed
         if let Some(depth_test) = self.entry_point.early_depth_test {
-            writeln!(self.out, "layout(early_fragment_tests) in;")?;
+            // If early depth test is supported for this version of GLSL
```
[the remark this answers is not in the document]
zicklag: I like that!

Gordon-F (approved):

src/back/glsl/features.rs
kvark: sounds reasobable

kvark (approved):

PR #1124, opened 2021-07-22
Add MSAA depth texture type

kvark:
Follows https://github.com/gpuweb/gpuweb/pull/1793

### DIFF
--- a/src/back/hlsl/image.rs
+++ b/src/back/hlsl/image.rs
@@ -101,8 +101,11 @@ impl<'a, W: Write> super::Writer<'a, W> {
         let dim_str = query.dim.to_hlsl_str();
         let class_str = match query.class {
             crate::ImageClass::Sampled { multi: true, .. } => "MS",
-            crate::ImageClass::Depth => "Depth",
-            _ => "",
+            crate::ImageClass::Depth { multi: true } => "DepthMS",
+            crate::ImageClass::Depth { multi: false } => "Depth",
+            crate::ImageClass::Sampled { multi: false, .. } | crate::ImageClass::Storage { .. } => {
+                ""
+            }
         };
         let arrayed_str = if query.arrayed { "Array" } else { "" };
         let query_str = match query.query {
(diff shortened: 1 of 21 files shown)


jimblandy (approved):
Looks good to me.

jimblandy:
Seems like `hlsl::Writer::write_expr`'s `ImageLoad` match arm needs to handle `ImageClass::Depth` specially, too?

jimblandy:
Does `back::glsl::Writer::collect_required_features` need to request `Features::MULTISAMPLED_TEXTURES` when it sees a multisampled depth texture?

jimblandy (changes requested):
Actually, I may have been too hasty. I think `back::glsl::Writer::collect_required_features` requires a new case for `ImageClass::Depth`, too.

jimblandy:
GitHub won't let me comment on lines outside patch, but here are a few more spots that need attention:

- [ ] In `back/hlsl/image.rs`, `Writer::write_wrapped_image_query_functions` may need to omit the mipmap level argument to `GetDimensions` for `Depth { multi: true }` images.
- [ ] In `valid/expression.rs`, in two separate spots, `Validator::validate_expression` mysteriously uses `ImageClass::Depth { .. }` as a pattern on current master, even though `Depth` has no fields yet. I didn't even know Rust permitted this. But if you were counting on rustc error messages to find patterns you needed to update, this would hide those patterns from you. The two spots are computing `can_level` values for `ImageSample` and `ImageQuery` expressions, and they do care about the `multi` flag for `Sampled`, so I'd assume they would want to check it on `Depth` as well.

src/back/hlsl/image.rs
```diff
@@ -129,127 +132,125 @@ impl<'a, W: Write> super::Writer<'a, W> {
...
+            write!(self.out, "{}", INDENT)?;
+            self.write_expr(module, image, func_ctx)?;
+            write!(self.out, ".GetDimensions(")?;
+            match query {
+                Iq::Size { level: Some(_) } => {
+                    write!(self.out, "{}, ", MIP_LEVEL_PARAM)?;
+                }
+                _ => match class {
+                    crate::ImageClass::Sampled { multi: true, .. }
+                    | crate::ImageClass::Depth { multi: true } => {}
```
jimblandy, line 216: Ahh, here is the only actually new line in this entire diff hunk.

src/valid/expression.rs
```diff
@@ -454,16 +452,11 @@ impl super::Validator {
                 }
 
                 // check level properties
-                let can_level = match class {
-                    crate::ImageClass::Sampled { multi, .. } => !multi,
-                    crate::ImageClass::Storage { .. } => unreachable!(),
-                    crate::ImageClass::Depth { .. } => true,
-                };
                 match level {
                     // require `can_level` here?
```
jimblandy, line 456: This comment probably needs to come out.

jimblandy (approved):
Looks good - just one comment that should be deleted now.

PR #1125, opened 2021-07-22
hlsl-out: write StructuredBuffer

kvark:


src/back/hlsl/writer.rs
```diff
@@ -352,29 +352,51 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                "b"
+            }
+            crate::StorageClass::Storage => {
+                let (prefix, register) =
+                    if global.storage_access.contains(crate::StorageAccess::STORE) {
+                        ("RW", "u")
+                    } else {
+                        ("", "t")
+                    };
+                write!(self.out, "{}StructuredBuffer<", prefix)?;
```
Gordon-F, line 380: `write` macro inside `let _ = {}` confusing a little bit. Are you trying to avoid extra heap allocation with `format!`?

src/back/hlsl/writer.rs
```diff
@@ -352,29 +352,51 @@ impl<'a, W: Write> Writer<'a, W> {
...
+            }
+            crate::StorageClass::WorkGroup => {
+                write!(self.out, "groupshared ")?;
+                self.write_type(module, global.ty)?;
+                ""
+            }
+            crate::StorageClass::Uniform => {
+                // constant buffer declarations are expected to be inlined, e.g.
+                // `cbuffer foo: register(b0) { field1: type1; }`
+                write!(self.out, "cbuffer ")?;
```
Gordon-F, line 370:
```suggestion
                write!(self.out, "cbuffer")?;
```

src/back/hlsl/writer.rs
kvark: sorry, I don't understand. This `write!` is not inside any `let _ = {}`

Gordon-F: Sorry, I mean a logic inside `let register_ty = match global.class {..}`.

Maybe we should split this logic ?

```rust
let register_ty = match global.class {
        crate::StorageClass::Function => unreachable!("Function storage class"),
        crate::StorageClass::Uniform => "b",
        crate::StorageClass::Storage => {
            if global.storage_access.contains(crate::StorageAccess::STORE) {
                "u"
            } else {
                "t"
            }
        }
....
```

```rust
        match global.class {
            crate::StorageClass::Function => unreachable!("Function storage class"),
            crate::StorageClass::Private => {
                write!(self.out, "static ")?;
                self.write_type(module, global.ty)?;
            }
            crate::StorageClass::Uniform => {
                // constant buffer declarations are expected to be inlined, e.g.
                // `cbuffer foo: register(b0) { field1: type1; }`
                write!(self.out, "cbuffer ")?;
            }
.....
```

kvark: We could. I just don't want matching on the storage class multiple times

Gordon-F: Ok, let's keep it as is :)

PR #1132, opened 2021-07-23
hlsl: remove the boolean casts for unary operators

kvark:
Those bool casts on vectors were wrong. And on scalars I think they aren't necessary.

### DIFF
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -100,6 +100,9 @@ impl<'a, W: Write> Writer<'a, W> {
             }
         }
 
+        // Extra newline for readability
+        writeln!(self.out)?;
+
         // Save all entry point output types
         let ep_results = module
             .entry_points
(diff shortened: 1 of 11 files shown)


kvark:
@Gordon-F looks like it works fine without the boolean casts? We need this to get Water example running.

src/back/hlsl/writer.rs
```diff
@@ -100,6 +100,9 @@ impl<'a, W: Write> Writer<'a, W> {
             }
         }
 
+        // Extra newline for readability
+        writeln!(self.out)?;
```
grovesNL, line 104: @kvark these seem to be placed at the beginning of the file, was this supposed to go somewhere else?

kvark: These are placed after the constants. Previously, each constant had its own newline.

grovesNL: The diff shows it placed at the beginning of the files though, unless the diff is unrelated

kvark: Oh, that's probably because those tests don't have constants. That's ok. We shouldn't optimize the verbosity for small shaders, but rather focus on bigger ones.

PR #1135, opened 2021-07-23
Fixes for zero init variables and variables with array type for hlsl and glsl backends

Gordon-F:
This PR contains:

1. Allow to translate from bool to another type in validator
1. [hlsl-out] Allow sint array size
1. [hlsl-out] Fix local variable writing with array type
1. [glsl-out] Fix array type writing
1. [glsl-out] Fix zero initialization for local variables

### DIFF
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -799,6 +799,10 @@ impl<'a, W: Write> Writer<'a, W> {
             // The leading space is important
             self.write_type(module, local.ty)?;
             write!(self.out, " {}", self.names[&func_ctx.name_key(handle)])?;
+            // Write size for array type
+            if let TypeInner::Array { size, .. } = module.types[local.ty].inner {
+                self.write_array_size(module, size)?;
+            }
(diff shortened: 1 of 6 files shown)


src/back/hlsl/writer.rs
```diff
@@ -499,14 +499,18 @@ impl<'a, W: Write> Writer<'a, W> {
 
         // Write the array size
         // Writes nothing if `ArraySize::Dynamic`
-        // Panics if `ArraySize::Constant` has a constant that isn't an uint
+        // Panics if `ArraySize::Constant` has a constant that isn't an sint or uint
         match size {
             crate::ArraySize::Constant(const_handle) => {
                 match module.constants[const_handle].inner {
                     crate::ConstantInner::Scalar {
```
kvark, line 506: we should just call `module.constants[const_handle].to_array_length().unwrap()`

kvark (approved):
Nice!
Interestingly, I was doing similar changes yesterday in HLSL world :)

PR #1144, opened 2021-07-27
[glsl-out] split reflection_names into a map for types and another for globals

### LINKED ISSUE #1137: [glsl-out] ImageMultipleSamplers error for multiple samplers due to name collision
Given the WGSL code
```rust
[[group(0), binding(1)]]
var texture_a: texture_2d<f32>;
[[group(0), binding(2)]]
var sampler_a: sampler;
[[group(0), binding(3)]]
var texture_b: texture_2d<f32>;
[[group(0), binding(4)]]
var sampler_b: sampler;

[[stage(fragment)]]
fn main() -> [[location(0)]] vec4<f32> {
    let a = textureSample(texture_a, sampler_a, vec2<f32>(0.0));
    let b = textureSample(texture_b, sampler_b, vec2<f32>(0.0));
    return vec4<f32>(1.0);
}
```
naga fails to convert it to GLSL due to `Conflicting samplers for _group_0_binding_3`.

The relevant code is here https://github.com/gfx-rs/naga/blob/526a29e2b84801cef9d068f9d1d4648336916ea0/src/back/glsl/mod.rs#L2441-L2457

Logging the `global` and `tex_name` values gives to
```
GlobalVariable { name: Some("texture_a"), class: Handle, binding: Some(ResourceBinding { group: 0, binding: 1 }), ty: [1], init: None, storage_access: (empty) }
_group_0_binding_3
GlobalVariable { name: Some("texture_b"), class: Handle, binding: Some(ResourceBinding { group: 0, binding: 3 }), ty: [1], init: None, storage_access: (empty) }
_group_0_binding_3
```

The `tex_name` is looked up using only the global's `ty` which is the same for both both textures.

jakobhellermann:
fixes #1137, not sure whether this is the right approach

The issue in #1137 was that the `reflection_names` map was per type but different globals shared the same type, so the first one got overwritten.
This PR splits the `reflection_names` into a map of `Handle<GlobalVariable> => String` and the old `Handle<Type> => String`.

I'm not sure if the globals map is even needed though, because the usage [here](https://github.com/gfx-rs/naga/compare/master...jakobhellermann:fix-duplicate-sampler?expand=1#diff-cc33bd3c8c18000d3ee7b6f5578e8fa07ffd08d87b0aa3d9525a107937fb3d3fR2435) could just be `let tex_name = self.get_global_name(handle, global)` and the global map could be removed.

The last change to this code was in https://github.com/gfx-rs/naga/pull/606, @Gordon-F can you take a look and comment on which approach you think is best?


zicklag:
This fixes the WGPU Bevy problem I ran into! :heart:

src/back/glsl/mod.rs
```diff
@@ -312,7 +312,9 @@ pub struct Writer<'a, W> {
     /// (generated by a [`Namer`](crate::proc::Namer))
     names: crate::FastHashMap<NameKey, String>,
     /// A map with all the names needed for reflections
-    reflection_names: crate::FastHashMap<Handle<crate::Type>, String>,
+    reflection_names_structs: crate::FastHashMap<Handle<crate::Type>, String>,
```
Gordon-F, line 315: Maybe `reflection_names_buffers` or `reflection_names_uniforms`?

Gordon-F (approved):
@jakobhellermann Thank you! LGTM.

src/back/glsl/mod.rs
kvark: why do we ever need to reflect the *type* names? they shouldn't be exposed anywhere in the GL API

Gordon-F: We need names only for uniforms here - https://github.com/gfx-rs/wgpu/blob/master/wgpu-hal/src/gles/device.rs#L38.

> they shouldn't be exposed anywhere in the GL API

Performance? We already got this info.

jakobhellermann: I renamed it to `reflection_names_uniforms`.

kvark: @Gordon-F that lookup goes with a variable handle, not a type handle. I'm asking why we need to look up any reflection based on the type.

src/back/glsl/mod.rs
```diff
@@ -2455,7 +2457,7 @@ impl<'a, W: Write> Writer<'a, W> {
             match self.module.types[var.ty].inner {
                 crate::TypeInner::Struct { .. } => match var.class {
                     crate::StorageClass::Uniform | crate::StorageClass::Storage => {
-                        let name = self.reflection_names[&var.ty].clone();
+                        let name = self.reflection_names_uniforms[&var.ty].clone();
```
kvark, line 2460: why are we looking up `var.ty` instead of looking up `handle` directly in that other map?

jakobhellermann: I believe this is because the `write_struct` function, where [`reflection_names_uniforms.insert` is called](https://github.com/gfx-rs/naga/blob/ab2fa435e6c7e3f0026cb753ddcf1a4a16dcee7e/src/back/glsl/mod.rs#L1241), has no access to the global variable handle, only the type handle.

Gordon-F: Yep, we need a block struct name, that will be generated as uniform buffer. 

https://github.com/gfx-rs/naga/blob/ab2fa435e6c7e3f0026cb753ddcf1a4a16dcee7e/src/back/glsl/mod.rs#L1233-L1239

Gordon-F: If we will use the global variable handle here, we still need to check the variable type and ensure that the variable is a struct with a uniform storage class.

If we want to change this, we need to change block struct writing.

kvark: ok I see. So the block name shows up in the API, not the variable name, and the block name comes from the type. Ok

kvark (approved):
let's rebase and land this

jakobhellermann:
> let's rebase and land this

I rebased it.

PR #1146, opened 2021-07-27
[glsl-in] Add implicit conversion for initializer

### LINKED ISSUE #900: [glsl-in] Implicit conversions
This is a tracking issue for implicit conversions in the glsl frontend

- [x] In function casting
This is one is the easier of the two, we have a list of conversion priority and when there is a type conflict we try to catch to the one with higher priority

<details>
<summary>example</summary>
  
```glsl
void main() {
  // float has priority greater than int
  float a = 1; // This works and 1 is casted to float
  float b = 2.0 * 1; // This also works and 1 is casted to float
  int c = 1.0; // This doesn't work because float has higher priority than int
  uint d = 1; // this works because uint has a higher priority than int
  int e = 1u; // The last line implies that this doesn't work and it's true
  vec2 f = ivec(2); // This works the same way, we only check the scalar kind
}
```
</details>

- [x] function call casting
This one is a little stranger, the idea is that if the signature matches any known definition/prototype we use it, otherwise we try a cast for all known overloads, then we can have three scenarios
- if none satisfy the priority rule we throw a error saying that there's no known function with that signature
- if we have one result it's business as usual we do the casts and call the selected overload
- if we get more a error is thrown saying that the resolution is ambiguous

<details>
<summary>example</summary>
  
```glsl
// Signature match call the second overload
void test(float a) {}
void test(int a) {}

void main() {
  test(1);
}
```

```glsl
// No signature match and no valid cast, throw error
void test(int a) {}
void test(uint a) {}

void main() {
  test(1.0);
}
```

```glsl
// No signature match but one overload satisfies the cast rules
void test(float a) {}
void test(int a) {}

void main() {
  test(1u);
}
```

```glsl
// No signature match but multiple overloads satisfy the cast rules, throw error
void test(float a) {}
void test(uint a) {}

void main() {
  test(1);
}
```
</details>

The first issue is a simple integer comparison but the second one I'm not sure about the best implementation so if anyone has any idea please leave a comment.

Our current implementation has a `HashMap` of `FunctionSignature` pointing to a function.

`FunctionSignature` is defined has
```
struct FunctionSignature {
  name: String,
  parameters: Vec<Handle<Type>>
}
```

JCapucho:
Should be the last place needing implicit conversions.

Closes #900

### DIFF
--- a/src/front/glsl/parser.rs
+++ b/src/front/glsl/parser.rs
@@ -522,7 +522,14 @@ impl<'source, 'program, 'options> Parser<'source, 'program, 'options> {
             ))
         } else {
             let expr = self.parse_assignment(ctx, body)?;
-            Ok(ctx.lower_expect(self.program, expr, false, body)?)
+            let (mut init, init_meta) = ctx.lower_expect(self.program, expr, false, body)?;
+
+            let scalar_components = ast::scalar_components(&self.program.module.types[ty].inner);
+            if let Some((kind, width)) = scalar_components {
+                ctx.implicit_conversion(self.program, &mut init, init_meta, kind, width)?;
+            }
+
+            Ok((init, init_meta))
         }
     }
 


kvark (approved):
uh those implicit conversions

PR #1148, opened 2021-07-28
[glsl-in] Implement layouting for structs

JCapucho:
Implements the rules described in section 7.6.2.2 of the OpenGl spec to calculate the offsets of struct members

### DIFF
--- a/src/front/glsl/offset.rs
+++ b/src/front/glsl/offset.rs
@@ -1,0 +1,145 @@
+//! Module responsible for calculating the offset and span for types.
+//!
+//! There exists two types of layouts std140 and std430 (there's technically
+//! two more layouts, shared and packed. Shared is not supported by spirv. Packed is
+//! implementation dependent and for now it's just implemented as an alias to
+//! std140).
+//!
+//! The OpenGl spec (the layout rules are defined by the OpenGl spec in section
+//! 7.6.2.2 as opposed to the GLSL spec) uses the term basic machine units which are
+//! equivalent to bytes.
+use super::{ast::StructLayout, error::ErrorKind, SourceMetadata};
+use crate::{Arena, Constant, Handle, Type, TypeInner};
+
+/// Struct with information needed for defining a struct member.
+///
+/// Returned by [`calculate_offset`](calculate_offset)
+#[derive(Debug)]
+pub struct TypeAlignSpan {
+    /// The handle to the type, this might be the same handle passed to
+    /// [`calculate_offset`](calculate_offset) or a new such a new array type
+    /// with a different stride set.
+    pub ty: Handle<Type>,
+    /// The alignment required by the type.
+    pub align: u32,
+    /// The size of the type.
+    pub span: u32,
+}
+
+/// Returns the type, alignment and span of a struct member according to a [`StructLayout`](StructLayout).
+///
+/// The functions returns a [`TypeAlignSpan`](TypeAlignSpan) which has a `ty` member
+/// this should be used as the struct member type because for example arrays may have to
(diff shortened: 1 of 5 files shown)


tests/out/wgsl/246-collatz-comp.wgsl
```diff
@@ -1,6 +1,6 @@
 [[block]]
 struct PrimeIndices {
-    indices: [[stride(4)]] array<u32>;
+    indices: [[stride(16)]] array<u32>;
```
kvark, line 3: is this expected? an array of `u32` having the stride of 16 now?

src/front/glsl/offset.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// The size of the type.
+    pub span: u32,
+}
+
+/// Returns the type, alignment and span of a struct member according to a [`StructLayout`](StructLayout).
+///
+/// The functions returns a [`TypeAlignSpan`](TypeAlignSpan) which has a `ty` member
+/// this should be used as the struct member type because for example arrays may have to
+/// change the stride and as such need to have a different type.
+pub fn calculate_offset(
```
kvark, line 34: I'm having a hard time understanding what this does. It looks to be mutating the type arena, adding types that are computed based on the offsets/strides/spans of the relevant std140/std430 rules. My question is - why aren't the types already written down with respect to those rules to start with? I.e. I'd expect this either not mutating the types, or actually adjusting the existing types. But instead, it's adding new types into the arena.

tests/out/wgsl/246-collatz-comp.wgsl
JCapucho: From my understanding of the spec not defining the layout like what the collatz shader does implies the `shared` layout which essentially means everything is left for the compiler but you cannot discard unused members and the layout must be the same for all shaders with this exact definition.

In Naga's case this means it defaults for `std140` since it follows this rules (`std430` despite generating smaller strides, isn't supported in uniform blocks and there's no logic to pick depending on the block type) and `std140` has this weird rule that the stride must be rounded up to a multiple of the alignment of a `vec4` (16 bytes).

src/front/glsl/offset.rs
JCapucho: The reason why we write new types instead of reusing old ones or writing them already with the correct layout , is that for the same syntactic type (ie. the one the user wrote like `float[2]`) there can be a multitude of layouts for example:

- `packed` - compiler dependent (naga uses `std140`)
- `shared` - compiler dependent but maintains compatibility with other shader programs  (naga uses `std140`)
- `std140`
- `std430`

According to spec these can all be incompatible but since we use `std140` for all compiler dependent layouts, our only distinction is between `std140` and `std430`.

The OpenGl spec has this weird rule for `std140` that the base alignment for structs and arrays must be rounded up to a multiple of the alignment of a `vec4` (16 bytes) and also the stride, `std430` removes it. But the layouting rules are recursive so now we have this problem:
```glsl
// The parser defaults to using std140
// But since the rules are recursive a new struct type with a different layout may be used 
struct BunchaFloats {
    float v[5]; // Stride defaults to 16
};

layout(std140) buffer TestStd140 {
    float test_float; // +0x00
    BunchaFloats test_struct; // +0x10
};

layout(std430) buffer TestStd430 {
    float test_float; // +0x00
    BunchaFloats test_struct; // +0x04
};
```
As you can see we only define one struct and one array but need two struct types and two array types.

kvark: makes sense, thank you for a good explanation!

kvark (approved):

PR #1152, opened 2021-07-29
[glsl-in] Texture function improvements

JCapucho:
Extracts the `function_call` to `function_or_constructor_call` to reduce the indentation, adds implicit conversion for texture functions and implements `textureGrad`.


kvark (approved):
let's make clippy happy

PR #1158, opened 2021-08-01
[glsl-in] Add support for shadow sampling

JCapucho:
glsl has no concept of depth images, so the frontend when parsing a constructor for shadow combined samplers will convert the image from sampled to depth, further more to support this when using textures passed trough function arguments a bitset is used that upon function call will be checked and will also convert the image type if needed.

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -163,6 +163,30 @@ impl<'a> Program<'a> {
         Ok(context.typifier.get(handle, &self.module.types))
     }
 
+    /// Invalidates the cached type resolution for `handle` forcing a recomputation
+    pub fn invalidate_expression<'b>(
+        &'b self,
+        context: &'b mut Context,
+        handle: Handle<Expression>,
+        meta: SourceMetadata,
+    ) -> Result<(), ErrorKind> {
+        let resolve_ctx = ResolveContext {
+            constants: &self.module.constants,
+            types: &self.module.types,
+            global_vars: &self.module.global_variables,
+            local_vars: context.locals,
+            functions: &self.module.functions,
+            arguments: context.arguments,
+        };
+
+        context
+            .typifier
+            .invalidate(handle, context.expressions, &resolve_ctx)
+            .map_err(|error| {
+                ErrorKind::SemanticError(meta, format!("Can't resolve type: {:?}", error).into())
+            })
+    }
+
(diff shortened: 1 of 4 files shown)


src/front/glsl/ast.rs
```diff
@@ -36,6 +39,10 @@ pub struct FunctionDeclaration {
     pub defined: bool,
     /// Wheter or not this function returns void (nothing)
     pub void: bool,
+    /// [`BitSet`](BitSet) where each bit indicates if the argument in the
+    /// corresponding position should be treated as a depth image instead of a
+    /// regular image
+    pub depth_set: BitSet,
```
kvark, line 45: would it make sense to just make it a boolean field in some `struct Parameter`? there is already `parameters: Vec<>` here to store this

src/front/glsl/functions.rs
```diff
@@ -1351,4 +1336,143 @@ impl Program<'_> {
...
+        } else {
+            Err(ErrorKind::SemanticError(
+                image_meta,
+                "Type is not an image".into(),
+            ))
+        }
+    }
+}
+
+/// Helper function to cast a expression holding a sampled a image to a
```
kvark, line 1421: "a sampled a image"

src/front/glsl/functions.rs
```diff
@@ -1351,4 +1336,143 @@ impl Program<'_> {
...
+                })
+            }
+            ImageClass::Depth { .. } => {}
+            _ => return Err(ErrorKind::SemanticError(meta, "Not a texture".into())),
+        },
+        _ => return Err(ErrorKind::SemanticError(meta, "Not a texture".into())),
+    };
+
+    // Add a new expression to not have problems with the Typifier
+    // caching the old type
```
kvark, line 1468: This may be a bit of problem. So the new expressions are going to be typified correctly, but the old ones are still incorrect, and we are fine with this?

kvark (commented):
looks nice!

src/front/glsl/functions.rs
JCapucho: It's definitely not the best solution and I wanted to discuss with you, how do you feel about adding a method to invalidate a handle in the typifier?

kvark: yes, that sounds good

kvark (approved):

PR #1162, opened 2021-08-03
[glsl-in] Fixes and improvements

JCapucho:
All the fixes are contained in their own commits and can be reviewed individually

- Use the same splatting logic of `max` for `min`
- Fix missing implicit conversion on `mix`
- Parse `restrict`, `readonly` and `writeonly` (no logic yet, only parsing)
- Accept precision qualifiers in locals
- Handle `pow` in constant evaluation
- Fix the span of a `gvec3`

### DIFF
--- a/src/front/glsl/lex.rs
+++ b/src/front/glsl/lex.rs
@@ -3,7 +3,7 @@ use super::{
     token::{SourceMetadata, Token, TokenValue},
     types::parse_type,
 };
-use crate::FastHashMap;
+use crate::{FastHashMap, StorageAccess};
 use pp_rs::{
     pp::Preprocessor,
     token::{Punct, Token as PPToken, TokenValue as PPTokenValue},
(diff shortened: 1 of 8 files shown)


src/front/glsl/lex.rs
```diff
@@ -76,6 +76,9 @@ impl<'a> Iterator for Lexer<'a> {
                     "highp" => TokenValue::PrecisionQualifier(Precision::High),
                     "mediump" => TokenValue::PrecisionQualifier(Precision::Medium),
                     "lowp" => TokenValue::PrecisionQualifier(Precision::Low),
+                    "restrict" => TokenValue::Restrict,
+                    "readonly" => TokenValue::ReadOnly,
```
kvark, line 80: are we just ignoring them?

src/front/glsl/token.rs
```diff
@@ -56,6 +56,11 @@ pub enum TokenValue {
     Uniform,
     Buffer,
     Const,
+
+    Restrict,
+    ReadOnly,
```
kvark, line 61: Could be `StorageAccess(crate::StorageAccess)`

src/front/glsl/lex.rs
JCapucho: I had the logic in a separate branch, I've added the commit.

kvark (approved):
Great!

PR #1163, opened 2021-08-03
[glsl-in] Add support for non constant global init

JCapucho:
Glsl allows things such as

```glsl
layout (location = 0) in vec4 a;
vec4 b = a;
```

In this case `b` depends on `a` which can't be known at compile time, so it's necessary to add a small prologue to the entry point to initialize it.

The prologue for global initialization is added after the prologue that assigns from the entry point arguments to globals because the parser utilizes this globals as opposed to the arguments.


kvark (approved):
That's quite extensive

PR #1164, opened 2021-08-03
[glsl-in] Remove support for multiple entry points

JCapucho:
As discussed in the matrix room, this PR remove support for multiple entry points the tl:dr for this decision is to much work and pitfalls with the only known case of anyone using it being the test.


kvark (approved):

kvark:
RIP multiple EPs

PR #1165, opened 2021-08-04
[glsl-in] Fixes and improvements #2

JCapucho:
All the fixes are contained in their own commits and can be reviewed individually

- Respect user defined interpolation
- Lower the init expression in for loops
- Use std430 for `buffer` objects (as defined in the spec)
- Reduce number of produced vector resizes (makes the output more readable)
- Matrix initializers with only one scalar produce diagonal matrices
- `mix` when using a boolean selector has inverted argument order (compared to the IR and SPV)

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -925,6 +925,22 @@ impl<'function> Context<'function> {
 
         Ok(())
     }
+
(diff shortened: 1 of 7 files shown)


kvark (approved):
Reviewed commits individually

PR #1170, opened 2021-08-05
[spirv-in] Fix Storage Access for Storage Buffers

zicklag:
I was getting a validation error when parsing a simple RustGPU-generated spirv file. I think, according to the spec, that the storage buffer storage class should have load and store access.

I know pretty much nothing about SPIRV, so I could be totally wrong, but this makes the SPIRV file in question validate and translate successfully with naga.


kvark:
Thank you for filing this! While reviewing, I found a few places that needed to be fixed and wrote #1171 which replaces this PR.

PR #1174, opened 2021-08-08
[glsl-in] Parser refractor and public api cleanup

JCapucho:
This PR makes a major refractor to the public api and internals of the glsl frontend.

# Changes

- The `Parser` is merged with `Program` struct and a new struct `ParsingContext` is used, the rationale behind this decision is to introduce reusability to the `Parser` and to please the borrow checker when handling preprocessor directives. The `ParsingContext` holds the lexer and has all the parsing methods but now the  `Parser` is also passed, this is necessary because `peek` might encounter a preprocessor directive and need to modify the `Parser` so the borrows need to be disjoint.

- The lexer now returns a `LexerResult`, this remove the token buffer and is needed for proper parsing of preprocessor directives otherwise snippets like the following would fail.

```glsl
float a = 1.0 +
    #extension a: enabled
    2.0;
```

- Error reporting now tries to accumulate as much errors as possible before returning, this helps with  iteration times for the end user and was partially needed because the new way the preprocessor directives handle errors would require a change in the signature of  `peek` and `next`, which in my opinion would be a worse solution.

- The parsing was split up into modules, the rationale behind this decision is that the `parser` module was getting to big and convoluted, furthermore adding new parsing or documentation would have made the file hard to traverse.

- Finally the public api was cleaned to only show the `Parser` and any type which it might require or produce and all of them were properly documented


src/front/glsl/context.rs
```diff
@@ -0,0 +1,783 @@
+use crate::{
+    front::{
+        glsl::{
```
kvark, line 3: we generally include the same-frontend stuff as `super::`

src/front/glsl/parser_tests.rs
```diff
@@ -305,19 +394,23 @@ fn functions() {
     )
     .unwrap();
 
-    parse_program(
+    parse(
+        &mut parser,
```
kvark, line 398: can we make `parse` as a method on the parser, for convenience?

kvark (commented):
That's quite extensive!
Do we understand where the extra 800 LOC come from?

src/front/glsl/parser_tests.rs
JCapucho: Sorry I don't quite understand what you're proposing, can you elaborate?

JCapucho: Okay I  think I've figured out what you meant.

kvark (approved):
:shipit:

PR #1180, opened 2021-08-10
[glsl-in] Don't clone HirExpr when lowering

JCapucho:
Introduces a new type `StmtContext` which now holds the `Arena<HirExpr>`, this type is needed because lowering requires a disjoint borrow otherwise a clone is needed.

This type is reused in between lowers, this needed a new method on the `Arena` equal to it's `Vec` counterpart `clear`.

### DIFF
--- a/src/front/glsl/context.rs
+++ b/src/front/glsl/context.rs
@@ -29,9 +29,9 @@ pub struct Context {
     pub lookup_global_var_exps: FastHashMap<String, VariableReference>,
     pub samplers: FastHashMap<Handle<Expression>, Handle<Expression>>,
 
-    pub hir_exprs: Arena<HirExpr>,
     pub typifier: Typifier,
     emitter: Emitter,
+    stmt_ctx: Option<StmtContext>,
 }
 
 impl Context {
@@ -51,9 +51,9 @@ impl Context {
             ),
             samplers: FastHashMap::default(),
 
-            hir_exprs: Arena::default(),
             typifier: Typifier::new(),
             emitter: Emitter::default(),
+            stmt_ctx: Some(StmtContext::new()),
         };
 
         this.emit_start();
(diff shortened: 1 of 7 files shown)


src/front/glsl/context.rs
```diff
@@ -302,19 +359,22 @@ impl Context {
...
+    fn lower_inner(
         &mut self,
+        stmt: &StmtContext,
         parser: &mut Parser,
         expr: Handle<HirExpr>,
         lhs: bool,
         body: &mut Block,
     ) -> Result<(Option<Handle<Expression>>, SourceMetadata)> {
-        let HirExpr { kind, meta } = self.hir_exprs[expr].clone();
+        let &HirExpr { ref kind, meta } = &stmt.hir_exprs[expr];
```
kvark, line 371: is `&` needed on both sides?

src/front/glsl/context.rs
```diff
@@ -29,9 +29,9 @@ pub struct Context {
     pub lookup_global_var_exps: FastHashMap<String, VariableReference>,
     pub samplers: FastHashMap<Handle<Expression>, Handle<Expression>>,
 
-    pub hir_exprs: Arena<HirExpr>,
     pub typifier: Typifier,
     emitter: Emitter,
+    stmt_ctx: Option<StmtContext>,
```
kvark, line 34: does this need to be an option? That prevents reusing the allocation of the arena.

kvark (commented):
LGTM, just one question

src/front/glsl/context.rs
JCapucho: The `Option` is what allows the `Arena` to be reused, you can only have one `StmtContext` when you call `stmt_ctx` is takes if available from the `Context` and panics otherwise and later you call `lower` and it returns to the `Context`.

src/front/glsl/context.rs
JCapucho: No, it isn't

src/front/glsl/context.rs
kvark: oh I see, so this is to fight the borrow checker

Issue #866, opened 2021-05-15
"Expected separator" on hex literal


Nehliin:
Hello again, I ran into another issue with my glsl -> wgsl conversion. Wgsl doesn't seem to accept any of the following:
`let mask: u8 = 0xFFu;`
`let mask: u8 = 0xFF;`

Both gives the following error even though the values matches the regex in the specification:
```
  thread 'main' panicked at 'Failed to parse shader: error: expected separator, found 'xFFu'
     ┌─ wgsl:37:21
     │
  37 │     let mask: u8 = 0xFFu;
     │                     ^^^^ expected separator
```

kvark:
The error could have been more clear... WGSL doesn't know what `u8` is.
What Naga version are you using? Could you try the latest master?

Nehliin:
I don't think u8 is the issue here, I got the same error when using u32. I currently run whatever naga version that's used within wgpu 8.1. I've tested with latest naga master now and get the same issue.

kvark:
oh it doesn't expect the `u` suffix for hex literals, I see

Nehliin:
Even without the `u` suffix I get the same error :/ 
`Failed to parse WGSL code: expected separator, found 'xFF'`

hcsch:
I also stumbled across this across this wanting a hexadecimal `u32` constant.

My code
```wgsl
let test: u32 = 0x0u;
```
gets me the following error
```
error: the type of `test` is expected to be [1]
  ┌─ wgsl:1:5
  │
1 │ let test: u32 = 0x0u;
  │     ^^^^ definition of `test`

Could not parse WGSL
```

I suppose similar to the above issues, `0` is consumed as a whole number token (i.e. `x0u` is not included).
Due to the missing `u` suffix this way, the number literal is interpreted as an `i32` literal.

Here are some test cases for all integer and float literals that match the spec (and their current errors for those that fail):
```wgsl
// Regexes for the literals are taken from the working draft at
// https://www.w3.org/TR/2021/WD-WGSL-20210806/#literals

// matches for /^(-?[0-9]*\.[0-9]+|-?[0-9]+\.[0-9]*)((e|E)(\+|-)?[0-9]+)?$/
let dec_float_lit_0 : f32 = -1.;
let dec_float_lit_1 : f32 = -.1;
let dec_float_lit_2 : f32 = 42.1234;
let dec_float_lit_3 : f32 = -1.E3;
let dec_float_lit_4 : f32 = -.1e-5;
[...]

PR #1184, opened 2021-08-10
Implement lexing and parsing for all WGSL number literal types

hcsch:
I tried to implement lexing for all different types of number literals
as described by the current working draft and editor's draft
(they are the same for the literal definitions at the time of writing).
This should bring naga closer to being able to parse all kinds of number literals,
partially resolving #866 for example.

The next step would be to also allow for parsing of these new kinds of number literals,
as the currently used `FromStr::parse` won't work for hexadecimal literals (float or integer), see rust-lang/rfcs#1098.
The relevant code for this seems to be here:
https://github.com/gfx-rs/naga/blob/e384bce771db1d7dd156e9e29afe9d9235984924/src/front/wgsl/mod.rs#L1027-L1060

I have added the test cases I had written up for #866 to `tests/in/operators.wgsl` since they seem to fit in there
(though the file name might need some bikeshedding, should they stay there).
I have not yet adjusted any of the output files for the test, as I'm somewhat unsure about what they'd look like.

While I have tried my best to make the code easy to understand, it has become somewhat unwieldy due to the state machinery required to (hopefully) correctly lex the literals.
This state machine code also still needs to see some testing / fuzzing, which I haven't done myself so far.
I have tested how naga fails for `operators.wgsl` and my small tests seem to indicate that the literals are lexed correctly.

For example, with the following WGSL code:
```wgsl
let hex_float_lit_0 : f32 = -0xa.p1;
```
the modified code reports the following error:
```
expected floating-point literal, found `-0xa.p1`
```
wheras the old code reports the following:
```
the type of `hex_float_lit_0` is expected to be [1]
```

An alternative to lexing the literals this way, with a manually written state machine, would be to use the regexes given in the WGSL spec directly (after correcting them, since they use `+` ambiguously for example) with some existing regex engine like https://github.com/rust-lang/regex, this would pull in additional dependencies, but might be worth it, to reduce maintenance efforts, should the spec change it's definitions.

### DIFF
--- a/src/front/wgsl/lexer.rs
+++ b/src/front/wgsl/lexer.rs
@@ -13,47 +13,328 @@ fn consume_any(input: &str, what: impl Fn(char) -> bool) -> (&str, &str) {
     input.split_at(pos)
 }
 
+/// Tries to skip a given prefix in the input string.
+/// Returns whether the prefix was present and could therefore be skipped,
+/// the remaining str and the number of *bytes* skipped.
+fn try_skip_prefix<'a, 'b>(input: &'a str, prefix: &'b str) -> (bool, &'a str, usize) {
+    if let Some(input_without_prefix) = input.strip_prefix(prefix) {
+        (true, input_without_prefix, prefix.len())
+    } else {
+        (false, input, 0)
+    }
+}
+
+enum NLDigitState {
+    Nothing,
+    LeadingZero,
+    DigitBeforeDot,
+    OnlyDot,
+    DigitsThenDot,
+    DigitAfterDot,
+    Exponent,
+    SignAfterExponent,
+    DigitAfterExponent,
+}
+
+struct NumberLexerState {
+    minus: bool,
+    hex: bool,
+    leading_zeros: usize,
+    digit_state: NLDigitState,
+    uint_suffix: bool,
+}
+
+impl NumberLexerState {
+    // TODO: add proper error reporting, possibly through try_into_token function returning Result
+
+    pub fn is_valid_number(&self) -> bool {
+        match *self {
+            Self {
+                minus: false, // No negative zero for integers.
+                hex,
+                leading_zeros,
+                digit_state: NLDigitState::LeadingZero,
+                ..
+            } => hex || leading_zeros == 1, // No leading zeros allowed in non-hex integers, "0" is always allowed.
+            Self {
+                minus,
+                hex,
+                leading_zeros,
+                digit_state: NLDigitState::DigitBeforeDot,
+                uint_suffix,
+            } => {
+                (hex || leading_zeros == 0) // No leading zeros allowed in non-hex integers.
+                                              // In this state the number has non-zero digits,
+                                              // i.e. it is not just "0".
+                    && (minus ^ uint_suffix) // Either a negative number, or and unsigned integer, not both.
+            }
+            Self {
+                digit_state:
+                    NLDigitState::DigitsThenDot
+                    | NLDigitState::DigitAfterDot
+                    | NLDigitState::DigitAfterExponent,
+                uint_suffix: false, // These states are only valid for floating point numbers.
+                ..
+            } => true,
+            _ => false,
+        }
+    }
+
+    pub fn is_float(&self) -> bool {
+        match self {
+            Self {
+                digit_state:
+                    NLDigitState::DigitsThenDot
+                    | NLDigitState::DigitAfterDot
+                    | NLDigitState::DigitAfterExponent,
+                ..
+            } => true,
+            _ => false,
+        }
+    }
+}
+
 fn consume_number(input: &str) -> (Token, &str) {
-    //Note: I wish this function was simpler and faster...
-    let mut is_first_char = true;
-    let mut right_after_exponent = false;
+    let (minus, working_substr, minus_offset) = try_skip_prefix(input, "-");
+
+    let (hex, working_substr, hex_offset) = try_skip_prefix(working_substr, "0x");
+
+    let mut state = NumberLexerState {
+        minus,
+        hex,
+        leading_zeros: 0,
+        digit_state: NLDigitState::Nothing,
+        uint_suffix: false,
+    };
 
     let mut what = |c| {
-        if is_first_char {
-            is_first_char = false;
-            c == '-' || ('0'..='9').contains(&c) || c == '.'
-        } else if c == 'e' || c == 'E' {
-            right_after_exponent = true;
-            true
-        } else if right_after_exponent {
-            right_after_exponent = false;
-            ('0'..='9').contains(&c) || c == '-'
-        } else {
-            ('0'..='9').contains(&c) || c == '.'
+        match state {
+            NumberLexerState {
+                hex,
+                digit_state: NLDigitState::Nothing,
+                uint_suffix: false,
+                ..
+            } => match c {
+                '0' => {
+                    state.digit_state = NLDigitState::LeadingZero;
+                    state.leading_zeros += 1;
+                    true
+                }
+                '1'..='9' => {
+                    state.digit_state = NLDigitState::DigitBeforeDot;
+                    true
+                }
+                'a'..='f' | 'A'..='F' if hex => {
+                    state.digit_state = NLDigitState::DigitBeforeDot;
+                    true
+                }
+                '.' => {
+                    state.digit_state = NLDigitState::OnlyDot;
+                    true
+                }
+                _ => false,
+            },
+
+            NumberLexerState {
+                hex,
+                digit_state: NLDigitState::LeadingZero,
+                uint_suffix: false,
+                ..
+            } => match c {
+                '0' => {
+                    state.digit_state = NLDigitState::LeadingZero;
+                    state.leading_zeros += 1;
+                    true
+                }
+                '1'..='9' => {
+                    state.digit_state = NLDigitState::DigitBeforeDot;
+                    true
+                }
+                'a'..='f' | 'A'..='F' if hex => {
+                    state.digit_state = NLDigitState::DigitBeforeDot;
+                    true
+                }
+                '.' => {
+                    state.digit_state = NLDigitState::DigitsThenDot;
+                    true
+                }
+                'e' | 'E' if !hex => {
+                    state.digit_state = NLDigitState::Exponent;
+                    true
+                }
+                'p' | 'P' if hex => {
+                    state.digit_state = NLDigitState::Exponent;
+                    true
+                }
+                'u' => {
+                    // We stay in NLDigitState::LeadingZero.
+                    state.uint_suffix = true;
+                    true
+                }
+                _ => false,
+            },
+
+            NumberLexerState {
+                hex,
+                digit_state: NLDigitState::DigitBeforeDot,
+                uint_suffix: false,
+                ..
+            } => match c {
+                '0'..='9' => {
+                    state.digit_state = NLDigitState::DigitBeforeDot;
+                    true
+                }
+                'a'..='f' | 'A'..='F' if hex => {
+                    state.digit_state = NLDigitState::DigitBeforeDot;
+                    true
+                }
+                '.' => {
+                    state.digit_state = NLDigitState::DigitsThenDot;
+                    true
+                }
+                'e' | 'E' if !hex => {
+                    state.digit_state = NLDigitState::Exponent;
+                    true
+                }
+                'p' | 'P' if hex => {
+                    state.digit_state = NLDigitState::Exponent;
+                    true
+                }
+                'u' => {
+                    // We stay in NLDigitState::LeadingZero.
+                    state.uint_suffix = true;
+                    true
+                }
+                _ => false,
+            },
+
+            NumberLexerState {
+                hex,
+                digit_state: NLDigitState::OnlyDot,
+                uint_suffix: false,
+                ..
+            } => match c {
+                '0'..='9' => {
+                    state.digit_state = NLDigitState::DigitAfterDot;
+                    true
+                }
+                'a'..='f' | 'A'..='F' if hex => {
+                    state.digit_state = NLDigitState::DigitAfterDot;
+                    true
+                }
+                _ => false,
+            },
+
+            NumberLexerState {
+                hex,
+                digit_state: NLDigitState::DigitsThenDot | NLDigitState::DigitAfterDot,
+                uint_suffix: false,
+                ..
+            } => match c {
+                '0'..='9' => {
+                    state.digit_state = NLDigitState::DigitAfterDot;
+                    true
+                }
+                'a'..='f' | 'A'..='F' if hex => {
+                    state.digit_state = NLDigitState::DigitAfterDot;
+                    true
+                }
+                'e' | 'E' if !hex => {
+                    state.digit_state = NLDigitState::Exponent;
+                    true
+                }
+                'p' | 'P' if hex => {
+                    state.digit_state = NLDigitState::Exponent;
+                    true
+                }
+                _ => false,
+            },
+
+            NumberLexerState {
+                digit_state: NLDigitState::Exponent,
+                uint_suffix: false,
+                ..
+            } => match c {
+                '0'..='9' => {
+                    state.digit_state = NLDigitState::DigitAfterExponent;
+                    true
+                }
+                '-' | '+' => {
+                    state.digit_state = NLDigitState::SignAfterExponent;
+                    true
+                }
+                _ => false,
+            },
+
+            NumberLexerState {
+                digit_state: NLDigitState::SignAfterExponent | NLDigitState::DigitAfterExponent,
+                uint_suffix: false,
+                ..
+            } => match c {
+                '0'..='9' => {
+                    state.digit_state = NLDigitState::DigitAfterExponent;
+                    true
+                }
+                _ => false,
+            },
+
+            NumberLexerState {
+                uint_suffix: true, ..
+            } => false, // Scanning is done once we've reached a type suffix.
         }
     };
-    let pos = input.find(|c| !what(c)).unwrap_or_else(|| input.len());
-    let (value, rest) = input.split_at(pos);
 
-    let mut rest_iter = rest.chars();
-    let ty = rest_iter.next().unwrap_or(' ');
-    match ty {
-        'u' | 'i' | 'f' => {
-            let width_end = rest_iter
-                .position(|c| !('0'..='9').contains(&c))
-                .unwrap_or_else(|| rest.len() - 1);
-            let (width, rest) = rest[1..].split_at(width_end);
-            (Token::Number { value, ty, width }, rest)
-        }
-        // default to `i32` or `f32`
-        _ => (
+    let pos = working_substr
+        .find(|c| !what(c))
+        .unwrap_or_else(|| working_substr.len());
+    let (value, rest) = input.split_at(pos + minus_offset + hex_offset);
+
+    // NOTE: This code can use string slicing,
+    //       because number literals are exclusively ASCII.
+    //       This means all relevant characters fit into one byte
+    //       and using string slicing (which slices UTF-8 bytes) works for us.
+
+    if state.is_valid_number() {
+        (
             Token::Number {
-                value,
-                ty: if value.contains('.') { 'f' } else { 'i' },
+                value: if state.uint_suffix {
+                    &value[..value.len() - 1]
+                } else {
+                    value
+                },
+                ty: if state.uint_suffix {
+                    'u'
+                } else if state.is_float() {
+                    'f'
+                } else {
+                    'i'
+                },
                 width: "",
             },
             rest,
-        ),
+        )
+    } else {
+        // TODO: A syntax error can already be recognized here, possibly report it at this stage.
+
+        // Return known incorrect token for now.
+        (
+            Token::Number {
+                value: if state.uint_suffix {
+                    &value[..value.len() - 1]
+                } else {
+                    value
+                },
+                ty: if state.uint_suffix {
+                    'u'
+                } else if state.is_float() {
+                    'f'
+                } else {
+                    'i'
+                },
+                width: "",
+            },
+            rest,
+        )
     }
 }
 
--- a/src/front/wgsl/lexer.rs
+++ b/src/front/wgsl/lexer.rs
@@ -37,7 +37,7 @@ enum NLDigitState {
 }
 
 struct NumberLexerState {
-    minus: bool,
+    _minus: bool,
     hex: bool,
     leading_zeros: usize,
     digit_state: NLDigitState,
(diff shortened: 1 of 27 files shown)


tests/in/operators.wgsl
```diff
@@ -1,3 +1,35 @@
+// Regexes for the literals are taken from the working draft at
```
kvark, line 1: these should just go into `front/wgsl/tests.rs`

src/front/wgsl/lexer.rs
```diff
@@ -13,47 +13,328 @@ fn consume_any(input: &str, what: impl Fn(char) -> bool) -> (&str, &str) {
...
+                width: "",
+            },
+            rest,
+        )
+    } else {
+        // TODO: A syntax error can already be recognized here, possibly report it at this stage.
+
+        // Return known incorrect token for now.
+        (
             Token::Number {
```
kvark, line 321: can we have a helper method for this? looks like it's getting repeated

src/front/wgsl/lexer.rs
```diff
@@ -13,47 +13,328 @@ fn consume_any(input: &str, what: impl Fn(char) -> bool) -> (&str, &str) {
...
+            } => true,
+            _ => false,
+        }
+    }
+
+    pub fn is_float(&self) -> bool {
+        match self {
+            Self {
+                digit_state:
+                    NLDigitState::DigitsThenDot
```
kvark, line 87: fancy pattern matching! I wonder if it breaks our MSRV

kvark (commented):
Looks very elaborate!

src/front/wgsl/lexer.rs
hcsch: Ah yes, I kinda forgot about checking for MSRV compatibility. I just built the code with 1.43 and indeed [or-patterns](https://github.com/rust-lang/rust/issues/54883) and [str_strip](https://github.com/rust-lang/rust/issues/67302) are too new

src/front/wgsl/lexer.rs
hcsch: I could just remove the check for validity at that point in the code and return the Token, incorrect or not. The two branches are indeed the same.

tests/in/operators.wgsl
hcsch: There's also the `test_tokens` test in `front/wgsl/lexer.rs`. I'm not quite sure which one is more fitting for these test cases, but I'd hazard a guess the lexer one might be the more fitting one. Feel free to correct me though

src/front/wgsl/lexer.rs
hcsch: str_strip has a minimum version of 1.45 and or-patterns have 1.53 as the minimum for stable.
str_strip usage I could easily replace, but for removing the or-patterns I'd probably have to switch to if-else-if-else chains,
which may make the code harder to read still.
I assume bumping to 1.53 is not something you wanna do if avoidable, correct @kvark?

hcsch: Actually, it seems I can replace the or-patterns relatively easily after all, the only real downside is having to use match arm guards and giving up the exhaustiveness check this way (requiring the addition of a fallback arm).

hcsch:
I just noticed that I broke the lexing of the width of the integer literal (i.e. the number after a type suffix, e.g. `32` after a `u`), the WGSL spec doesn't really have this concept though (at least for now).
Should I remove the `width` attribute from `Token::Number`, or should I make the lexing code accept that?

kvark:
WGSL has types defined as "xW", where W is the number of bits. We definitely need to parse that.

hcsch:
> WGSL has types defined as "xW", where W is the number of bits. We definitely need to parse that.

I think I might be missing something in the spec. I was looking at https://gpuweb.github.io/gpuweb/wgsl/#literals so far. Could you perhaps point me to the right section in the spec, that describes this part of the language?
As far as I can tell (from reading parts of the naga code) the `width` field describes the width in bytes of a number. Currently the WGSL language only seems to support 4-byte wide numeric scalar types (`i32`, `u32`, `f32`). Am I correct in these assumptions?

kvark:
Ok, yeah, the suffixes are not type names, they just look the same. Parsing a literal can just return the width of 4 (in bytes).

kvark:
Just return an empty string there?

kvark (approved):

hcsch:
With 9ff29df2d41757988a4845331884a2cdaebb56e3 the only tests that still fail, fail because of the parsing code currently not being capable of handling hexadecimal integers and floats. And with c5817f0374f69b8e2243d5b6b88c2a4ad1811b6c the code compiles with Rust 1.43 (at least on my machine).
I'd still like some input for whether or not to keep the now unused syntactical verification code. Other than that I'm pretty happy with the code as it now.

kvark:
> syntactical verification code

which code is this?

hcsch:
> > syntactical verification code
> 
> which code is this?

The following are only needed if the state machine should also check for validity due to stuff like negatives zeros (not allowed), floats with a uint suffix, etc.:
https://github.com/hcsch/naga/blob/c5817f0374f69b8e2243d5b6b88c2a4ad1811b6c/src/front/wgsl/lexer.rs#L41
https://github.com/hcsch/naga/blob/c5817f0374f69b8e2243d5b6b88c2a4ad1811b6c/src/front/wgsl/lexer.rs#L51-L74

kvark:
I don't have a clear idea on where to draw the line in responsibilities between the new state machine and the parser. Keeping this code around isn't bad, since it's fairly isolated. So it seems to me that we can proceed and then see if it can be simplified?

hcsch:
There are still two TODO comments left with regards to the verification code. [One where the verification code could be called](https://github.com/hcsch/naga/blob/c5817f0374f69b8e2243d5b6b88c2a4ad1811b6c/src/front/wgsl/lexer.rs#L295) and [one suggesting that the verification code should probably use `Result`s](https://github.com/hcsch/naga/blob/c5817f0374f69b8e2243d5b6b88c2a4ad1811b6c/src/front/wgsl/lexer.rs#L49). Should I remove them so as to not pollute greps over the code-base or are they alright?

kvark:
No need to remove. Keeping them is fine.

hcsch:
I just took a quick look at the parsing side of things:
Hexadecimal integers are ok, with the small caveat of [`i32::from_str_radix`](https://doc.rust-lang.org/std/primitive.i32.html#method.from_str_radix) and [`u32::from_str_radix`](https://doc.rust-lang.org/std/primitive.u32.html#method.from_str_radix) not supporting the case of a `0x` being between the minus sign and the digits (requiring a string allocation to remove the `0x` while keeping the `-`, manual parsing, other trickery, or the use of a crate that does account for this).
Hex floats will probably require the use of an existing crate or porting of some parsing algorithm from another language. The format doesn't seem too complicated to parse and convert into a value, but then again, I'm kinda careful around anything to do with floats (and their wonderful edge-cases).

hcsch:
> I don't have a clear idea on where to draw the line in responsibilities between the new state machine and the parser. Keeping this code around isn't bad, since it's fairly isolated. So it seems to me that we can proceed and then see if it can be simplified?

How would you like the code to be simplified? I'm not sure the state machine part of the code (the giant match that is) can be simplified much further without accepting a broader or a narrower syntactical definition of the literals.
One avenue of simplification could be to parse the number literals into Rust types in the lexer already (since we already have the structure of the literal available here: hex or not?, decimal digits or not?, etc.). Searching around, some lexers seem to do this (i.e. have the token contain the value, not the string), but that would require returning a `Result<Token, SomeErrorType>` rather than just a `Token`.
This parsing of literals at the lexing stage would match what the lexer does for booleans.

kvark:
Anything block you on this?

hcsch:
> Anything block you on this?

Well I guess I'd like to hear from you how you'd like to see the code structured (see previous comment from me). Personally I'd just keep the lexing code as-is and add some modifications to the parsing code to allow for hex integers. Hex float parsing could be done with something like [hexf-parse](https://crates.io/crates/hexf-parse/0.2.1), though v0.2.1 requires Rust ≥1.45, v0.1.0 is Rust ≥ 1.15. I haven't yet looked at the changes between these versions.
Then there is the String allocation for parsing of negative hexadecimal numbers with `i/u32::from_str_radix`, but I guess that is probably acceptable.

hcsch:
There's also `next_uint_literal` and the like here in the lexer code:
https://github.com/gfx-rs/naga/blob/448ea65940354e9eea27bb8942886e47e117aacd/src/front/wgsl/lexer.rs#L294-L320
These do their own parsing, unaware of hexadecimal representations and should probably replaced with functions in the parsing code. They only seem to be used for integer literals as parameters to attributes like `location`.

hcsch:
I think I'll just write up a commit with the changes I think this needs to allow for parsing & lexing of hex integers (hex floats can wait a bit). If something is off with that, you can just tell me then ^^

kvark:
Yes, that sounds good, thank you!

hcsch:
Mmm... might take me a bit to make a reasonably correct parsing impl for integer literals because of the fun with [non-negative / positive / possibly uint or sint, but no larger than `i32::MAX` attribute values](https://www.w3.org/TR/WGSL/#attributes). Currently naga handles this semi-correctly, parsing with those lexer parsing methods which ignore suffixes (sometimes uint literals are allowed, like for [array element counts](https://www.w3.org/TR/WGSL/#element-count), for attributes it mostly isn't, with the exception of [workgroup_size](https://www.w3.org/TR/WGSL/#attribute-workgroup_size)).
I'm gonna leave value restrictions (with the exception of non-negativity for fields that can be both uint or sint, but no larger than `f32::MAX` and not negative, since I need i32s there) to the validation stage.

hcsch:
Ok, with e6efcfa030c97d7904df170e9784e784fece6cee I've implemented parsing for hexadecimal integers. Now there are only two open points for this PR:
- [ ] Hex float parsing. Do we use a crate for this, implement something of our own for this PR or defer it to a future PR and error on hex floats for now (what my code currently does)?
- [ ] Non-hex negative signed zeros, which aren't allowed by the spec (currently there is a test case for this erroring, but not error is implemented for this yet). Do we drop this restriction or do we error on inputs violating it?

kvark:
For hex parsing, the group hasn't figured out the full story yet - https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+hex+
The main concern here is friction: group decides to parse with one set of restrictions, but the crate disagrees. So we'd either have to duplicate some code, or tightly coordinate with crates authors.
If you know some of the authors, that would help. Do you have a recommendation?

hcsch:
> For hex parsing, the group hasn't figured out the full story yet - https://github.com/gpuweb/gpuweb/issues?q=is%3Aissue+is%3Aopen+hex+
> The main concern here is friction: group decides to parse with one set of restrictions, but the crate disagrees. So we'd either have to duplicate some code, or tightly coordinate with crates authors.
> If you know some of the authors, that would help. Do you have a recommendation?

First of all, thanks for bringing this to my attention, I was not aware of the fact that there is still bikeshedding happening at the time. I don't know any of the authors of the spec, I am in fact rather new to WGSL in general. I just wanted to use wgpu with WGSL shaders ported from my current GLSL shaders (since WGSL and SPIR-V are the two shader langs with first class support in wgpu).

Does naga already make guarantees about the backwards compatibility with regards to parsing and lexing? Since the spec still seems to be in flux, especially with something important like gpuweb/gpuweb#775, this doesn't really seem possible at them moment.
I understand that you worry about friction between the changing spec and the implementation, especially with regards to number literals, since they are a core part of a language (and stuff like `b-5` possibly changing meaning, probably necessitates some breakage of existing shaders).
At the moment it seems to me, that floats are much more a target of friction than hex integers (with the exception of `x` vs `X`, which is an easy fix). And I doubt we can just not support (regular, non-hex) floats and wait till the spec has finalized them :P
Hex float however don't seem all that commonly used in shaders at the moment (to me), but that is more a vague feeling than anything close to a quantitative study on my side.

I think one thing to at least make people aware of the friction / possible changes would be to prominently display the current spec version that the implementation is based on in the `README.md` and similar places. That's not really a solution, but I at least makes it possible for people using naga to see why something might not match the newest spec (until after a future update), or why something changed after an update.

Sadly, I don't think the spec has something like the Redline editions of the IEEE specs (i.e. a page where a fully rendered spec has changed sections marked up accordingly). The next best thing is probably looking at the changes to the sources in the git repo.

I believe my implementation of the lexing is relatively easy to modify in terms of changes similar to accepting the exponent marker case-insensitively. Harder changes would be in the greediness of number parsing or the `-` changing to not being part of a number in some cases (or in general).

The best scenario for naga would probably be an author or somebody working closely with the authors of the spec writing issues / pinging contributors to naga in case of changes. But that person would likely have to know the naga implementation of the spec in order to know how the changes affect naga.

I can't really offer much valuable advice as I'm still rather inexperience in the world of specs, spec implementations and especially implementations of specs in the drafting stage, but I hope my thoughts above are at least somewhat helpful.

kvark:
> I don't know any of the authors of the spec

Oh sorry, I should have been more clear. I meant the author of hex-related Rust libraries. As for WGSL spec, we are actively participating in it, so we are technically authors.

> Does naga already make guarantees about the backwards compatibility with regards to parsing and lexing? 

We want to have semver covering sanely written user code. I.e. releasing a compatible Naga version shouldn't break valid WGSL in the sense of "valid" that it had with previous Naga.
In `master`, we are free to break on each commit if needs to be.

> I think one thing to at least make people aware of the friction / possible changes would be to prominently display the current spec version that the implementation is based on in the README.md and similar places.

That's an interesting proposal! I think it would help in earlier stages. These days, we have just a few breaking WGSL changes ahead, and it should be mostly stable, so it seems to me that we can get away without tracking the spec revision exactly. If you still want to have it in the readme, that's fine by me as well :)

hcsch:
> > I don't know any of the authors of the spec
> 
> Oh sorry, I should have been more clear. I meant the author of hex-related Rust libraries. As for WGSL spec, we are actively participating in it, so we are technically authors.

Ah ok, makes sense that you are also involved in the spec development in that way.
With regards to the hex-related Rust libraries, I also currently don't know any of the authors of them. I had picked https://crates.io/crates/hexf-parse/0.2.1 since it seemed relevant and relatively convenient, but it also seems like it 1) doesn't see too much traffic and 2) might not be all that maintained (possibly due to 1).
I suppose, if we do wanna support hex floats now, the best way might be to to get in contact with the authors of existing crates, or have a hex crate that someone from the naga team has at least some control over (so as to guarantee that it is maintained).
I would be interested in writing a hex float parsing crate, should that be necessary, but I don't have very much experience with high performance parsers and floats do have a lot of edge cases. Should the writing of a new crate be necessary, it probably is best to just port an already written, battle-tested algorithm from another language like C, C++, python or similar (with all of its tradeoffs), but that would need some researching.
I also think that it probably is best to have hex float parsing in a separate crate, since that is likely non-trivial (and other might benefit from it as well).

> > I think one thing to at least make people aware of the friction / possible changes would be to prominently display the current spec version that the implementation is based on in the README.md and similar places.
> 
> That's an interesting proposal! I think it would help in earlier stages. These days, we have just a few breaking WGSL changes ahead, and it should be mostly stable, so it seems to me that we can get away without tracking the spec revision exactly. If you still want to have it in the readme, that's fine by me as well :)

I was also thinking along these lines. If you believe the spec is not changing all that much anymore, then its probably fine to not have the spec draft version prominently displayed. Though should any significant change occur it would probably be good to have a note about it in the `README.md` or a changelog or similar for a while.

kvark:
I'm open to consider what your preference is wrt using or not using a crate, and picking one. My only wish is to make it as minimal as possible.

hcsch:
> I'm open to consider what your preference is wrt using or not using a crate, and picking one. My only wish is to make it as minimal as possible.

Alright, then I'll look around some, see what is already available and then I'll throw together some code to parse hex floats as specified by the spec right now. I'm really not all that sure about the complexity of parsing hex float yet (especially since atm we don't have to support any NaNs and Infs from what I can tell), I might end up just adding a sub-module to the wgsl frontend in naga if it turns out to be simple enough.

hcsch:
I've now implemented parsing of hex floats with the hexf-parse crate. Version 0.2.1 of hexf-parse *is* compatible with 1.43 and the `README.md`/code/documentation/CI has been updated accordingly in the [repo](https://github.com/lifthrasiir/hexf). I also added some fuzzing code that checks for panics and compares with `libc::strtod` which the seemingly very responsive maintainers / collaborators have rather promptly merged :D.

hcsch:
I haven't updated the test output files in `tests/out`, since there are some changes from the master branch that have not yet been applied to the files. It'd probably be good if that was done in a commit separate from this PR, onto which I could then rebase.

kvark:
> I haven't updated the test output files in tests/out, since there are some changes from the master branch that have not yet been applied to the files. It'd probably be good if that was done in a commit separate from this PR, onto which I could then rebase.

This should not happen. I recently found and fixed a case for GLSL outputs. If you rebase, the test outputs should be all matching.

hcsch:
Ah yes, I've rebased onto master now and now the test outputs are up to data again. My changes do affect them at the moment in exactly one way:
My code parses the `f32` literals as `f32`s instead of `f64`s as with the old code. The `f32` are converted to `f64`s afterwards though, since that is what `ScalarValue::Float` uses. This causes something like `0.1` to become `0.10000000149011612` in the output, since the writer uses the doubles to write out the values and those show the imprecision of the `f32`s the value was parsed into.
Should I parse the values as `f64`s again to allow for prettier output or do I keep using `f32`s?
The benefit of parsing as `f32`s is having over-/underflows detected by naga, the downside is the above and that usable precision is lost in the case of a target language with 64-bit floats (though I'm pretty sure most if not all shader langs have `f32`s by default).

There's also the option of casting back down to an `f32` on the back-end side for float output, where appropriate, this shouldn't result in any loss of precision from the original `f32` value and would cause the value to be printed as intended.

kvark:
Those `f64` are internal, and the assumption is that we can always just cast it to a smaller size, like `f32`. So the example of 0.1 seems fine?

hcsch:
Yeah, I mean when interpreted as an `f32` the value written out by the current wgsl back-end will end up being the same. I just wanted to make sure you are aware of the change in the output files. 1a4021cd4571b250742caca645f0f82a2226b9f9 has them as the current code will produce them.

Other than that, I'd say the code is fine now. Do you have any points you'd like me to address still?

src/front/wgsl/lexer.rs
```diff
@@ -13,48 +13,306 @@ fn consume_any(input: &str, what: impl Fn(char) -> bool) -> (&str, &str) {
...
+                }
+                'a'..='f' | 'A'..='F' if hex => {
+                    state.digit_state = NLDigitState::DigitBeforeDot;
+                    true
+                }
+                '.' => {
+                    state.digit_state = NLDigitState::OnlyDot;
+                    true
+                }
+                _ => false,
```
kvark, line 122: nit: would it be easier to just change this to `return false` to avoid adding `true` to each and every handled case?

src/front/wgsl/mod.rs
```diff
@@ -158,9 +188,25 @@ impl<'a> Error<'a> {
                             }
                         }
                         ExpectedToken::Identifier => "identifier".to_string(),
-                        ExpectedToken::Float => "floating point literal".to_string(),
-                        ExpectedToken::Uint => "non-negative integer literal".to_string(),
-                        ExpectedToken::Sint => "integer literal".to_string(),
+                        ExpectedToken::Number { ty, width } => {
+                            let literal_ty_str = ty.map(|ty| match ty {
```
kvark, line 192: can we do `match ty { Some(..) => "...", None => ".." }` instead of the `map`/`unwrap_or`?

src/front/wgsl/mod.rs
```diff
@@ -923,6 +977,226 @@ pub enum Scope {
...
+    let parsed_val = if hex {
+        parse_hexf32(word, false).map_err(BadFloatError::ParseHexfError)
+    } else {
+        word.parse::<f32>().map_err(BadFloatError::ParseFloatError)
+    };
+
+    parsed_val.map_err(|e| Error::BadFloat(span, e))
+}
+
+fn parse_sint_literal<'a>(lexer: &mut Lexer<'a>, width: Bytes) -> Result<i32, Error<'a>> {
```
kvark, line 1040: can we move all of these into a sub-module?

kvark (commented):
Looks great, thank you! Just a few organizational things.

hcsch:
Should I squash some commits together or even squash them all into one after you give your stamp of approval? Some commits like the removal of Rust 1.43 incompatible code are prime candidates for squashing, including the cleanup commits just now (which I mostly did so you can get a better idea of what I changed).

kvark (approved):

tests/out/wgsl/operators.wgsl
```diff
@@ -7,7 +7,7 @@ fn builtins() -> vec4<f32> {
     let s2_: vec4<f32> = select(vec4<f32>(0.0, 0.0, 0.0, 0.0), vec4<f32>(1.0, 1.0, 1.0, 1.0), true);
     let s3_: vec4<f32> = select(vec4<f32>(1.0, 1.0, 1.0, 1.0), vec4<f32>(0.0, 0.0, 0.0, 0.0), vec4<bool>(false, false, false, false));
     let m1_: vec4<f32> = mix(vec4<f32>(0.0, 0.0, 0.0, 0.0), vec4<f32>(1.0, 1.0, 1.0, 1.0), vec4<f32>(0.5, 0.5, 0.5, 0.5));
-    let m2_: vec4<f32> = mix(vec4<f32>(0.0, 0.0, 0.0, 0.0), vec4<f32>(1.0, 1.0, 1.0, 1.0), 0.1);
+    let m2_: vec4<f32> = mix(vec4<f32>(0.0, 0.0, 0.0, 0.0), vec4<f32>(1.0, 1.0, 1.0, 1.0), 0.10000000149011612);
```
kvark, line 10: I'm curious about this bit. This is `f32`, and the backend knows it's `f32`, so it sounds like there is a missing cast in the backend from f64 to f32 before printing it?

kvark (changes requested):

tests/out/wgsl/operators.wgsl
hcsch: Yes, there are two places where a cast would be necessary. I even tried out adding those casts, but then decided not to commit them for now since that makes it somewhat inconsistent with the other back-ends.
Would you like for me to add them to the wgsl backend? For the others I'd first have to check what float types are used, but they should all be `f32`s (or an equivalent type) as well, right?

hcsch: Though if that is the case, then the question is why an `f64` is used in the first place. Does one of the input languages allow for `f64`s? At least for that language (if output in it is supported) I wouldn't be able to add the cast.

kvark: we just picked a biggest size. f64 are supported on some platforms.
Yes, you are right, we'd need this change on all textual backends...

hcsch: Should I add those casts to all textual back-ends then? I'd also add a TODO comment or something like that, since just casting blindly is a potential future footgun in that if an input lang starts supporting `f64`s, then the cast will still be there (unless it is manually removed)

kvark: Wait, I'm confused. This change is about parsing WGSL literals. Why does it affect WGSL outputs?

hcsch: Because previously the `f32` literals were parsed as `f64`s and in my changes I switched to parsing as `f32`s and then casting to `f64` (so over-/underflow errors work).

hcsch: See https://github.com/gfx-rs/naga/blob/93db57c12b4a5eff48bdd00c494efa5ec89567ad/src/front/wgsl/mod.rs#L1120-L1123 for how it is done currently, without my changes. The parse call here uses the type from `ScalarValue::Float`, i.e. `f64`.

kvark (approved):

tests/out/wgsl/operators.wgsl
kvark: thank you for explaining!

PR #1185, opened 2021-08-11
[spv-out]: Use OpCompositeConstruct's schmear behavior.

jimblandy:
In `back::spv::Writer::write_texture_coordinates`, `OpCompositeConstruct` can concatenate scalars and vectors, so when combining coordinates with an array index, there is no need to extract the coordinate vector's components individually: once the index has been converted to the appropriate component type, it can be combined with the coordinates in a single instruction.

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -110,79 +110,82 @@ impl<'w> BlockContext<'w> {
     ) -> Result<Word, Error> {
         let coordinate_id = self.cached[coordinates];
 
-        Ok(if let Some(array_index) = array_index {
-            let coordinate_scalar_type_id =
-                self.get_type_id(LookupType::Local(LocalType::Value {
-                    vector_size: None,
-                    kind: crate::ScalarKind::Float,
-                    width: 4,
-                    pointer_class: None,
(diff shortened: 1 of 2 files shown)


jimblandy:
Here's an example of the kind of change this commit causes:

```diff
              %bounds3 = OpImageQuerySize %v3_int %arr_source
               %limit3 = OpISub %v3_int %bounds3 %v3_one
 
-           %coords_x3 = OpCompositeExtract %int %coords 0
-           %coords_y3 = OpCompositeExtract %int %coords 1
-           %combined3 = OpCompositeConstruct %v3_int %coords_x3 %coords_y3 %index
+           %combined3 = OpCompositeConstruct %v3_int %coords %index
      %clamped_coords3 = OpExtInst %v3_int %glsl450 UMin %combined3 %limit3
               %texel3 = OpImageRead %v4_f32 %source %clamped_coords3
 ```

kvark (approved):
So this is purely a quality improvement of SPV output, right?
Looks good!

jimblandy:
> So this is purely a quality improvement of SPV output, right?

Yes, I believe so.

PR #1188, opened 2021-08-11
[spv-out] Writer::write_texture_coordinates: Fix result type.

### LINKED ISSUE #1186: spv-out: `Writer::write_texture_coordinates` always produces float coordinates, for arrayed images
The SPIR-V `OpImageSample...` instructions require floating-point coordinate vectors, but `OpImageFetch`, `OpImageRead`, and `OpImageWrite` require integer coordinate vectors. However, if an array index is provided, `back::spv::Writer::write_texture_coordinates` always produces a vector whose components are floating point.

The `test/in/image.wgsl` never tries doing an image read, write, or fetch on an arrayed texture, so this isn't covered by the test suite.

jimblandy:
Some SPIR-V texture access instructions take coordinates as integers, others as
floats. The types of coordinates in Naga expressions generally match those in
SPIR-V, but Naga indices for arrayed textures are always integers, whereas
SPIR-V combines coordinates and array indices into a single vector, so indices
need to be cast to match the coordinate component type.

This commit makes `write_texture_coordinates` properly cast array indices to
match the coordinates' component type before combining them all into a single
result vector.

Fixes #1186.

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -92,22 +92,34 @@ impl<'w> BlockContext<'w> {
 
     /// Extend texture coordinates with an array index, if necessary.
     ///
-    /// SPIR-V image read and write instructions take the coordinates of the
-    /// texel to access as a vector. If the image is arrayed, the array index
-    /// must be supplied as the final component of the coordinate vector.
+    /// Whereas [`Expression::ImageLoad`] and [`ImageSample`] treat the array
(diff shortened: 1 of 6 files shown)


kvark (approved):

kvark:
Looks good! Let's make CI happy :)

jimblandy:
D'oh. Fixed doc link (and my local version of the CI).

kvark:
And now we have a merge conflict?

PR #1193, opened 2021-08-13
Various HLSL fixes

kvark:
revealed from vange-rs testing on latest wgpu
The last commit fixes a regression from #1189


kvark:
This is not the code I'm most proud of, but I think it's necessary at this point as we approaching wgpu-0.10.

PR #1195, opened 2021-08-13
[glsl-in] Builtin redefinition and calls fixes

JCapucho:
# Overview

This PR adds the ability to redefine builtins and to support this a more spec compliant way to select the correct function overload in calls.

# Motivation

The glsl spec states in Chapter 8. Built-In Functions:

> User  code  can  replace  built-in  functions  with  their  own  if  they  choose,  by  simply  redeclaring  and
defining  the  same  name  and  argument  list.  Because  built-in  functions  are  in  a  more  outer  scope
than  user  built-in  functions,  doing  this  will  hide  all  built-in  functions  with  the  same  name  as  the
redeclared function.

Testing it with `glslang` the behavior is basically as if all builtin are prototypes included in the beginning of every program (and glslang implements it this way even doing the parsing from what I could find).

# Implementation

To reduce startup time and allocations builtins are "injected" as needed, basically if the user calls/declare a certain function `x` the compiler will check if there has been any attempt to interact with a function named `x`, if not the compiler will check if there are any builtin functions with the name `x`, if there are the compiler will create all prototypes for the builtin otherwise the name is initialized in the lookup and the compilation moves on.

Now that there's more overlap in prototypes, function calling also needed to be improved in accordance with the spec which says:

> After these comparisons are performed, each pair of matching functions are compared. A function
declaration A is considered a better match than function declaration B if
> - for  at  least  one  function  argument,  the  conversion  for  that  argument  in  A  is  better  than  the
corresponding conversion in B; and
> - there  is  no  function  argument  for  which  the  conversion  in  B  is  better  than  the  corresponding
conversion in A

# Benefits

- Spec compliance
- Type checking for arguments
- Implicit conversions handled in function call

# Downsides

- Significantly more code

# Unresolved problems

- ~~Since some prototypes have double arguments, a double type is added to the arena even if it isn't used, this causes issues with spirv where doubles are considered a capability and the user must enable them.~~ (Now it only adds builtins with double arguments if needed)


src/front/glsl/builtins.rs
```diff
@@ -0,0 +1,1650 @@
...
+    Error, ErrorKind, Parser, Result, SourceMetadata,
+};
+use crate::{
+    BinaryOperator, Block, Expression, Handle, ImageClass, ImageDimension, ImageQuery,
+    MathFunction, Module, RelationalFunction, SampleLevel, Span, Type, TypeInner, VectorSize,
+};
+
+/// Helper function, to create a function prototype for a builtin
+fn add_builtin(
+    module: &mut Module,
```
kvark, line 13: nit: you can have it on `impl crate::Module {...}` for convenience

src/front/glsl/builtins.rs
```diff
@@ -0,0 +1,1650 @@
...
+                                kind: Sk::Float,
+                                width: 4,
+                            },
+                            None => TypeInner::Scalar {
+                                kind: Sk::Float,
+                                width: 4,
+                            },
+                        }],
+                        Builtins::MathFunction(match name.as_str() {
+                            "sin" => MathFunction::Sin,
```
kvark, line 546: these matches, aren't we already doing them elsewhere?

src/front/glsl/builtins.rs
```diff
@@ -0,0 +1,1650 @@
...
+        void: false,
+    }
+}
+
+impl Parser {
+    /// Injects a builtin into the [`lookup_function`](Self::lookup_function)
+    ///
+    /// This is done to not add a large startup cost and not increase memory
+    /// usage if it isn't needed.
+    pub(crate) fn inject_builtin(&mut self, name: String) {
```
kvark, line 48: this isn't used anywhere?

src/front/glsl/builtins.rs
```diff
@@ -0,0 +1,1650 @@
...
+            }
+            // Not a builtin (or at lest not yet handled)
+            _ => {}
+        }
+    }
+}
+
+/// A compiler defined builtin function
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum Builtins {
```
kvark, line 1250: the name should be singular (`Builtin`) or better something that doesn't conflict with IR names

src/front/glsl/builtins.rs
JCapucho: We were doing them in `functions.rs` but I moved them here.

src/front/glsl/builtins.rs
JCapucho: It's used in `function_call`, `add_prototype` and `add_function` otherwise the tests wouldn't pass.

kvark (approved):

PR #1214, opened 2021-08-16
[wgsl-out] Write pointers types and loads

### LINKED ISSUE #1211: [glsl-in] [wgsl-out] GLSL inout parameters not correct compiled to WGSL
Not sure if this is glsl-in or wgsl-out but:

GLSL Input:

```glsl
#version 440 core

struct Mat4x3 { vec4 mx; vec4 my; vec4 mz; };
void Fma(inout Mat4x3 d, Mat4x3 m, float s) { d.mx += m.mx * s; d.my += m.my * s; d.mz += m.mz * s; }
```

WGSL Output:

```wgsl
struct Mat4x3_ {
    mx: vec4<f32>;
    my: vec4<f32>;
    mz: vec4<f32>;
};

fn Fma(d: Mat4x3_, m: Mat4x3_, s: f32) {
    var m1: Mat4x3_;
    var s1: f32;

    m1 = m;
    s1 = s;
    let _e6: Mat4x3_ = d;
    let _e8: Mat4x3_ = m1;
    let _e10: f32 = s1;
    d.mx = (_e6.mx + (_e8.mx * _e10));
    let _e14: Mat4x3_ = d;
    let _e16: Mat4x3_ = m1;
    let _e18: f32 = s1;
    d.my = (_e14.my + (_e16.my * _e18));
    let _e22: Mat4x3_ = d;
    let _e24: Mat4x3_ = m1;
    let _e26: f32 = s1;
    d.mz = (_e22.mz + (_e24.mz * _e26));
    return;
}
```

Tint doesn't like this:
```
Tint WGSL reader failure:
Parser: 143:7 error: cannot assign to value of type 'vec4<f32>'
    d.mx = (_e6.mx + (_e8.mx * _e10));
```

I believe d should be a pointer. Not 100% sure.

JCapucho:
Closes #1211 

Writes pointer types with the new syntax, adds support for the function storage class and writes dereferences where needed.

(Will follow up with dereference support for wgsl-in)

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -6,11 +6,6 @@ use crate::{
 };
 use std::fmt::Write;
 
-// This is a hack: we need to pass a pointer to an atomic,
-// but generally the backend isn't putting "&" in front of every pointer.
-// Some more general handling of pointers is needed to be implemented here.
-const ATOMIC_REFERENCE: &str = "&";
-
 /// Shorthand result used internally by the backend
 type BackendResult = Result<(), Error>;
 
@@ -537,9 +532,16 @@ impl<W: Write> Writer<W> {
                 )?;
             }
             TypeInner::Pointer { base, class } => {
+                let (storage, maybe_access) = storage_class_str(class);
+                if let Some(class) = storage {
+                    write!(self.out, "ptr<{}, ", class)?;
+                    if let Some(access) = maybe_access {
+                        write!(self.out, ", {}", access)?;
+                    }
+                }
                 self.write_type(module, base)?;
-                if let Some(storage_class) = storage_class_str(class) {
-                    write!(self.out, "<{}>", storage_class)?;
+                if storage.is_some() {
+                    write!(self.out, ">")?;
                 }
             }
             _ => {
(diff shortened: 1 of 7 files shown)


src/back/wgsl/writer.rs
```diff
@@ -973,10 +980,30 @@ impl<W: Write> Writer<W> {
...
-
                 let base_ty_res = &func_ctx.info[base].ty;
                 let mut resolved = base_ty_res.inner_with(&module.types);
+
+                let deref = match *resolved {
+                    TypeInner::Pointer { .. } => {
+                        match func_ctx.expressions[base] {
+                            // Variables have a pointer return type but shouldn't be dereferenced
+                            Expression::LocalVariable(_) | Expression::GlobalVariable(_)
+                            // Access chains might have pointer types but glsl considers them as references
```
kvark, line 991: this mentions `glsl`, but we are in the WGSL backend, it shouldn't be influenced by GLSL

src/back/wgsl/writer.rs
```diff
@@ -1169,15 +1196,28 @@ impl<W: Write> Writer<W> {
...
+                        crate::TypeInner::Pointer { base, .. } => (
+                            true,
+                            match module.types[base].inner {
+                                crate::TypeInner::Atomic { .. } => true,
+                                _ => false,
+                            },
+                        ),
+                        _ => (false, false),
+                    };
+                is_pointer &= match func_ctx.expressions[pointer] {
```
kvark, line 1210: can we have this piece of logic moved out to something
```rust
impl crate::Expression {
  fn is_pointer(&self) -> bool {..}
}
```

src/back/wgsl/writer.rs
```diff
@@ -1169,15 +1196,28 @@ impl<W: Write> Writer<W> {
...
+                    // Variables have a pointer return type but shouldn't be dereferenced
+                    Expression::LocalVariable(_) | Expression::GlobalVariable(_)
+                    // Access chains might have pointer types but glsl considers them as references
+                    | Expression::AccessIndex {..} | Expression::Access {..} => false,
+                    _ => true,
                 };
                 if is_atomic {
                     write!(self.out, "atomicLoad({}", ATOMIC_REFERENCE)?;
```
kvark, line 1218: do we still need `ATOMIC_REFERENCE`, given that your change fixes the pointers?

JCapucho: The wgsl spec doesn't state it clearly but I think that global variables return a `ref` instead of a `ptr` so we would still need this, but it's really hard to figure out what the spec is saying.

kvark: Here is how it's defined: referencing a global is indeed a `ref`. But atomic functions work on pointers, and from IR point of view the argument is a pointer. So something here needs to be checking for `should_deref`, I suppose

kvark:
That looks right, just need to cover a few more spaces (as indicated by CI failures)

kvark (approved):

PR #1223, opened 2021-08-18
Fix WGSL vector splat constructor's type handling

### LINKED ISSUE #1222: [wgsl-in] Vector splat constructur precondition not checked
The following code is incorrectly accepted, as `0` is an expression that has the type `i32`, which does not match the `T` of the `vec2<f32>`:
```wgsl
fn main() {
    let z = vec2<f32>(0);
}
```

Another even more interesting, but probably highly related issue is the the type of the variable somehow being affected by this:
```wgsl
fn main() {
    let z = vec2<f32>(0);
    let z1: vec2<f32> = z;
}
```
This code results in the following error:
```wgsl
[2021-08-18T04:57:33Z ERROR naga::front::wgsl] Given type Vector { size: Bi, kind: Float, width: 4 } doesn't match expected Vector { size: Bi, kind: Sint, width: 4 }
error: the type of `z1` is expected to be [1]
  ┌─ wgsl:3:9
  │
3 │     let z1: vec2<f32> = z;
  │         ^^ definition of `z1`

Could not parse WGSL
```
This clearly indicates that the variable `z` has the inferred type `vec2<i32>`, which is very wrong.

hcsch:
Previously the constructor just used the type of the scalar argument
regardless of whether or not it actually matched the vector splat
constructors target type. This resulted in e.g. `vec2<f32>(0)`
erroneously getting the type of a bi-vector of Sints with width 4.

Now the splat constructor checks that the kind and width of the scalar
argument is the same as that of the target specified in the
constructor's type parameter.

For example, this code
```wgsl
fn main() {
    let z = vec2<f32>(0);
    let z_norm_sq = dot(z, z);
}
```
now results in the following error:
```
error: cannot cast a i32 to a vec2<f32>
  ┌─ wgsl:2:22
  │
2 │     let z = vec2<f32>(0);
  │                      ^^^ cannot cast a i32 to a vec2<f32>

Could not parse WGSL
```

Fixes #1222.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -1746,12 +1746,19 @@ impl Parser {
                 ty_resolution.inner_with(ctx.types),
                 ctx.typifier.get(last_component, ctx.types),
             ) {
-                (&crate::TypeInner::Vector { size, .. }, &crate::TypeInner::Scalar { .. }) => {
-                    crate::Expression::Splat {
-                        size,
-                        value: last_component,
-                    }
-                }
+                (
+                    &crate::TypeInner::Vector {
+                        size, kind, width, ..
+                    },
+                    &crate::TypeInner::Scalar {
+                        kind: arg_kind,
+                        width: arg_width,
+                        ..
+                    },
+                ) if arg_kind == kind && arg_width == width => crate::Expression::Splat {
+                    size,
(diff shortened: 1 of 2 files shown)


kvark (approved):
Very nice, thank you!

PR #1227, opened 2021-08-18
[wgsl-in] Add support for the dereference operator

### LINKED ISSUE #1130: [wgsl-in] there is no indirection/dereference operator
The spec [defines](https://gpuweb.github.io/gpuweb/wgsl/#indirection-expr) the familiar syntax `*p` to convert a pointer to a reference, but naga does not support it:

```
fn deref(p: ptr<storage, u32>) {
	let x = *p;
}
```

```
error: expected expression, found '*'
  ┌─ wgsl:2:10
  │
2 │     let x = *p;
  │             ^ expected expression
```

JCapucho:
Closes #1130

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -668,6 +668,7 @@ impl<'a, 'temp> StatementContext<'a, 'temp, '_> {
             arguments: self.arguments,
             block,
             emitter,
+            rhs: true,
         }
     }
 }
@@ -689,6 +690,9 @@ struct ExpressionContext<'input, 'temp, 'out> {
     functions: &'out Arena<crate::Function>,
     block: &'temp mut crate::Block,
     emitter: &'temp mut super::Emitter,
+    /// Wether or not the current expression is in the right hand side of an
+    /// assignemnt statement or equivalent
+    rhs: bool,
 }
 
 impl<'a> ExpressionContext<'a, '_, '_> {
@@ -705,7 +709,20 @@ impl<'a> ExpressionContext<'a, '_, '_> {
             arguments: self.arguments,
             block: self.block,
             emitter: self.emitter,
+            rhs: self.rhs,
         }
+    }
+
+    fn reborrow_rhs(&mut self) -> ExpressionContext<'a, '_, '_> {
+        let mut reborrow = self.reborrow();
+        reborrow.rhs = true;
+        reborrow
+    }
+
+    fn reborrow_lhs(&mut self) -> ExpressionContext<'a, '_, '_> {
+        let mut reborrow = self.reborrow();
+        reborrow.rhs = false;
+        reborrow
     }
 
     fn resolve_type(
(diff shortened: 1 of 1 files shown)


src/front/wgsl/mod.rs
```diff
@@ -2219,6 +2226,7 @@ impl Parser {
         &mut self,
         lexer: &mut Lexer<'a>,
         mut context: ExpressionContext<'a, '_, '_>,
+        rhs: bool,
```
kvark, line 2229: Boolean arguments are an anti-pattern. Let's make this an enum, please?

src/front/wgsl/mod.rs
```diff
@@ -705,9 +709,22 @@ impl<'a> ExpressionContext<'a, '_, '_> {
...
         }
     }
 
+    fn reborrow_rhs(&mut self) -> ExpressionContext<'a, '_, '_> {
+        let mut reborrow = self.reborrow();
+        reborrow.rhs = true;
+        reborrow
+    }
+
+    fn reborrow_lhs(&mut self) -> ExpressionContext<'a, '_, '_> {
```
kvark, line 722: I feel rather strongly that we shouldn't have all 3 of `reborrow()`, `reborrow_rhs()` and `reborrow_lhs()`. First, because the old `reborrow()` becomes rather exceptional. It's almost never that we really want to inherit this flag.
Second, because most of reborrows should be "rhs", so it should be the default.

How do you feel about:
  - `reborrow()` having the RHS semantics. Most of the usage stays unchanged therefore.
  - `reborrow_lhs` having the LHS semantics.
  - `reborrow_lhs_inherit` for inheriting the LHS from the parent, if any

kvark (commented):
Eventually, we need to implement "the load rule" as it's spelled in WGSL spec. So basically, we'd keep stuff as a "reference" until a certain point, where we emit the load. Right now, even with your PR, we aren't properly implementing this load rule. For example, passing stuff by pointer into some function wouldn't work correctly.

src/front/wgsl/mod.rs
JCapucho: For me it's fine, I'll change it

JCapucho: Actually, after trying to implement it I think this adds more confusion and is more error prone for example let's say I have this snippet

```rust
fn test(p: ptr<function, Struct>) {
   (*p).field = 1;
}
```

This is valid wgsl and for it to work in `(*p).field` it starts as lhs trough `parse_singular_expression` it then must go trough `parse_primary_expression` still as lhs, as it's a expression in parentheses it will go trough `parse_general_expression` it then goes trough 2 more calls until reaching `parse_singular_expression` where we finally see the deref, in all of this the lhs must be kept in context, it's more error prone in my opinion to know where it must be inherited than where it must cleared.

JCapucho:
I don't see why it wouldn't work for functions

src/front/wgsl/mod.rs
kvark: there is a a very limited set of things that need to inherit the LHS status, it's precisely the "postfix" things and nothing else.

kvark (approved):

PR #1242, opened 2021-08-19
[spv-out] Request Sample1D capability when texture_1d is used.

### LINKED ISSUE #1241: SPIR-V back end doesn't add `Image1D` capability when it ought
The following input generates SPIR-V that spirv-val doesn't like:

```
[[group(0), binding(0)]]
var image_1d: texture_1d<f32>;

fn test_textureLoad_1d(coords: i32, level: i32) -> vec4<f32> {
   return textureLoad(image_1d, coords, level);
}
```

`spirv-val` says:

```
error: line 9: Operand 3 of TypeImage requires one of these capabilities: Sampled1D Image1D 
  %3 = OpTypeImage %float 1D 0 0 0 1 Unknown
```

jimblandy:
Fixes #1241.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -703,7 +703,12 @@ impl Writer {
                 class,
             } => {
                 let kind = match class {
-                    crate::ImageClass::Sampled { kind, multi: _ } => kind,
+                    crate::ImageClass::Sampled { kind, multi: _ } => {
+                        if dim == crate::ImageDimension::D1 {
+                            self.check(&[spirv::Capability::Sampled1D])?;
+                        }
+                        kind
+                    }
                     crate::ImageClass::Depth { multi: _ } => crate::ScalarKind::Float,
                     crate::ImageClass::Storage { format, .. } => {
                         let required_caps: &[_] = match dim {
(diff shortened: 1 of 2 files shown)


tests/in/image-1d.wgsl
```diff
@@ -0,0 +1,13 @@
+// Check that we request the `Sample1D` SPIR-V capability if we use `texture_1d`.
```
kvark, line 1: Do we need to create a separate test file for each small feature being tested? It would be easier to navigate the tests if you can at least scroll through them.
Perhaps, we could just add an image to `image.wgsl` or `globals.wgsl` instead?

jimblandy: In general, yes, I'd love to gather tests into coherent groups. But the comment just below this explains why this specific test can't be combined with others.

jimblandy: For example, `image.wgsl` does include a `texture_1d`. But it also includes a `texture_storage_1d`, which is why it didn't catch the problem.

kvark: how about globals.wgsl then?

kvark (approved):

kvark (approved):

PR #1247, opened 2021-08-19
Texture bounds checks

jimblandy:
Here's the work-in-progress for bounds checks on `ImageLoad` and `ImageStore` for the SPIR-V back end.

Status: all three `IndexBoundsCheckPolicy` variants, for all texture types:
- [X] for `ImageLoad`
- [x] for `ImageStore`
- [x] Address review comments and squash

### DIFF
--- a/src/back/spv/image.rs
+++ b/src/back/spv/image.rs
@@ -1,0 +1,553 @@
+//! Generating SPIR-V for image operations.
+
+use super::{Block, BlockContext, Error, Instruction, LocalType, LookupType};
+use crate::arena::Handle;
+use spirv::Word;
+
+impl<'w> BlockContext<'w> {
+    /// Extend image coordinates with an array index, if necessary.
+    ///
+    /// Whereas [`Expression::ImageLoad`] and [`ImageSample`] treat the array
+    /// index as a separate operand from the coordinates, SPIR-V image access
+    /// instructions include the array index in the `coordinates` operand. This
+    /// function builds a SPIR-V coordinate vector from a Naga coordinate vector
+    /// and array index.
+    ///
+    /// If `array_index` is `Some(expr)`, then this function constructs a new
+    /// vector that is `coordinates` with `array_index` concatenated onto the
+    /// end: a `vec2` becomes a `vec3`, a scalar becomes a `vec2`, and so on.
+    ///
+    /// Naga's `ImageLoad` and SPIR-V's `OpImageRead`, `OpImageFetch`, and
+    /// `OpImageWrite` all use integer coordinates, while Naga's `ImageSample`
+    /// and SPIR-V's `OpImageSample...` instructions all take floating-point
+    /// coordinate vectors. The array index, always an integer scalar, may need
+    /// to be converted to match the component type of `coordinates`.
+    ///
+    /// If `array_index` is `None`, this function simply returns the id for
+    /// `coordinates`.
+    ///
+    /// [`Expression::ImageLoad`]: crate::Expression::ImageLoad
+    /// [`ImageSample`]: crate::Expression::ImageSample
+    fn write_image_coordinates(
+        &mut self,
+        coordinates: Handle<crate::Expression>,
+        array_index: Option<Handle<crate::Expression>>,
+        block: &mut Block,
+    ) -> Result<Word, Error> {
+        use crate::TypeInner as Ti;
+        use crate::VectorSize as Vs;
+
+        let coordinate_id = self.cached[coordinates];
+
+        // If there's no array index, the image coordinates are exactly the
+        // `coordinate` field of the `Expression::ImageLoad`. No work is needed.
+        let array_index = match array_index {
+            None => return Ok(coordinate_id),
+            Some(ix) => ix,
+        };
+
+        // Find the component type of `coordinates`, and figure out the size the
+        // combined coordinate vector will have.
+        let (component_kind, result_size) = match *self.fun_info[coordinates]
+            .ty
+            .inner_with(&self.ir_module.types)
+        {
+            Ti::Scalar { kind, width: 4 } => (kind, Vs::Bi),
+            Ti::Vector {
+                kind,
+                width: 4,
+                size: Vs::Bi,
+            } => (kind, Vs::Tri),
+            Ti::Vector {
+                kind,
+                width: 4,
+                size: Vs::Tri,
+            } => (kind, Vs::Quad),
+            Ti::Vector { size: Vs::Quad, .. } => {
+                return Err(Error::Validation("extending vec4 coordinate"));
+            }
+            ref other => {
+                log::error!("wrong coordinate type {:?}", other);
+                return Err(Error::Validation("coordinate type"));
+            }
+        };
+
+        // Convert the index to the coordinate component type, if necessary.
+        let array_index_i32_id = self.cached[array_index];
+        let reconciled_array_index_id = if component_kind == crate::ScalarKind::Sint {
+            array_index_i32_id
+        } else {
+            let component_type_id = self.get_type_id(LookupType::Local(LocalType::Value {
+                vector_size: None,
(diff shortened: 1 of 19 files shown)


jimblandy:
(I'll definitely be squashing some of these commits before promoting this from 'draft'.)

src/back/spv/selection.rs
```diff
@@ -0,0 +1,250 @@
...
+//! When there is no merged value to produce, you can pass `()` for `merge_type`
+//! and the merge values. In this case the `finish` method returns `()`.
+
+#![allow(dead_code)] // until texture bounds checks
+
+use super::{Block, BlockContext, Instruction};
+use spirv::Word;
+
+/// A private struct recording what we know about the selection construct so far.
+pub(super) struct Selection<M: MergeTuple> {
```
kvark, line 59: In terms of Valid Usage, is it pretty much required that the same `&mut Block` is passed to all of the methods here? If that's true, maybe it makes sense to just borrow it mutably for the duration of `Selection` lifetime?

src/back/spv/image.rs
```diff
@@ -33,36 +60,48 @@ impl<'w> BlockContext<'w> {
...
 
         // Find the component type of `coordinates`, and figure out the size the
         // combined coordinate vector will have.
-        let (component_kind, result_size) = match *self.fun_info[coordinates]
-            .ty
-            .inner_with(&self.ir_module.types)
-        {
-            Ti::Scalar { kind, width: 4 } => (kind, Vs::Bi),
+        let (component_kind, size) = match *inner_ty {
+            Ti::Scalar { kind, width: 4 } => (kind, Some(Vs::Bi)),
```
kvark, line 202: this never returns `None` in the second part

src/back/spv/image.rs
```diff
@@ -133,6 +176,255 @@ impl<'w> BlockContext<'w> {
...
+    }
+
+    /// Write code to restrict coordinates for an image reference.
+    ///
+    /// First, clamp the level of detail or sample index to fall within bounds.
+    /// Then, obtain the image size, possibly using the clamped level of detail.
+    /// Finally, use an unsigned minimum instruction to force all coordinates
+    /// into range.
+    ///
+    /// Return a pair `(COORDS, LEVEL_OR_SAMPLE)`, where `bCOORDS` are a
```
kvark: is "b" in "bCOORDS" accidental here?

src/back/spv/image.rs
```diff
@@ -133,6 +176,255 @@ impl<'w> BlockContext<'w> {
...
+    /// coordinate vector (including the array index, if any), and
+    /// `LEVEL_OR_SAMPLE` is an level of detail or sample index guaranteed, all
```
kvark: "an level"

src/back/spv/image.rs
```diff
@@ -133,6 +176,255 @@ impl<'w> BlockContext<'w> {
...
+            coordinates.type_id,
+            coordinate_limit_id,
+            coordinate_bounds_id,
+            ones,
+        ));
+
+        let restricted_coordinate_id = self.gen_id();
+        block.body.push(Instruction::ext_inst(
+            self.writer.gl450_ext_inst_id,
+            spirv::GLOp::UMin,
```
kvark, line 464: might be worth noting (in a comment) that using unsigned min allows us to handle negative values

src/back/spv/image.rs
```diff
@@ -142,70 +434,74 @@ impl<'w> BlockContext<'w> {
...
+                )
+            },
+        )?;
+
+        let result_id = if result_type_id == access_type_id {
+            // The instruction produced the type we expected. We can use
+            // its result as-is.
+            access_id
+        } else {
+            // For `ImageClass::Depth` images, Vulkan gave us four components,
```
kvark, line 696: nit: should say "SPIR-V" instead of "Vulkan"

tests/in/bounds-check-image-restrict.param.ron
```diff
@@ -0,0 +1,5 @@
+(
+        image_bounds_check_restrict: true,
```
kvark, line 2: some padding is inconsistent here

src/back/spv/image.rs
```diff
@@ -28,6 +28,111 @@ struct TextureCoordinates {
...
+        })
+    }
+
+    /// Write an instruction to access a given texel of this image.
+    ///
+    /// This assumes that `coordinates_id` has already had the image array index,
+    /// if any, folded in, as done by `write_texture_coordinates`.
+    ///
+    /// Return the value id produced by the instruction.
+    fn write(
```
kvark, line 102: alternatively, this could just return `Instruction` and avoid receiving `id_gen` and `block`

tests/in/bounds-check-image-rzsw.wgsl
```diff
@@ -0,0 +1,55 @@
+[[group(0), binding(0)]]
```
kvark, line 1: A bit unfortunate that this source file is exactly the same. Maybe we need some way of having multiple parameter files separately

kvark (commented):
Amazing work overall. Now I see the reason it's so complicated :)

src/back/spv/selection.rs
```diff
@@ -0,0 +1,250 @@
...
+    ///
+    /// Use `final_values` as the merged values contributed by the current
+    /// block, and transition to the merge block, emitting `OpPhi` instructions
+    /// to produce the merged values. This must be the same shape as the
+    /// `merge_types` argument passed to [`Selection::start`].
+    ///
+    /// Return the SPIR-V ids of the merged values. This value has the same
+    /// shape as the `merge_types` argument passed to `Selection::start`.
+    pub(super) fn finish(self, ctx: &mut BlockContext, block: &mut Block, final_values: M) -> M {
+        match self {
```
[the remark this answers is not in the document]
jimblandy: The idea was to show that I actually used every component of the `Selection` in the non-trivial case. Not important to me, if you think it's distracting I'll cheerfully change it.

src/back/spv/selection.rs
jimblandy: That's a good idea. Valid Usage requires using this in a properly nested way with respect to other control structures, and having `Selection` take ownership of the `&mut Block` would help enforce that. But in between calls to `start`, `if_true`, and `finish`, we need that `&mut Block` for writing the body instructions of the conditional. So `Selection` would need a method for reborrowing the block out of it.

This suggestion also made me realize that `Selection::current_label` would go stale if somebody switched blocks between calls to `Selection` methods - which is certainly necessary in some cases. As it turns out, `current_label` is unnecessary, so I took it out.

src/back/spv/image.rs
jimblandy: Yes, but it's always used in contexts that require an `Option`, and wrapping it early lets us use the structure field shorthand when we make the return value at the bottom of the function.

src/back/spv/image.rs
jimblandy: This ends up needing to work this way in the versions that support stores.

tests/in/bounds-check-image-rzsw.wgsl
jimblandy: That sounds like a good idea. I'll file it as a follow-up issue, and handle it in a separate PR, if that's all right.

jimblandy: Filed as #1284.

jimblandy: (feel free to 'unresolve' if you'd like this handled in this PR)

jimblandy:
Okay, I think this version has the review comments to date addressed.

src/back/spv/selection.rs
jimblandy: This is implemented in the latest push. Let me know what you think.

kvark (approved):

PR #1249, opened 2021-08-19
[glsl-in] Allow dynamic indexing on constant variables

### LINKED ISSUE #1237: [glsl-in] cannot index into const array
The following:

```glsl
#version 450

void main() {
    const vec2 positions[3] = {
        vec2(0, 0.5),
        vec2(-0.5, -0.5),
        vec2(0.5, -0.5)
    };

    gl_Position = vec4(positions[gl_VertexIndex], 0, 1);
}
```

produces the error (naga 0.6.0):

```
    ERROR: Variable cannot be used in LHS position
     --> shaders/basic.vert line 10:24-33
      |     gl_Position = vec4(positions[gl_VertexIndex], 0, 1);
                               ^^^^^^^^^
```

I wouldn't expect this to throw an error since it is allowed by `glslangValidator`. Though, if the `const` is removed, the shader compiles successfully.

The issue seems to originate from [here](https://github.com/gfx-rs/naga/blob/4e181d6af4b758c11482b633b3a54b58bfbd4a59/src/front/glsl/context.rs#L556) where `lhs` is incorrectly set to `true` (the `position` variable is obviously only used in the RHS of the expression).

JCapucho:
Previously we always set the lhs flag when lowering to generate a
pointer so that dynamic indexing would work, this would produce an error
on constant variables since they can't be in a lhs context.

Now we introduce an enum which distinguishes not only between lhs and
rhs but also in array base, if lowering in a lhs context the base is
also lowered in a lhs context but if lowering in rhs the base is lowered
in a special array base context which bypasses the mutability check.

Fixes #1237

### DIFF
--- a/src/front/glsl/context.rs
+++ b/src/front/glsl/context.rs
@@ -15,6 +15,28 @@ use crate::{
 };
 use std::{convert::TryFrom, ops::Index};
 
+/// The position at which an expression is, used while lowering
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum ExprPos {
+    /// The expression is in the left hand side of an assignment
+    Lhs,
+    /// The expression is in the right hand side of an assignment
+    Rhs,
+    /// The expression is an array being indexed, needed to allow constant
+    /// arrays to be dinamically indexed
+    ArrayBase,
+}
+
+impl ExprPos {
+    /// Returns an lhs position if the current position is lhs otherwise ArrayBase
+    fn maybe_array_base(&self) -> Self {
+        match *self {
+            ExprPos::Lhs => *self,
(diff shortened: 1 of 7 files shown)


src/front/glsl/context.rs
```diff
@@ -347,7 +378,7 @@ impl Context {
         lhs: bool,
```
kvark, line 378: let's switch this to the enum?

src/front/glsl/context.rs
```diff
@@ -326,7 +357,7 @@ impl Context {
...
```
kvark, line 357: and this

kvark (approved):

PR #1255, opened 2021-08-20
[wgsl-in] [wgsl-out] Implement bitcast()

### LINKED ISSUE #1229: Are Bitcast Expressions Not Supported in WGSL?
Looking at the [Reinterpretation of Representation Expressions](https://gpuweb.github.io/gpuweb/wgsl/#bitcast-expr) section of the WGSL spec, if I'm interpreting it correctly, it seems like you should be able call `bitcast<f32>(my_i32_value)` to interpret the bits of an `i32` as an `f32`, but it seems that naga doesn't recognize the `bitcast` expression. Am I understanding the spec correctly?

I have a situation where I need to be able to bitcast from i32 to f32 and vice-versa, but I couldn't find a way to do that in WGSL with naga. If this is just something that needs to be implemented I'd be glad to do it.

zicklag:
Resolves #1229.

Implements `bitcast<type>(expression)` for the WGSL frontend and backend.

let me know if I got this right! It works in my project. :)

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -1257,6 +1257,36 @@ impl Parser {
             }
         } else {
             match name {
+                "bitcast" => {
+                    let _ = lexer.next();
+                    lexer.expect_generic_paren('<')?;
+                    let ((ty, _access), type_span) = lexer.capture_span(|lexer| {
+                        self.parse_type_decl(lexer, None, ctx.types, ctx.constants)
+                    })?;
+                    lexer.expect_generic_paren('>')?;
+
+                    lexer.open_arguments()?;
+                    let expr = self.parse_general_expression(lexer, ctx.reborrow())?;
+                    lexer.close_arguments()?;
+
+                    let kind = match ctx.types[ty].inner {
+                        crate::TypeInner::Scalar { kind, .. } => kind,
+                        crate::TypeInner::Vector { kind, .. } => kind,
+                        _ => {
+                            return Err(Error::BadTypeCast {
+                                from_type: format!("{:?}", ctx.resolve_type(expr)?),
+                                span: type_span,
+                                to_type: format!("{:?}", ctx.types[ty].inner),
+                            })
+                        }
+                    };
+
+                    crate::Expression::As {
+                        expr,
+                        kind,
+                        convert: None,
+                    }
+                }
                 "select" => {
                     let _ = lexer.next();
                     lexer.open_arguments()?;
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -1145,7 +1145,11 @@ impl<W: Write> Writer<W> {
                 let name = &self.names[&NameKey::GlobalVariable(handle)];
                 write!(self.out, "{}", name)?;
             }
-            Expression::As { expr, kind, .. } => {
+            Expression::As {
+                expr,
+                kind,
+                convert,
+            } => {
                 let inner = func_ctx.info[expr].ty.inner_with(&module.types);
                 match *inner {
                     TypeInner::Matrix { columns, rows, .. } => {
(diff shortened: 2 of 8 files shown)


src/front/wgsl/mod.rs
```diff
@@ -1257,6 +1257,72 @@ impl Parser {
...
             match name {
+                "bitcast" => {
+                    let _ = lexer.next();
+                    lexer.expect_generic_paren('<')?;
+                    let ((ty, _access), type_span) = lexer.capture_span(|lexer| {
+                        self.parse_type_decl(lexer, None, ctx.types, ctx.constants)
+                    })?;
+                    lexer.expect_generic_paren('>')?;
+
+                    let expr = self.parse_general_expression(lexer, ctx.reborrow())?;
```
JCapucho, line 1268: You need to call `lexer.{open,close}_arguments()` otherwise you'll allow code like
```rust
var a = bitcast<f32>1;
```

JCapucho (changes requested):
There's a small problem in the frontend code

src/front/wgsl/mod.rs
```diff
@@ -1257,6 +1257,72 @@ impl Parser {
...
+
+                    // Create error that will be returned if type matching fails
+                    let err = Err(Error::BadTypeCast {
```
kvark, line 1271: let's not do this, please. This error involves 2 string heap allocations, which we don't want to pay for is the code is good in the first place. Instead you can do something like `let good = match  {.. }; if !good { return Err(...) };`

src/front/wgsl/mod.rs
```diff
@@ -1257,6 +1257,72 @@ impl Parser {
...
+                                expr,
+                                kind,
+                                convert: None,
+                            },
+                            _ => return err,
+                        },
+                        crate::TypeInner::Vector { size, kind, .. } => {
+                            match *ctx.resolve_type(expr)? {
+                                crate::TypeInner::Vector { .. } => {
+                                    let count = match size {
```
kvark, line 1289: you can just do `size as u32`

src/front/wgsl/mod.rs
```diff
@@ -1257,6 +1257,72 @@ impl Parser {
...
+                    // Create error that will be returned if type matching fails
+                    let err = Err(Error::BadTypeCast {
+                        from_type: format!("{:?}", ctx.resolve_type(expr)?),
+                        span: type_span,
+                        to_type: format!("{:?}", ctx.types[ty].inner),
+                    });
+
+                    match ctx.types[ty].inner {
+                        crate::TypeInner::Scalar { kind, .. } => match *ctx.resolve_type(expr)? {
+                            crate::TypeInner::Scalar { .. } => crate::Expression::As {
```
kvark, line 1279: We don't need to validate here at all, leave it to IR validator.
If we know this is `bitcast`, all we really need is the `kind`, which is extracted from the type. There should be no need to match any expression types here

src/front/wgsl/mod.rs
```diff
@@ -1257,6 +1257,72 @@ impl Parser {
...
+                        crate::TypeInner::Vector { size, kind, .. } => {
+                            match *ctx.resolve_type(expr)? {
+                                crate::TypeInner::Vector { .. } => {
+                                    let count = match size {
+                                        crate::VectorSize::Bi => 2,
+                                        crate::VectorSize::Tri => 3,
+                                        crate::VectorSize::Quad => 4,
+                                    };
+
+                                    // For vectors, cast each field individually
```
kvark, line 1295: we shouldn't need this either. Generally we are trying to make WGSL the textual form of our IR. So if WGSL supports this on vectors, we'll need to support it in the IR.

tests/in/bitcast.wgsl
```diff
@@ -0,0 +1,8 @@
+[[stage(vertex)]]
+fn main(
+  [[location(0)]] pos : vec4<i32>) -> [[builtin(position)]] vec4<f32> {
+  let example1: f32 = bitcast<f32>(pos.x);
```
kvark, line 4: let's add it to `operators.wgsl`, which already has `fn builtins()`, instead of adding a new test

kvark (changes requested):

src/front/wgsl/mod.rs
zicklag: OK, cool, that makes it simpler.

You always come at my pull requests with like a PR shrink ray. :smile:

src/front/wgsl/mod.rs
zicklag: After trying it, it actually looks like the IR is already fine with us casting vectors using the `As` expression. It seems like OpenGL needs a fix to properly cast vectors because `intBitsToFloat` doesn't support vectors, but that can be a separate issue/PR.

kvark (approved):
Great, thank you!

PR #1256, opened 2021-08-20
[spv-out] Clean up capability handling.

jimblandy:
Remove 'forbidden_caps'.

Accumulate capabilities actually used separately from the permitted
capabilities, so that the latter can be retained across Writer resets, while the
former is cleared between modules.

### DIFF
--- a/src/back/spv/mod.rs
+++ b/src/back/spv/mod.rs
@@ -55,8 +55,8 @@ const BITS_PER_BYTE: crate::Bytes = 8;
 pub enum Error {
     #[error("target SPIRV-{0}.{1} is not supported")]
     UnsupportedVersion(u8, u8),
-    #[error("one of the required capabilities {0:?} is missing")]
-    MissingCapabilities(Vec<Capability>),
+    #[error("using {0} requires at least one of the capabilities {1:?}, but none are available")]
+    MissingCapabilities(&'static str, Vec<Capability>),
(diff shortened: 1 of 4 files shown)


src/back/spv/mod.rs
```diff
@@ -385,8 +385,18 @@ pub struct Writer {
...
-    forbidden_caps: Option<&'static [Capability]>,
+
+    /// The set of capabilities modules are permitted to use.
+    ///
+    /// This is initialized from `Options::capabilities`.
+    capabilities_available: Option<crate::FastHashSet<Capability>>,
+
+    /// The set of capabilities used by this module.
+    ///
+    /// If `capabilities_available` is `Some`, then this is always a subset of
```
kvark, line 396: what's the motivation for collecting this subset if `capabilities_available` is provided?

kvark (commented):
Looks very nice, just one concern here

src/back/spv/mod.rs
jimblandy: We use it to emit `OpCapability` instructions.

kvark: I don't think it helps in any way versus just passing `capabilities_available` (the superset).

kvark (approved):

kvark:
Published in 0.6.2

PR #1260, opened 2021-08-20
Add support for inverse hyperbolic trignometric functions

JCapucho:
Hlsl and wgsl don't support them directly so a polyfill is used, taken from the msl spec.

`asinh` = `log(x + sqrt(x * x + 1.0))`
`acosh` = `log(x + sqrt(x * x - 1.0))`
`atanh` = `0.5 * log((1.0 + x) / (1.0 – x))`


tests/snapshots.rs
```diff
@@ -499,6 +499,16 @@ fn convert_spv_shadow() {
     convert_spv("shadow", true, Targets::IR | Targets::ANALYSIS);
 }
 
+#[cfg(feature = "spv-in")]
+#[test]
+fn convert_spv_inverse_hyperbolic_trig_functions() {
+    convert_spv(
+        "inv-hyperbolic-trig-functions",
+        true,
+        Targets::HLSL | Targets::WGSL | Targets::IR,
```
kvark, line 508: let's exclude IR from there

src/back/hlsl/writer.rs
```diff
@@ -1635,76 +1635,105 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
-                    //Mf::Inverse =>,
-                    Mf::Transpose => "transpose",
-                    Mf::Determinant => "determinant",
-                    // bits
-                    Mf::CountOneBits => "countbits",
-                    Mf::ReverseBits => "reversebits",
-                    _ => return Err(Error::Unimplemented(format!("write_expr_math {:?}", fun))),
-                };
+                match fun {
+                    Mf::Asinh | Mf::Acosh => {
```
kvark, line 1639: Generally we'd want to avoid double-matching on `fun`, because it's repetitive and error-prone.
Instead, we should do something like this:
```
enum Function {
  Asincosh { is_sin: bool },
  Atanh,
  Regular(&'static str),
}
```

This way, we'd construct it in the first match, and then do a match on the result specifically.

src/back/wgsl/writer.rs
```diff
@@ -1240,75 +1240,101 @@ impl<W: Write> Writer<W> {
...
-                    Mf::Transpose => "transpose",
-                    Mf::Determinant => "determinant",
-                    // bits
-                    Mf::CountOneBits => "countOneBits",
-                    Mf::ReverseBits => "reverseBits",
-                    _ => {
-                        return Err(Error::UnsupportedMathFunction(fun));
+                // NOTE: If https://github.com/gpuweb/gpuweb/issues/1622 ever is
+                // accepted, replace this with the builtin functions
+                match fun {
```
kvark, line 1245: same here

kvark (approved):
Thank you!

PR #1264, opened 2021-08-20
Fix some clippy suggestions

sigaloid:
Passes cargo check and cargo test. 

PS: Running cargo test changes some files in tests/out (`#version 310 es`, seems to upgrade the shader to 310), but I didn't include them here because they are not changed by this PR. I think these should be under the .gitignore, as it's the shader output of the tests? Might be helpful to track the changes of the output shaders over time, though, in which case you can simply clone the repo and run cargo test, it edits these files: 
image

### DIFF
--- a/src/back/spv/layout.rs
+++ b/src/back/spv/layout.rs
@@ -135,10 +135,8 @@ impl Instruction {
             inst_index += 1;
         }
 
-        let mut op_index = 0;
-        for i in inst_index..wc as usize {
+        for (op_index, i) in (inst_index..wc as usize).enumerate() {
             assert_eq!(words[i], self.operands[op_index]);
-            op_index += 1;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/back/spv/layout.rs
```diff
@@ -135,10 +135,8 @@ impl Instruction {
             inst_index += 1;
         }
 
-        let mut op_index = 0;
-        for i in inst_index..wc as usize {
+        for (op_index, i) in (inst_index..wc as usize).enumerate() {
```
kvark, line 138: might as well do `for (operand, word) in self.operands.iter().zip(words[inst_index..wc as usize].iter()) { .. }`

kvark (approved):

src/back/spv/layout.rs
sigaloid: I just tried that however it gives an error about indexing with a &i32. 😔

kvark:
I'm addressing the cargo test GLSL problem shortly....

Issue #1017, opened 2021-06-23
Naga panics translating valid WGSL to SPIR-V


jimblandy:
This may be a duplicate, but filing it just in case: The following WGSL program passes validation, but panics in `back/spv/writer.rs`:

```
fn loop_switch_continue(x: i32) {
   loop {
       switch (x) {
           case 1: {
               continue;
           }
       }
   }
}
```

The failing code is the `unwrap` [here](https://github.com/gfx-rs/naga/blob/64b9e4501576bf5fecdb60c6375b191becbaa992/src/back/spv/writer.rs#L2583):

```
crate::Statement::Continue => {
    block.termination =
        Some(Instruction::branch(loop_context.continuing_id.unwrap()));
}
```

jimblandy:
I think it'll be a challenge to generate SPIR-V for this. SPIR-V §2.11, "Structured Control Flow", includes this rule:

> each header block must strictly dominate its merge block, unless the merge block is unreachable in the CFG

Control flow in shaders is required to be structured, which means that the WGSL `switch` statement gets translated into a SPIR-V `OpSelectionMerge` followed by an `OpSwitch`. The rule quoted above implies that the `continue` cannot simply be translated into an `OpBranch` to the `loop`'s continue target.

I'll see what `glslangValidate` does with similar GLSL code.

jimblandy:
Okay, I was misunderstanding the rule:

Although the header block must strictly dominate the merge block, that does not imply that all paths much reach the merge block, as I had incorrectly inferred. There is no requirement that the merge block post-dominate the header block.

jimblandy:
Given this GLSL:

```
#version 450

int f(int q) { return q + 1; }

void main() {
  int i;
  for (i = 0; i < 10; i++) {
       switch (i & 3) {
           case 1: {
               continue;
           }
       }

       i = f(i);
   }
}
```

we get the following SPIR-V for the body of `main`:

```
         %19 = OpLabel
               OpLoopMerge %21 %22 None
               OpBranch %23

         %23 = OpLabel
         %24 = OpLoad %int %i
         %27 = OpSLessThan %bool %24 %int_10
               OpBranchConditional %27 %20 %21

         %20 = OpLabel
         %28 = OpLoad %int %i
         %30 = OpBitwiseAnd %int %28 %int_3
               OpSelectionMerge %32 None
               OpSwitch %30 %32 1 %31

         %31 = OpLabel
               OpBranch %22

         %32 = OpLabel
[...]

PR #1265, opened 2021-08-20
[hlsl-out] Implement switch statement

### LINKED ISSUE #1261: [hlsl-out] Unimplemented Switch
One of the shaders that is in my dependency tree ([wgpu_glyph shader](https://github.com/hecrj/wgpu_glyph/blob/master/src/shader/glyph.wgsl)) uses a switch statement, which causes a panic because of unimplemented switch writer in hlsl. It's especially problematic because DX12 is the default wgpu backend on Windows.

https://github.com/gfx-rs/naga/blob/dfcb79880f3a42781efc72ae47ea291cbea8efec/src/back/hlsl/writer.rs#L1208

Gordon-F:
Fix #1261

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1441,11 +1441,6 @@ impl<'a, W: Write> Writer<'a, W> {
                     for sta in case.body.iter() {
                         self.write_stmt(sta, ctx, indent + 2)?;
                     }
-
-                    // Write `break;` if the block isn't fallthrough
-                    if !case.fall_through {
-                        writeln!(self.out, "{}break;", INDENT.repeat(indent + 2))?;
-                    }
                 }
 
                 // Only write the default block if the block isn't empty
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -1204,8 +1204,63 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                 self.temp_access_chain = chain;
                 self.named_expressions.insert(result, res_name);
             }
-            Statement::Switch { .. } => {
-                return Err(Error::Unimplemented(format!("write_stmt {:?}", stmt)))
+            Statement::Switch {
+                selector,
+                ref cases,
+                ref default,
+            } => {
+                // Start the switch
+                write!(self.out, "{}", INDENT.repeat(indent))?;
+                write!(self.out, "switch(")?;
+                self.write_expr(module, selector, func_ctx)?;
+                writeln!(self.out, ") {{")?;
+
+                // Write all cases
+                let indent_str_1 = INDENT.repeat(indent + 1);
+                let indent_str_2 = INDENT.repeat(indent + 2);
+
+                for case in cases {
+                    writeln!(self.out, "{}case {}: {{", &indent_str_1, case.value)?;
+
+                    if case.fall_through {
+                        // Generate each fallthrough case statement in a new block. This is done to
+                        // prevent symbol collision of variables declared in these cases statements.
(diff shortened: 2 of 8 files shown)


src/back/hlsl/writer.rs
```diff
@@ -1204,8 +1204,53 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                    writeln!(self.out, "{}case {}: {{", &indent_str_1, case.value)?;
+
+                    for sta in case.body.iter() {
+                        self.write_stmt(module, sta, func_ctx, indent + 2)?;
+                    }
+
+                    if case.fall_through {
+                        writeln!(self.out, "{}/* fallthrough */", &indent_str_2)?;
+                        writeln!(self.out, "{}{{", &indent_str_2)?;
+                        writeln!(self.out, "{}break;", INDENT.repeat(indent + 3))?;
```
kvark, line 1232: I don't understand. This goes into `break` and the other branch goes into `break` too?

kvark (commented):
Thank you!
we need to add `switch` test to `control_flow.wgsl` to see how/if it works

kvark:
We'll need to release a patch after this is merged, for iced

src/back/hlsl/writer.rs
Gordon-F: Sorry, misunderstood `tint` logic in fallthrough case. We should generate it in new block.

tests/snapshots.rs
```diff
@@ -425,7 +425,9 @@ fn convert_wgsl() {
         ),
         (
             "control-flow",
-            Targets::SPIRV | Targets::METAL | Targets::GLSL | Targets::HLSL | Targets::WGSL,
+            // TODO: SPIRV https://github.com/gfx-rs/naga/issues/1017
+            //Targets::SPIRV |
```
Gordon-F, line 429: Unfortunately, `spv-out` panic with the updated test. So, I decided to remove it for a while.

src/back/glsl/mod.rs
```diff
@@ -1441,11 +1441,6 @@ impl<'a, W: Write> Writer<'a, W> {
                     for sta in case.body.iter() {
                         self.write_stmt(sta, ctx, indent + 2)?;
                     }
-
-                    // Write `break;` if the block isn't fallthrough
-                    if !case.fall_through {
-                        writeln!(self.out, "{}break;", INDENT.repeat(indent + 2))?;
```
Gordon-F, line 1447: That's not required, since `return` is present in IR for all cases.

src/back/hlsl/writer.rs
```diff
@@ -1204,8 +1204,63 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                            func_ctx,
+                            indent + 2 + usize::from(case.fall_through),
+                        )?;
+                    }
+
+                    if !case.fall_through {
+                        writeln!(self.out, "{}break;", &indent_str_2)?;
+                    }
+
+                    if case.fall_through {
```
kvark, line 1244: any reason this isn't just `if/else`?

tests/snapshots.rs
kvark: what's the panic exactly?

Gordon-F: I added test case here - https://github.com/gfx-rs/naga/issues/1017
```log
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/back/spv/block.rs:1552:83
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:515:5
```

kvark (approved):

kvark:
Included in v0.6.1

PR #1267, opened 2021-08-20
Fix round in backends and add support for roundEven in frontends

JCapucho:
According to https://github.com/gpuweb/gpuweb/issues/1381 `MathFunction::Round` always goes to the even number when the fractional part is `.5`. But in spirv, glsl and msl the default `round` leaves this case as implementation defined, to fix this in spirv and glsl we now use `roundEven` and in msl `rint`.

### DIFF
--- a/src/front/glsl/builtins.rs
+++ b/src/front/glsl/builtins.rs
@@ -953,8 +953,8 @@ fn inject_common_builtin(
     float_width: crate::Bytes,
 ) {
     match name {
-        "ceil" | "round" | "floor" | "fract" | "trunc" | "sqrt" | "inversesqrt" | "normalize"
-        | "length" | "isinf" | "isnan" => {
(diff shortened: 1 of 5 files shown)


src/front/glsl/builtins.rs
```diff
@@ -980,6 +980,7 @@ fn inject_common_builtin(
                 let fun = match name {
                     "ceil" => MacroCall::MathFunction(MathFunction::Ceil),
                     "round" => MacroCall::MathFunction(MathFunction::Round),
```
kvark, line 982: nit: was it intentional to not use "|" here?

kvark (approved):

src/front/glsl/builtins.rs
JCapucho: No just me being dumb :)

PR #1271, opened 2021-08-21
[glsl-out] Modulo operator on float must be converted to mod function

### LINKED ISSUE #1266: [glsl-out] Operands to "%" must be integral
### Example:
This valid glsl:
```glsl
float out = mod(1.0,1.0);
```
Translates to this wgls
```rs
let out: f32 = 1.0 % 1.0;
```
And this wgls translates to this glsl:
```glsl
float out = 1.0 % 1.0;
```
The resulting shader does not work.

When run with wgpu it outputs:
`Internal error in VERTEX shader: 0(27) : error C1021: operands to "%" must be integral`

JCapucho:
Fixes #1266


kvark (approved):

kvark:
Included in v0.6.1

PR #1273, opened 2021-08-21
[spv-out] Quick tests for capability requests.

jimblandy:
Add a new test, `spirv-capabilities.rs`, for quickly assessing which SPIR-V features a small code fragment requires.

Also, fix the processing of cube array images:

- Using 1D images should require either the `Sampled1D` or `Image1D` capability.

- Using cube array images should require either the `SampledCubeArray` or
`ImageCubeArray` capability.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1310,6 +1310,11 @@ impl Writer {
         self.logical_layout.in_words(words);
         Ok(())
     }
+
+    /// Return the set of capabilities the last module written used.
+    pub fn get_capabilities_used(&self) -> &crate::FastHashSet<spirv::Capability> {
+        &self.capabilities_used
+    }
 }
 
 #[test]
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -719,35 +719,10 @@ impl Writer {
                 arrayed,
                 class,
             } => {
-                let kind = match class {
-                    crate::ImageClass::Sampled { kind, multi: _ } => {
-                        if dim == crate::ImageDimension::D1 {
-                            self.require_any("1D sampled images", &[spirv::Capability::Sampled1D])?;
-                        }
-                        kind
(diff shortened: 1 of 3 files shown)


jimblandy:
I guess `#[track_caller]` wasn't stabilized until 1.46.

src/back/spv/writer.rs
```diff
@@ -754,6 +729,29 @@ impl Writer {
...
+                        if sampled {
+                            self.require_any("sampled 1D images", &[spirv::Capability::Sampled1D])?;
+                        } else {
+                            self.require_any("1D storage images", &[spirv::Capability::Image1D])?;
+                        }
+                    }
+                    spirv::Dim::DimCube if arrayed => {
+                        if sampled {
+                            self.require_any(
+                                "sampled 1D images",
```
kvark, line 743: says "1D" here?

src/back/spv/writer.rs
```diff
@@ -754,6 +729,29 @@ impl Writer {
...
+                                &[spirv::Capability::SampledCubeArray],
+                            )?;
+                        } else {
+                            self.require_any(
+                                "1D storage images",
```
kvark, line 748: and here?

src/back/spv/writer.rs
```diff
@@ -1311,6 +1309,11 @@ impl Writer {
         self.logical_layout.in_words(words);
         Ok(())
     }
+
+    /// Return the set of capabilities the last module written used.
+    pub fn get_capabilities_used(&self) -> crate::FastHashSet<spirv::Capability> {
```
kvark, line 1314: why is this not returning a reference? if the client wants to clone, they can do it

kvark (approved):

src/back/spv/writer.rs
jimblandy: why is there no D'oh! emoji

PR #1276, opened 2021-08-22
HLSL: rewrite handling of interface matching rules

kvark:
This PR largerly rewrites the logic responsible to match input/output interfaces for HLSL. It's a follow-up to #1189
The new code is more isolated and robust. It works on structures that are used as both host-shared and entry point i/o.


kvark:
Included in v0.6.1

PR #1279, opened 2021-08-22
[glsl-in] Add support for derivative functions

### LINKED ISSUE #1234: [glsl-in] dFdx and dFdy are currently unsupported
naga 0.6 does not seem to support the built-in dFdx and dFdy functions (see https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/dFdx.xhtml).

JCapucho:
Fixes #1234


kvark (approved):
nice!

Issue #1277, opened 2021-08-22
[glsl-in] Infinite loop when first token of a statement is invalid.


fsh:
Something like the following hits an infinite loop in `parse_compound_statement()` because `parse_statement()` silently ignores invalid start tokens.
```
void main () {
 /
}
```

I fixed this locally by replacing the default case with
```
          _ => {
            let Token { value, meta } = self.bump(parser)?;
            return Err(Error {
              kind: ErrorKind::InvalidToken(
                value,
                vec![
                  TokenValue::RightBrace.into(),
                  // TODO: add all possible statement start tokens here? (has to be a better way)
                ],
              ),
              meta,
            });
          }
 ```

But since I'm not familiar with the rest of the parser I don't know if it's correct, i.e. are there valid cases where parse_statement() has to be a no-op or act like a silent peek? Any comment on that, or should I submit a PR?

JCapucho:
If you can submit a PR it would be helpful :)

JCapucho:
This was fixed in #1280

Issue #1232, opened 2021-08-19
[glsl-in] Prefix increment/decrement trigger infinite loop in parser


StephanvanSchaik:
When using the prefix increment or decrement operator in GLSL and using naga 0.6 to parse the GLSL shader, the parse function seems to end up getting stuck in an infinite loop.

Here are some minimal test cases to trigger this exact behaviour:

```
#version 450

void main() {
    uint index = 0;
    
    ++index;
}
```

```
#version 450

void main() {
    uint index = 0;
    
    --index;
}
```

Using the postfix operators: `index++;` and `index--;` seems to work fine. Similarly, you can also use `index += 1;` and `index -= 1;` without issues.

PR #1280, opened 2021-08-22
[glsl-in] Avoid infinite loop when parsing invalid statements

fsh:
Tentative fix for https://github.com/gfx-rs/naga/issues/1277

### DIFF
--- a/src/front/glsl/parser/functions.rs
+++ b/src/front/glsl/parser/functions.rs
@@ -493,34 +493,15 @@ impl<'source> ParsingContext<'source> {
 
                 meta
             }
-            TokenValue::Plus
-            | TokenValue::Dash
-            | TokenValue::Bang
-            | TokenValue::Tilde
-            | TokenValue::LeftParen
-            | TokenValue::Increment
-            | TokenValue::Decrement
-            | TokenValue::Identifier(_)
-            | TokenValue::TypeName(_)
-            | TokenValue::IntConstant(_)
-            | TokenValue::BoolConstant(_)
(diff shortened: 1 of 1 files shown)


src/front/glsl/parser/functions.rs
```diff
@@ -511,7 +511,17 @@ impl<'source> ParsingContext<'source> {
...
+                let Token { value, meta } = self.bump(parser)?;
+                return Err(Error {
+                    kind: ErrorKind::InvalidToken(
+                        value,
+                        // TODO: Add all tokens that can start a statement?
+                        vec![TokenValue::RightBrace.into()],
+                    ),
+                    meta,
+                });
+            }
```
JCapucho, line 524: We can just force a expression parsing if we don't recognized a token since they'll both result in a error, but we get the benefit that the error already has the expected tokens and we don't get cases like #1232 where the initial token wasn't recognized as valid for an expression.

JCapucho (changes requested):
We can do this in a more robust way

src/front/glsl/parser/functions.rs
fsh: Sure. So just replace all the `TokenValue::Plus | TokenValue::Dash | TokenValue::Bang | ...` that currently forwards to expression parsing, collapsing it to simply `_ => { <attempt to parse expression> }`?

JCapucho: Exactly

JCapucho (approved):
Thank you

kvark:
Published in 0.6.2

PR #1285, opened 2021-08-23
[spv-in] track function parameter sampling flags across handle permutation

### LINKED ISSUE #1275: Error compiling the shader "Naga parsing: InvalidGlobalVar(FunctionArgument(2))"
I am trying to compile and run the triangle example for [https://github.com/fintelia/smaa-rs](https://github.com/fintelia/smaa-rs). 
When running this it compiles fine, but fails the fragment shader at runtime with a validation error:
``` 
[2021-08-21T23:53:28Z ERROR wgpu::backend::direct] wgpu error: Validation Error
    
    Caused by:
          note: label = `smaa.pipeline.edge_detect`
        Internal error in FRAGMENT shader: Error compiling the shader "Naga parsing: InvalidGlobalVar(FunctionArgument(2))"
```

I am running on macOS 10.15 and the spirV has been compiles with shaderrc. Below I have included the spirV that fails to run.

[shader.smaa.shader.edge_detect.frag.spv.zip](https://github.com/gfx-rs/naga/files/7026373/shader.smaa.shader.edge_detect.frag.spv.zip)

I will try to provide more information if needed.

kvark:
Fixes #1275

Problem was the interaction between these things:
  - associated data for each function, tracking how each of the parameters is sampled
  - topological sort of the function call graph

The latter was reshuffling the function handles, while the former stored a separate array with data addressable by handles. So after the topological sort they got de-synchronized.

### DIFF
--- a/src/front/spv/function.rs
+++ b/src/front/spv/function.rs
@@ -1,6 +1,6 @@
 use crate::arena::{Arena, Handle};
 
-use super::{flow::*, Error, FunctionInfo, Instruction, LookupExpression, LookupHelper as _};
+use super::{flow::*, Error, Instruction, LookupExpression, LookupHelper as _};
 use crate::front::Emitter;
 
 pub type BlockId = u32;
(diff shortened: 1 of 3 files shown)


kvark:
Included in v0.6.1

PR #1293, opened 2021-08-25
Derive Clone for Type, TypeInner, and Constant

zicklag:


kvark:
I'm sorry, this is an intentional restriction that I'm trying to keep in the code. Experience shows that cloning these composite types is a good way to hide overhead, since they contain heap allocations. In Naga, all the overhead like that needs to be explicit and visible.

zicklag:
Gotcha, that makes sense.

My motivation is that that I was using the `Typifier` in a custom `naga` frontend of mine and I needed to create a new naga local variable that's type was calculated from `typifier.resolve()`. The only way I could figure to create a type from the `resolve` call was to clone the `TypeInner` that it returned and the use `module.types.fetch_or_append()` to get a handle to a type with that `TypeInner`.

I think I'm fine in my case with the performance implications of the clone, but I'm not sure if there's a better way around it. If there isn't another way to do it I'd probably just have to roll my own function that manually clones the `TypeInner` just by doing a `match` and then cloning the contents of each variant.

Do you have any thoughts on how I might be able to do that?

JCapucho:
Without more context is hard to help but if you just need the same type reusing the handle would be ideal, if you need to perform some changes to the type and then append it as a new one it depends on what changes you are trying but without context I can't really come with a one size fits all solution.

zicklag:
I would reuse the handle, but the typifier didn't return a handle, it returned a `&TypInner`. Maybe this is just something we could extend the `Typifier` API with, to add a function that would return a handle instead of a `&TypeInner`.

As I understand, internally the `Typeifier` may resolve the type of an expression to either a `Handle<Type>` or a `TypeInner` that doesn't happen to be added to the `Arena<Type>` yet, so maybe we just need a function that will always resolve to a handle, adding the type to the arena if necessary?

PR #1301, opened 2021-08-26
Implement fuzzing for the GLSL parser

hcsch:
With arbitrary 1.0.2 out (brings support for `HashMap`s with a custom hasher), fuzzing for the GLSL parser can be added relatively easily. This does that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,6 @@ rose_tree = { version ="0.3", optional = true }
 pp-rs = { version = "0.2.1", optional = true }
 hexf-parse = { version = "0.2.1", optional = true }
 
-# For random GLSL Options for fuzzing
-arbitrary = { version = "1.0.2", optional = true, features = ["derive"] }
-
 [features]
 default = []
 dot-out = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,9 @@ rose_tree = { version ="0.3", optional = true }
 pp-rs = { version = "0.2.1", optional = true }
 hexf-parse = { version = "0.2.1", optional = true }
 
+# For random GLSL Options for fuzzing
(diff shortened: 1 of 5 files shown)


hcsch:
Is there anything you'd like to see changed? The pull request should be mergable as-is. The build error with the most recent commit was a network error in the CI pipeline (i.e. a retry should fix the issue). Strangely enough, in my logged in session the commit itself failed along with the PR as a whole, but in a new session in a private window, the commit passed, only the PR as a whole failed, which is a bit... confusing
image

kvark:
The PR needs to be changed so that the fuzzer doesn't run on MSRV

hcsch:
> The PR needs to be changed so that the fuzzer doesn't run on MSRV

I'm afraid I don't quite understand what you want me to do, sorry.

Would you like for me to revert 3be60a81eb4a6d590a547e64018e98e4b17ae7ae, or would you like me to add a fuzzing stage somewhere in the CI pipeline that uses nightly (which I think is probably better suited for another separate PR, especially since I'm not yet familiar with GH workflows)?
Unless I'm missing something, the CI pipeline (or anything in this repo to do with building naga) currently doesn't touch/run any fuzzing code. More specifically, nothing in this PR as it is at the moment (with 3be60a81eb4a6d590a547e64018e98e4b17ae7ae) should change how `naga` compiles, and `fuzz/` is an entirely different crate (not even in the same cargo workspace as `naga`) that is at present not yet being compiled anyways (though you can run `cargo fuzz run glsl_parser` for example, but that will require nightly, because cargo-fuzz currently requires nightly).

If I've misunderstood you, I'd appreciate if you could clear up the misunderstanding for me.

kvark:
Looks like we are green!

kvark (approved):

PR #1307, opened 2021-08-29
[spv-out] Avoid generating duplicate OpTypeImage instructions.

### LINKED ISSUE #1305: SPIR-V image type deduplication isn't right
Naga generates bad SPIR-V for the following WGSL:

```
[[group(0), binding(0)]] var texture: texture_storage_2d<rgba32float, write>;
[[group(0), binding(0)]] var texture: texture_storage_2d<rgba32float, read>;
```

`spirv-val` rejects this with the error:
```
error: line 14: Duplicate non-aggregate type declarations are not allowed. Opcode: TypeImage id: 5
  %5 = OpTypeImage %float 2D 0 0 0 2 Rgba32f
```

Indeed, there are two identical `OpTypeImage` instructions in the generated SPIR-V:

```
          %3 = OpTypeImage %float 2D 0 0 0 2 Rgba32f
          %5 = OpTypeImage %float 2D 0 0 0 2 Rgba32f
```

This is because the SPIR-V backend deduplicates equal `LocalType` values, but `LocalType` equality requires `ImageClass::Storage::access`, to match, even though that has no influence on the `OpImageType` generated.

There may be other aspects of a Naga `Image` type that do not affect the generated `OpImageType`. Any two `Image` types that would generate the same `OpImageType` must have equal `LocalType` representations.

jimblandy:
(This may be easier to review one commit at a time.)

Fixes #1305.

Ensure that two `back::spv::LocalType::Image` values are sure to be equal (and
hash equal) whenever they would generate the same `OpTypeImage` instruction, so
that the usual duplicate removal via `Writer::lookup_type` works. Accomplish
this by changing the contents of `LocalType::Image` to more closely match the
operands of `OpTypeImage` instructions.

Previously, `LocalType::Image` included a `ImageClass::Storage::access` value,
which did not affect the `OpTypeImage` instruction generated. If two
`LocalType::Image` values differered only in their `access`, then they would get
separate entries in `Writer::lookup_type`, two identical `OpTypeImage`
instructions would be generated, and SPIR-V validation would fail.


jimblandy:
(Last force-push fixes a clippy warning on one of the intermediate commits. This should be all right to land without squashing.)

jimblandy:
Tests are a good thing.

src/back/spv/mod.rs
```diff
@@ -153,6 +153,70 @@ impl Function {
     }
 }
 
+fn map_dim(dim: crate::ImageDimension) -> spirv::Dim {
```
kvark, line 156: any reason to not `impl From<>`?

kvark (approved):

src/back/spv/instructions.rs
```diff
@@ -219,67 +219,26 @@ impl super::Instruction {
         instruction
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub(super) fn type_image(
         id: Word,
         sampled_type_id: Word,
         dim: spirv::Dim,
+        depth: bool,
```
kvark, line 227: This is internal API, so it's not that important, but generally we shouldn't accept booleans.
Creating a bitflags type for these would be nice.

kvark (approved):
That's way more code than I thought we'd have for this, but it's good to have it fixed nevertheless. Thank you!

src/back/spv/instructions.rs
jimblandy: I did this, it was a great cleanup. Noted for the future.

src/back/spv/mod.rs
jimblandy: This was pre-existing. I'd much prefer a `From` impl. Done.

kvark (approved):

kvark:
Published in 0.6.2

PR #1308, opened 2021-08-30
[glsl-in] Allow code after terminator statements

JCapucho:


src/front/glsl/builtins.rs
```diff
@@ -1418,6 +1418,7 @@ impl MacroCall {
         body: &mut Block,
         args: &mut [Handle<Expression>],
         meta: SourceMetadata,
+        finished: bool,
```
kvark, line 1421: can we add it to `Context` instead of having a separate argument?

block_ctx
```diff
@@ -0,0 +1,883 @@
+BlockContext {
```
kvark, line 1: is this file included accidentally?

dis
```diff
@@ -0,0 +1,47 @@
+; SPIR-V
```
kvark, line 1: same here

test-block-ctx/block_ctx.Fragment-main.txt
```diff
@@ -0,0 +1,113 @@
+BlockContext {
```
kvark, line 1: also a leftover?

kvark (commented):
I think we can make it even better. Keeping the flag around in the Context still requires checking for it on each statement. What if we instead just cut away all of the statements added after return? I.e.
```rust
fn parse_block() {
...
if stmt == Return {
  return_statements = Some(statements.len());
}
... 
if let Some(count) = return_statements {
  statements.drain(count..); // just flush the rest of statements out
}
}
```

It seems to me that this would require only a single place being responsible for this, instead of spreading the tentacles of this logic through the code

block_ctx
JCapucho: yeah, sorry accidentally added some test files from the spirv parser PR

JCapucho:
The method you propose would incur more allocations, If that's not a problem in your opinion then I'll change it.

kvark:
Not a problem. We are talking about GLSL code that does weird stuff, it can suffer :)

JCapucho:
My poor baby :(

kvark:
Sorry, I didn't mean GLSL in general. I mean specifically GLSL code that does anything after `return`. Your baby is great!

JCapucho:
@kvark implemented your suggestion

kvark (approved):

PR #1310, opened 2021-08-31
Fix GLSL output for non-fallthrough switch cases

### LINKED ISSUE #1309: [glsl-out] Switches don't include break
I have this wgsl shader (part of [wgpu_glyph](https://github.com/hecrj/wgpu_glyph/blob/master/src/shader/glyph.wgsl)):
```rs
switch (i32(input)) {
        case 0: {
            
        }
        case 1: {
            
        }
        case 2: {
            
        }
        case 3: {
           
        }
}
```
Before this change: https://github.com/gfx-rs/naga/commit/02c74b50021379092a9d4567fd11aeb0879c5405#diff-cc33bd3c8c18000d3ee7b6f5578e8fa07ffd08d87b0aa3d9525a107937fb3d3f

it used to output this glsl:
```glsl
switch(int(input)) {
            case 0:
                break;
            case 1:
                break;
            case 2:
                break;
            case 3:
                break;
}
```
Now it outputs:
```glsl
switch(int(input)) {
            case 0:

            case 1:

            case 2:

            case 3:
}
```

This breaks wgpu_glyph on OpenGL backend. ~~I'm not sure if this is correct behavior for wgsl, does it brake by default? Judging by other backend, I always thought so.~~

hcsch:
Fixes #1309 by partially reverting 02c74b50021379092a9d4567fd11aeb0879c5405.
Also cleans up the GLSL output by only inserting a `break` where a case is not already broken out of at the end of the case body and inserts a comment indicating explicitly fallthrough cases.

Also adds a test case for switch statements not in the last statement position of a function, because those might have return statements inserted, which previously might have made it harder to notice some missing `break`s that should have been inserted.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1442,8 +1442,18 @@ impl<'a, W: Write> Writer<'a, W> {
                         self.write_stmt(sta, ctx, indent + 2)?;
                     }
 
-                    // Write `break;` if the block isn't fallthrough
-                    if !case.fall_through {
+                    // Write fallthrough comment if the case is fallthrough,
+                    // otherwise write a break, if the case is not already
+                    // broken out of at the end of its body.
+                    if case.fall_through {
+                        writeln!(self.out, "{}/* fallthrough */", INDENT.repeat(indent + 2))?;
+                    } else if !matches!(
+                        case.body.last(),
+                        Some(&Statement::Break)
+                            | Some(&Statement::Continue)
+                            | Some(&Statement::Return { .. })
+                            | Some(&Statement::Kill)
+                    ) {
                         writeln!(self.out, "{}break;", INDENT.repeat(indent + 2))?;
                     }
                 }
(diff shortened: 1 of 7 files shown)


src/back/glsl/mod.rs
```diff
@@ -1441,6 +1441,21 @@ impl<'a, W: Write> Writer<'a, W> {
...
                     }
+
+                    // Write fallthrough comment if the case is fallthrough,
+                    // otherwise write a break, if the case is not already
+                    // broken out of at the end of its body.
+                    if case.fall_through {
+                        writeln!(self.out, "{}/* fallthrough */", INDENT.repeat(indent + 2))?;
+                    } else if !matches!(
+                        case.body.last(),
+                        Some(&Statement::Break)
```
kvark, line 1452: we should probably add a helper method here so that this looks like `statement.is_jump()` or something similar

kvark (approved):
Nice, thank you!

PolyMeilex:
Thank you!

src/back/glsl/mod.rs
hcsch: I was considering adding something like that, but I wasn't very sure about what to call it and whether this one use would justify the addition. Since this PR is merged already, should I do a follow up PR for this?

kvark:
Published in 0.6.2

PR #1312, opened 2021-08-31
Update WGSL grammar for pointer access.

### LINKED ISSUE #1311: [wgsl-in] Issues with ptr<storage, T>
When using a storage class pointer type, the information is lost about the StorageAccess mode:

```
[2021-08-31T20:01:23Z ERROR naga::front::wgsl] Given type Pointer { base: [10], class: Storage { access: (empty) } } doesn't match expected Pointer { base: [10], class: Storage { access: LOAD } }
error: the type of `ref` is expected to be [23]
    ┌─ wgsl:180:9
    │
180 │     var ref: ptr<storage, Type> = &u_storage.value;
    │         ^^^ definition of `ref`

Could not parse WGSL
```

This appears to be a separate issue than implementing pointer memory access modes.

kvark:
Fixes #1311
Comes with a small test, which revealed a number of issues in the backends.


encounter:
I found a couple of issues while testing, possibly unrelated to this PR specifically, but:

```
[[block]]
struct Storage {
    bar: f32;
};
[[group(0), binding(0)]]
var<storage, read> foo: Storage;

fn read_from_storage(baz: ptr<storage, Storage>) -> f32 {
    return (*baz).bar;
}

[[stage(vertex)]]
fn vs_main() -> [[builtin(position)]] vec4<f32> {
    return vec4<f32>(read_from_storage(&foo));
}
```

For GLSL, this emits invalid syntax:

```
float read_from_storage(inout Storage_block_1Vs {
    float bar;
} baz) {
    Storage _e3 = baz;
    return _e3.bar;
}
```

For HLSL, this hits an `Unimplemented` error:

```
Pointer access of FunctionArgument(0)
```

kvark:
@encounter thank you for pro-actively testing this!
Storage-class pointer arguments are actually forbidden by the spec for user-defined functions. The new commit adds the relevant validation as well as a small test.

encounter:
I missed that, thanks for the explanation! (A shame, though, storage-class pointers would help in my shader)

kvark:
Indeed. I'll raise this with the group (again).

jimblandy:
Okay,  455a7e5 is the remainder of this PR after #1332 and #1360. I'll review tomorrow.

tests/snapshots.rs
```diff
@@ -548,8 +550,8 @@ fn convert_spv_inverse_hyperbolic_trig_functions() {
 }
 
 #[cfg(all(feature = "spv-in", feature = "spv-out"))]
-#[test]
-fn convert_spv_pointer_access() {
+//#[test] //TODO: fix storage class in function parameter
+fn _convert_spv_pointer_access() {
```
jimblandy, line 554: Do we have a follow-up issue filed for this?

tests/wgsl-errors.rs
```diff
@@ -639,13 +639,21 @@ fn invalid_functions() {
         if function_name == "unacceptable_unsized" && argument_name == "arg"
     }
 
-    // A *valid* way to pass an unsized value.
```
jimblandy, line 642: Could we keep a test here that validates successfully?

```
    check_validation_error! {
        "
        struct Unsized { data: array<f32>; };
        fn acceptable_pointer_to_unsized(arg: ptr<workgroup, Unsized>) { }
        ":
        Ok(_)
    }

```

jimblandy (changes requested):
Looks good! A few small issues here.

src/back/msl/writer.rs
```diff
@@ -1383,7 +1383,14 @@ impl<W: Write> Writer<W> {
...
                 crate::Statement::Emit(ref range) => {
                     for handle in range.clone() {
-                        let expr_name = if let Some(name) =
+                        let info = &context.expression.info[handle];
+                        let ptr_class = info
+                            .ty
+                            .inner_with(&context.expression.module.types)
+                            .pointer_class();
+                        let expr_name = if ptr_class.is_some() {
+                            None // don't bake pointer expressions (just yet)
```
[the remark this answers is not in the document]
kvark: probably just not anticipated by the backend yet, therefore ("just yet")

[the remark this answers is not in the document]
jimblandy: Okay, sounds fine then.

tests/snapshots.rs
```diff
@@ -225,15 +225,16 @@ fn write_output_spv(
...
+                .expect("SPV EP write failed");
             let dis = rspirv::dr::load_words(spv)
                 .expect("Produced invalid SPIR-V")
                 .disassemble();
             let path = format!("spv/{}.{}.spvasm", file_name, ep.name);
             fs::write(destination.join(path), dis).unwrap();
         }
     } else {
-        let spv = spv::write_vec(module, info, &options, None).unwrap();
+        let spv = spv::write_vec(module, info, &options, None).expect("SPV write failed");
```
jimblandy, line 237: Oh, these changes might conflict with #1427. Oh well!

jimblandy (approved):
Looks great. I had one concern about the WGSL snapshot output.

tests/snapshots.rs
kvark: I'll remove it

PR #1314, opened 2021-09-01
Make span copyable

JCapucho:


kvark (changes requested):

src/span.rs
```diff
@@ -1,38 +1,32 @@
...
-        match *self {
-            Self::Unknown => self.clone_from(other),
-            Self::ByteRange(ref mut self_range) => {
-                if let Self::ByteRange(ref other_range) = *other {
-                    self_range.start = self_range.start.min(other_range.start);
-                    self_range.end = self_range.end.max(other_range.end);
-                }
-            }
-        }
+    pub fn subsume(&mut self, other: Self) {
```
kvark, line 11: I think this is incorrect. If I have an empty span, like the 0..0 returned by `Default`, then the result of `subsume` will always have `start == 0`, but the behavior we want is that the result would be the other span.

src/span.rs
```diff
@@ -1,38 +1,65 @@
...
-                    self_range.start = self_range.start.min(other_range.start);
-                    self_range.end = self_range.end.max(other_range.end);
-                }
-            }
+    /// Modifies `self` to contain the smallest `Span` possible that
+    /// contains both `self` and `other`
+    pub fn subsume(&mut self, other: Self) {
+        if !self.is_defined() {
+            // self isn't defined so use other
+            return *self = other;
```
kvark, line 23: matter of choice, I could suggest
```rust
*self =
        if !self.is_defined() {
            // self isn't defined so use other
            other
        } else if !other.is_defined() {
            *self
        } else {
          Span {...}
        };
```
That would remove the early returns

kvark (approved):

src/arena.rs
```diff
@@ -260,15 +260,18 @@ impl<T> Arena<T> {
         self.data.clear()
     }
 
-    pub fn get_span(&self, handle: Handle<T>) -> &Span {
+    pub fn get_span(&self, handle: Handle<T>) -> Span {
         #[cfg(feature = "span")]
         {
-            return self.span_info.get(handle.index()).unwrap_or(&Span::Unknown);
+            return *self
```
kvark, line 266: no need for `return` here

PR #1319, opened 2021-09-02
[wgsl-out] Reduce heap allocation

Gordon-F:


src/back/wgsl/writer.rs
```diff
@@ -292,71 +292,91 @@ impl<W: Write> Writer<W> {
...
-                // Add a separator between args
-                let separator = if index < attributes.len() - 1 {
-                    ", "
-                } else {
-                    ""
-                };
-                attributes_str = format!("{}{}{}", attributes_str, attribute_str, separator);
+
+            // Only write a comma if isn't the last element
+            if index != attributes.len().saturating_sub(1) && need_last_comma {
```
kvark, line 371: Instead of doing the fancy `saturating_sub`, we could just do `index + 1`.

kvark (commented):
Thank you!
Noted a few things to refactor, overall looks like a good improvement!

One of the main offenders of heap allocations, as noticed by @jimblandy , is `INDENT.repeat(indent)`, which returns a string. If you want to bring the backend to a really nice state, perhaps, you could help with refactoring this by using a thing like in MSL:
```rust
#[derive(Clone)]
struct Level(usize);
impl Level {
    fn next(&self) -> Self {
        Level(self.0 + 1)
    }
}
impl Display for Level {
    fn fmt(&self, formatter: &mut Formatter<'_>) -> Result<(), FmtError> {
        (0..self.0).try_for_each(|_| formatter.write_str(back::INDENT))
    }
}
```

Gordon-F:
> If you want to bring the backend to a really nice state, perhaps, you could help with refactoring this by using a thing like in MSL:

This would be a really nice improvement. I will do it. But I think it would be better to implement it in a separate PR and for all backends.

kvark (approved):

kvark:
Published in 0.6.3

PR #1323, opened 2021-09-03
Refactoring indent logic for backends

Gordon-F:
Follow up #1319.

Refactoring indent logic for hlsl, glsl, wgsl backends.

### DIFF
--- a/src/back/mod.rs
+++ b/src/back/mod.rs
@@ -20,6 +20,24 @@ const INDENT: &str = "    ";
 #[allow(dead_code)]
 const BAKE_PREFIX: &str = "_e";
 
+#[derive(Clone, Copy)]
+#[allow(dead_code)]
+struct Level(usize);
+
+#[allow(dead_code)]
+impl Level {
+    fn next(&self) -> Self {
+        Level(self.0 + 1)
+    }
+}
+
+#[allow(dead_code)]
+impl std::fmt::Display for Level {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        (0..self.0).try_for_each(|_| formatter.write_str(INDENT))
+    }
+}
+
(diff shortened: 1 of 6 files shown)


src/back/mod.rs
```diff
@@ -20,6 +20,28 @@ const INDENT: &str = "    ";
...
+#[allow(dead_code)]
+struct Level(usize);
+
+#[allow(dead_code)]
+impl Level {
+    fn next(&self) -> Self {
+        Level(self.0 + 1)
+    }
+
+    fn extend(&self, size: usize) -> Self {
```
kvark, line 33: I don't think we really need this one. It's never a case that we are jumping 2 indents at once.
Generally, it should be just `let l2 = level.next().next();` instead.

kvark (commented):
Very nice, thank you!
Just one tiny thing

src/back/mod.rs
Gordon-F: I don't think that this:
```rust
writeln!(self.out, "{}/* fallthrough */", level.next().next())?;
```
better then this:
```rust
writeln!(self.out, "{}/* fallthrough */", level.extend(2))?;
```


Also we have one case in hlsl backend:
```rust
level.extend(2 + usize::from(case.fall_through)
```

Without `extend` it should be:
```rust
back::Level(level.0 + 2 + usize::from(case.fall_through))
```

kvark: when you are writing code, you never jump 2 tabs.
This first piece should look like this:
```rust
                for case in cases {
                    let level = level.next(); // <------------- here
                    writeln!(self.out, "{}case {}:", level, case.value)?;

                    for sta in case.body.iter() {
                        self.write_stmt(sta, ctx, level.next())?;
                    }
...
```

Gordon-F: Make sense. Thank you for explain!

kvark (approved):

kvark:
Published in 0.6.3

PR #1325, opened 2021-09-04
[glsl-out] Fix select order

JCapucho:
When there is a vector as the condition for `Expression::Select` we use a `mix` in the glsl backend but the first argument is the reject and the second the accept this was inverted.

This should probably be backported as can cause surprising results.


kvark (approved):

kvark:
Published in 0.6.3

PR #1326, opened 2021-09-04
[spv-in] Handle structs with no offset decorations

JCapucho:
From the llvm discourse on whether spirv with missing offset decorations
> For non-external storage classes, the struct does not need to be explicitly laid out.

Seems like `shaderc` does this and I had shader where all offsets were 0, this PR fixes it by using a similar strategy to glsl frontend to calculate span/offsets.


kvark (approved):

kvark:
Published in 0.6.3

PR #1327, opened 2021-09-04
[spv-in] Divide the depth ref in projection sample

JCapucho:
Despite the spec not making any allusion to this, glslang produces spirv that expects this behavior and spirv_cross reads them this way.


kvark (approved):
Nice!

kvark:
Published in 0.6.3

PR #1332, opened 2021-09-08
[wgsl-out] Correct production of `*` and `&` operators.

### LINKED ISSUE #1322: WGSL back end generates incorrect code for address-of expressions
(Bug description revised).

Passing the following WGSL through Naga IR:

```
fn f(i: ptr<function, i32>) {}

fn g() {
  var x = 10;
  f(&x);
}
```

produces the following incorrect output:

```
fn f(i: ptr<function, i32>) {
    return;
}

fn g() {
    var x: i32 = 10;

    f(x);
    return;
}
```

Specifically, the passing of `x` to `f` is ill-typed: the `&` operator has been dropped.

---

Original, incorrect bug report:

Passing the following WGSL through Naga IR:

```
fn f(i: i32) -> i32 {
  var x = i;
  var p = &x;
  return i;
}
```

produces the following incorrect output:

```
fn f(i: i32) -> i32 {
    var x: i32;
    var p: ptr<function, i32>;

    x = i;
    p = x;
    return i;
}
```

Specifically, the assignment of `x` to `p` is ill-typed: the `&` operator has been dropped.

jimblandy:
Fixes #1322.

This still needs tests, so, marking it as a draft, but comments welcome.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -22,6 +22,85 @@ enum Attribute {
     WorkGroupSize([u32; 3]),
 }
 
+/// The WGSL form that `write_expr_with_indirection` should use to render a Naga
+/// expression.
+///
+/// Sometimes a Naga `Expression` alone doesn't provide enough information to
+/// choose the right rendering for it in WGSL. For example, one natural WGSL
+/// rendering of a Naga `LocalVariable(x)` expression might be `&x`, since
+/// `LocalVariable` produces a pointer to the local variable's storage. But when
+/// rendering a `Store` statement, the `pointer` operand must be the left hand
+/// side of a WGSL assignment, so the proper rendering is `x`.
+///
+/// The caller of `write_expr_with_indirection` must provide an `Expected` value
+/// to indicate how ambiguous expressions should be rendered.
+#[derive(Clone, Copy, Debug)]
+enum Indirection {
+    /// Render pointer-construction expressions as WGSL `ptr`-typed expressions.
+    ///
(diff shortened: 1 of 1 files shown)


src/back/wgsl/writer.rs
```diff
@@ -22,6 +22,85 @@ enum Attribute {
...
+/// - The plain form of `LocalVariable(x)` is simply `x`, which is a reference
+///   to the local variable's storage.
+///
+/// - The plain form of `GlobalVariable(g)` is simply `g`, which is usually a
+///   reference to the global variable's storage. However, globals in the
+///   `Handle` storage class are immutable, and `GlobalVariable` expressions for
+///   those produce the value directly, not a pointer to it. Such
+///   `GlobalVariable` expressions are `Ordinary`.
+///
+/// - `Access` and `AccessIndex` are `Reference` when their `base` operand is a
```
kvark, line 68: so they just inherit the indirection from the base

src/back/wgsl/writer.rs
```diff
@@ -938,8 +1031,25 @@ impl<W: Write> Writer<W> {
             return Ok(());
         }
 
+        // If the plain form of the expression is not what we need, emit the
+        // operator necessary to correct that.
+        let plain = plain_form_indirection(expr, module, func_ctx);
+        match (requested, plain) {
```
kvark, line 1037: this is really clean, well done!

kvark (approved):

kvark:
I'd feel good just landing it. This isn't less tested than what we already have :)

jimblandy:
All right, we'll land it as is.

PR #1333, opened 2021-09-08
Infer types of module-scope let declarations

kvark:
Matches https://github.com/gpuweb/gpuweb/pull/2053


kvark:
Published in 0.6.3

PR #1334, opened 2021-09-08
spv-out: option to skip OpName for varyings

### LINKED ISSUE #1331: [spv-out] Generated SPIR-V with duplicate OpName fails on Adreno drivers
This one is fun! Apparently OpName is semantic, somehow, on Adreno drivers.

```wgsl
struct VertexOutput {
    [[builtin(position)]] pos: vec4<f32>;
    [[location(0)]] lighting: vec3<f32>;
};

[[stage(vertex)]]
fn vs_main_test(
    [[location(0)]] pos: vec3<f32>,
    [[location(1)]] lighting: vec3<f32>,
) -> VertexOutput {
    return VertexOutput(vec4<f32>(pos, 1.0), lighting);
}
```

Adreno rejects this SPIR-V simply because the stage input `lighting` has the same `OpName` as the stage output `lighting`. 

Renaming the input `lighting` to `in_lighting` resolves this issue, and a `spirv-dis --raw-id` diff shows that the only difference is the `OpName` for that field.

However, the `pos`/`pos` `OpName`s do _not_ cause an issue, indicating it's only a problem with non-builtin stage inputs and outputs. This example is also OK:

```wgsl
struct VertexOutput {
    [[builtin(position)]] pos: vec4<f32>;
    [[location(0)]] vertex_index: u32;
};

[[stage(vertex)]]
fn vs_main_test(
    [[builtin(vertex_index)]] vertex_index: u32,
    [[location(0)]] pos: vec3<f32>,
) -> VertexOutput {
    return VertexOutput(vec4<f32>(pos, 1.0), vertex_index);
}
```

kvark:
Fixes #1331 (need testing/confirmation!)

### DIFF
--- a/src/back/spv/mod.rs
+++ b/src/back/spv/mod.rs
@@ -511,6 +511,9 @@ bitflags::bitflags! {
         const DEBUG = 0x1;
         /// Flip Y coordinate of `BuiltIn::Position` output.
         const ADJUST_COORDINATE_SPACE = 0x2;
+        /// Emit `OpLabel` for input/output locations.
+        /// Some drivers treat it as semantic, not allowing any conflicts.
+        const LABEL_VARYINGS = 0x4;
     }
 }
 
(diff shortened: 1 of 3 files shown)


kvark:
Published in 0.6.3

PR #1346, opened 2021-09-10
[spv-in] Move essential code out of debug_assert

JCapucho:
Some idiot (me) decided to put essential code inside of `debug_assert` and then was surprised why it didn't work in release.


jimblandy (approved):
LGTM

PR #1348, opened 2021-09-11
Output parentheses around dereference

magcius:
Regression from ce676cf13090b87e1405640ea8ee43e1a1666005.

We need to output (*d).mx rather than *d.mx, at least according to Tint.
wgsl-in seems to handle *d.mx just fine, which is likely a separate bug.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -1034,11 +1034,11 @@ impl<W: Write> Writer<W> {
         // If the plain form of the expression is not what we need, emit the
         // operator necessary to correct that.
         let plain = plain_form_indirection(expr, module, func_ctx);
-        match (requested, plain) {
-            (Indirection::Ordinary, Indirection::Reference) => write!(self.out, "&")?,
-            (Indirection::Reference, Indirection::Ordinary) => write!(self.out, "*")?,
-            (_, _) => {}
-        }
+        let opened_paren = match (requested, plain) {
+            (Indirection::Ordinary, Indirection::Reference) => { write!(self.out, "&")?; false },
+            (Indirection::Reference, Indirection::Ordinary) => { write!(self.out, "(*")?; true },
+            (_, _) => { false },
+        };
 
         let expression = &func_ctx.expressions[expr];
 
@@ -1542,6 +1542,10 @@ impl<W: Write> Writer<W> {
             // Nothing to do here, since call expression already cached
             Expression::CallResult(_) | Expression::AtomicResult { .. } => {}
         }
+
+        if opened_paren {
+            write!(self.out, ")")?
+        };
 
         Ok(())
     }
--- a/tests/in/glsl/fma.frag
+++ b/tests/in/glsl/fma.frag
@@ -1,0 +1,9 @@
+#version 440 core
+
+struct Mat4x3 { vec4 mx; vec4 my; vec4 mz; };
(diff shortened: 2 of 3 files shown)


kvark:
Thank you! Let's double-check with WGSL spec first.

jimblandy:
@magcius, can you give an example of input which provokes incorrect output?

jimblandy:
On master, the following input:

    struct S {
        member: i32;
    };
    
    fn f(p: ptr<function, S>) -> i32 {
      return (*p).member;
    }

produces the following output:

    struct S {
        member: i32;
    };
    
    fn f(p: ptr<function, S>) -> i32 {
        let _e1: S = *p;
        return _e1.member;
    }

This seems correct. In fact, given the way `back::wgsl` treats `Load` expressions, I don't see how we would be generating `(*p).x` at all - the `Load` would always be 'baked' into a `let` of a temporary variable, as shown above.

jimblandy:
Oh, I see there's a test case included in the PR. My bad.

jimblandy:
Okay, I see it's about stores:

    struct S {
        member: i32;
    };
    
    fn f(p: ptr<function, S>) {
      (*p).member = 10;
    }

produces:

    struct S {
        member: i32;
    };
    
    fn f(p: ptr<function, S>) {
        *p.member = 10;
        return;
    }

WGSL says that `*p.member` is a `unary_expression` containing the `singular_expression` `p.member`, so it is parsed as `*(p.member)`, which is indeed incorrect.

jimblandy:
Okay, I see where I went wrong. Before my patch, the comment on `write_expr_with_indirection` claiming that it produces a WGSL grammar `singular_expression` was correct. But after my patch, it produces a `unary_expression`, since that's the production that includes `*` and `&` operators. So that comment needs to be fixed.

src/back/wgsl/writer.rs
```diff
@@ -1034,11 +1034,11 @@ impl<W: Write> Writer<W> {
...
         // operator necessary to correct that.
         let plain = plain_form_indirection(expr, module, func_ctx);
-        match (requested, plain) {
-            (Indirection::Ordinary, Indirection::Reference) => write!(self.out, "&")?,
-            (Indirection::Reference, Indirection::Ordinary) => write!(self.out, "*")?,
-            (_, _) => {}
-        }
+        let opened_paren = match (requested, plain) {
+            (Indirection::Ordinary, Indirection::Reference) => { write!(self.out, "&")?; false },
+            (Indirection::Reference, Indirection::Ordinary) => { write!(self.out, "(*")?; true },
```
jimblandy, line 1039: The same problem will arise with `&` operators, in theory. So both of these need to be parethesized.

jimblandy (changes requested):

src/back/wgsl/writer.rs
magcius: Yeah, I didn't know if `&foo.bar` was valid. I can easily fix.

magcius:
I tried to do this, but it seems it fails validation.

magcius:
Put this back so that it only parenthsizes * for now while I try to investigate why parenthesizing `&at` appears to make the reference become non-atomic.

jimblandy:
Lol, with the parenthesization, this validates:

    atomicStore(&at, 2u);

but this does not:

    atomicStore((&at), 2u);

"Parenthesis of Death"

jimblandy:
Since we've filed #1352, we can land this without the fix for `&`.

jimblandy (approved):
Thank you for the fix!

PR #1350, opened 2021-09-11
Fix sampling texture array types

magcius:


src/front/glsl/builtins.rs
```diff
@@ -1670,12 +1670,19 @@ fn texture_call(
     meta: SourceMetadata,
 ) -> Result<Handle<Expression>> {
     if let Some(sampler) = ctx.samplers.get(&image).copied() {
+        let mut array_index = comps.array_index;
+
+        if let Some(mut array_index_expr) = array_index {
```
JCapucho, line 1675: You can use `ref mut` here and remove the reference afterwards and the assignment

src/front/glsl/context.rs
```diff
@@ -834,6 +834,25 @@ impl Context {
             .and_then(|(kind, width)| type_power(kind, width)))
     }
 
+    pub fn conversion(
```
JCapucho, line 837: Very nice, this part really needed to refractored

JCapucho (changes requested):
Did just light review since I won't be able to properly review the code until monday

src/front/glsl/builtins.rs
magcius: Thanks, that helps clean it up a lot.

src/front/glsl/types.rs
```diff
@@ -131,14 +131,14 @@ pub fn parse_type(type_name: &str) -> Option<Type> {
 
                 let (dim, arrayed, class) = match size {
                     "1D" => (ImageDimension::D1, false, sampled(false)),
-                    "1DArray" => (ImageDimension::D1, false, sampled(false)),
+                    "1DArray" => (ImageDimension::D1, true, sampled(false)),
                     "2D" => (ImageDimension::D2, false, sampled(false)),
-                    "2DArray" => (ImageDimension::D2, false, sampled(false)),
+                    "2DArray" => (ImageDimension::D2, true, sampled(false)),
                     "2DMS" => (ImageDimension::D2, true, sampled(true)),
```
kvark, line 137: why does this one have "true" for the arrayed?

tests/in/glsl/sampler2DArray.frag
```diff
@@ -0,0 +1,12 @@
+#version 440 core
```
kvark, line 1: I don't think we should have a separate test per sampler type. Can we just call it `samplers.frag`?

kvark (commented):
Looks reasonable, a few notes here

src/front/glsl/types.rs
magcius: Nice catch!

tests/in/glsl/sampler2DArray.frag
magcius: Done.

kvark (approved):

PR #1363, opened 2021-09-14
Clean up some old snapshot files.

jimblandy:
`tests/out/glsl/operators.main.Fragment.glsl` is no longer generated because the function in question changed to a Compute entry point.

`tests/out/wgsl/multiple_entry_points-glsl.wgsl` is no longer generated because the input test was removed in 057dc310.


kvark (approved):
Thank you!

PR #1364, opened 2021-09-15
Enable span tracking by default in the Naga CLI.

jimblandy:


cli/Cargo.toml
```diff
@@ -14,7 +14,7 @@ name = "naga"
 path = "src/main.rs"
 
 [dependencies]
-naga = { path = "../", features = ["wgsl-in", "wgsl-out", "glsl-in", "glsl-out", "spv-in", "spv-out", "msl-out", "hlsl-out", "dot-out", "glsl-validate"] }
+naga = { path = "../", features = ["span", "wgsl-in", "wgsl-out", "glsl-in", "glsl-out", "spv-in", "spv-out", "msl-out", "hlsl-out", "dot-out", "glsl-validate"] }
```
kvark, line 17: should this be just `all-featues = true` or something?

kvark (approved):

cli/Cargo.toml
jimblandy: That would be good, but I don't think there is any `all-features` key in Cargo manifests:

    warning: /home/jimb/rust/naga/cli/Cargo.toml: unused manifest key: dependencies.naga.all-features

PR #1365, opened 2021-09-15
Correctly recognize defined spans.

jimblandy:
Without this change, the `to_range` method method returns `None` even when spans are enabled.


kvark (approved):
Oh wow...

PR #1367, opened 2021-09-15
[wgsl-in] Refuse to apply postfix expressions to WGSL pointers.

### LINKED ISSUE #1356: [wgsl-in] Missing parentheses around dereference
https://github.com/gfx-rs/naga/pull/1348 fixed this for wgsl-out, but the following still incorrectly validates:

```
[[stage(vertex)]]
fn main() -> [[builtin(position)]] vec4<f32> {
    var v: vec4<f32> = vec4<f32>(1.0, 1.0, 1.0, 1.0);
    let pv = &v;
    let a = *pv[3]; // Problematic line
    return v;
}
```

```
cargo run -- test.wgsl
```

[According to the WGSL grammar](https://www.w3.org/TR/WGSL/#expression-grammar), [] should bind more tightly than *, so `*pv[3]` should not pass validation. The correct code would be `(*pv)[3]`.

naga @ https://github.com/gfx-rs/naga/commit/0e66930aff1340d5c636df616d573cc326702158

jimblandy:
Fixes #1356.

Output for the test case in that issue is now:

    error: the value indexed by a `[]` subscripting expression must not be a pointer
      ┌─ wgsl:5:14
      │
    5 │     let a = *pv[3]; // Problematic line
      │              ^^ expression is a pointer

    Could not parse WGSL

Also, record identifier definitions as TypedExpression values.

Both a `var` binding and a `let` binding of a pointer to a variable create entries in the `lookup_ident` tables whose expression is a `LocalVariable`. However, the `let` should be a pointer, whereas the `var` should be a reference. This patch changes the tables to hold `TypedExpression` values, thus preserving the distinction.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -165,9 +165,9 @@ pub enum Error<'a> {
     MissingType(Span),
     InvalidAtomicPointer(Span),
     InvalidAtomicOperandType(Span),
+    Pointer(&'static str, Span),
     NotPointer(Span),
-    AddressOfNotReference(Span),
-    AssignmentNotReference(Span),
+    NotReference(&'static str, Span),
     Other,
 }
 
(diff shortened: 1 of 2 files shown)


kvark (approved):
High quality stuff, as always!
Took me a bit to figure out what was wrong :)

PR #1368, opened 2021-09-15
[wgsl-out] Properly parenthesize address-of expressions.

### LINKED ISSUE #1352: WGSL back end does not properly parenthesize address-of expressions.
In the WGSL back end, `write_expr_with_indirection` claims to produce code that matches the `singular_expression` production, but when it emits a prefix `&` operator, the output actually only matches `unary_expression`. This results in incorrect output for, as an example, assignments to members of pointed-to structs.

For example, given the input:

```
struct S {
    member: i32;
};

fn f(p: ptr<function, S>) {
  (*p).member = 10;
}
```

Re-rendering this as WGSL produces `*p.member = 10;`, instead of the correct `(*p).member = 10;`.

It would be easy to fix this as part of #1348. However, fixing this is blocked by #1351, because properly parenthesized address-of expressions confuse `parse_atomic_pointer`.

jimblandy:
Now that the WGSL front end distinguishes pointers and references correctly, we can land this.

Fixes #1352.

### DIFF
--- a/tests/out/wgsl/access.wgsl
+++ b/tests/out/wgsl/access.wgsl
@@ -19,7 +19,7 @@ fn foo([[builtin(vertex_index)]] vi: u32) -> [[builtin(position)]] vec4<f32> {
     let matrix: mat4x4<f32> = bar.matrix;
     let arr: array<vec2<u32>,2> = bar.arr;
     let b: f32 = bar.matrix[3][0];
-    let a: i32 = bar.data[(arrayLength(&bar.data) - 2u)];
+    let a: i32 = bar.data[(arrayLength((&bar.data)) - 2u)];
     bar.matrix[1][2] = 1.0;
     bar.matrix = mat4x4<f32>(vec4<f32>(0.0), vec4<f32>(1.0), vec4<f32>(2.0), vec4<f32>(3.0));
     bar.arr = array<vec2<u32>,2>(vec2<u32>(0u), vec2<u32>(1u));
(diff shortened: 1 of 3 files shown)


tests/out/wgsl/access.wgsl
```diff
@@ -33,23 +33,23 @@ fn foo([[builtin(vertex_index)]] vi: u32) -> [[builtin(position)]] vec4<f32> {
...
+    let _e12: i32 = atomicAnd((&bar.atom), 5);
     tmp = _e12;
-    let _e15: i32 = atomicOr(&bar.atom, 5);
+    let _e15: i32 = atomicOr((&bar.atom), 5);
     tmp = _e15;
-    let _e18: i32 = atomicXor(&bar.atom, 5);
+    let _e18: i32 = atomicXor((&bar.atom), 5);
     tmp = _e18;
-    let _e21: i32 = atomicMin(&bar.atom, 5);
+    let _e21: i32 = atomicMin((&bar.atom), 5);
```
kvark, line 47: I assume this is a separate issue that we are over-parenthesizing

kvark (approved):

tests/out/wgsl/access.wgsl
jimblandy: Yeah. Since we decide what operators to emit dynamically, no fixed parenthesization rule can be ideal. To get it right in general, we'd have to pass in information about the precedence of the surrounding operators, or return information about the precedence of the expression generated.

kvark: I think there is a nice middle ground without using precedence - stuff that we do in MSL backend. Basically, when emitting an expression we are passing a flag "am I already bound by some scope?". If the flag is false, then the expression may spawn "(" ")" around it if needed.

jimblandy: Okay, that sounds like basically saying, "Am I a primary expression?" If you're an assignment rhs, or a function argument, or something, then, no parens needed. Anything else, parens are needed. So it's a coarse-grained version of what I had in mind. `bool` is better than Yet Another Enum.

PR #1369, opened 2021-09-15
[spv-in] Move function data into BlockContext

JCapucho:
This allows removing many arguments from functions dealing with
instruction parsing


kvark (approved):
Looks nice!

Issue #1354, opened 2021-09-12
[glsl-in] Support more texture function variations


magcius:
* texelFetchOffset
* textureGatherOffset
* textureGatherOffsets
* textureGradOffset
* textureLodOffset
* textureOffset
* textureProjOffset
* textureProjGrad
* textureProjLod
* textureProjGradOffset
* textureProjLodOffset

Did I miss any?

magcius:
Most of the above now work. Still need:

*    texelFetchOffset
*    textureGatherOffset
*    textureGatherOffsets

PR #1371, opened 2021-09-15
[glsl-in] Add support for a large number of texture sample overrides

magcius:
Partially solves #1354. Still need to write tests.

### DIFF
--- a/src/front/glsl/builtins.rs
+++ b/src/front/glsl/builtins.rs
@@ -39,6 +39,26 @@ impl Module {
     }
 }
 
+fn make_coords_arg(number_of_components: usize, kind: Sk) -> TypeInner {
+    let width = 4;
+
+    match number_of_components {
+        1 => TypeInner::Scalar {
+            kind,
+            width,
+        },
+        _ => TypeInner::Vector {
+            size: match number_of_components {
+                2 => VectorSize::Bi,
+                3 => VectorSize::Tri,
+                _ => VectorSize::Quad,
+            },
+            kind,
+            width,
+        },
+    }
+}
+
 /// Inject builtins into
 ///
 /// This is done to not add a large startup cost and not increase memory
@@ -118,31 +138,79 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
                 ))
             }
         }
-        "texture" | "textureLod" => {
+        "texture" | "textureGrad" | "textureGradOffset" | "textureLod" | "textureLodOffset" | "textureOffset"
+        | "textureProj" | "textureProjGrad" | "textureProjGradOffset" | "textureProjLod"
+        | "textureProjLodOffset" | "textureProjOffset" => {
             // bits layout
-            // bits 0 trough 1 - dims
+            // bits 0 through 1 - dims
             // bit 2 - shadow
             // bit 3 - array
-            // bit 4 - !bias
-            //
-            // 0b11111 is the latest since there are not 3D arrayed shadowed
-            // textures and cube arrayed shadows textures always have a
-            // compare argument
-            for bits in 0..(0b100000) {
+            // bit 4 - extra variant
+            // bit 5 - bias
+
+            for bits in 0..(0b1000000) {
                 let dim = bits & 0b11;
                 let shadow = bits & 0b100 == 0b100;
                 let arrayed = bits & 0b1000 == 0b1000;
-                let bias = bits & 0b10000 == 0b00000;
+                let variant = bits & 0b10000 == 0b10000;
+                let bias = bits & 0b100000 == 0b100000;
+
                 let builtin = match name {
-                    "texture" => MacroCall::Texture,
-                    _ => MacroCall::TextureLod,
+                    // texture(gsampler, gvec P, [float bias]);
+                    "texture" => MacroCall::Texture { proj: false, offset: false, shadow, level_type: TextureLevelType::None, },
+                    // textureGrad(gsampler, gvec P, gvec dPdx, gvec dPdy);
+                    "textureGrad" => MacroCall::Texture { proj: false, offset: false, shadow, level_type: TextureLevelType::Grad, },
+                    // textureGradOffset(gsampler, gvec P, gvec dPdx, gvec dPdy, ivec offset);
+                    "textureGradOffset" => MacroCall::Texture { proj: false, offset: true, shadow, level_type: TextureLevelType::Grad, },
+                    // textureLod(gsampler, gvec P, float lod);
+                    "textureLod" => MacroCall::Texture { proj: false, offset: false, shadow, level_type: TextureLevelType::Lod, },
+                    // textureLodOffset(gsampler, gvec P, float lod, ivec offset);
+                    "textureLodOffset" => MacroCall::Texture { proj: false, offset: true, shadow, level_type: TextureLevelType::Lod, },
+                    // textureOffset(gsampler, gvec+1 P, ivec offset, [float bias]);
+                    "textureOffset" => MacroCall::Texture { proj: false, offset: true, shadow, level_type: TextureLevelType::None, },
+                    // textureProj(gsampler, gvec+1 P, [float bias]);
+                    "textureProj" => MacroCall::Texture { proj: true, offset: false, shadow, level_type: TextureLevelType::None, },
+                    // textureProjGrad(gsampler, gvec+1 P, gvec dPdx, gvec dPdy);
+                    "textureProjGrad" => MacroCall::Texture { proj: true, offset: false, shadow, level_type: TextureLevelType::Grad, },
+                    // textureProjGradOffset(gsampler, gvec+1 P, gvec dPdx, gvec dPdy, ivec offset);
+                    "textureProjGradOffset" => MacroCall::Texture { proj: true, offset: true, shadow, level_type: TextureLevelType::Grad, },
+                    // textureProjLod(gsampler, gvec+1 P, float lod);
+                    "textureProjLod" => MacroCall::Texture { proj: true, offset: false, shadow, level_type: TextureLevelType::Lod, },
+                    // textureProjLodOffset(gsampler, gvec+1 P, gvec dPdx, gvec dPdy, ivec offset);
+                    "textureProjLodOffset" => MacroCall::Texture { proj: true, offset: true, shadow, level_type: TextureLevelType::Lod, },
+                    // textureProjOffset(gsampler, gvec+1 P, ivec offset, [float bias]);
+                    "textureProjOffset" => MacroCall::Texture { proj: true, offset: true, shadow, level_type: TextureLevelType::None, },
+                    _ => unreachable!(),
                 };
 
-                // Shadow, arrayed or both 3D images are not allowed
-                if (shadow || arrayed) && dim == 0b11
-                    || (builtin == MacroCall::TextureLod
-                        && (dim != 0b00 && shadow && arrayed || dim == 0b10 && shadow || !bias))
-                {
+                // Parse out the variant settings.
+                let proj = matches!(builtin, MacroCall::Texture { proj: true, .. });
+                let grad = matches!(builtin, MacroCall::Texture { level_type: TextureLevelType::Grad, .. });
+                let lod = matches!(builtin, MacroCall::Texture { level_type: TextureLevelType::Lod, .. });
+                let offset = matches!(builtin, MacroCall::Texture { offset: true, .. });
+
+                let supports_variant = proj && !shadow;
+                if variant && !supports_variant {
+                    continue;
+                }
+
+                let supports_bias = matches!(builtin, MacroCall::Texture { level_type: TextureLevelType::None, .. }) && !shadow;
+                if bias && !supports_bias {
+                    continue;
+                }
+
+                // Proj doesn't work with arrayed, Cube or 3D samplers
+                if proj && (arrayed || dim == 0b10 || dim == 0b11) {
+                    continue;
+                }
+
+                // 3DArray and 3DShadow are not valid texture types
+                if dim == 0b11 && (arrayed || shadow) {
+                    continue;
+                }
+
+                // It seems that textureGradOffset(samplerCube) is not defined by GLSL for some reason...
+                if dim == 0b10 && grad && offset {
                     continue;
                 }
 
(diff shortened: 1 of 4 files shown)


src/front/glsl/builtins.rs
```diff
@@ -118,34 +138,65 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
+                    break;
+                }
+
+                // 3DArray is not a valid texture type
+                if dim == 0b11 && arrayed {
                     continue;
                 }
 
+                // 3DShadow is not a valid texture type
+                if dim == 0b11 && shadow {
```
JCapucho, line 196: You can make the check in one if, `dim == 0b11 && (shadow || arrayed)`

src/front/glsl/builtins.rs
```diff
@@ -165,184 +216,84 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
+                    0b00 => 1, // 1D
+                    0b01 => 2, // 2D
+                    0b10 => 3, // Cube
+                    _ => 3, // 3D
                 };
+                let mut num_coords = num_coords_from_dim;
 
-                let mut args = vec![image, vector];
+                if shadow {
+                    num_coords += 1;
```
JCapucho, line 228: You can make these as `num_coords += shadow as u8`

JCapucho (changes requested):
Did a light review and found some issues

Cargo.toml
```diff
@@ -32,7 +32,7 @@ hexf-parse = { version = "0.2.1", optional = true }
 [features]
 default = []
 dot-out = []
-glsl-in = ["pp-rs"]
+glsl-in = ["pp-rs", "codespan-reporting"]
```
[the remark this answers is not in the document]
magcius: All I know is that without this, `cargo test --features=glsl-in` fails to compile.

src/front/glsl/builtins.rs
```diff
@@ -118,34 +138,65 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
+                    "textureProjLodOffset" => MacroCall::Texture { proj: true, offset: true, level_type: TextureLevelType::Lod, },
+                    // textureProjOffset(gsampler, gvec+1 P, ivec offset, [float bias]);
+                    "textureProjOffset" => MacroCall::Texture { proj: true, offset: true, level_type: TextureLevelType::None, },
+                    _ => unreachable!(),
                 };
 
-                // Shadow, arrayed or both 3D images are not allowed
-                if (shadow || arrayed) && dim == 0b11
-                    || (builtin == MacroCall::TextureLod
-                        && (dim != 0b00 && shadow && arrayed || dim == 0b10 && shadow || !bias))
```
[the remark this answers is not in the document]
magcius: I believe it's spec-compliant now (after a few fixes)

src/front/glsl/builtins.rs
magcius: This is now changed.

src/front/glsl/builtins.rs
```diff
@@ -165,184 +216,84 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
-                            size,
-                            kind: Sk::Float,
-                            width,
-                        }
-                    }
+                let num_coords_from_dim = match dim {
+                    0b00 => 1, // 1D
+                    0b01 => 2, // 2D
+                    0b10 => 3, // Cube
+                    _ => 3, // 3D
```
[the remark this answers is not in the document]
magcius: min(3) but sure

magcius:
Added a large test suite.

magcius:
Validation fail is: ```
error: unknown type: 'texture_depth_1d'
```

Which implies we shouldn't support shadow samplers on 1D samplers either? Currently the IR doesn't have a problem with it.

**Comment by JCapucho** [comment]
wgsl doesn't indeed specify a 1D depth target, I think the wgsl-out should error on this but pinging @kvark to know what he thinks about it

File: tests/in/glsl/samplers.frag [hunk 6]
```diff
@@ -1,12 +1,229 @@
...
+    // d = textureLodOffset(sampler2DArrayShadow(tex2DArrayShadow, sampShadow), vec4(coord, 1.0), 3.0, ivec2(5));
+    d = textureOffset(sampler2DArrayShadow(tex2DArrayShadow, sampShadow), vec4(coord, 1.0), ivec2(5));
+}
+
+void testTexCube(in vec3 coord) {
+    vec4 c;
+    c = texture(samplerCube(texCube, samp), coord);
+    c = texture(samplerCube(texCube, samp), coord, 2.0);
+    c = textureGrad(samplerCube(texCube, samp), coord, vec3(4.0), vec3(4.0));
+    c = textureGradOffset(samplerCube(texCube, samp), coord, vec3(4.0), vec3(4.0), ivec3(5));
```
JCapucho, line 179: This isn't allowed by the glsl spec and `glslang` will yell at you, `textureGradOffset` doesn't support `Cube` textures.

src/front/glsl/builtins.rs
```diff
@@ -118,31 +138,83 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
+                        continue;
+                    }
+                }
+
+                if bias {
+                    let supports_bias = matches!(builtin, MacroCall::Texture { level_type: TextureLevelType::None, .. }) && !shadow;
+                    if !supports_bias {
+                        continue;
+                    }
+                }
```
JCapucho, line 204: nit: I don't like this nested `if`s you can do `variant && (proj && ! shadow)` or keep the `supports_variant` but move it out of the `if` (compilers will almost certainly inline this) and rust defines short circuiting so there's no additional computations if the first check doesn't pass (it's [defined](https://doc.rust-lang.org/reference/expressions/operator-expr.html#lazy-boolean-operators) at the language level so you can always expect it).

src/front/glsl/builtins.rs
```diff
@@ -118,31 +138,83 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
+                    continue;
+                }
+
+                // 3DArray and 3DShadow are not valid texture types
+                if dim == 0b11 && (arrayed || shadow) {
+                    continue;
+                }
+
+                // Naga IR restriction -- explicit LOD is not allowed on shadow samplers.
+                if lod && shadow {
```
JCapucho, line 217: Check how spv-in handles this https://github.com/gfx-rs/naga/blob/8aae069f22ac94ec46329a9adbfaf341cd3e7adf/src/front/spv/image.rs#L455-L467

It assumes `Zero` and throws an warning, we should do the same

src/front/glsl/builtins.rs
```diff
@@ -165,184 +237,79 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
+                } else if shadow {
+                    num_coords += 1;
+                } else if proj {
+                    if variant && num_coords == 4 {
+                        // Normal form already has 4 components, no need to have a variant form.
+                        continue;
+                    } else if variant {
+                        num_coords = 4;
+                    } else {
+                        num_coords += 1;
```
JCapucho, line 254: nit: When dealing with this kind of `else if` with the same variables I like to use a `match` because it looks like a truth table, something like:

```rust
match (shadow, proj) {
  (true, true) => num_coords = 4,
  (true, false) => num_coords += 1,
  (false, true) => ...,
  (false, false) => {}
}
```

src/front/glsl/builtins.rs
```diff
@@ -165,184 +237,79 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
+                // Special case: texture(gsamplerCubeArrayShadow) kicks the shadow compare ref to a separate argument,
+                // since it would otherwise take five arguments. It also can't take a bias, nor can it be proj/grad/lod/offset
+                // (presumably because nobody asked for it, and implementation complexity?)
+                if num_coords >= 5 {
+                    if lod || grad || offset || proj {
+                        continue;
+                    }
+                    if bias {
+                        continue;
+                    }
```
JCapucho, line 269: maybe merge these two `if`s?

src/front/glsl/builtins.rs
```diff
@@ -165,184 +237,79 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
 
-                let vector = TypeInner::Vector {
-                    size: VectorSize::Quad,
-                    kind: Sk::Float,
-                    width,
-                };
+                    assert!(dim == 0b10 && shadow && arrayed);
```
JCapucho, line 271: This one should also be a `debug_assert`

src/front/glsl/builtins.rs
```diff
@@ -1429,88 +1400,113 @@ impl MacroCall {
...
-                    MacroCall::TextureLod => SampleLevel::Exact(args[2]),
-                    _ => unreachable!(),
-                };
-                texture_call(ctx, args[0], level, comps, body, meta)
-            }
-            MacroCall::TextureGrad(shadow) => {
-                let comps = parser.coordinate_components(ctx, args[0], args[1], meta, body)?;
-                let level = match shadow {
+            MacroCall::Texture{ proj, offset, level_type } => {
+                let coords = match proj {
```
JCapucho, line 1404: nit: this could be rewritten as
```rust
let mut coord = args[1];
if proj {
...
}
```
I think it symbolizes better that argument 1 is the coordinate vector and reduces one level of indentation

src/front/glsl/builtins.rs
```diff
@@ -1429,88 +1400,113 @@ impl MacroCall {
...
-                        SourceMetadata::none(),
-                        body,
-                    );
-                    ctx.vector_resize(size, base, SourceMetadata::none(), body)
+
+                let texture_offset = match offset {
+                    true => {
+                        let offset_arg = args[num_args];
+                        num_args += 1;
+                        Some(parser.solve_constant(ctx, offset_arg, meta)?)
```
JCapucho, line 1496: Not blocking this, but maybe we could put a `map_err` here and provide a better error message explaining the offset must be a constant expression.

src/front/glsl/builtins.rs
```diff
@@ -1429,88 +1400,113 @@ impl MacroCall {
...
-                        right,
-                    },
-                    SourceMetadata::none(),
-                    body,
-                );
-                let comps = parser.coordinate_components(ctx, args[0], coords, meta, body)?;
-                texture_call(ctx, args[0], level, comps, body, meta)
+
+                // Now go back and look for optional bias arg (if available)
+                if level_type == TextureLevelType::None {
```
JCapucho, line 1502: nit: `if let`

JCapucho (changes requested):
There are still some problems but most comments are just stylistic changes

src/front/glsl/builtins.rs
magcius: I think there's code to do that, but it's not quite working, let me look into why.

Cargo.toml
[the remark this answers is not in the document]
JCapucho: I just tested this and it works on master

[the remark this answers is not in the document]
magcius: Just retested on master by running `cargo test --features=glsl-in -p naga`, this is what I see:

```
test src\front\glsl\token.rs - front::glsl::token::SourceMetadata (line 26) ... FAILED

failures:

---- src\front\glsl\token.rs - front::glsl::token::SourceMetadata (line 26) stdout ----
error[E0433]: failed to resolve: use of undeclared crate or module `codespan_reporting`
 --> src\front\glsl\token.rs:28:5
  |
5 | use codespan_reporting::diagnostic::Label;
  |     ^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `codespan_reporting`

error[E0433]: failed to resolve: use of undeclared type `Label`
 --> src\front\glsl\token.rs:31:13
  |
8 | let label = Label::primary(file, meta);
  |             ^^^^^ use of undeclared type `Label`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0433`.
Couldn't compile the test.

failures:
    src\front\glsl\token.rs - front::glsl::token::SourceMetadata (line 26)

test result: FAILED. 3 passed; 1 failed; 4 ignored; 0 measured; 0 filtered out; finished in 0.71s

error: test failed, to rerun pass '--doc'
```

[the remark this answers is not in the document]
JCapucho: Oh ok I see the problem

src/front/glsl/builtins.rs
magcius: I find this much more difficult to read.

Cargo.toml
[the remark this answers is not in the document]
JCapucho: Fixed on master, can you rebase and remove this

src/front/glsl/builtins.rs
JCapucho: understandable, ignore this then

tests/in/glsl/samplers.frag
magcius: That's a fun exception! lol

Cargo.toml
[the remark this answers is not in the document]
magcius: Will do, thanks!

src/front/glsl/builtins.rs
magcius: I don't quite understand what this is asking for?

src/front/glsl/builtins.rs
magcius: I have an ugly way to this, please suggest something better!

src/front/glsl/builtins.rs
magcius: Ah, it was only set up to work on Grad, it never worked on Lod. Fixed.

src/front/glsl/builtins.rs
JCapucho: Use `if let TextureLevelType::None = level_type`

src/front/glsl/builtins.rs
```diff
@@ -118,31 +138,79 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
+                    continue;
+                }
+
+                // 3DArray and 3DShadow are not valid texture types
+                if dim == 0b11 && (arrayed || shadow) {
+                    continue;
+                }
+
+                // It seems that textureGradOffset(samplerCube) is not defined by GLSL for some reason...
+                if dim == 0b10 && (grad && offset) {
```
JCapucho, line 213: nit: the braces are not needed

src/front/glsl/builtins.rs
magcius: Ah, thank you. Fixed.

src/front/glsl/builtins.rs
```diff
@@ -1429,88 +1393,127 @@ impl MacroCall {
...
-                let left = if let VectorSize::Bi = size {
-                    ctx.add_expression(
-                        Expression::AccessIndex { base, index: 0 },
+            MacroCall::Texture { proj, offset, shadow, level_type } => {
+                let mut coords = args[1];
+
+                if proj {
+                    let size = match *parser.resolve_type(ctx, args[1], meta)? {
+                        TypeInner::Vector { size, .. } => size,
+                        _ => {
```
JCapucho, line 1402: nit: this is unreachable by design, maybe use `VectorSize::Bi` as default and add a comment saying it's unreachable as the functions are type checked

src/front/glsl/builtins.rs
```diff
@@ -1429,88 +1393,127 @@ impl MacroCall {
...
-                    body,
-                );
-                let left = if let VectorSize::Bi = size {
-                    ctx.add_expression(
-                        Expression::AccessIndex { base, index: 0 },
+            MacroCall::Texture { proj, offset, shadow, level_type } => {
+                let mut coords = args[1];
+
+                if proj {
+                    let size = match *parser.resolve_type(ctx, args[1], meta)? {
```
JCapucho, line 1400: nit: you can use `coords` here instead of `args[1]`

src/front/glsl/builtins.rs
```diff
@@ -1429,88 +1393,127 @@ impl MacroCall {
...
+
+                // Parse out explicit texture level.
+                let mut level = match level_type {
+                    TextureLevelType::None => SampleLevel::Auto,
+
+                    TextureLevelType::Lod => {
+                        num_args += 1;
+
+                        if shadow {
+                            log::debug!(
```
JCapucho, line 1463: We should follow spv-in and use warn

src/front/glsl/builtins.rs
```diff
@@ -1429,88 +1393,127 @@ impl MacroCall {
...
+
+                let texture_offset = match offset {
+                    true => {
+                        let offset_arg = args[num_args];
+                        num_args += 1;
+                        match parser.solve_constant(ctx, offset_arg, meta) {
+                            Ok(v) => Some(v),
+                            Err(e) => {
+                                parser.errors.push(e);
+                                None
```
JCapucho, line 1501: Very nice, non exiting error reporting is always good to have when it doesn't break further error handling

JCapucho (approved):
Everything looks good minus some nits, if you can fix them it would be great otherwise we can merge and fix them later

src/front/glsl/builtins.rs
magcius: This code came straight from the older proj code, I believe. Should I just use `unreachable!()`?

src/front/glsl/builtins.rs
magcius: It would be nice to have a helper for this eventually.

src/front/glsl/builtins.rs
magcius: This was pre-existing, but fixed.

src/front/glsl/builtins.rs
magcius: Decided to use unreachable!()

JCapucho:
Everything looks good to me, thank you.

PR #1372, opened 2021-09-15
spv-out: option to emit PointSize

### LINKED ISSUE #1070: Validation error for point_list
Since there is no `point_size` in WGSL, I omit it from a shader in our tests:
```wgsl
        struct VertexOutput {
            //[[builtin(pointSize]] point_size: f32;
        };

        [[stage(vertex)]]
        fn vs_main([[builtin(vertex_index)]] vertex_index : u32) -> VertexOutput {
            var positions: array<vec3<f32>, 4> = array<vec3<f32>, 4>(
                vec3<f32>(-0.5, -0.5, 0.0),
                vec3<f32>(-0.5,  0.5, 0.0),
                vec3<f32>( 0.5, -0.5, 0.2),
                vec3<f32>( 0.5,  0.5, 0.2),
            );
            var out: VertexOutput;
            out.position =  vec4<f32>(positions[vertex_index], 1.0);
            //out.point_size = 16.0;
            return out;
        }

        [[stage(fragment)]]
        fn fs_main() -> [[location(0)]] vec4<f32> {
            return vec4<f32>(1.0, 0.499, 0.0, 1.0);
        }
``` 

Earlier, this kindof worked. Only points of 1px, but ah well... However, with the wgpu-native from latest master I now get:
```
VALIDATION [UNASSIGNED-CoreValidation-Shader-PointSizeMissing (0xf3693078)]
	Validation Error: [ UNASSIGNED-CoreValidation-Shader-PointSizeMissing ] Object 0: VK_NULL_HANDLE, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xf3693078 | Pipeline topology is set to POINT_LIST, but PointSize is not written to in the shader corresponding to VK_SHADER_STAGE_VERTEX_BIT.
	objects: (type: PIPELINE, hndl: 0x0, name: ?)
```

kvark:
Fixes #1070


tests/snapshots.rs
```diff
@@ -54,21 +79,8 @@ struct Parameters {
...
-    #[cfg_attr(not(feature = "spv-out"), allow(dead_code))]
     #[serde(default)]
-    spv_debug: bool,
-    #[cfg_attr(not(feature = "spv-out"), allow(dead_code))]
-    #[serde(default)]
-    spv_adjust_coordinate_space: bool,
-    #[cfg_attr(not(feature = "spv-out"), allow(dead_code))]
-    #[serde(default)]
-    spv_separate_entry_points: bool,
+    spv: SpirvOutParameters,
```
jimblandy, line 83: Should this still be `#[cfg_attr(not(feature = "spv-out"), allow(dead_code)]`?

tests/in/access.param.ron
```diff
@@ -1,7 +1,9 @@
 (
-	spv_version: (1, 1),
-	spv_debug: true,
-	spv_adjust_coordinate_space: false,
+	spv: (
+		version: (1, 1),
+		debug: true,
+		adjust_coordinate_space: false,
```
jimblandy, line 5: We should omit the values that are the default, right?

tests/in/extra.param.ron
```diff
@@ -1,6 +1,8 @@
 (
 	god_mode: true,
-	spv_version: (1, 0),
+	spv: (
+		version: (1, 2),
```
jimblandy, line 4: This change doesn't seem to make any substantive difference in the output. Should we just let it default?

jimblandy (approved):
Looks good to me!

tests/snapshots.rs
kvark: nicely spotted!

tests/in/access.param.ron
kvark: we *could* but not necessarily should :)

tests/in/extra.param.ron
kvark: I thought it's going to be better in the long term if the "extra" test targeted the highest SPIR-V version, since it would need to use some capabilities not available in early versions anyway.

PR #1380, opened 2021-09-16
[glsl-in] Use local for constant indexing

### LINKED ISSUE #1343: [glsl-in] Shader rejected with "The expression ... may only be indexed by a constant"
Tested using naga 7681f4a

[This shader](https://github.com/gfx-rs/naga/files/7144218/stable_bifurcation.json.frag.txt) is accepted by glslangValidator, but with naga it leads to a complaint about indexing via a non-constant. From the error message I couldn't work out which part of the shader is being complained about.

JCapucho:
Constants that are to be dynamically indexed now are first transformed
into locals with them as the initializer.

Closes #1343


kvark (approved):

kvark:
Would be great to have this tested.

PR #1381, opened 2021-09-16
[wgsl-in]: Correctly compare pointer types.

### LINKED ISSUE #1318: Naga complains of type error on spec example code
This shader:
```
fn compute()
{
    var uv: vec2<f32>;
    uv.y = 1.25;      // Place 1.25 into the second component of uv.

    let uv_x_ptr: ptr<function,f32> = &uv.x;
    *uv_x_ptr = 2.5;   // Place 2.5 into the first component of uv.
}
```
Fails to compile on naga 0.6.0 with the type error:

```
[2021-09-02T04:23:09Z ERROR naga::front::wgsl] Given type Pointer { base: [2], class: Function } doesn't match expected ValuePointer { size: None, kind: Float, width: 4, class: Function }
error: the type of `uv_x_ptr` is expected to be [3]
  ┌─ wgsl:7:9
  │
7 │     let uv_x_ptr: ptr<function,f32> = &uv.x;
  │         ^^^^^^^^ definition of `uv_x_ptr`

Could not parse WGSL
```

Ignoring the fact that I don't know what `[3]` is supposed to be, this code as far as I am aware should compile, it's drawn from the working draft spec [here](https://www.w3.org/TR/WGSL/#assignment).

This seems like a pretty simple write into a pointer, and I really can't imagine why it wouldn't compile.

(as an aside, naga should support some kind of `--version` option, if only for debugging and issue reporting purposes.)

jimblandy:
Treat `TypeInner::ValuePointer` and `TypeInner::Pointer` as equivalent by
converting them to a canonical form before comparison.

Support `ValuePointer` in WGSL type output.

Fixes #1318.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -630,15 +630,64 @@ impl<W: Write> Writer<W> {
             }
             TypeInner::Pointer { base, class } => {
                 let (storage, maybe_access) = storage_class_str(class);
+                // Everything but `StorageClass::Handle` gives us a `storage` name, but
+                // Naga IR never produces pointers to handles, so it doesn't matter much
+                // how we write such a type. Just write it as the base type alone.
                 if let Some(class) = storage {
                     write!(self.out, "ptr<{}, ", class)?;
-                    if let Some(access) = maybe_access {
-                        write!(self.out, ", {}", access)?;
-                    }
                 }
                 self.write_type(module, base)?;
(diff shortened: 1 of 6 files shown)


kvark:
This is definitely a welcome improvement! Let's chat about #1385 briefly before we proceed.

jimblandy:
This has been merged into master now. Not sure why GitHub doesn't think so.

PR #1384, opened 2021-09-17
[wgsl-out] Correct handling of named pointer expressions.

### LINKED ISSUE #1382: Naga produces incorrect WGSL for let-bound pointers
The following WGSL input generates incorrect WGSL output:

```
fn f() {
   var v: vec2<i32>;
   let px = &v.x;
   *px = 10;
}
```

The output omits the `*`:

```
fn f() {
    var v: vec2<i32>;

    let px: ptr<function, i32> = (&v.x);
    px = 10;
    return;
}
```

This is because the fix for #1332 does not properly address named expressions.

jimblandy:
Depends on #1381.

Treat expressions in `Function::named_expressions` like WGSL `let` declarations,
assuming that the Load Rule was applied to the rhs of the declaration, meaning
that their values are always `Indirection::Ordinary`.

Split `write_expr_plain_form` out from `write_expr_with_indirection`, to clean
up the parenthesis generation: no more `opened_paren` variable, just function
calls. This makes the early return for named expressions neater.

Fixes #1382.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -51,56 +51,6 @@ enum Indirection {
     Reference,
 }
 
-/// Return the sort of indirection that `expr`'s plain form evaluates to.
-///
-/// An expression's 'plain form' is the most general rendition of that
-/// expression into WGSL, lacking `&` or `*` operators:
-///
-/// - The plain form of `LocalVariable(x)` is simply `x`, which is a reference
-///   to the local variable's storage.
-///
-/// - The plain form of `GlobalVariable(g)` is simply `g`, which is usually a
-///   reference to the global variable's storage. However, globals in the
-///   `Handle` storage class are immutable, and `GlobalVariable` expressions for
-///   those produce the value directly, not a pointer to it. Such
-///   `GlobalVariable` expressions are `Ordinary`.
-///
-/// - `Access` and `AccessIndex` are `Reference` when their `base` operand is a
-///   pointer. If they are applied directly to a composite value, they are
-///   `Ordinary`.
-///
-/// Note that `FunctionArgument` expressions are never `Reference`, even when
-/// the argument's type is `Pointer`. `FunctionArgument` always evaluates to the
-/// argument's value directly, so any pointer it produces is merely the value
-/// passed by the caller.
-fn plain_form_indirection(
-    expr: Handle<crate::Expression>,
-    module: &Module,
-    func_ctx: &back::FunctionCtx<'_>,
-) -> Indirection {
-    use crate::Expression as Ex;
-    match func_ctx.expressions[expr] {
-        Ex::LocalVariable(_) => Indirection::Reference,
-        Ex::GlobalVariable(handle) => {
-            let global = &module.global_variables[handle];
-            match global.class {
-                crate::StorageClass::Handle => Indirection::Ordinary,
-                _ => Indirection::Reference,
-            }
-        }
-        Ex::Access { base, .. } | Ex::AccessIndex { base, .. } => {
-            let base_ty = func_ctx.info[base].ty.inner_with(&module.types);
-            match *base_ty {
-                crate::TypeInner::Pointer { .. } | crate::TypeInner::ValuePointer { .. } => {
-                    Indirection::Reference
-                }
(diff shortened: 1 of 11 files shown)


tests/in/pointers.wgsl
```diff
@@ -0,0 +1,5 @@
+fn f() {
+   var v: vec2<i32>;
```
kvark, line 2: can we just roll this into `access.wgsl` instead?

src/back/hlsl/writer.rs
```diff
@@ -874,7 +874,7 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
             }
             _ => {
                 return Err(Error::Unimplemented(format!(
-                    "write_value_type {:?}",
+                    "back::hlsl::Writer::write_value_type {:?}",
```
kvark, line 877: is this necessary?

src/back/wgsl/writer.rs
```diff
@@ -630,20 +580,69 @@ impl<W: Write> Writer<W> {
...
+                    return Err(Error::Unimplemented(format!(
+                        "ValuePointer to StorageClass::Handle {:?}",
+                        inner
+                    )));
                 }
             }
             _ => {
                 return Err(Error::Unimplemented(format!(
-                    "write_value_type {:?}",
+                    "back::wgsl::Writer::write_value_type {:?}",
```
kvark, line 645: and this

src/proc/mod.rs
```diff
@@ -130,6 +132,50 @@ impl super::TypeInner {
...
+
+    /// Return the canoncal form of `self`, or `None` if it's already in
+    /// canonical form.
+    ///
+    /// Certain types have multiple representations in `TypeInner`. This
+    /// function converts all forms of equivalent types to a single
+    /// representative of their class, so that simply applying `Eq` to the
+    /// result indicates whether the types are equivalent, as far as Naga IR is
+    /// concerned.
+    pub fn canonical_form(&self, types: &crate::Arena<crate::Type>) -> Option<crate::TypeInner> {
```
kvark, line 144: Can we really have the `Eq` on types working? I know we are doing it in the validator, and it was always a little bit concerning. For example, today the IR may have 2 different types (A and B) that are `u32`, and comparing them would work because they are compared by value. But wrapping them into anything will not work, e.g `array<A, 3>` vs `array<B, 3>` would compare to false because the base type is just a handle, and the handle is different...

So this function here warrants a much larger discussion, which I filed https://github.com/gfx-rs/naga/issues/1385 for. Perhaps, at least adding a comment here saying "this only works for a specific kind of canonicalization"?

kvark (changes requested):
Looks great! Just a few notes

tests/in/pointers.wgsl
jimblandy: I tried that at first. But we generate all the output formats for `access.wgsl`, and we don't have a pointer story yet for GLSL, HLSL, or even MSL. So I put this in its own file so that we could just generate WGSL and SPIR-V.

src/back/hlsl/writer.rs
jimblandy: I added this because the other back end generated an identical error message, and I ended up jumping to the conclusion that I was looking at the wrong code, which befuddled me for a few minutes.

tests/in/pointers.wgsl
kvark: ok, fair

src/back/hlsl/writer.rs
kvark: Where did you see the error? Is it from snapshots?
I just don't think this scales well. We don't want to have long prefixes for each and every error in each backend.
Instead, the `snapshot.rs` could just `println!("Testing MSL")`, and that's it.

jimblandy: Okay, that's a good alternative.

jimblandy: I've implemented this in #1427. I'll remove the change here.

src/proc/mod.rs
jimblandy: This should be fixed by #1419.

kvark (approved):

PR #1387, opened 2021-09-17
[glsl-in] Improve pre/postfix handling

### LINKED ISSUE #1344: [glsl-in] Increment on float not supported
Tested using naga 7681f4a

GLSL (unfortunately!) allows increment and decrement operators on floats, e.g.:

```
#version 450

void main() {
  float x = 0.0;
  x++;
}
```

Naga complains about these:

error: Increment/decrement operations must operate in integers.

JCapucho:
Add support for float, vector and matrices targets.

Fix prefix and postfix being inverted (one was returning the value
of the other).

Remove an unneeded local indirection for prefix handling.

Add tests.

Closes #1344

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -123,9 +123,13 @@ pub enum HirExprKind {
         tgt: Handle<HirExpr>,
         value: Handle<HirExpr>,
     },
-    IncDec {
-        increment: bool,
+    /// A prefix/postfix operator like `++`
+    PrePostfix {
+        /// The operation to be performed
+        op: BinaryOperator,
+        /// Whether this is a postfix or a prefix
         postfix: bool,
+        /// The target expression
         expr: Handle<HirExpr>,
     },
 }
--- a/src/front/glsl/context.rs
+++ b/src/front/glsl/context.rs
@@ -750,102 +750,108 @@ impl Context {
 
                 value
             }
-            HirExprKind::IncDec {
-                increment,
-                postfix,
-                expr,
-            } => {
-                let op = match increment {
-                    true => BinaryOperator::Add,
-                    false => BinaryOperator::Subtract,
-                };
-
+            HirExprKind::PrePostfix { op, postfix, expr } if ExprPos::Lhs != pos => {
                 let pointer = self
                     .lower_expect_inner(stmt, parser, expr, ExprPos::Lhs, body)?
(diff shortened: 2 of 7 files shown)


kvark (commented):
I don't know what's going on here. Do you want to have a stronger review, or confident otherwise?

src/front/glsl/ast.rs
```diff
@@ -123,9 +123,13 @@ pub enum HirExprKind {
         tgt: Handle<HirExpr>,
         value: Handle<HirExpr>,
     },
-    IncDec {
-        increment: bool,
+    /// A prefix/postfix operator like `++`
+    Infix {
```
magcius, line 127: This is the exact opposite of infix!

Infix is the + in 2+2, postfix is 2+, prefix is +2

JCapucho: hmm, you're right english isn't my first language so I confused the meaning, doesn't seem like english has a direct equivalent of the word I initially thought of.

JCapucho:
I'm confident in the code and there are tests backing it, but I always prefer a strong review, the important bits are in the changes to `context.rs`, the others changes are renaming from `IncDec` to `PrePostfix` which is in my opinion more clear, passing directly the `BinaryOperator` instead of a boolean saying whether it was increment or decrement and adding some documentation to it.

The lowering for `Infix` which was the part that changed in `context.rs` now works by first from the type of the target expression getting some information, the `TypeInner` (this is just like cloning), a `ConstantInner` which always has the value of `1` but this is based on the kind and width of the target's type, an optional `VectorSize` which symbolizes the `rows` if the type is matrix or the `size` if it's a vector and yet another optional which is the matrix columns if the type is a matrix. This is all wrapped in a `Option` so that types or kinds that aren't allowed are treated in the next match as an error and the loaded value is returned.

Then we add the `ConstantInner` to the arena and add an expression to it, then we check to see if `rows` is some if it's we must splat the result, furthermore if `columns` is some we need  to compose a matrix.

Finally we add the `Binary` expression flush the `Emit` and add a `Store` and return the expression to be used, if it's postfix we want to use `left` since that contains the target value before the operation, otherwise it's an infix and we return `value` which is the final value.

magcius:
The patch looks sensible to me.

JCapucho:
@magcius it should now be infix free

PR #1388, opened 2021-09-17
[glsl-in] Cast switch selector to int if it's uint

### LINKED ISSUE #1342: [glsl-in] Shader rejected with "The `switch` value ... is not an integer scalar"
Tested using naga 7681f4a

[This shader](https://github.com/gfx-rs/naga/files/7144201/matrices_smart_loops.json.frag.txt) is accepted by glslangValidator, but with naga it leads to a complaint that the value associated with a switch is not an integer scalar.

I think the problem is that it's a uint, rather than an int. A uint is allowed in GLSL.

JCapucho:
Closes #1342

### DIFF
--- a/src/front/glsl/parser/functions.rs
+++ b/src/front/glsl/parser/functions.rs
@@ -153,12 +153,20 @@ impl<'source> ParsingContext<'source> {
                 let end_meta;
 
                 self.expect(parser, TokenValue::LeftParen)?;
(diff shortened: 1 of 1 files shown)


kvark (approved):

kvark:
we'll probably need to support `uint` selectors as well in the IR

PR #1389, opened 2021-09-17
[glsl-in] Add test for array constant dynamic indexing

JCapucho:
Folllow up to #1380

I didn't find a test shader to add this to, so I created a new one


kvark (approved):
Thank you!

PR #1390, opened 2021-09-17
[wgsl-in] Disallow identifiers starting with _

### LINKED ISSUE #1386: naga should reject identifiers beginning with underscore
This should fail compilation.  But Naga allows it.
```
let _123 = 123;
```

JCapucho:
Closes #1386


magcius:
We'll need to change wgsl-out and change all of the wgsl-out tests as well, no?

JCapucho:
Yes I'm also working on it, but I think they each deserve their own PR since the frontend isn't tied to the backend

kvark (approved):

PR #1392, opened 2021-09-17
Avoid case fallthrough when last statement is break, continue, kill or return

DDoS:
Naga is generating `break; fallthrough;` statements at the end of each case for any switch statement in my shaders (SPIR-V compiled from GLSL). The tint WGSL compiler used by Chromium rejects these shaders because the `fallthrough;` statement is obviously unreachable code. These changes prevent generating a `fallthrough` when the last statement of the case body always exits the switch.

I've never written Rust before, let me know if this idiomatic code.

### DIFF
--- a/src/front/spv/function.rs
+++ b/src/front/spv/function.rs
@@ -570,11 +570,19 @@ impl<'function> BlockContext<'function> {
                                     .map(|&(value, body_idx)| {
                                         let body = lower_impl(blocks, bodies, body_idx);
 
+                                        // Handle simple cases that would make a fallthrough statement unreachable code
+                                        let fall_through = match body.last() {
+                                            Some(&crate::Statement::Break)
+                                            | Some(&crate::Statement::Continue)
+                                            | Some(&crate::Statement::Kill)
+                                            | Some(&crate::Statement::Return { .. }) => false,
+                                            _ => true,
+                                        };
+
                                         crate::SwitchCase {
                                             value,
                                             body,
-                                            // TODO
-                                            fall_through: true,
+                                            fall_through,
                                         }
                                     })
                                     .collect(),


src/front/spv/function.rs
```diff
@@ -570,11 +570,19 @@ impl<'function> BlockContext<'function> {
                                     .map(|&(value, body_idx)| {
                                         let body = lower_impl(blocks, bodies, body_idx);
 
+                                        // Handle simple cases that would make a fallthrough statement unreachable code
+                                        let fall_through = match body.last() {
+                                            Some(&crate::Statement::Break) => false,
+                                            Some(&crate::Statement::Continue) => false,
+                                            Some(&crate::Statement::Kill) => false,
+                                            Some(&crate::Statement::Return { value: _ }) => false,
```
JCapucho, line 578: You can replace `value: _` with `..`

src/front/spv/function.rs
```diff
@@ -570,11 +570,19 @@ impl<'function> BlockContext<'function> {
                                     .map(|&(value, body_idx)| {
                                         let body = lower_impl(blocks, bodies, body_idx);
 
+                                        // Handle simple cases that would make a fallthrough statement unreachable code
+                                        let fall_through = match body.last() {
+                                            Some(&crate::Statement::Break) => false,
+                                            Some(&crate::Statement::Continue) => false,
+                                            Some(&crate::Statement::Kill) => false,
```
JCapucho, line 577: Instead of doing
```rust
arm => false,
arm => false,
arm => false,
```
You can do
```rust
arm | arm | arm => false
```

JCapucho (changes requested):
Thank you for your contribution, there are just some styling fixes nothing major

DDoS:
I made the requested changes. Is there a way to auto format the code?

JCapucho:
you should be able to run `cargo fmt` in the terminal

DDoS:
Running  `cargo fmt` only changed the file `src/front/glsl/builtins.rs`, so I assume my code is already correctly formatted.

JCapucho (approved):
lgtm

PR #1393, opened 2021-09-17
Don't output identifiers starting with _

JCapucho:
Follow up to #1390 

Changes the bake prefix to `NAGA_e` instead of `_e` and adds a check to the namer, most of the changes are test outputs.

This is needed because wgsl doesn't allow identifiers starting with _


kvark:
this is somewhat unfortunate. The code littered with NAGA doesn't look nearly as good.
How about just `e`?

JCapucho:
@kvark updated the PR to use `e`

kvark (approved):

kvark:
Can we limit this to WGSL-out only? It seems to me that "_e" is still better for backends that allow it.

JCapucho:
Sure I was just reusing the code that was already there, but I can add a wgsl-out only bake prefix

kvark:
Let's do it. Thank you!

JCapucho:
@kvark done

kvark (approved):
thank you!

PR #1398, opened 2021-09-19
Add support for multi view extensions

JCapucho:
Adds support for [`VK_KHR_multiview`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_multiview.html).

Adds the `ViewIndex` builtin which is only supported in spirv and glsl (2/5 backends), spirv-cross has two types of handling for this, for metal it uses a variable with the value 0, for hlsl it throws an error. This PR errors on both backends.

I understand that since this is not part of webgpu it's not a priority, but I needed it for a project which consumed spirv and used it in both vulkan and gles for stereo rendering and spirv-cross was not liking being cross compiled so I ended up patching naga and using it, so feel free to close.


kvark (approved):
Looks pretty clean!

PR #1399, opened 2021-09-19
Allow unsigned integers in switch

### LINKED ISSUE #1029: [wgsl-in] Naga complains about unsigned literal in switch statement
Naga (1f42d4f) rejects this WGSL (which was generated by Tint, via the example from issue #1028):

```
[[stage(fragment)]]
fn main() {
  switch(0u) {
    default: {
      if (false) {
        break;
      }
    }
  }
  return;
}
```

I get this output:

```
$ RUST_BACKTRACE=1 cargo run temp.wgsl blah.spv
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/naga temp.wgsl blah.spv`
Entry point main at Fragment is invalid: 
	The `switch` value [1] is not an integer scalar
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cli/src/main.rs:272:59
stack backtrace:
   0: rust_begin_unwind
             at /rustc/24bdc6d73a75dce9a7013ebc7c037013ff4ea099/library/std/src/panicking.rs:515:5
   1: core::panicking::panic_fmt
             at /rustc/24bdc6d73a75dce9a7013ebc7c037013ff4ea099/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/24bdc6d73a75dce9a7013ebc7c037013ff4ea099/library/core/src/panicking.rs:50:5
   3: core::option::Option<T>::unwrap
             at /rustc/24bdc6d73a75dce9a7013ebc7c037013ff4ea099/library/core/src/option.rs:388:21
   4: naga::main
             at ./cli/src/main.rs:272:45
   5: core::ops::function::FnOnce::call_once
             at /rustc/24bdc6d73a75dce9a7013ebc7c037013ff4ea099/library/core/src/ops/function.rs:227:5
```

JCapucho:
The wgsl spec defines that the switch selector must be a scalar integer type, this encompasses both signed and unsigned integers. This PR adds support for unsigned integers.

Furthermore the spec defines that the type of the selector and the case value type must match this PR also adds validation for that.

Closes #1029

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -887,6 +887,13 @@ impl<W: Write> Writer<W> {
                 let all_fall_through = cases
                     .iter()
                     .all(|case| case.fall_through && case.body.is_empty());
+                let type_postfix = match *func_ctx.info[selector].ty.inner_with(&module.types) {
+                    crate::TypeInner::Scalar {
+                        kind: crate::ScalarKind::Uint,
+                        ..
+                    } => "u",
+                    _ => "",
+                };
 
                 let l2 = level.next();
                 if !cases.is_empty() {
@@ -896,11 +903,11 @@ impl<W: Write> Writer<W> {
                         }
                         if !all_fall_through && case.fall_through && case.body.is_empty() {
                             write_case = false;
-                            write!(self.out, "{}, ", case.value)?;
+                            write!(self.out, "{}{}, ", case.value, type_postfix)?;
                             continue;
                         } else {
                             write_case = true;
-                            writeln!(self.out, "{}: {{", case.value)?;
+                            writeln!(self.out, "{}{}: {{", case.value, type_postfix)?;
                         }
 
                         for sta in case.body.iter() {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1260,7 +1260,7 @@ pub use block::Block;
 pub struct SwitchCase {
     /// Value, upon which the case is considered true.
     pub value: i32,
-    /// Body of the cae.
+    /// Body of the case.
     pub body: Block,
     /// If true, the control flow continues to the next case in the list,
     /// or default.
--- a/tests/in/control-flow.wgsl
+++ b/tests/in/control-flow.wgsl
@@ -32,6 +32,12 @@ fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
         }
     }
 
+	// switch with unsigned integer selectors
+	switch(0u) {
+		case 0u: {
+		}
+	}
+
     // non-empty switch in last-statement-in-function position
     switch (pos) {
         case 1: {
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1427,11 +1427,18 @@ impl<'a, W: Write> Writer<'a, W> {
                 write!(self.out, "switch(")?;
                 self.write_expr(selector, ctx)?;
                 writeln!(self.out, ") {{")?;
+                let type_postfix = match *ctx.info[selector].ty.inner_with(&self.module.types) {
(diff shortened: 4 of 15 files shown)


src/lib.rs
```diff
@@ -1251,15 +1251,24 @@ pub enum Expression {
...
 /// A case for a switch statement.
 // Clone is used only for error reporting and is not intended for end users
 #[derive(Clone, Debug)]
 #[cfg_attr(feature = "serialize", derive(Serialize))]
 #[cfg_attr(feature = "deserialize", derive(Deserialize))]
 pub struct SwitchCase {
     /// Value, upon which the case is considered true.
-    pub value: i32,
-    /// Body of the cae.
+    pub value: SwitchCaseValue,
```
kvark, line 1270: is this really needed?
  1. different cases inside the same switch must have the same type
  2. both i32 and u32 can be losslessly converted to each other

So it seems to me that we could just keep `i32` (or `u32`) in here, and just interpret it differently based on some bit inside the switch itself.

kvark (changes requested):
Got a few questions about the IR

src/lib.rs
JCapucho: My fear is how you would check that they have the same type?

src/back/mod.rs
```diff
@@ -311,3 +311,12 @@ impl crate::TypeInner {
         }
     }
 }
+
+impl std::fmt::Display for crate::SwitchCaseValue {
```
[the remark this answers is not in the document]
JCapucho: I can move the implementation elsewhere but it's useful since all backends use the C-like syntax for their integer literals, except spirv which is a binary representation.

src/lib.rs
kvark: WGSL frontend can check this. The IR would just assume they are the same type

JCapucho: And the backends would need to check the type of the selector and write the postfix as needed?

kvark: correct, yes

JCapucho:
@kvark updated to always use `i32`

src/back/wgsl/writer.rs
```diff
@@ -880,13 +880,22 @@ impl<W: Write> Writer<W> {
...
+                    .scalar_kind()
+                    == Some(crate::ScalarKind::Uint);
                 writeln!(self.out, ") {{")?;
 
                 // Write all cases
                 let mut write_case = true;
                 let all_fall_through = cases
                     .iter()
                     .all(|case| case.fall_through && case.body.is_empty());
+                let postfix = || match uint {
```
kvark, line 895: using closures here seems like an overkill. Couldn't we just do:
```rust
let type_postfix = match func_ctx.info[selector].ty.inner_with(&module.types) {
  crate::TypeInner::Scalar { kind: crate::ScalarKind::Uint, .. } => "u",
  _ => "",
};
```
I think doing this in other places would also improve things a little.

kvark (commented):
Thank you! just one last thing

src/back/wgsl/writer.rs
JCapucho: ok I realized what you meant now

tests/in/control-flow.wgsl
```diff
@@ -32,6 +32,13 @@ fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
         }
     }
 
+	// switch with unsigned integer selectors
+	switch(0u) {
+		case 0u: {
+			break;
```
kvark, line 38: why do we `break` here? We can do it, but it seems irrelevant to the PR.

kvark (changes requested):

kvark (approved):

PR #1401, opened 2021-09-20
Better dead code detection

### LINKED ISSUE #1019: Naga doesn't detect unreachable code
The following test, if added to `wgsl-errors.rs`, should fail, but it passes:

```
check_validation_error! {
    "
    fn dead_code_after_if(condition: bool) -> i32 {
        if (condition) {
            return 1;
        } else {
            return 2;
        }
        return 3;
    }
    ",
    "
    fn dead_code_after_block() -> i32 {
        {
            return 1;
        }
        return 2;
    }
    ":
    Ok(_)
}
```

JCapucho:
Introduces a new enum `FinishState` which is passed with the `valid_stages` trough `BlockInfo` for parent scope analysis. The enum has three states

- `None` - the block didn't have any instruction that would cause execution to finish in that scope
- `Finished` - the block had an instruction that caused execution to end in that scope
- `Returned` - the block had an instruction that caused execution to end in the function scope, possibly the invocation (`Kill`)

If the current `FinishState` is `None` now further statements are allowed.

The tree states are needed to compute `FinishState` in branching/looping constructs, if the current statement is an
- `If` - if both blocks are `Returned`, the parent scope also is, otherwise if both aren't `None` the parent scope is `Finished`, else the parent scope is `None`.
- `Switch` - the same logic as `If` applies but instead of it applying to just the `accept` and `reject` it applies to all cases including the default case.
- `Loop` - the loop has special logic, if the loop body is `Returned` than so is the parent otherwise it's `None`.

Closes #1019

### DIFF
--- a/src/valid/function.rs
+++ b/src/valid/function.rs
@@ -135,6 +135,32 @@ bitflags::bitflags! {
     }
 }
 
+#[derive(Clone, Copy, PartialEq)]
+enum FinishState {
+    /// The block didn't finish execution
+    None,
+    /// The block finished execution locally
+    Finished,
+    /// The block returned ending execution in function
+    Returned,
+}
+
+impl FinishState {
+    /// Merges two states to determine the combined state
+    pub fn merge(&mut self, other: FinishState) {
+        *self = match (*self, other) {
+            (FinishState::Returned, FinishState::Returned) => FinishState::Returned,
+            (FinishState::None, _) | (_, FinishState::None) => FinishState::None,
+            _ => FinishState::Finished,
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


kvark:
Thank you for tackling this!
Unfortunately, we can't just decide on our own what WGSL code is valid or not, we have to follow the spec.
This PR needs to match https://github.com/gpuweb/gpuweb/pull/2096

JCapucho:
Oh ok I didn't find any thing when searching, from what I understand only compound statements propagate their state is that right?

JCapucho:
Okay I'll close this one and open a new PR with just the compound changes

PR #1405, opened 2021-09-20
wgsl-out: vector bitcast

DDoS:
Bitcasting a vector to a vector is valid according to the WGSL spec: https://www.w3.org/TR/WGSL/#bitcast-expr

The current wgsl-out code generates an unsigned-integer-to-float  conversion expression when compiling `uintBitsToFloat(uvec3)`, instead of the expected bitcast.


JCapucho:
Run `cargo test` to update the snapshot tests and them to the PR

JCapucho (approved):
Looks good to me

PR #1409, opened 2021-09-21
[glsl-in] Use Span instead of SourceMetadata

JCapucho:
Removes the old `SourceMetadata` and replaces it with the crate wide `Span`.

They were but equal conceptually and structurally so there was no reason to keep both of them around, this removes a whole bunch of conversions.


kvark (approved):
If that's not nice I don't know what is!

PR #1416, opened 2021-09-21
[glsl-in] Type constructors improvements

JCapucho:
Fixes some issues around type constructors:

- Allow use of constructors for user types
- Allow array constructors

Each commit passes CI and can be reviewed separately


kvark (approved):
wow, looks impressive

PR #1418, opened 2021-09-23
[glsl-in] Use intermediate local if storage class isn't function

### LINKED ISSUE #1341: [glsl-in] Shader rejected with "Argument ... value ... doesn't match the type ..."
Tested using naga 7681f4a21f554412dad494809855edb8367f8994

[This shader](https://github.com/gfx-rs/naga/files/7144188/binarysearch_tree.json.frag.txt) is accepted by glslangValidator, but with naga it leads to a type error.

JCapucho:
Automatically spills to a local variable function call arguments to
parameters expecting a pointer where the argument storage class isn't
function since the storage classes wouldn't match.

Closes #1341

### DIFF
--- a/src/front/glsl/functions.rs
+++ b/src/front/glsl/functions.rs
@@ -9,7 +9,8 @@ use super::{
 use crate::{
     front::glsl::types::type_power, proc::ensure_block_returns, Arena, Block, Constant,
     ConstantInner, EntryPoint, Expression, FastHashMap, Function, FunctionArgument, FunctionResult,
-    Handle, LocalVariable, ScalarKind, ScalarValue, Span, Statement, StructMember, Type, TypeInner,
+    Handle, LocalVariable, ScalarKind, ScalarValue, Span, Statement, StorageClass, StructMember,
+    Type, TypeInner,
 };
 use std::iter;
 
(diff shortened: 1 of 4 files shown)


src/front/glsl/functions.rs
```diff
@@ -549,47 +550,63 @@ impl Parser {
...
+                                inner: TypeInner::Vector { size, kind, width },
+                            },
+                            Span::default(),
+                        ),
+                        handle,
+                    ),
+                    // If the argument is a pointer whose storage class isn't `Function` an
+                    // indirection trough a local variable is needed to align the storage
+                    // classes of the call argument and the overload parameter
+                    TypeInner::Pointer { base, class } if class != StorageClass::Function => (
```
kvark, line 572: seems like we should handle `TypeInner::ValuePointer` as well here?

tests/in/glsl/expressions.frag
```diff
@@ -74,7 +74,11 @@ void testArrayConstructor() {
 	float tree[1] = float[1](0.0);
 }
 
+float global;
+void privatePointer(inout float a) {}
+
 out vec4 o_color;
 void main() {
+	privatePointer(global);
```
kvark, line 82: nit: something happened with indentation

JCapucho: Sorry was using tabs and didn't notice

kvark (approved):
Please rebase and land

Issue #1385, opened 2021-09-17
Comparing types


kvark:
Today, the validation code does "best effort" to compare types. It does so by resolving the `TypeInner` and comparing it by value. This isn't enough for pointers and arrays though (structural composite types), since the `TypeInner` would refer to the handle of the base type, which can be different.

jimblandy:
#1419 changes the type arenas to use `IndexSet`, ensuring that each type occurs only once in the arena, so `Eq` on `Handle<Type>` is the same as `Eq` on `Type`.

PR #1419, opened 2021-09-24
Ensure that each distinct type occurs only once in `Module::types`

jimblandy:
Each commit should be CI-safe. I recommend reviewing commits separately.

-   Use `UniqueArena` for types.

    Ensure that each distinct type occurs only once in `Module::types`, so that we
    can use `Eq` on `Type` or `TypeInner` for type equivalence, without being
    confused by differing `Handle<Type>` values that point to identical types.

    This removes a number of duplicate types from the ir snapshots.

-   Simplify interpolation defaulting.

    Replace `Module::apply_common_default_interpolation` with a simpler function
    that handles a single `Binding` at a time. In exchange for the simplicity, the
    function must be called at each point function arguments, function results, and
    struct members are prepared. (Any missed spots will be caught by the verifier.)

    This approach no longer requires mutating types in the arena, a prerequisite for
    properly handling type identity.

    Applying defaults to struct members when the struct declaration is parsed does
    have a disadvantage, compared to the old whole-module pass: at struct parse
    time, we don't yet know which pipeline stages the struct will be used in. The
    best we can do is apply defaults to anything with a `Location` binding. This
    causes needless qualifiers to appear in some output. However, it seems that our
    back end languages all tolerate such qualifiers.

-   Consolidate Handle construction code.

-   [glsl-in]: Register new types as necessary during constant solving.

-   Rename `UniqueArena` methods to more closely resemble `HashSet`.

    `UniqueArena::fetch_or_append` becomes `insert`.

    `UniqueArena::try_get` becomes `get_handle`, by analogy with `get`, that takes a
    type as a key.


src/front/spv/function.rs
```diff
@@ -341,14 +341,10 @@ impl<I: Iterator<Item = u32>> super::Parser<I> {
...
-                        if let Some(crate::Binding::Location {
-                            interpolation: ref mut interpolation @ None,
-                            ..
-                        }) = arg.binding
-                        {
-                            *interpolation = Some(crate::Interpolation::Perspective);
-                            // default
-                        }
+                        crate::proc::apply_default_interpolation(
+                            &mut arg.binding,
```
kvark, line 345: can this look like:
```rust
arg.binding.apply_default_interpolation(&module.types[arg.ty].inner);
```
?

src/proc/mod.rs
```diff
@@ -8,6 +8,7 @@ mod terminator;
 mod typifier;
 
 pub use index::IndexableLength;
+pub use interpolator::apply_default_interpolation;
```
kvark, line 11: this logic could also live in `src/front/mod.rs` (or `src/front/interpolator.rs`), which is more specific, right?

kvark:
That's wonderful, I was looking forward to this change! And well executed, too 👍 🎖️ . Just one note to address.

src/front/spv/function.rs
jimblandy: It can, but because we can't define methods on `Option<Binding>`, the code looks like this:

```
if let Some(ref mut binding) = arg.binding {
    binding.apply_default_interpolation(&module.types[arg.ty].inner);
}
```

I can make that change.

src/proc/mod.rs
jimblandy: Sure. Moved to `src/front/interpolator.rs`.

jimblandy:
That last forced push just adds a "fixes #1385" note to the commit message. No code change.

kvark (approved):

PR #1420, opened 2021-09-24
[Typifier] Some relational functions output vectors

JCapucho:
Previously the typifier flagged all relational functions as producing a
scalar boolean but with the exception of `all` and `any`, according to
the wgsl spec, all other relational functions output a type the same
size as the input.

Furthermore all target languages support this out of the box:
- MSL - see the official spec at page 143, table 6.3
- HLSL - [`isinf`](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-isinf), [`isfinite`](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-isfinite), [`isnan`](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-isnan), from what I could find there's no `isnormal` function so this seems like a problem in the backend.
- GLSL - [`isinf`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/isinf.xhtml), [`isnan`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/isnan.xhtml), the backend only uses this two and negates their result to produce the others.
- SPIRV - [relevant section of the specification](https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#_a_id_relational_and_logical_a_relational_and_logical_instructions)
- WGSL - [relevant table in the specification](https://gpuweb.github.io/gpuweb/wgsl/#value-testing-builtin-functions)

### DIFF
--- a/src/proc/typifier.rs
+++ b/src/proc/typifier.rs
@@ -626,10 +626,34 @@ impl<'a> ResolveContext<'a> {
             }
             crate::Expression::Select { accept, .. } => past(accept).clone(),
             crate::Expression::Derivative { axis: _, expr } => past(expr).clone(),
-            crate::Expression::Relational { .. } => TypeResolution::Value(Ti::Scalar {
-                kind: crate::ScalarKind::Bool,
-                width: crate::BOOL_WIDTH,
-            }),
+            crate::Expression::Relational { fun, argument } => match fun {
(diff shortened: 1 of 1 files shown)


kvark (approved):
Nice catch!

PR #1425, opened 2021-09-26
[spv-in] preserve case order when reusing blocks

### LINKED ISSUE #1403: Incorrect shader passthrough of SPIR-V with fallthrough switch
Passing SPIR-V shaders to naga and back alters their effects.

This bug affects the `stage` shader, a term I use to describe converting an integer encoded texel to its full 4-component floating point format. This makes heavy use of switch-case for essentially choosing one conversion function based on input parameters. [https://github.com/HeroicKatora/stealth-paint/blob/wgpu-0.10/src/shaders/stage.frag](https://github.com/HeroicKatora/stealth-paint/blob/e3d7b586102851df08482b3df8c112c58e703a8d/src/shaders/stage.frag)

The critical section is here, which is utilized while decoding Oklab colors to rgb.

```glsl
vec4 parts_normalize(vec4 components, uint parts) {
  switch (parts) {
// .. lines skipped
  case SAMPLE_PARTS_Lab:
  case SAMPLE_PARTS_LCh:
    return vec4(c.xyz, 1.0);
  case SAMPLE_PARTS_LabA:
  case SAMPLE_PARTS_LChA:
    return c.xyzw;
  }
  return BIT_DECODE_FAIL; // vec4(1.0, 0.0, 0.0, 1.0)
}
```

Expected:
oklab-good

Actual:
oklab-bad

To understand why we must have a small understanding of `Lab` color spaces. The first component is `lightness` and the second and third are chromaticities. So, full white would correspond to `1.0, 0.0, 0.0`. Familiar? Yes. Indeed this render pipeline includes invoking the above `parts_normalize` with `parts = SAMPLE_PARTS_LChA`. However this seems to instead reach to statement _below_ and returns `BIT_DECODE_FAIL`. (I've validated that by switching the default return value to `c.xyzw` as well).

I have also uploaded a fix that simply does not use fallthrough behavior at all, which also works. [This is the full patch](https://github.com/HeroicKatora/stealth-paint/pull/15/commits/a24928ce657a5aa6507f9f889db6bd06cde4c1bb).

```glsl
vec4 parts_normalize(vec4 components, uint parts) {
  switch (parts) {
// .. lines skipped
  case SAMPLE_PARTS_Lab:
    return vec4(c.xyz, 1.0);
  case SAMPLE_PARTS_LCh:
    return vec4(c.xyz, 1.0);
  case SAMPLE_PARTS_LabA:
   return c.xyzw;
  case SAMPLE_PARTS_LChA:
    return c.xyzw;
  }
  return BIT_DECODE_FAIL; // vec4(1.0, 0.0, 0.0, 1.0)
}
```

## Additional information

Reproduce: Run the test suite [here](https://github.com/HeroicKatora/stealth-paint), see `tests/debug/oklab.png.crc` output (a png image).

```bash
# Enables SPIR-V passthrough, works okay.
STEALTH_PAINT_PASSTHROUGH=1 STEALTH_PAINT_BLESS=1 cargo test
# Goes through Naga, fails
STEALTH_PAINT_BLESS=1 cargo test
```

Version information:
```toml
naga = { git = "https://github.com/gfx-rs/naga", rev = "60d0bf850b5bfeb87aeb5a0d4db6bb4059f80290" }
```
[Cargo.lock.gz](https://github.com/gfx-rs/naga/files/7198423/Cargo.lock.gz)

JCapucho:
Spir-v allows in the case list of an `OpSwitch` for target block ids to be reused, this caused issues because our frontend didn't check this and emitted them all in the same order they appeared.

With this PR now first they are collected to an `indexmap` (which would've already been introduced with jimb's changes) and then added to the body making sure that cases that share the same target block are contiguous.

Closes #1403

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -539,6 +539,16 @@ pub struct Parser<I> {
     options: Options,
     index_constants: Vec<Handle<crate::Constant>>,
     index_constant_expressions: Vec<Handle<crate::Expression>>,
+
+    /// Maps for a switch from a case target to the respective body and associated literals that
+    /// use that target block id.
+    ///
+    /// Used to preserve allocations between instruction parsing.
+    switch_cases: indexmap::IndexMap<
+        spirv::Word,
+        (BodyIndex, Vec<i32>),
+        std::hash::BuildHasherDefault<fxhash::FxHasher>,
+    >,
 }
 
 impl<I: Iterator<Item = u32>> Parser<I> {
@@ -571,6 +581,7 @@ impl<I: Iterator<Item = u32>> Parser<I> {
             options: options.clone(),
             index_constants: Vec::new(),
             index_constant_expressions: Vec::new(),
+            switch_cases: indexmap::IndexMap::default(),
         }
     }
 
@@ -2593,12 +2604,24 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                         ref other => unimplemented!("Unexpected selector {:?}", other),
                     };
 
-                    let mut cases = Vec::new();
+                    // Clear past switch cases to prevent them from entering this one
+                    self.switch_cases.clear();
+
                     for _ in 0..(inst.wc - 3) / 2 {
                         let literal = self.next()?;
                         let target = self.next()?;
 
                         let case_body_idx = ctx.bodies.len();
+
+                        // Check if any previous case already used this target block id, if so
+                        // group them together to reorder them later so that no weird
+                        // falltrough cases happen.
+                        if let Some(&mut (_, ref mut literals)) = self.switch_cases.get_mut(&target)
+                        {
+                            literals.push(literal as i32);
+                            continue;
+                        }
+
                         let mut body = Body::with_parent(body_idx);
 
                         if let Some(info) = ctx.mergers.get(&target) {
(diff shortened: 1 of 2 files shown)


src/front/spv/mod.rs
```diff
@@ -2608,7 +2631,34 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+
+                    // Loop trough the collected target blocks creating a new case for each
+                    // literal pointing to it, only one case will have the true body and all the
+                    // others will be empty falltrough so that they all execute the same body
+                    // without duplicating code.
+                    //
+                    // Since `switch_cases` is an indexmap the order of insertation is preserved
+                    // this is needed because spir-v defines falltrough order in the switch
+                    // instruction.
+                    let mut cases = Vec::new();
```
kvark, line 2648: nit: could use `with_capacity` since we know exactly

src/front/spv/mod.rs
```diff
@@ -2608,7 +2631,34 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
+                    for &(case_body_idx, ref literals) in self.switch_cases.values() {
+                        let value = literals[0];
+
+                        for literal in literals.iter().skip(1).copied() {
```
kvark, line 2652: nit: the `.copied()` here is strange. Wouldn't it be easier to just do `for &literal in ...` instead?

kvark (commented):
Looks good! just a few nits

kvark (approved):

PR #1427, opened 2021-09-27
When writing snapshots, reliably identify which writer has failed.

jimblandy:
This implements @kvark's suggestion in his review of #1384.

### DIFF
--- a/tests/snapshots.rs
+++ b/tests/snapshots.rs
@@ -130,7 +130,7 @@ fn check_targets(module: &naga::Module, name: &str, targets: Targets) {
 
     let info = naga::valid::Validator::new(naga::valid::ValidationFlags::all(), capabilities)
         .validate(module)
-        .unwrap();
+        .expect("Naga module validation failed");
 
     #[cfg(feature = "serialize")]
     {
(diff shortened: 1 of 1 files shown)


kvark (commented):
Do you see any advantage of doing this versus just `printf`?
I think having `unwrap` in testing infrastructure is actually better than propagating errors up, since we get stack traces.

jimblandy:
> Do you see any advantage of doing this versus just printf?

I thought about that, but then I thought the test output would get really long - number of files x number of output forms.

Backtraces are nice, but ideally they should point at the spot where the error was thrown - that's what I always want to know, anyway. The very popular `anyhow` crate seems to be the new standard, and it supports capturing backtraces when errors are created.

kvark:
> I thought about that, but then I thought the test output would get really long - number of files x number of output forms.

Prints from tests are only actually emitted on failure, unless you explicitly ask for them. So with `println!` by default you get exactly zero noise. The extra info only shows up when stuff goes wrong.

jimblandy:
Okay, this version just adds some printfs.

kvark (approved):
thanks!

PR #1435, opened 2021-09-28
Emit std430 layout qualifiers for storage buffers

### LINKED ISSUE #1428: Emit std140/std430 layout qualifiers in GLSL
Copied from https://github.com/gfx-rs/wgpu/issues/2005

GLSL backend doesn't produce `layout(std430)` for storage buffers.

```rust
[[block]]
struct Data {
  data: array<f32>;
};
[[group(0), binding(0)]] var<storage, read_write> dest : Data;
fn main(
) {
  let val: f32 = 1.;
  dest.data[0] = val;
  dest.data[1] = val;
}
```
Produced GLSL:
```glsl
#version 310 es

precision highp float;
precision highp int;

buffer Data_block_0Cs {
    float data[];
} _group_0_binding_0;


void main() {
    _group_0_binding_0.data[0] = 1.0;
    _group_0_binding_0.data[1] = 1.0;
    return;
}
```

Without `std430`, the array becomes 16-byte aligned, which is incorrect here.

Gordon-F:
Fix #1428


src/back/glsl/mod.rs
```diff
@@ -816,13 +820,27 @@ impl<'a, W: Write> Writer<'a, W> {
         if self.options.version.supports_explicit_locations() {
             if let Some(ref br) = global.binding {
                 match self.options.binding_map.get(br) {
-                    Some(binding) => write!(self.out, "layout(binding = {}) ", binding)?,
+                    Some(binding) => {
+                        let layout = if self.options.version.supports_std430_layout() {
+                            if let crate::StorageClass::Storage { .. } = global.class {
+                                "std430, "
```
kvark, line 826: looks like we are going to be writing `layout(std430)` both here and at line 842 down from here?

src/back/glsl/mod.rs
```diff
@@ -816,13 +820,27 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                            } else {
+                                ""
+                            }
+                        } else {
+                            ""
```
kvark, line 831: let's also write `std140` for uniform buffers

kvark (changes requested):

kvark (approved):
Thanks!

PR #1436, opened 2021-09-28
[glsl-out] Es: convert 1D images to 2D

JCapucho:
glsl es doesn't support 1d samplers so they must be converted to 2d samplers and the respective operations on them patched.

### DIFF
--- a/src/back/glsl/features.rs
+++ b/src/back/glsl/features.rs
@@ -22,8 +22,6 @@ bitflags::bitflags! {
         /// Adds support for image load and early depth tests
         const IMAGE_LOAD_STORE = 1 << 8;
         const CONSERVATIVE_DEPTH = 1 << 9;
-        /// Isn't supported in ES
-        const TEXTURE_1D = 1 << 10;
         /// Interpolation and auxiliary qualifiers. Perspective, Flat, and
         /// Centroid are available in all GLSL versions we support.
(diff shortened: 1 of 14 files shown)


kvark (approved):
Looks reasonable, thank you!
We might want to consolidate the checks a little bit better. There is a number of places doing 
`if coord_dim == 1 && self.options.version.is_es()`

JCapucho:
@kvark I've rebased the PR and cleaned it up, but it seems like msl doesn't support offsets for 1d images and that trips up the CI.

kvark:
@JCapucho sampling 1D with an offset is not a part of WGSL spec:
https://gpuweb.github.io/gpuweb/wgsl/#texturesample

So we need to remove this from "image.wgsl" out into some "image-ext.wgsl" potentially, or not test this at all

JCapucho:
@kvark I've removed it for now.

Is this something we want to support (sampling with offset for 1d texutes)? if so do you want me to open an issue for msl or otherwise a issue for validation.

kvark:
We'd probably want to have it at some point, but it feels super low priority. Unless you have a use case, that is :)
If/when we do this, we'll add a validation capability for this, and check accordingly.

kvark (approved):

kvark:
We'll need wgpu-hal collaboration here

PR #1437, opened 2021-09-29
Compile the validation out of an executable

### LINKED ISSUE #1213: Compile the validation out of an executable
For tools like glsl-to-wgsl, which gets deployed on the web as a WASM module, validation isn't actually needed. If anything is wrong, the translation may fail with a panic, or produce bad WGSL, neither of which is critical for this case.

Gordon-F:
This PR will disable shader validation as much as possible by default. Currently in draft status, because I think that implementation is not good enough.

Fix #1213

### DIFF
--- a/.github/workflows/lazy.yml
+++ b/.github/workflows/lazy.yml
@@ -149,3 +149,14 @@ jobs:
                echo "Result: $(expr $FILE_COUNT - $SUCCESS_RESULT_COUNT) / $FILE_COUNT" > counter
            done
            cat counter
+  check_snapshots_without_validation:
+    name: Check snapshots (without validation)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions-rs/cargo@v1
+        name: Test all (without validation)
+        with:
+          command: test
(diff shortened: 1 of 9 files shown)


.github/workflows/pipeline.yml
```diff
@@ -39,6 +39,13 @@ jobs:
           args: --all-features --workspace
       - name: Check snapshots
         run: git diff --exit-code -- tests/out
+      - uses: actions-rs/cargo@v1
+        name: Test all (without validation)
```
kvark, line 43: why do we test all without validation?

Gordon-F: We need this to make sure, that backend generates exactly the same shader without the validation feature. `Naga` is collecting `ModuleInfo` inside the validation module, so, theoretically, we can accidentally break shader generation.

kvark: Can this be done in Lazy CI? Chance that it breaks are rather low.

kvark (approved):

kvark:
Just in time for the release!

PR #1442, opened 2021-10-01
Properly type-check `Expression::Select` arguments.

jimblandy:
Require the `accept` and `reject` arguments to `select` to be scalars or vectors, per WGSL spec.


kvark (approved):

kvark:
rebase please

PR #1443, opened 2021-10-01
Forbid unsized structs as final members of structs.

jimblandy:
WGSL says:

> - The last member of the structure type defining the store type for a variable ... may be a runtime-sized array.
>
> - A runtime-sized array must not be used as the store type or contained within a store type in any other cases.

Thus, a struct whose final member is a struct whose final member is a runtime-sized array is verboten.

### DIFF
--- a/src/valid/type.rs
+++ b/src/valid/type.rs
@@ -473,9 +473,13 @@ impl super::Validator {
                         handle,
                     );
 
-                    // only the last field can be unsized
+                    // The last field may be an unsized array.
                     if !base_info.flags.contains(TypeFlags::SIZED) {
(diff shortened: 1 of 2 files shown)


src/valid/type.rs
```diff
@@ -473,9 +473,13 @@ impl super::Validator {
                         handle,
                     );
 
-                    // only the last field can be unsized
+                    // The last field may be an unsized array.
                     if !base_info.flags.contains(TypeFlags::SIZED) {
```
kvark, line 477: oh snap. Looks like the whole `SIZED` isn't useful any more.

kvark (approved):

PR #1445, opened 2021-10-03
[glsl-in] Fix array index parsing for 1D images

JCapucho:
The array index for 1D images was being parsed as the first element of the coordinate vector which is the coordinate itself.

The fix is just changing a number and we already had tests for this so I don't know how it managed to stay broken for so long


kvark (approved):
Nice one!

PR #1448, opened 2021-10-05
[glsl-in] Always use uint for array size

### LINKED ISSUE #1446: [glsl-in] array length is inferred as `Sint` in type and `Uint` in expression
The code
```glsl
void main() {
    vec4 offset[2] = vec4[](vec4(0.0), vec4(0.0));
}
```
fails to compile because of
```glsl
	Local variable [1] 'offset' is invalid
	Initializer doesn't match the variable type
```

Looking at the types, the only difference is the type of the array length constant:

Local variable `offset` array length:
```rust
[1]: Constant {
	name: None,
	specialization: None,
	inner: Scalar {
		width: 4,
		value: Sint(
			2,
		),
	},
}
```
Initializer expression type array length:
```rust
[3]: Constant {
	name: None,
	specialization: None,
	inner: Scalar {
		width: 4,
		value: Uint(
			2,
		),
	},
}
```

The token for the 2 in `offset[2]` as returned by `pp-rs` is
```rust
Integer {
    value: 2,
    signed: true,
    width: 32,
}
```

JCapucho:
Closes #1446

### DIFF
--- a/tests/out/wgsl/bevy-pbr-frag.wgsl
+++ b/tests/out/wgsl/bevy-pbr-frag.wgsl
@@ -23,8 +23,8 @@ struct CameraPosition {
 struct Lights {
     AmbientColor: vec4<f32>;
     NumLights: vec4<u32>;
-    PointLights: [[stride(48)]] array<PointLight,10>;
-    DirectionalLights: [[stride(32)]] array<DirectionalLight,1>;
+    PointLights: [[stride(48)]] array<PointLight,10u>;
+    DirectionalLights: [[stride(32)]] array<DirectionalLight,1u>;
 };
 
 [[block]]
--- a/src/front/glsl/parser/types.rs
+++ b/src/front/glsl/parser/types.rs
@@ -24,7 +24,18 @@ impl<'source> ParsingContext<'source> {
                 return Ok(Some((ArraySize::Dynamic, meta)));
             }
 
-            let (constant, _) = self.parse_constant_expression(parser)?;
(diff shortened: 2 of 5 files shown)


tests/out/wgsl/bevy-pbr-frag.wgsl
```diff
@@ -23,8 +23,8 @@ struct CameraPosition {
 struct Lights {
     AmbientColor: vec4<f32>;
     NumLights: vec4<u32>;
-    PointLights: [[stride(48)]] array<PointLight,10>;
-    DirectionalLights: [[stride(32)]] array<DirectionalLight,1>;
+    PointLights: [[stride(48)]] array<PointLight,10u>;
```
kvark, line 26: this change shouldn't be needed. WGSL allows signed/unsigned. The frontend should generate only unsigned.

kvark (changes requested):

tests/out/wgsl/bevy-pbr-frag.wgsl
JCapucho: That's what is happening here, the frontend always generates uints and the wgsl backend picks up the uints and writes them with the correct suffix, there's nothing that can be done here to prevent this (besides changing the backend)

kvark: ah sorry, dummy me, you are right

kvark (approved):

PR #1449, opened 2021-10-05
Add insert/extractBits and Pack/Unpack

cwfitzgerald:
Implements the following functions for all frontends and all backends (except hlsl... cause ugh).

- `extractBits`
- `insertBits`
- `pack4x8snorm`
- `pack4x8unorm`
- `pack2x16snorm`
- `pack2x16unorm`
- `pack2x16float`
- `unpack4x8snorm`
- `unpack4x8unorm`
- `unpack2x16snorm`
- `unpack2x16unorm`
- `unpack2x16float`

Should there be tests for some/all of these, if so, how should I go about adding them.


kvark (commented):
Can we have a `tests/in/bits.wgsl` with all these?

src/proc/typifier.rs
```diff
@@ -781,7 +782,21 @@ impl<'a> ResolveContext<'a> {
...
-                    Mf::ReverseBits => res_arg.clone(),
+                    Mf::ReverseBits |
+                    Mf::ExtractBits |
+                    Mf::InsertBits => res_arg.clone(),
+                    // data packing
+                    Mf::Pack4x8snorm |
+                    Mf::Pack4x8unorm |
+                    Mf::Pack2x16snorm |
+                    Mf::Pack2x16unorm |
+                    Mf::Pack2x16float => TypeResolution::Value(Ti::Scalar { kind: crate::ScalarKind::Uint, width: 4 }),
```
kvark, line 793: are you sure `cargo fmt` is ran here?

kvark (approved):
Beautiful!

src/proc/typifier.rs
cwfitzgerald: Absolutely not :)

Issue #1410, opened 2021-09-21
Glsl's `mod` should not be translated to metal `fmod`


thenlevy:
Translating this glsl [fragment shader](https://github.com/thenlevy/ensnano/blob/3aa82992518018f2c61944ceadbf3867bd563397/src/flatscene/view/background.frag) to spriv (using `glslc`) then to metal (using `naga` for the spirv -> metal ) or directly to metal (using only `naga`for glsl -> spriv ) results in a metal shader where the `mod` instruction has been translated into a `metal::fmod` instruction.

According to https://stackoverflow.com/questions/7610631/glsl-mod-vs-hlsl-fmod this is incorrect as `mod(x,y)` and `fmod(x, y)` have different values when `x/y <= 0`

PS:  I don't think that this is a glsl-in issue because on platforms that don't use `metal`, the shader works as expected.

JCapucho:
So what would be the appropriate here? `BinaryOperator::Remainder`?

teoxoy:
Has been resolved by https://github.com/gfx-rs/naga/pull/1867.

PR #1452, opened 2021-10-07
[glsl-out] Convert modulo operator on float to SPIR-V `OpFRem` equivalent function

### LINKED ISSUE #1451: [wgsl-in]/[glsl-out] Bug with vec types and modulo operator
I had some code in wgsl like this:
```rust
    let rgb = clamp(
        abs(
            ((hue * 6.0 + vec3<f32>(0.0, 4.0, 2.0)) % 6.0 - 3.0
        ) - 1.0,
        vec3<f32>(0.0),
        vec3<f32>(1.0)
    );
```
and hit this when trying to run on WebGL2:
```
Internal error in VERTEX shader: ERROR: 0:244: '%' : wrong operand types - no operation '%' exists that takes a left-hand operand of type 'highp 3-component vector of float' and a right operand of type 'const 3-component vector of float' (or there is no acceptable conversion)
```
It's specifically the calculated vec3 % constant vec3 that seems to be the problem. I can work around it by individually calculating each component, but I think this is a bug somewhere, perhaps in glsl-out?

Gordon-F:
Fix #1451

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -311,6 +311,16 @@ pub enum Error {
     Custom(String),
 }
 
+/// Binary operation with a different logic on the GLSL side
+enum BinaryOperation {
+    /// Vector comparison should use the function like `greaterThan()`, etc.
+    VectorCompare,
+    /// GLSL `%` is SPIR-V `OpUMod/OpSMod` and `mod()` is `OpFMod`, but [`BinaryOperator::Modulo`](crate::BinaryOperator::Modulo) is `OpFRem`
+    Modulo,
+    /// Any plain operation. No additional logic required
+    Other,
+}
+
 /// Main structure of the glsl backend responsible for all code generation
 pub struct Writer<'a, W> {
     // Inputs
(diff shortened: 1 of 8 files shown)


JCapucho:
This translation isn't correct see #1410

src/back/glsl/mod.rs
```diff
@@ -311,6 +311,16 @@ pub enum Error {
     Custom(String),
 }
 
+/// Binary operation with a different logic on the GLSL side
+enum BinaryOperation {
```
kvark, line 315: that's nice

src/back/glsl/mod.rs
```diff
@@ -2214,35 +2224,79 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                write!(self.out, "{}(", function.unwrap_or(""))?;
-                self.write_expr(left, ctx)?;
-
-                if function.is_some() {
-                    write!(self.out, ",")?
-                } else {
-                    write!(self.out, " {} ", super::binary_operation_str(op))?;
+                match function {
+                    BinaryOperation::VectorCompare => {
+                        match op {
```
kvark, line 2261: let's move the `write!` out of the match, which can return just an `&str`

src/back/glsl/mod.rs
```diff
@@ -2214,35 +2224,79 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                            Bo::NotEqual => write!(self.out, "notEqual")?,
+                            _ => {}
+                        };
+                    }
+                    BinaryOperation::Modulo | BinaryOperation::Other => {
+                        // nothing to do here
+                    }
+                }
+                write!(self.out, "(")?;
+                match function {
```
kvark, line 2276: let's merge this `match` with the previous one

kvark (approved):

PR #1459, opened 2021-10-12
Implement integral-bool casts in SPV, MSL, and HLSL backends

### LINKED ISSUE #1458: [msl-out] Metal backend giving module is not valid for validated wgsl input
I have some WGSL shader code that passed validation but that when outputting to metal gives "module is not valid". See here: https://gist.github.com/superdump/d8794eaf9a83771a2b4d59cd9f5b33ce

kvark:
Fixes #1458
Comes with a test!


kvark:
Published in 0.7.2

PR #1462, opened 2021-10-15
[msl-out] Delete Writer::runtime_sized_buffers.

jimblandy:
This is just a table mapping handles to their indices, but we can just fetch the index directly from the handle. We never iterate over its entries, but even if we did, we could at least use a HashSet instead of a HashMap.

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -307,7 +307,6 @@ pub struct Writer<W> {
     names: FastHashMap<NameKey, String>,
     named_expressions: crate::NamedExpressions,
     namer: proc::Namer,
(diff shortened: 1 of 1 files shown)


src/back/msl/writer.rs
```diff
@@ -1778,7 +1774,6 @@ impl<W: Write> Writer<W> {
             for (handle, var) in module.global_variables.iter() {
                 if needs_array_length(var.ty, &module.types) {
                     let idx = handle.index();
-                    self.runtime_sized_buffers.insert(handle, idx);
```
kvark, line 1781: oh right. I think this was intended to be different originally, but now it's redundant indeed

kvark (approved):

PR #1469, opened 2021-10-19
[spv-in]: Cull unused builtins inside structs

JCapucho:
Some compilers like shaderc introduce a full gl_PerVertex struct, this
includes gl_ClipDistance. Normally this isn't a problem since most
drivers optimize it away, but naga zero inits globals if they weren't
previously initialized. This causes gl_ClipDistance to be initialized to
zero which can be really bad for performance.

It only culls builtins inside structs since the problems only have been observed inside them and adding it to all expressions could be costly (would need a benchmark to be certain).

This is only observed in spirv backend since it's the only backend to zero initialize stuff.


kvark:
Published in 0.7.2

PR #1470, opened 2021-10-22
Use span information in analyzer errors (proof of concept)

ElectronicRU:
Related to #358 

Implements auxiliary span information in errors emitted by the analyzer, and prints the annotation out in the CLI.

P.S. Turns out traversing a tree is kinda hard. Bubbling errors up is much easier, though requires a bit of boilerplate.

### DIFF
--- a/src/span.rs
+++ b/src/span.rs
@@ -195,12 +195,6 @@ pub(crate) trait AddSpan: Sized {
     fn with_span_handle<T, A: SpanProvider<T>>(self, handle: Handle<T>, arena: &A) -> Self::Output;
 }
 
-/// Convenience trait for converting `Result<_, E>` into `Result<_, WithSpan<E>>`.
-pub(crate) trait AddSpanResult: Sized {
-    type Output;
-    fn with_span(self) -> Self::Output;
-}
-
 /// Trait abstracting over getting a span from an [`Arena`] or a [`UniqueArena`].
 pub(crate) trait SpanProvider<T> {
     fn get_span(&self, handle: Handle<T>) -> Span;
@@ -250,17 +244,6 @@ where
         arena: &A,
     ) -> WithSpan<Self> {
         WithSpan::new(self).with_handle(handle, arena)
-    }
-}
-
-impl<T, E> AddSpanResult for Result<T, E>
-where
-    E: Error,
-{
-    type Output = Result<T, WithSpan<E>>;
-
-    fn with_span(self) -> Self::Output {
-        self.map_err(|e| e.with_span())
     }
 }
 
--- a/src/valid/function.rs
+++ b/src/valid/function.rs
@@ -6,7 +6,7 @@ use super::{
     analyzer::{UniformityDisruptor, UniformityRequirements},
     ExpressionError, FunctionInfo, ModuleInfo,
 };
-use crate::span::{AddSpan as _, AddSpanResult as _, MapErrWithSpan as _, WithSpan};
+use crate::span::{AddSpan as _, MapErrWithSpan as _, WithSpan};
 
 #[cfg(feature = "validate")]
 use bit_set::BitSet;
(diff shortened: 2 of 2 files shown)


src/span.rs
```diff
@@ -1,3 +1,7 @@
+use crate::{Arena, Handle, UniqueArena};
+use std::borrow::Cow;
```
kvark, line 2: nit: let's follow the rest of the code base and have one `use` per crate

src/span.rs
```diff
@@ -66,3 +70,254 @@ impl From<Range<usize>> for Span {
         }
     }
 }
+
+#[cfg(feature = "span")]
+extern crate smallvec;
```
kvark, line 75: no need for this in Rust2018

src/span.rs
```diff
@@ -66,3 +70,254 @@ impl From<Range<usize>> for Span {
...
+    {
+        Self(span, Cow::from(description))
+    }
+}
+
+#[derive(Debug)]
+pub struct WithSpan<E> {
+    inner: E,
+    #[cfg(feature = "span")]
+    spans: smallvec::SmallVec<[(Span, Cow<'static, str>); 8]>,
```
kvark, line 92: Why are we using `SmallVec` here? It looks to be a wrapper around an error. If errors have heap allocations, that's totally cool. So here let's just use `Vec`.

src/span.rs
```diff
@@ -66,3 +70,254 @@ impl From<Range<usize>> for Span {
...
     }
 }
+
+#[cfg(feature = "span")]
+extern crate smallvec;
+
+pub struct SpanContext(Span, Cow<'static, str>);
+
+impl SpanContext {
+    pub fn new<S>(span: Span, description: S) -> Self
```
kvark, line 80: any reason this isn't `S: Into<Cow<...>>`?

src/span.rs
```diff
@@ -1,3 +1,7 @@
+use crate::{Arena, Handle, UniqueArena};
+use std::borrow::Cow;
+use std::error::Error;
+use std::fmt::{Display, Formatter};
```
kvark, line 4: let's just include `fmt` and refer to members as `fmt::Xxx` - that's what other modules do in this repo

src/span.rs
```diff
@@ -66,3 +70,254 @@ impl From<Range<usize>> for Span {
...
+/// Convenience trait for [`Error`] to be able to apply spans to anything
+pub(crate) trait AddSpan: Sized {
+    type Output;
+    fn with_span(self) -> Self::Output;
+    fn with_span_static(self, span: Span, description: &'static str) -> Self::Output;
+    fn with_span_context(self, span_context: SpanContext) -> Self::Output;
+    fn with_span_handle<T, A: SpanProvider<T>>(self, handle: Handle<T>, arena: &A) -> Self::Output;
+}
+
+pub(crate) trait AddSpanResult: Sized {
```
kvark, line 222: where is this used?

src/valid/interface.rs
```diff
@@ -4,6 +4,7 @@ use super::{
 };
 use crate::arena::{Handle, UniqueArena};
 
+use crate::span::{AddSpan, MapErrWithSpan, SpanProvider, WithSpan};
```
kvark, line 7: nit: when including a trait just for the methods, use `Trait as _`

tests/out/glsl/quad.fs_extra.Fragment.glsl
```diff
@@ -1,4 +1,4 @@
-#version 300 es
+#version 310 es
```
kvark, line 1: why are we changing this?

kvark (commented):
I noticed a few things that are really nits.
There are no comments about the general approach, which I'm still trying to understand. Could you describe it in a few sentences?

Another question, since this PR implements spans for validation errors, could you post an example of the new error?

Finally, what is left to do to close #358?

PR #1473, opened 2021-10-23
Add FindLsb / FindMsb

fintelia:
Currently untested. I welcome feedback on whether this approach makes sense and how to go about adding tests.


src/back/msl/writer.rs
```diff
@@ -1184,6 +1184,8 @@ impl<W: Write> Writer<W> {
                     Mf::ReverseBits => "reverse_bits",
                     Mf::ExtractBits => "extract_bits",
                     Mf::InsertBits => "insert_bits",
+                    Mf::FindLsb => return Err(Error::UnsupportedCall(format!("{:?}", fun))),
```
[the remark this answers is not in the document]
kvark: @fintelia ping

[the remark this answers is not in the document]
fintelia: Could you say a bit about how to inject an expression? All the other cases seem to just reference a single function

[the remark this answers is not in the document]
kvark: The `Distance` is also injecting an expression:
```rust
                    write!(self.out, "{}::abs(", NAMESPACE)?;
                    self.put_expression(arg, context, false)?;
                    write!(self.out, " - ")?;
                    self.put_expression(arg1.unwrap(), context, false)?;
                    write!(self.out, ")")?;
```

fintelia:
Ok, it seems that passing validation requires that all GLSL inputs can be written as valid WGSL. Unfortunately, WGSL doesn't yet have `ctz` / `clz` functions so I'm not sure the right way to achieve that without simply not testing the new functionality

src/back/wgsl/writer.rs
```diff
@@ -1563,6 +1567,16 @@ impl<W: Write> Writer<W> {
...
                     }
+                    Function::FindLsb => {
+                        write!(self.out, "((ctz(")?;
+                        self.write_expr(module, arg, func_ctx)?;
+                        write!(self.out, ") + 1) % 33) - 1")?;
+                    }
+                    Function::FindMsb => {
+                        write!(self.out, "((clz(")?;
+                        self.write_expr(module, arg, func_ctx)?;
+                        write!(self.out, ") + 1) % 33) - 1")?;
```
kvark, line 1578: something isn't right here. The implementations for `FindLsb` and `FindMsb` look the same?

fintelia: Note the "t" vs. "l" in `ctz` vs. `clz`

fintelia: Relatedly, since these functions aren't officially added to WGSL yet someone should seriously consider other names for them that are easier to tell apart!

kvark: Please comment about this on the upstream issue as well!

kvark:
I agree about the naming, it would be nice to settle this in the group.

fintelia:
Ok, at this point it seems that the main blocker is that WGSL's builtins for this functionality haven't been specified yet. What would be the right way to push that forward so this PR can be merged?

kvark:
If you want to use it now, we can land this with the expectation that we'll rename these methods once the group settles on them.

kvark:
@fintelia there seem to be a small problem left on Metal?

fintelia:
Ok, I think I managed to fix it. The issue was that there needed to be an unsigned integer -> signed integer cast, but to do that the vector width of the input needed to be known. I tried a bunch of ways to cause the type coercion to produce that without the generated code differing between them, but none ended up working. The current version checks the input width so it can insert a cast using one of `int`/`int2`/`int3`/`int4`.

src/back/msl/writer.rs
```diff
@@ -1204,6 +1221,22 @@ impl<W: Write> Writer<W> {
                     write!(self.out, " - ")?;
                     self.put_expression(arg1.unwrap(), context, false)?;
                     write!(self.out, ")")?;
+                } else if fun == Mf::FindLsb {
+                    write!(
+                        self.out,
+                        "(((1 + int{}({}::ctz(",
+                        argument_size_suffix, NAMESPACE
+                    )?;
+                    self.put_expression(arg, context, false)?;
```
kvark, line 1230: last argument should be `true` since we are inside brackets

src/back/msl/writer.rs
```diff
@@ -1204,6 +1221,22 @@ impl<W: Write> Writer<W> {
...
+                    write!(self.out, "))) % 33) - 1)")?;
+                } else if fun == Mf::FindMsb {
+                    write!(
+                        self.out,
+                        "(((1 + int{}({}::clz(",
+                        argument_size_suffix, NAMESPACE
+                    )?;
+                    self.put_expression(arg, context, false)?;
```
kvark, line 1238: also `true`

src/proc/typifier.rs
```diff
@@ -785,6 +785,16 @@ impl<'a> ResolveContext<'a> {
                     Mf::ReverseBits |
                     Mf::ExtractBits |
                     Mf::InsertBits => res_arg.clone(),
+                    Mf::FindLsb |
+                    Mf::FindMsb => match res_arg.inner_with(types)  {
```
kvark, line 789: let's do `match *res_arg.inner_with(types)` like we do in other places

tests/in/bits.wgsl
```diff
@@ -36,4 +36,20 @@ fn main() {
     u2 = extractBits(u2, 5u, 10u);
     u3 = extractBits(u3, 5u, 10u);
     u4 = extractBits(u4, 5u, 10u);
+    i = findLsb(i);
+    i2 = findLsb(i2);
+    i3 = findLsb(i3);
```
kvark, line 41: why do we need to call it so many times! it's not like we are checking the result anyway, we are only interested in code generation here. So calling once should be ok.

kvark (changes requested):
Thank you!

tests/in/bits.wgsl
fintelia: This is one call with each argument type. I can remove some if you'd like, but each one potentially generates different code.

kvark: I think having something like an signed int scalar and an unsigned int vector would do (i.e. 2 cases per function).

kvark (approved):

kvark:
@fintelia would you be interested to update the naming according to https://github.com/gpuweb/gpuweb/pull/2467 ?

fintelia:
I probably wouldn't be able to get to it for a while. If anyone else wants to make the update they should go ahead!

With the functions now fully specified in WGSL, someone should probably also make sure there's conformance tests for the various edge cases. I'm not fully confident that I understood the specs fully for each of the backends; testing on lots of platforms is the only way to be sure.

PR #1475, opened 2021-10-24
Allow new clippy::if_same_then_else

kvark:
Sometimes I'm wondering why clippy is so picky by default. In this case, there are many legitimate reasons to have duplicate code in "if/else if" branches.


kvark:
Published in 0.7.2

PR #1484, opened 2021-10-26
Handle empty variable names in namer

### LINKED ISSUE #1483: [wgsl-out] Certain SPIR-V Patterns Cause Invalid Accesses
When converting from spirv generated from the following glsl, the wgsl backend generates invalid code.

This is the the relevant snippet of the glsl:

```glsl
// Line 33
layout(set = 2, binding = 10) uniform TextureData {
    CPUMaterialData material;
};

void main() {
	// Line 47
	vec2 coords = vec2(material.uv_transform0 * vec3(i_coords0, 1.0));
}
```

Generates the following snippet of wgsl:

```wgsl
// Line 18
[[block]]
struct TextureData {
    material: CPUMaterialData;
};

[[group(2), binding(10)]]
// BUG: uniform with no name
var<uniform> : TextureData;

fn main1() {
	// line 36
	// BUG: invalid name before the .
	let e28: mat3x3<f32> = .material.uv_transform0;
}
```

Attached is the full glsl, spirv, spirv-dis, and wgsl generated. Conversion from glsl -> spirv done with glslc, spirv -> wgsl done with master naga.

[naga-bug.zip](https://github.com/gfx-rs/naga/files/7420503/naga-bug.zip)

I have assigned this as a wgsl-out backend issue, but could also be a spv-in issue.

Gordon-F:
Fix #1483

### DIFF
--- a/tests/in/glsl/empty-global-name.frag
+++ b/tests/in/glsl/empty-global-name.frag
@@ -1,0 +1,7 @@
+layout(set = 1, binding = 1) uniform TextureData {
+    vec4 material;
+};
+
+void main() {
+	vec2 coords = vec2(material.xy);
+}
--- a/tests/out/wgsl/empty-global-name-frag.wgsl
+++ b/tests/out/wgsl/empty-global-name-frag.wgsl
@@ -1,0 +1,21 @@
+[[block]]
+struct TextureData {
+    material: vec4<f32>;
+};
+
+[[group(1), binding(1)]]
+var<uniform> global: TextureData;
+
+fn main1() {
+    var coords: vec2<f32>;
+
+    let e2: vec4<f32> = global.material;
+    coords = vec2<f32>(e2.xy);
+    return;
+}
+
+[[stage(fragment)]]
+fn main() {
+    main1();
+    return;
+}


jimblandy:
The test case in this PR passes even without the patch. The unnamed `uniform` has a name of `None`, and `Namer::reset` supplies "global" as the fallback in that case.

jimblandy:
The `.spv` file provided in #1483 does trigger the bug, though.

jimblandy:
The fix in the PR, however, is correct: the SPIR-V file provokes a call like this:

```
call_or(Some(""), "global") -> ""
```

The `Some("")` case is exactly what the PR handles.

It's just the test case that is ineffectual.

Gordon-F:
> The test case in this PR passes even without the patch. The unnamed `uniform` has a name of `None`, and `Namer::reset` supplies `"global"` as the fallback in that case.

Yes, you are right, thank you. It's because `glsl-in` parse empty name as `None`, but `spv-in` parse as `Some("")`. I will make a PR with spv test instead of glsl.

kvark:
Published in 0.7.2

Issue #1498, opened 2021-10-29
[glsl-in] naga rejects use of clamp in constant initializer


afd:
This shader is legal in GLSL 4.5:

```
#version 450

const int x = clamp(1, 1, 1);

void main()
{
}
```

because builtins such as clamp are allowed in constant expressions if invoked with constant expression arguments (see "4.3.3 Constant Expressions" in the GLSL 4.5 spec).

naga (revision 9d2b357214dacce0fe30587844d4c7bf41b51773) rejects the example with:

```
error: const values must have an initializer
```

sjinno:
I would like to work on this issue if it's okay :) @jimblandy will be mentoring me.

PR #1501, opened 2021-11-01
Output an accurate error message when unimplemented math functions are called (globally)

sjinno:
An accurate error message wasn't output when unimplemented math functions were called.

In this case (#1498),
```
#version 450

const int x = clamp(1, 1, 1);

void main() {}
```
would say `error: const values must have an initializer` when it should actually be saying
```
error: Not implemented: Clamp
```

**Note:** This change only applies to unimplemented math functions that are called globally (or should I say externally?).

Feedback on wording/how code should be written is very much appreciated! :)


jimblandy:
(Taking review, since I'm mentoring @sjinno on this work.)

src/front/glsl/parser/declarations.rs
```diff
@@ -199,8 +199,16 @@ impl<'source> ParsingContext<'source> {
             // TODO: Should we try to make constants here?
             // This is mostly a hack because we don't yet support adding
             // bodies to entry points for variable initialization
-            let maybe_constant =
-                init.and_then(|(root, meta)| parser.solve_constant(ctx.ctx, root, meta).ok());
+            let maybe_constant = if let Some((root, meta)) = init {
+                let result = parser.solve_constant(ctx.ctx, root, meta);
+                match (result, ctx.external) {
```
jimblandy, line 204: This looks correct, but I think it might be clearer as a plain `match result`, with a guard expression on the `Err` case:

```
Err(err) if ctx.external => return Err(err),
```

It also deserves a little comment, something like:

Initializers for external declarations must be constants, so if `solve_constant` failed, that's an error the user should see. But if it's an initializer for a local variable, then failing to produce a constant just means we emit the expression as code; that is not an error.

jimblandy (changes requested):
This looks good - thank you very much!

There are just a few things I'd like changed.

src/front/glsl/parser.rs
```diff
@@ -368,7 +368,7 @@ impl Parser {
 
 pub struct DeclarationContext<'ctx> {
     qualifiers: Vec<(TypeQualifier, Span)>,
-    external: bool,
+    external: bool, // Indicates global declaration
```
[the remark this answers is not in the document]
sjinno: Okay!

src/front/glsl/parser/declarations.rs
sjinno: Thank you very much for the review! I'd like to use what you wrote as a comment. :)

sjinno: Currently `maybe_constant` assigns the returned value of `sove_constant` for both externally and internally declared math functions if implemented. 

So, should it also be
```
Ok(res) if ctx.external => Some(res)
```
so that internally declared math expressions are left as is?

src/front/glsl/parser/declarations.rs
```diff
@@ -199,8 +199,22 @@ impl<'source> ParsingContext<'source> {
...
+            // Initializers for external declarations must be constants,
+            // so if `solve_constant` failed, that's an error the user should see.
+            // But if it's an initializer for a local variable, then failing to
+            // produce a constant just means we emit the expression as code;
+            // that is not an error.
+            let maybe_constant = if let Some((root, meta)) = init {
+                let result = parser.solve_constant(ctx.ctx, root, meta);
+                match result {
+                    Ok(res) => Some(res),
+                    Err(err) if ctx.external => return Err(err),
```
JCapucho, line 212: This will disallow shaders like:
```rust
#version 460

layout (location = 0) in vec4 read;
layout (location = 0) out vec4 fragColor;
vec4 a = read;

void main()
{
	fragColor = vec4(0.4, 0.4, 0.8, 1.0);
}
```
Which are valid glsl, you need to also check if the declaration is constant

JCapucho (changes requested):
There is a problem with your current approach that makes valid shaders not pass

src/front/glsl/parser/declarations.rs
jimblandy: Oh, I didn't know that. Thanks for pointing it out.

sjinno: @JCapucho Thank you very much for the review! Just to clarify one thing to see if I'm following you, the code should assign `None` to `maybe_constant` instead of returning an error in that case?

In other words, is it only supposed to check if the expression of `root` is `Expression::Load{ .. }`?

src/front/glsl/parser/declarations.rs
```diff
@@ -199,8 +199,25 @@ impl<'source> ParsingContext<'source> {
...
+            let maybe_constant = if let Some((root, meta)) = init {
+                let expr = &ctx.ctx.expressions[root];
+                match *expr {
+                    Expression::Load { .. } => None,
+                    _ => match parser.solve_constant(ctx.ctx, root, meta) {
+                        Ok(res) => Some(res),
+                        Err(err) if ctx.external => return Err(err),
+                        _ => None,
+                    },
+                }
```
JCapucho, line 217: I don't know why but github didn't send my comment which said this wasn't the right way to check.

You need to check the declaration qualifiers not the type of expression, see `DeclarationContext::qualifiers`

src/front/glsl/parser/declarations.rs
```diff
@@ -199,8 +199,31 @@ impl<'source> ParsingContext<'source> {
...
+                    match *qualifier {
+                        TypeQualifier::StorageQualifier(StorageQualifier::Const) => {
+                            match parser.solve_constant(ctx.ctx, root, meta) {
+                                Ok(res) => Some(res),
+                                Err(err) if ctx.external => return Err(err),
+                                _ => None,
+                            }
+                        }
+                        _ => None,
+                    }
```
JCapucho, line 222: This won't work because the constant qualifier isn't required to be the first try something like:
```rust
let is_const = ctx.qualifiers.iter().find(|(qualifier,_)| match qualifier {
  TypeQualifier::StorageQualifier(StorageQualifier::Const) => true,
  _ => false
}).is_some();
```

JCapucho (approved):
Implementation wise it looks good to me, I'll let @jimblandy handle the rest since he is mentoring you

sjinno:
@JCapucho Thank you for your help!

JCapucho:
Since the author hasn't touched this PR in a while, I've gone ahead and rebased it and cleaned it up a little, the code is essentially the same, only that now the constants are also folded for non const declarations (as was previously done), I reworded the comments a bit as was requested by jimb in his review and changed the error message to be a bit more helpful.

I've also cleaned the git history since it picked up many unneeded commits.

@jimblandy it still needs an approval from you, so if you could do a review it would be very helpful.

jimblandy (approved):
Looks good to me - thank you very much!

PR #1502, opened 2021-11-01
[glsl-in] Implement `MathFunction::Clamp`

sjinno:
As the title says, added the implementation of `MathFunction::Clamp`. Feedback is very much appreciated! :)

### DIFF
--- a/src/front/glsl/constants.rs
+++ b/src/front/glsl/constants.rs
@@ -239,17 +239,16 @@ impl<'a> ConstantSolver<'a> {
                                 },
                                 width,
                             ),
-                            _ => return Err(ConstantSolvingError::InvalidMathArg),
+                            _ => {
+                                return Err(ConstantSolvingError::NotImplemented(format!(
+                                    "{:?} applied to vector values",
+                                    fun
+                                )))
+                            }
                         };
 
-                        Ok(self.constants.fetch_or_append(
-                            Constant {
-                                name: None,
-                                specialization: None,
-                                inner: ConstantInner::Scalar { width, value },
-                            },
-                            span,
-                        ))
+                        let inner = ConstantInner::Scalar { width, value };
+                        Ok(self.register_constant(inner, span))
                     }
                     _ => Err(ConstantSolvingError::NotImplemented(format!("{:?}", fun))),
                 }
--- a/src/front/glsl/constants.rs
+++ b/src/front/glsl/constants.rs
@@ -557,10 +557,10 @@ impl<'a> ConstantSolver<'a> {
     }
 }
 
-/// Helper function to implement the glsl `max` function for floats
+/// Helper function to implement the GLSL `max` function for floats.
 ///
-/// While rust does provide a `f64::max` method it has a different behavior than the
-/// glsl `max` for `NaN`s, in rust if any of the arguments is a `NaN` then the other
+/// While Rust does provide a `f64::max` method, it has a different behavior than the
+/// GLSL `max` for NaNs. In Rust, if any of the arguments is a NaN, then the other
 /// is returned.
 ///
 /// This leads to different results in the following example
@@ -569,7 +569,7 @@ impl<'a> ConstantSolver<'a> {
 /// std::f64::NAN.max(1.0);
 /// ```
 ///
-/// Rust will return 1.0 while glsl should return NaN
+/// Rust will return `1.0` while GLSL should return NaN.
 fn glsl_float_max(x: f64, y: f64) -> f64 {
     if x < y {
         y
@@ -578,10 +578,10 @@ fn glsl_float_max(x: f64, y: f64) -> f64 {
     }
 }
 
-/// Helper function to implement the glsl `min` function for floats
+/// Helper function to implement the GLSL `min` function for floats.
 ///
-/// While rust does provide a `f64::min` method it has a different behavior than the
-/// glsl `min` for `NaN`s, in rust if any of the arguments is a `NaN` then the other
+/// While Rust does provide a `f64::min` method, it has a different behavior than the
+/// GLSL `min` for NaNs. In Rust, if any of the arguments is a NaN, then the other
 /// is returned.
 ///
 /// This leads to different results in the following example
@@ -590,7 +590,7 @@ fn glsl_float_max(x: f64, y: f64) -> f64 {
 /// std::f64::NAN.min(1.0);
 /// ```
 ///
-/// Rust will return 1.0 while glsl should return NaN
+/// Rust will return `1.0` while GLSL should return NaN.
 fn glsl_float_min(x: f64, y: f64) -> f64 {
     if y < x {
         y
--- a/src/front/glsl/constants.rs
+++ b/src/front/glsl/constants.rs
@@ -234,7 +234,9 @@ impl<'a> ConstantSolver<'a> {
                                         ScalarValue::Float(a),
                                         ScalarValue::Float(b),
                                         ScalarValue::Float(c),
-                                    ) => ScalarValue::Float(a.max(b).min(c)),
+                                    ) => {
+                                        ScalarValue::Float(glsl_float_min(glsl_float_max(a, b), c))
+                                    }
                                     _ => return Err(ConstantSolvingError::InvalidMathArg),
                                 },
                                 width,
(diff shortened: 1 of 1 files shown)


src/front/glsl/constants.rs
```diff
@@ -225,6 +233,48 @@ impl<'a> ConstantSolver<'a> {
...
+                                    (
+                                        ScalarValue::Float(a),
+                                        ScalarValue::Float(b),
+                                        ScalarValue::Float(c),
+                                    ) => ScalarValue::Float(a.max(b).min(c)),
+                                    _ => return Err(ConstantSolvingError::InvalidMathArg),
+                                },
+                                width,
+                            ),
+                            _ => return Err(ConstantSolvingError::InvalidMathArg),
```
jimblandy, line 266: Generally, Naga IR follows the WGSL spec, and that says that the `clamp` function operates "component-wise when [the argument type] is a vector". It's fine not to bother handling vector arguments for now, but this should be a `NotImplemented` error, not an `InvalidMathArg` error.

src/front/glsl/constants.rs
```diff
@@ -225,6 +233,48 @@ impl<'a> ConstantSolver<'a> {
...
+                        };
+
+                        Ok(self.constants.fetch_or_append(
+                            Constant {
+                                name: None,
+                                specialization: None,
+                                inner: ConstantInner::Scalar { width, value },
+                            },
+                            span,
+                        ))
```
jimblandy, line 276: Rather than repeat this code for every `MathFunction` variant, it would be better to pull it out of the `match fun` so that all the `MathFunction`s can share it.

jimblandy (changes requested):

src/front/glsl/constants.rs
sjinno: Done, though I haven't handled vector arguments yet.

src/front/glsl/constants.rs
sjinno: Done.

src/front/glsl/constants.rs
```diff
@@ -201,6 +209,47 @@ impl<'a> ConstantSolver<'a> {
...
+                                    (
+                                        ScalarValue::Uint(a),
+                                        ScalarValue::Uint(b),
+                                        ScalarValue::Uint(c),
+                                    ) => ScalarValue::Uint(a.max(b).min(c)),
+                                    (
+                                        ScalarValue::Float(a),
+                                        ScalarValue::Float(b),
+                                        ScalarValue::Float(c),
+                                    ) => ScalarValue::Float(a.max(b).min(c)),
```
JCapucho, line 237: nit: Can we add just a comment with something like
```rust
// TODO: when the MSRV is higher or equal to 1.50 switch to f32::clamp
```

JCapucho, line 237: We need to figure out how the `NaN`s work with this. The glsl spec says
> min returns the minimum of the two parameters. It returns y if y is less than x, otherwise it returns x. 

while the rust standard library has

> If one of the arguments is NaN, then the other argument is returned.

I couldn't find how glsl defines the float ordering with `NaN`s, but standard `IEEE 754-2008` floats don't have a stable ordering between `NaN`s and any other number.

JCapucho (changes requested):
I have some concerns related to `NaN`s handling, that I would like addressed

src/front/glsl/constants.rs
sjinno: Yes.

sjinno: I'm so sorry for the delay. I've been meaning to get back to this, but I got a little too busy... But I am meaning to resume this unresolved issue, so I was wondering if there was any update on how the `NaN`s work with clamp. In the meantime, I'll also try digging into what I can find :)

JCapucho: Sorry at the time I didn't realize `clamp` panicked on `NaN` so we can't really use it.

JCapucho: I did found out how it worked, for `min` thanks to it's definition it's behavior is the same as rust. but for `max` this isn't the case so you'll need the following function
```rust
/// Helper function to implement the glsl `max` function for floats
///
/// While rust does provide a `f32::max` method it has a different behavior than the
/// glsl `max` for `NaN`s, in rust if any of the arguments is a `NaN` then the other
/// is returned.
///
/// This leads to different results in the following example
/// ```
/// max(NaN, 1.0)
/// ```
///
/// Rust will return 1.0 while glsl should return NaN
fn glsl_float_max(x: f32, y: f32) -> f32 {
    if x < y {
        y
    } else {
        x
    }
}
```

JCapucho: Wait sorry used some old notes, `min` is not equal between the two. you also need this function
```rust
/// Helper function to implement the glsl `min` function for floats
///
/// While rust does provide a `f32::min` method it has a different behavior than the
/// glsl `min` for `NaN`s, in rust if any of the arguments is a `NaN` then the other
/// is returned.
///
/// This leads to different results in the following example
/// ```
/// min(NaN, 1.0)
/// ```
///
/// Rust will return 1.0 while glsl should return NaN
fn glsl_float_min(x: f32, y: f32) -> f32 {
    if y < x {
        y
    } else {
        x
    }
}
```

sjinno: Thanks so much for this! I'm almost ready to push the new change, but I am wondering in which file I should be writing that function? `glsl/functions.rs`?

JCapucho: No it should be added in the same file since it concerns constant expression functions and not runtime functions

sjinno: Okay, thank you very much!

sjinno: I just pushed a new commit. Let me know if it works :) And thank you very much for your assist!
One thing maybe I should have gotten clarity on is that I changed `f32` to `f64` in the `glsl_float_min()` function, since `ScalarValue::Float` takes `f64`.

JCapucho: You forgot to also use the `max` modified variant

sjinno: Hopefully fixed it :)

JCapucho: Sorry didn't notice at first but you need to use `f64`s, when I posted the code I forgot that naga uses `f64`s

sjinno: Oh, no worries! That's what I was thinking too, but forgot to ask for clarity on it.

JCapucho (approved):
LGTM, just need to wait on @jimblandy approval

src/front/glsl/constants.rs
```diff
@@ -506,6 +557,48 @@ impl<'a> ConstantSolver<'a> {
     }
 }
 
+/// Helper function to implement the glsl `max` function for floats
+///
+/// While rust does provide a `f64::max` method it has a different behavior than the
+/// glsl `max` for `NaN`s, in rust if any of the arguments is a `NaN` then the other
+/// is returned.
```
jimblandy, line 564: Capitalize "Rust", and write "GLSL" in all caps. Also, this is a run-on sentence; please change ", in rust if" to ". In Rust, if"

src/front/glsl/constants.rs
```diff
@@ -506,6 +557,48 @@ impl<'a> ConstantSolver<'a> {
...
+///
+/// This leads to different results in the following example
+/// ```
+/// use std::cmp::max;
+/// std::f64::NAN.max(1.0);
+/// ```
+///
+/// Rust will return 1.0 while glsl should return NaN
```
jimblandy, line 572: Missing period at end of sentence.

src/front/glsl/constants.rs
```diff
@@ -506,6 +557,48 @@ impl<'a> ConstantSolver<'a> {
...
+fn glsl_float_max(x: f64, y: f64) -> f64 {
+    if x < y {
+        y
+    } else {
+        x
+    }
+}
+
+/// Helper function to implement the glsl `min` function for floats
```
jimblandy, line 581: Same comments for `glsl_float_min` as `glsl_float_max`.

jimblandy (approved):
Just some minor comments - other than that, this looks good.

PR #1505, opened 2021-11-02
 [glsl-in] Add a helper function to reduce repetition

sjinno:
In `src/fornt/glsl/constants.rs`, the following:

```
Ok(self.constants.fetch_or_append(
    Constant {
        name: None,
        specialization: None,
        inner,
    },
    span,
))
```

was called repetitively.

I added `ConstantSolver::register_constant(&mut self, ConstantInner, Span)`
that returns `Handle<Constant>` to reduce repetition.

### DIFF
--- a/src/front/glsl/constants.rs
+++ b/src/front/glsl/constants.rs
@@ -96,17 +96,11 @@ impl<'a> ConstantSolver<'a> {
                     }
                 };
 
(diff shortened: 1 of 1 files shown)


jimblandy (approved):
This looks good to me - thanks very much!

PR #1507, opened 2021-11-03
hlsl: respect array stride in storage buffers

### LINKED ISSUE #1504: HLSL back-end ignores stride on storage buffers
Shader playground: https://shader-playground.timjones.io/4e4be36000db47ee5a8fadfe4e44701c

While debugging https://github.com/LPGhatguy/crevice/runs/4081541660 for LPGhatguy/crevice#27, I've faced the unfortunate reality that copying into/out of interface arrays ignores the array stride altogether. This means WGPU code on DX backend will be cross-non-portable, erroneously so probably, with Vulkan and GL backends.

Arrays of vec3 are not exactly rare, so I propose we fix this ASAP, this looks to me like a serious compatibility issue.

kvark:
Fixes #1504

The code had to go through a bit of a refactor. The end result is very type-verbose, but I think it should also be cleaner to read.

### DIFF
--- a/src/back/hlsl/storage.rs
+++ b/src/back/hlsl/storage.rs
@@ -367,39 +367,22 @@ impl<W: fmt::Write> super::Writer<'_, W> {
         mut cur_expr: Handle<crate::Expression>,
         func_ctx: &FunctionCtx,
     ) -> Result<Handle<crate::GlobalVariable>, Error> {
+        enum AccessIndex {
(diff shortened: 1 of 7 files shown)


ElectronicRU:
thanks for picking up the slack, LG!

kvark:
I'm fairly confident in this code, merging.

kvark:
Published in 0.7.2

PR #1512, opened 2021-11-04
[WGSL] Only single bracket for attributes

mehmetoguzderin:
This PR implements single brackets for attributes which enables middleware developers to adopt syntax like [[ and ]], #[ and ], @[ and ], ![ and ] for their tagging purposes and significantly reduces horizontal bloat while enabling much faster authoring on digital and physical mediums.


mehmetoguzderin:
(closing as undesired by the group)

Issue #1272, opened 2021-08-21
Incorrect translation of spirv with float64 (dvec2)


dhardy:
KAS's [mandlebrot example](https://github.com/kas-gui/kas/tree/master/examples/mandlebrot) does not render with 64-bit shaders (`--features shader64`) since upgrading to wgpu v0.10 (leaving a black background, which is not the clear colour). It runs correctly with 32-bit shaders (almost identical code aside from the fragment shader).

An API trace is attached: [trace.tar.gz](https://github.com/gfx-rs/naga/files/7025630/trace.tar.gz)

Using `wgpu::Features::SPIRV_SHADER_PASSTHROUGH` the example works correctly, therefore I deduce the error lies with Naga rather than (other parts of) WGPU. The following patch achieves this:
```diff
diff --git a/examples/mandlebrot/mandlebrot.rs b/examples/mandlebrot/mandlebrot.rs
index bf1b202b..03e23cbb 100644
--- a/examples/mandlebrot/mandlebrot.rs
+++ b/examples/mandlebrot/mandlebrot.rs
@@ -45,10 +45,12 @@ impl Shaders {
         let vertex = device.create_shader_module(&include_spirv!("shader.vert.spv"));
         // Note: we don't use wgpu::include_spirv since it forces validation,
         // which cannot currently deal with double precision floats (dvec2).
-        let fragment = device.create_shader_module(&wgpu::ShaderModuleDescriptor {
-            label: Some("fragment shader"),
-            source: wgpu::util::make_spirv(FRAG_SHADER),
-        });
+        let fragment = unsafe {
+            device.create_shader_module_spirv(&wgpu::ShaderModuleDescriptorSpirV {
+                label: Some("fragment shader"),
+                source: wgpu::util::make_spirv_raw(FRAG_SHADER),
+            })
+        };
 
         Shaders { vertex, fragment }
     }
@@ -94,7 +96,9 @@ impl CustomPipeBuilder for PipeBuilder {
     fn device_descriptor() -> wgpu::DeviceDescriptor<'static> {
         wgpu::DeviceDescriptor {
             label: None,
-            features: wgpu::Features::PUSH_CONSTANTS | SHADER_FLOAT64,
+            features: wgpu::Features::PUSH_CONSTANTS
+                | wgpu::Features::SPIRV_SHADER_PASSTHROUGH
+                | SHADER_FLOAT64,
             limits: wgpu::Limits {
[...]

PR #1514, opened 2021-11-06
[wgsl-out] Write correct scalar kind when width != 4

fintelia:
Related to https://github.com/gfx-rs/naga/issues/1272#issuecomment-945044963

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -1424,17 +1424,18 @@ impl<W: Write> Writer<W> {
                         }
                     }
                     TypeInner::Scalar { width, .. } => {
+                        let scalar = scalar_kind_str(kind, convert.unwrap_or(width));
                         if convert.is_some() {
                             write!(
                                 self.out,
                                 "{}",
-                                scalar_kind_str(kind, convert.unwrap_or(width))
+                                scalar
                             )?
                         } else {
                             write!(
                                 self.out,
                                 "bitcast<{}>",
-                                scalar_kind_str(kind, convert.unwrap_or(width))
+                                scalar
                             )?
                         }
                     }
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -452,11 +452,11 @@ impl<W: Write> Writer<W> {
     /// Adds no trailing or leading whitespace
     fn write_value_type(&mut self, module: &Module, inner: &TypeInner) -> BackendResult {
         match *inner {
-            TypeInner::Vector { size, kind, .. } => write!(
+            TypeInner::Vector { size, kind, width } => write!(
                 self.out,
                 "vec{}<{}>",
                 back::vector_size_str(size),
-                scalar_kind_str(kind),
+                scalar_kind_str(kind, width),
             )?,
             TypeInner::Sampler { comparison: false } => {
                 write!(self.out, "sampler")?;
@@ -478,7 +478,7 @@ impl<W: Write> Writer<W> {
                     Ic::Sampled { kind, multi } => (
                         "",
                         if multi { "multisampled_" } else { "" },
-                        scalar_kind_str(kind),
+                        scalar_kind_str(kind, 4),
                         "",
                     ),
                     Ic::Depth { multi } => {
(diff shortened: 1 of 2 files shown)


src/back/wgsl/writer.rs
```diff
@@ -1353,9 +1353,9 @@ impl<W: Write> Writer<W> {
                             back::vector_size_str(rows)
                         )?;
                     }
-                    TypeInner::Vector { size, .. } => {
+                    TypeInner::Vector { size, width, .. } => {
                         let vector_size_str = back::vector_size_str(size);
-                        let scalar_kind_str = scalar_kind_str(kind);
+                        let scalar_kind_str = scalar_kind_str(kind, convert.unwrap_or(width));
```
JCapucho, line 1358: I know you didn't write the code for `Expression::As` but this seems wrong, we are using the type of the inner expression instead of the current expression, but maybe I'm reading this wrong.

JCapucho, line 1358: What's supposed to happen to code like this
```rust
let a: f64 = f64(0.0);
let b: f32 = bitcast<f32>(a);
```
or
```rust
let a: f32 = f32(0.0);
let b: f64 = bitcast<f64>(a);
```
From what I could see the wgsl spec doesn't define what happens if the width of the types isn't equal. Naga doesn't even like this code and uses the width of the inner expression, so the code doesn't compile
```
[2021-11-07T15:07:09Z ERROR naga::front::wgsl] Given type Scalar { kind: Float, width: 4 } doesn't match expected Scalar { kind: Float, width: 8 }
error: the type of `b` is expected to be [3]
  ┌─ wgsl:4:9
  │
4 │     let b: f32 = bitcast<f32>(a);
  │         ^ definition of `b`

Could not parse WGSL
```
I may have skipped something while reading the spec, so if @kvark could help us here

src/back/wgsl/writer.rs
```diff
@@ -1353,9 +1353,9 @@ impl<W: Write> Writer<W> {
                             back::vector_size_str(rows)
```
JCapucho, line 1353: You seem to be missing code for when the matrix is of scalar width 8

JCapucho (dismissed):
I gave it a quick look, but I have some concerns about certain parts of the current code and the wgsl spec that I would like clarified first before giving a more in depth review.

teoxoy:
I think we shouldn't move forward with this approach. WGSL only supports 32 bit numbers and only recently added 16 bit floating point support trough an extension.

We should instead error if we encounter anything that's not width == 4 and when we get to implement the f16 extension handle width == 2 then.

teoxoy:
I retract my previous comment. It looks like we already support non 32 bit numbers in the WGSL front-end and also handle validation of widths (depending on capabilities).

teoxoy:
@fintelia any updates on this?

fintelia:
The current status is that I totally lost track of this PR and if anyone else wants to take over they should feel free. Otherwise, if this is something that the gfx-rs maintainers want to go forward with, when I have a chance I'd be happy to fix the conflicts and try to get this ready to merge.

src/back/wgsl/writer.rs
jimblandy: The WGSL spec [only allows](https://gpuweb.github.io/gpuweb/wgsl/#bitcast-expr) conversions between values that are the same size. So the programs above should be rejected as type errors.

teoxoy: I think so far the assumption was that we always bitcast between types with same widths (i.e. scalar to scalar, vector to vector, matrix to matrix) but this does indeed not work with different widths.

Looks like we also do this in all the other backends. Will open an issue.

teoxoy (approved):
There are still some unknowns but let's merge this for now.

We'll go over non 4 widths at a broader scale once we look into implementing the f16 extension for WGSL (which will require for instance adding width data to matrices in the IR).

src/back/wgsl/writer.rs
```diff
@@ -1916,14 +1924,25 @@ const fn image_dimension_str(dim: crate::ImageDimension) -> &'static str {
...
+        (Sk::Sint, 2) => "i16",
+        (Sk::Sint, 8) => "i64",
+        (Sk::Sint, _) => "i32",
+        // Unsigned integer
+        (Sk::Uint, 1) => "u8",
+        (Sk::Uint, 2) => "u16",
+        (Sk::Uint, 8) => "u64",
+        (Sk::Uint, _) => "u32",
+        // Boolean
+        (Sk::Bool, _) => "bool",
```
teoxoy, line 1945:
```suggestion
    match (kind, width) {
        (Sk::Float, 8) => "f64",
        (Sk::Float, 4) => "f32",
        (Sk::Sint, 4) => "i32",
        (Sk::Uint, 4) => "u32",
        (Sk::Bool, 1) => "bool",
        _ => unreachable!(),
```

We are removing the non-32bit integers in https://github.com/gfx-rs/naga/pull/2146 since the validator never allows those to go through.

teoxoy (changes requested):

PR #1515, opened 2021-11-06
[spv-out] Fix word order for constants

fintelia:
Per the SPIR-V spec:

> For a numeric literal, the lower-order words appear first.


fintelia:
With this patch, [terra](github.com/fintelia/terra) can upgrade past wgpu 0.9 without needing to use SPIR-V passthrough :tada:

JCapucho (approved):
Looks good to me

kvark:
Published in 0.7.2

PR #1517, opened 2021-11-07
[glsl-in] Implement matrix to matrix constructor

### LINKED ISSUE #1497: [glsl-in] naga panics on an attempt to construct a mat2x4 from a mat3
As per "5.4.2 Vector and Matrix Constructors" in the GLSL 4.5 specification, the following is legal:

```
#version 450

mat2x4 m = mat2x4(mat3(1.0));

void main()
{
}
```

However, naga (version 9d2b357214dacce0fe30587844d4c7bf41b51773) panics:

```
thread 'main' panicked at 'index out of bounds: the len is 3 but the index is 3', src/front/glsl/constants.rs:147:32
```

JCapucho:
> If a matrix is constructed from a matrix, then each component
(column i, row j) in the result that has a corresponding component
(column i, row j) in the argument will be initialized from there. All
other components will be initialized to the identity matrix.

This was previously not implemented causing a matrix to be composed from a matrix which wouldn't pass validation.

Closes #1497

### DIFF
--- a/src/front/glsl/functions.rs
+++ b/src/front/glsl/functions.rs
@@ -55,7 +55,8 @@ impl Parser {
         match fc {
             FunctionCallKind::TypeConstructor(ty) => {
                 if args.len() == 1 {
-                    self.constructor_single(ctx, body, ty, args, meta).map(Some)
+                    self.constructor_single(ctx, body, ty, args[0], meta)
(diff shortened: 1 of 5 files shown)


kvark (approved):

kvark:
Please rebase and merge

PR #1518, opened 2021-11-07
[glsl-in] Don't panic on UB integer operations

### LINKED ISSUE #1496: [glsl-in] naga panics on shader with obvious undefined behaviour
This shader has a statically identifiable undefined behaviour, due to shifting by a negative value:

```
#version 450

void main()
{
  int x = 1 << -1;
}
```

It causes naga (9d2b357214dacce0fe30587844d4c7bf41b51773) to panic:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/front/glsl/constants.rs:327:14
```

Perhaps this is fine, but I thought I'd report it in case it would be easy to have naga report a graceful error message instead.

JCapucho:
Forces wrapping for all operations, except that for division/remainder when `rhs == 0` the result will be `0`.

Closes #1496

### DIFF
--- a/src/front/glsl/constants.rs
+++ b/src/front/glsl/constants.rs
@@ -293,16 +293,6 @@ impl<'a> ConstantSolver<'a> {
         target_width: crate::Bytes,
         span: crate::Span,
     ) -> Result<Handle<Constant>, ConstantSolvingError> {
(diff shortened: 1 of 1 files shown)


kvark (approved):

kvark:
Published in 0.7.2

PR #1520, opened 2021-11-09
[glsl-in] Allow constructors from non-scalar to scalar

### LINKED ISSUE #1493: [glsl-in] naga disallows constructing a float from a matrix
This GLSL is valid according to "5.4.1 Conversion and Scalar Constructors" of the GLSL 4.5 spec, which says:

> Scalar constructors with non-scalar parameters can be used to take the first element from a non-scalar.

```
#version 450

const float f = float(mat2(1.0));

void main()
{
}
```

It is rejected by naga revision 9d2b357214dacce0fe30587844d4c7bf41b51773 with:

```
error: Can't resolve type: IncompatibleOperands("Matrix { columns: Bi, rows: Bi, width: 4 } as Float")
```

The shader is accepted by glslangValidator.

naga does currently allow constructing a float from e.g. a vec2.

JCapucho:
Glsl defines in section 5.4.1:

> Scalar constructors with non-scalar parameters can be used to take the first element from a non-scalar. For example, the constructor float(vec3) will select the first component of the vec3

From my testing this only applies if the non-scalar argument is either a matrix or a vector.

Closes #1493

### DIFF
--- a/src/front/glsl/functions.rs
+++ b/src/front/glsl/functions.rs
@@ -117,15 +117,42 @@ impl Parser {
 
                 ctx.add_expression(Expression::Splat { size, value }, meta, body)
             }
-            TypeInner::Scalar { kind, width } => ctx.add_expression(
-                Expression::As {
-                    kind,
-                    expr: value,
-                    convert: Some(width),
-                },
-                meta,
-                body,
-            ),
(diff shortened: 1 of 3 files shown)


kvark (approved):
oh GLSL, y u so complex

PR #1522, opened 2021-11-10
[spv-in] Permit pointers to runtime arrays only in StorageBuffer.

### LINKED ISSUE #1519: SPIR-V front end generates invalid Naga types for pointers to unsized arrays
Naga IR permits `Dynamic`-sized arrays only in the `Storage` storage class, but the SPIR-V front end creates Naga `Pointer` types to `Dynamic`-sized arrays in the `Uniform` storage class. I think these types are not used, but their presence prevents us from tightening up the validation of pointer types as much as I'd like.

For example, the Naga IR for `tests/in/spv/shadow.spv` includes the following types:

        [19]: Type {
            name: None,
            inner: Array {
                base: [18],
                size: Dynamic,
                stride: 96,
            },
        },
        ...
        [22]: Type {
            name: None,
            inner: Pointer {
                base: [19],
                class: Uniform,
            },
        },

SPIR-V allows runtime-sized arrays to appear only as the last member of a struct in the `StorageBuffer` storage class, or in the `Uniform` storage class when the struct type is decorated with `BufferBlock`. Naga's SPIR-V front end notices the `BufferBlock` decoration, and assigns all `GlobalVariable`s of such types to the Naga `Storage` storage class, regardless of the class given in the `OpVariable` instruction. In this way, the SPIR-V front end respects Naga's requirement that all `Dynamic`-sized arrays appear in `Storage`.

However, in the SPIR-V, all the pointers derived from such variables respect the class given in the `OpVariable`, so there are plenty of pointers to runtime-sized arrays in the `Uniform` storage class floating around in code like `tests/in/spv/shadow.spv`. I imagine some of these must be provoking the SPIR-V front end to also emit a Naga `Uniform` `Pointer` type to a `Dynamic`-sized array.

jimblandy:
Fixes #1519.

[spv-in] Change `shadow.spv` test input to use `StorageBuffer`.

The ecosystem around Naga will generally not be able to use Vulkan adapters that don't support the `SPV_KHR_storage_buffer_storage_class` (which was incorporated into SPIR-V 1.3), so we can assume it is present.

Changing the test not to use runtime-sized arrays in the `Uniform` storage class will allow Naga to tighten up some validation checks.


jimblandy:
I'm not sure this is the sort of fix you had in mind in #1519. Happy to revise.

kvark (approved):

PR #1524, opened 2021-11-12
[spv-out] Fix scalar-times-matrix operations

parasyte:
Converting some GLSL to WGSL, I ran into a panic in the SpirV backend: `not implemented: Mul (Scalar, Matrix)`. The fix is super simple. Tests are included.

Interesting, the HLSL backend already does this same transformation. Cool!

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -346,6 +346,10 @@ impl<'w> BlockContext<'w> {
                         (Dimension::Matrix, Dimension::Scalar { .. }) => {
                             spirv::Op::MatrixTimesScalar
                         }
+                        (Dimension::Scalar, Dimension::Matrix { .. }) => {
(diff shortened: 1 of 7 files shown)


parasyte:
The MSL validation error is interesting. Can anyone provide pointers on that?

tests/in/scalar-times-matrix.wgsl
```diff
@@ -0,0 +1,11 @@
+[[stage(vertex)]]
```
kvark, line 1: let's not have a separate test file for each operation, instead put it into `operations.wgsl`

kvark (changes requested):
Thank you for the PR!

parasyte:
@kvark I've addressed your feedback. Thanks for the review!

kvark (approved):
thank you!

PR #1529, opened 2021-11-14
[valid] Check that switches have a default case

Bobo1239:
From the WGSL spec: "Each switch statement must have exactly one default clause."

Noticed this because [this switch](https://github.com/hecrj/wgpu_glyph/blob/cd6f154db0e7dd8231fdd34fcac470f975c825d7/src/shader/glyph.wgsl#L35-L52) doesn't get rejected by naga but Chrome/Tint reports:
```
Tint WGSL reader failure:
Parser: 35:5 error: switch statement must have a default clause
    switch (i32(input.vertex_index)) {
    ^^^^^^
```


kvark (approved):
Wow, I didn't realize it's mandatory

PR #1530, opened 2021-11-14
Forbid specializable constants as array lengths.

jimblandy:
Only the SPIR-V front end can produce such array types, and it seems that our back ends don't support this case well.

### DIFF
--- a/src/proc/index.rs
+++ b/src/proc/index.rs
@@ -49,10 +49,6 @@ pub enum IndexableLength {
     /// Values of this type always have the given number of elements.
     Known(u32),
 
-    /// The value of the given specializable constant is the number of elements.
-    /// (Non-specializable constants are reported as `Known`.)
-    Specializable(crate::Handle<crate::Constant>),
-
     /// The number of elements is determined at runtime.
     Dynamic,
 }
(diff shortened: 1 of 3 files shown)


src/proc/index.rs
```diff
@@ -49,10 +49,6 @@ pub enum IndexableLength {
     /// Values of this type always have the given number of elements.
     Known(u32),
 
-    /// The value of the given specializable constant is the number of elements.
-    /// (Non-specializable constants are reported as `Known`.)
-    Specializable(crate::Handle<crate::Constant>),
```
kvark, line 54: Did you confirm with the latest working group notes that this is not going to be needed ever? I recall that the group was about to agree on supporting this.

jimblandy: There's quite a saga here.

- gpuweb/gpuweb#1272 was the PR that made array lengths pipeline-overridable.
- However, that PR was rejected in favor of gpuweb/gpuweb#2031, which allowed array lengths to be module-scoped constants, but not overridable. This was merged on August 11th.
- However, two days after that, gpuweb/gpuweb#2024 started reintroducing the idea.
- gpuweb/gpuweb#2238 pursued some follow-up.

I don't yet understand the last two. The conversations are long; I'll need to get back to them later.

I still think we should remove this enum variant for now. Including it requires all back ends to handle it, which they then just punt on by calling `to_array_length` instead of `indexable_length`. Even if we wanted to handle it, we don't have spec language yet to tell us exactly what to do.

We can always get everything back from version control when the time comes.

kvark: ok, my recollection was that Apple agreed to make the workgroup-class array sizes overridable, but I was missing for a while after that. I trust your judgement, let's proceed.

PR #1532, opened 2021-11-15
Msl bounds checks

### LINKED ISSUE #1531: Metal output needs index bounds checks
Metal output should incorporate bounds checks as directed by a `BoundsCheckPolicy`, as SPIR-V output does.

jimblandy:
Fixes #1531.

Remaining work:

- [x] fix macOS validation errors
- [x] Don't forget cast to unsigned in bounds checks
- [x] Generate MSL for `policy-mix.wgsl` too
- [x] handle AccessIndex on dynamic arrays
- [x] bother to check policy for each index in `analyze_function`
    - [x] consider caching policy decisions in `IndexInfo`? 
- [x] make `choose_checks` reuse a Vec to save allocation
- (leave for followup) Unify `GuardedIndex` and `back::spv`'s `MaybeKnown`
- [x] produce zero values for non-scalar loads


jimblandy:
Because GitHub doesn't understand dependent PRs, this just incorporates a bunch of commits that I think could be reviewed and landed separately. I'll sort these out tomorrow.

src/back/msl/writer.rs
```diff
@@ -671,55 +671,57 @@ impl<W: Write> Writer<W> {
         Ok(())
     }
 
-    fn put_array_length(
+    /// Write the length of the dynamically sized array at the end of `handle`.
+    ///
+    /// `expr` must be the handle of a global variable whose final member is a dynamically
```
kvark, line 678: comment is outdated?

src/back/msl/writer.rs
```diff
@@ -671,55 +671,57 @@ impl<W: Write> Writer<W> {
...
+            crate::TypeInner::Array { base, stride, .. } => (
+                context.module.types[base]
+                    .inner
+                    .span(&context.module.constants),
+                stride,
+            ),
+            _ => return Err(Error::Validation),
+        };
+
+        // When the stride length is larger than the span, the final element's stride of
```
kvark, line 706: nice comments!

kvark (commented):
Looks great!

src/back/msl/writer.rs
```diff
@@ -846,6 +860,12 @@ impl<W: Write> Writer<W> {
             crate::Expression::GlobalVariable(handle) => {
                 let name = &self.names[&NameKey::GlobalVariable(handle)];
                 write!(self.out, "{}", name)?;
+                log::trace!(
```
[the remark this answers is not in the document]
jimblandy: No - I'll take out all the logging before landing.

src/back/msl/writer.rs
```diff
@@ -1203,14 +1223,83 @@ impl<W: Write> Writer<W> {
...
+        );
+        if !comparisons.is_empty() {
+            if !is_scoped {
+                write!(self.out, "(")?;
+            }
+
+            self.put_condition(&comparisons, pointer, context)?;
+            write!(self.out, " ? ")?;
+            self.put_unchecked_load(pointer, context, false)?;
+            write!(self.out, " : 0")?;
```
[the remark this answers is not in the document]
jimblandy: This is more explicit in the newer version of the patch.

src/back/msl/writer.rs
```diff
@@ -1739,6 +1879,29 @@ impl<W: Write> Writer<W> {
...
+            context.expression.function,
+            context.expression.info,
+        );
+        if !comparisons.is_empty() {
+            write!(self.out, "{}if (", level)?;
+            self.put_condition(&comparisons, pointer, &context.expression)?;
+            writeln!(self.out, ")")?;
+            self.put_unchecked_store(pointer, value, level.next(), context)
+        } else {
+            self.put_unchecked_store(pointer, value, level, context)
```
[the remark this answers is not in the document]
jimblandy: I've changed this to always put the store in brackets. That did make it awkward to move the `put_unchecked_store` out of the `if`. Since this was "nit", I didn't worry about it - but if you think it would still look better to hoist out that call (and add plumbing to indent it properly), we can totally do that.

src/proc/index.rs
```diff
@@ -138,6 +147,277 @@ impl BoundsCheckPolicies {
...
+        policies: BoundsCheckPolicies,
+    ) -> IndexInfo {
+        use crate::Expression as Ex;
+
+        log::trace!("analyze_function: {:?}", policies);
+
+        let mut guarded_indices = HandleBitSet::default();
+
+        // Don't bother scanning if we never need `ReadZeroSkipWrite`.
+        if policies.index == BoundsCheckPolicy::ReadZeroSkipWrite
```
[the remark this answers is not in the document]
jimblandy: I think we can just treat texel coordinates like everything else. We will need to check the policy for them.

src/proc/index.rs
```diff
@@ -138,6 +147,277 @@ impl BoundsCheckPolicies {
...
+    /// [`Load`]: crate::Expression::Load
+    /// [`Store`]: crate::Statement::Store
+    pub fn choose_checks(
+        &self,
+        mut pointer: Handle<crate::Expression>,
+        module: &crate::Module,
+        function: &crate::Function,
+        info: &crate::valid::FunctionInfo,
+    ) -> Vec<GuardedAccess> {
+        let mut accesses = vec![];
```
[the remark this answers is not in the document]
jimblandy: The new approach avoids all this.

jimblandy:
Rebased, and a lot simpler.

src/back/msl/writer.rs
```diff
@@ -2,9 +2,11 @@ use super::{sampler as sm, Error, LocationMode, Options, PipelineOptions, Transl
 use crate::{
     arena::Handle,
     back,
+    proc::index,
     proc::{self, NameKey, TypeResolution},
```
kvark, line 6: please merge with the other `proc::`

src/back/msl/writer.rs
```diff
@@ -1265,9 +1213,376 @@ impl<W: Write> Writer<W> {
...
+                    write!(self.out, "{}{}", level, prefix)?;
+                    check_written = true;
+                }
+
+                // Check that the index falls within bounds. Do this with a single
+                // comparison, by casting the index to `uint` first, so that negative
+                // indices become large positive values.
+                write!(self.out, "{}::uint(", NAMESPACE)?;
+                self.put_index(index, context, true)?;
+                self.out.write_str(") < ")?;
```
kvark, line 1305: let's not mix different styles of writing. Since `write!` is used in other places, let's just use it here as well

tests/snapshots.rs
```diff
@@ -78,10 +78,26 @@ struct Parameters {
...
-        Ok(string) => ron::de::from_str(&string).expect("Couldn't parse param file"),
-        Err(_) => Parameters::default(),
+    let params = {
+        let params =
+            match fs::read_to_string(format!("{}/{}/{}.param.ron", root, BASE_DIR_IN, name)) {
+                Ok(string) => ron::de::from_str(&string).expect("Couldn't parse param file"),
+                Err(_) => Parameters::default(),
+            };
+
+        // Make MSL output parameters use params.bounds_check_policies.
```
kvark, line 88: why aren't we doing the same thing as we do for SPIR-V? it passes the bounds check policies separately

tests/snapshots.rs
```diff
@@ -458,10 +474,11 @@ fn convert_wgsl() {
             "globals",
             Targets::SPIRV | Targets::METAL | Targets::GLSL | Targets::HLSL | Targets::WGSL,
         ),
-        ("bounds-check-zero", Targets::SPIRV),
+        ("bounds-check-zero", Targets::SPIRV | Targets::METAL),
```
kvark, line 477: horray!

tests/in/bounds-check-restrict.wgsl
```diff
@@ -0,0 +1,63 @@
+// Tests for `naga::back::BoundsCheckPolicy::Restrict`.
+
+[[block]]
+struct Globals {
+    a: array<f32, 10>;
+    v: vec4<f32>;
+    m: mat3x4<f32>;
```
kvark, line 7: do we have a test ensuring that the policy is omitted for cases where we already know the index is in bounds?

src/back/msl/writer.rs
```diff
@@ -1265,9 +1213,376 @@ impl<W: Write> Writer<W> {
...
+            match limit {
+                index::IndexableLength::Known(limit) => {
+                    write!(self.out, "{}u)", limit - 1)?;
+                }
+                index::IndexableLength::Dynamic => {
+                    let global = base
+                        .originating_global(context.function)
+                        .ok_or(Error::Validation)?;
+                    self.put_dynamic_array_length(global, context)?;
+                    write!(self.out, " - 1)")?;
```
kvark, line 1488: this is a little unfortunate, since `put_dynamic_array_length` already adds "+1".
Perhaps, a better way to approach this would be to change the semantics of `put_dynamic_array_length` into `put_last_index_of_dynamic_array`? This would end up with the following changes:
  - the `arrayLength()` implementations would then append "+1", while checking for `is_scoped` to see if "("/")" are needed. This would be a slight improvement by itself!
  - the in-bounds checks could do "xxx <= last_index" instead of the current "xxx < last_index + 1"
  - the last index output would just be "last_index" instead of the current "last_index + 1 - 1"

kvark (changes requested):
Almost ready to go, amazing work!
Just a few last notes

tests/snapshots.rs
jimblandy: Yeah!!

tests/in/bounds-check-restrict.wgsl
jimblandy: Good idea. Added.

src/back/msl/writer.rs
jimblandy: Great ideas. Done.

src/proc/mod.rs
```diff
@@ -313,6 +313,33 @@ impl crate::Expression {
...
+    /// return a handle for that global.
+    ///
+    /// If the expression does not ultimately access a global variable, return
+    /// `None`.
+    pub fn originating_global(
+        mut self,
+        func: &crate::Function,
+    ) -> Option<crate::Handle<crate::GlobalVariable>> {
+        loop {
+            self = match func.expressions[self] {
```
kvark, line 330: whaaat, nasty stuff!
Please, let's not add methods with any logic on handles. They are supposed to be just data.
Instead, we can do it with `impl crate::Function`

kvark (commented):
A few uber-nits left, almost good to go!

src/proc/mod.rs
jimblandy: LOL - sorry, I had forgotten about that principle. I'll try to remember it in the future.

kvark (approved):

PR #1534, opened 2021-11-16
Check that stores are permitted by the pointer's storage access.

### LINKED ISSUE #1533: Naga validation doesn't reject stores to read-only variables
[EDIT: I originally misidentified which rule was being broken here. The lead para has been changed.]

Naga fails to reject assignments to variables in the `storage` storage class declared with a default access mode. WGSL says that the default for such variables should be `read`, not `read_write`. For example, the following WGSL input is not rejected:

```
[[block]]
struct Globals {
    i: i32;
};

[[group(0), binding(0)]]
var<storage> globals: Globals;

fn store(v: i32) {
    globals.i = v;
}
```

jimblandy:
Fixes #1533.

### DIFF
--- a/src/proc/mod.rs
+++ b/src/proc/mod.rs
@@ -184,6 +184,21 @@ impl super::TypeInner {
     }
 }
 
+impl super::StorageClass {
+    pub fn access(self) -> crate::StorageAccess {
+        use crate::StorageAccess as Sa;
+        match self {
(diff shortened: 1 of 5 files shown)


kvark (approved):
Nice!

kvark:
Published in 0.7.2

PR #1535, opened 2021-11-16
Introduce `HandleBitSet` type.

jimblandy:
This is a prerequisite for #1532.

### DIFF
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -6,6 +6,7 @@ use std::{cmp::Ordering, fmt, hash, marker::PhantomData, num::NonZeroU32, ops};
 type Index = NonZeroU32;
 
 use crate::Span;
+use bit_set::BitSet;
 use indexmap::set::IndexSet;
 
 /// A strongly typed reference to an arena item.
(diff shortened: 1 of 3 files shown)


src/arena.rs
```diff
@@ -575,3 +576,94 @@ where
...
     }
 }
+
+/// A set of arena handles, represented as a bit vector.
+///
+/// This represents a set of `Handle`s simply by storing a bit for every
+/// `Handle` index. The underlying bit vector is sized dynamically, so storage
+/// depends on the highest handle index that has been inserted.
+///
+/// A `HandleBitSet<T>` will only hold `Handle<T>` values, so it's a bit more
```
kvark, line 586: is the strict typing useful here, considering that it's only used in the validator and the backends? I.e. it's hardly a part of a public API

src/arena.rs
```diff
@@ -575,3 +576,94 @@ where
...
+/// strictly typed than using `BitSet` directly.
+///
+/// This is a good representation for a handle set if you expect it to typically
+/// contain many of the arena's handles. In that situation, `HandleBitSet` will use
+/// a fraction of the memory of a `HashSet`. But if you expect the set to hold only
+/// a small proportion of the arena's handles, `HashSet` would be a better choice.
+pub struct HandleBitSet<T> {
```
kvark, line 593: if we are to have it, perhaps we can move it somewhere closer to the place where it's needed (i.e. a submodule `valid/handle_set.rs`).

jimblandy:
We can leave this out. I thought it was kind of nice to have the typing, but even with the full patch it ended up playing a very minor role.

PR #1536, opened 2021-11-16
Tighten checks on pointers to unsized values, and pointer arguments

### LINKED ISSUE #1513: Naga should reject pointers to unsized structs in the wrong storage class
Naga accepts the following code:

```
struct Unsized { data: array<f32>; };
fn acceptable_pointer_to_unsized(arg: ptr<workgroup, Unsized>) { }
```

It should reject it, because unsized arrays may only appear in the `storage` storage class, never the `workgroup`storage class. [_edit_: They can't be placed in the `uniform` storage class.]

(This code is a test case in `tests/wgsl-errors.rs`.)

jimblandy:
Pointers should not be `DATA`: they can never be stored in anything. (Function arguments are not storage; they're like `let` bindings.)

Un-`SIZED` values may only live in the `Storage` storage class, so creating pointers to them in other storage classes is meaningless.

The `ARGUMENT` flag should be set only on pointers in those storage classes that are permitted to be passed to functions.

See comments in code for details.

Fixes #1513.

[spv-out] Don't support arguments pointing to unsized types.

Functions with such arguments could never have been called anyway, and now they are forbidden in validation.

### DIFF
--- a/tests/out/wgsl/pointers.wgsl
+++ b/tests/out/wgsl/pointers.wgsl
@@ -14,22 +14,16 @@ fn f() {
     return;
 }
 
-fn index_dynamic_array(p: ptr<workgroup, DynamicArray>, i: i32, v_1: u32) -> u32 {
-    let old: u32 = (*p).array_[i];
-    (*p).array_[i] = v_1;
-    return old;
-}
-
-fn index_unsized(i_1: i32, v_2: u32) {
-    let val: u32 = dynamic_array.array_[i_1];
-    dynamic_array.array_[i_1] = (val + v_2);
+fn index_unsized(i: i32, v_1: u32) {
+    let val: u32 = dynamic_array.array_[i];
+    dynamic_array.array_[i] = (val + v_1);
     return;
 }
 
-fn index_dynamic_array_1(i_2: i32, v_3: u32) {
-    let p_1: ptr<storage, array<u32>, read_write> = (&dynamic_array.array_);
-    let val_1: u32 = (*p_1)[i_2];
-    (*p_1)[i_2] = (val_1 + v_3);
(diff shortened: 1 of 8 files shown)


tests/out/wgsl/pointers.wgsl
```diff
@@ -11,9 +6,3 @@ fn f() {
     return;
 }
 
-fn index_dynamic_array(p: ptr<workgroup, DynamicArray>, i: i32, v_1: u32) -> u32 {
```
kvark, line 14: can we just switch this to "storage" instead of removing?

kvark (changes requested):

tests/out/wgsl/pointers.wgsl
jimblandy: Well, that's the thing: you can't pass `ptr<storage, T>` to user-defined functions. There is no way to pass a pointer to any unsized type to a function.

What we can do, instead, is test using pointers to dynamically sized arrays, within a function. I can add a test for that.

kvark: Right, yes, we should have this test still. Thank you!

jimblandy: This change is done, but it depends on #1543.

jimblandy:
Note, the most recent push incorporates the changes in #1543, because GitHub doesn't know how to handle PRs that extend other PRs.

tests/out/wgsl/pointers.wgsl
jimblandy: (See `tests/in/pointers.wgsl`, for the new pointer-to-unsized test.)

kvark (approved):

PR #1537, opened 2021-11-16
Move `BoundsCheckPolicy`/`Policies` into `proc`, from `back`.

jimblandy:
I'll be adding more back-end-agnostic indexing code to `proc/index.rs`.

An alternative to this would be to create `src/back/index.rs`.


kvark (approved):
No strong feelings about this :)

kvark:
there is a conflict, please proceed when resolved

PR #1539, opened 2021-11-16
Switch to rustc-hash

bjorn3:
See gfx-rs/rspirv#221 for the rationale behind this change


kvark (approved):
That's an interesting change!

PR #1541, opened 2021-11-16
Add view_index wgsl builtin

LaylBongers:
This builtin isn't part of the specification, but is a simple addition to add multi-view support for WGSL. Is there an extension mechanism we can make use of here? I'm too unfamiliar with WGSL what the best approach here is to take.

Related to: https://github.com/gfx-rs/wgpu/issues/2186

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -1841,6 +1841,7 @@ fn builtin_str(built_in: crate::BuiltIn) -> Option<&'static str> {
         Bi::SampleIndex => Some("sample_index"),
         Bi::SampleMask => Some("sample_mask"),
         Bi::PrimitiveIndex => Some("primitive_index"),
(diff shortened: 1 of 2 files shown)


LaylBongers:
Given that this is already behind a `Capability`, maybe this should be fine anyways, I'll take it off draft status.

kvark (approved):
Might as well put it into `src/back/wgsl/` too?

LaylBongers:
Updated the commit so it's added to the backend for WGSL as well

kvark:
Last question, is this really going to be a "view" index? Or could it be more specifically a "layer" index instead?

LaylBongers:
The feature this maps to in OpenGL and Vulkan is specifically called "view index". For example this is `gl_ViewIndex` in GLSL. It works only with the "multiview" feature.

kvark (approved):

kvark:
Published in 0.7.3

PR #1543, opened 2021-11-18
[wgsl-in] Don't be confused by pointers to globals.

jimblandy:
The WGSL front end knows that globals in the `Handle` storage class do not produce references, but it attempts to manage this in two places which trip over each other. As a consequence, referring to a `let` binding holding a pointer to a global flips the variable's type from WGSL `ptr` to WGSL reference, with bewildering consequences.


kvark (approved):

kvark:
please rebase and land

Issue #1481, opened 2021-10-26
[wgsl-in] Keywords should not be allowed for identifiers


Herschel:
This code passes validation in Naga:
```
[[group(0), binding(0)]]
var sampler: sampler;
```

The [WGSL spec](https://www.w3.org/TR/WGSL/#identifiers) says:
 > The form of an identifier is defined via pattern-matching, except that an identifier must not have the same spelling as a keyword or as a reserved word.

Tint rejects with this error:
```
d:\projects\naga\color.wgsl:2:5 error: expected identifier for variable declaration
var sampler: sampler;
    ^^^^^^^
```

naga @ https://github.com/gfx-rs/naga/commit/5cf11ab73429f4b64cda6617459e7eca379f4135

jimblandy:
I think this was fixed in [2107b20](https://github.com/gfx-rs/naga/commit/2107b205619dfa329214bfcbe1b6b5dd38445966). If that's not the case, please re-open.

PR #1549, opened 2021-11-24
[wgsl-in] Don't allow keywords to be used as identifiers

### LINKED ISSUE #1546: [wgsl-in] naga should not accept keywords as identifiers
The code `let sampler = 0;` is accepted by naga, but according to the spec, an identifier [must not have the same spelling as a keyword or as a reserved word](https://gpuweb.github.io/gpuweb/wgsl/#identifiers), and `sampler` [is a keyword](https://gpuweb.github.io/gpuweb/wgsl/#keyword-summary).

Gordon-F:
Fix #1546, #1481.


src/front/wgsl/keywords.rs
```diff
@@ -0,0 +1,133 @@
+pub const RESERVED: &[&str] = &[
```
Gordon-F, line 1: I think back and front should use the same reserved list. Currently just copy-paste.

kvark: good point! We need a place to share them

tests/wgsl-errors.rs
```diff
@@ -13,6 +13,7 @@ fn check(input: &str, snapshot: &str) {
                 diff::Result::Right(r) => println!("+{}", r),
             }
         }
+        println!("Raw output:\n{}", output);
```
kvark, line 16: debug stuff?

Gordon-F: Yep, easier copy-paste for tests actual\expected stuff.

tests/wgsl-errors.rs
```diff
@@ -531,6 +531,114 @@ fn postfix_pointers() {
...
+
+"###,
+    );
+
+    // struct
+    check(
+        r#"
+            struct array {};
+        "#,
+        r###"error: Name ` array` is a reserved keyword
```
Gordon-F, line 618: Avoiding extra spaces would be nice

src/lib.rs
```diff
@@ -209,6 +209,7 @@ mod arena;
 pub mod back;
 mod block;
 pub mod front;
+pub(crate) mod keywords;
```
kvark, line 212: `pub(crate)` doesn't make sense in this place. Just remove it, leave only `mod keywords`.

kvark (approved):

PR #1550, opened 2021-11-24
[wgsl] Remove IsInf and IsNan

### LINKED ISSUE #1542: Remove IsInf and IsNan
See https://github.com/gpuweb/gpuweb/pull/2311

Gordon-F:
Fix #1542


kvark (approved):
We may need to do something more elaborate in frontends (like replace "is_inf" by "false" unconditionally), but we can do it later

Issue #1480, opened 2021-10-26
[wgsl-in] Redefinition incorrectly passes validation


Herschel:
Naga allows reusing a name in the same scope:

```
fn foo() {}
fn foo() {}

[[block]]
struct Globals {
    bar: vec4<f32>;
};

[[group(0), binding(0)]]
var<uniform> a: Globals;
[[group(0), binding(1)]]
var<uniform> a: Globals;
```
This passes validation in Naga, but this is not allowed by the [WGSL spec](https://www.w3.org/TR/WGSL/#declaration-and-scope):
 > A declaration must not introduce a name when that identifier is already in scope with the same end scope as another instance of that name.

Tint rejects the code with this message:
```
d:\projects\naga\color.wgsl:2:1 error: redefinition of 'foo'
fn foo() {}
^^

d:\projects\naga\color.wgsl:1:1 note: previous definition is here
fn foo() {}
^^
```

I hit this because [learn wgpu book](https://github.com/sotrh/learn-wgpu/blob/master/code/beginner/tutorial3-pipeline/src/shader.wgsl) uses `main` for both vertex and fragment entry points. Tint disagrees with this, so I was caught by surprise when my shader worked in wgpu desktop but not Chrome.

naga @ https://github.com/gfx-rs/naga/commit/5cf11ab73429f4b64cda6617459e7eca379f4135

jimblandy:
I think this was fixed in [dba6beb](https://github.com/gfx-rs/naga/commit/dba6beb941aaf89436b57f14d6fc405da612e6e5). If that's not the case, please re-open.

PR #1551, opened 2021-11-24
[wgsl-in] Don't allow redefinition of module scope identifiers

Gordon-F:
I guess fix for #1480.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -171,6 +171,10 @@ pub enum Error<'a> {
     NotPointer(Span),
     NotReference(&'static str, Span),
     ReservedKeyword(Span),
+    Redefinition {
+        previous: Span,
+        current: Span,
+    },
     Other,
 }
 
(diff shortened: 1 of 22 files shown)


kvark (commented):
We need to forbid any kind of name collision at the module scope. Functions are one of many. Does it make sense to have this logic finding collisions for functions only, still? Perhaps, we need just a set of strings that are forbidden for names, which we grow as we scan the module.

src/front/wgsl/mod.rs
```diff
@@ -1204,6 +1215,8 @@ impl std::error::Error for ParseError {
 
 pub struct Parser {
     scopes: Vec<(Scope, usize)>,
+    module_scope_variable_identifiers: FastHashMap<String, Span>,
```
kvark, line 1218: do these need to be separate from function identifiers?

Gordon-F: Maybe I misread the spec, but I can't find confirmation that module scope functions and variables cannot have the same name. If they can't have the same identifier, we can merge 2 hashmaps.

kvark: Yes, pretty sure they are in the same namespace. Let's merge the maps.

kvark (approved):
Thank you!

Gordon-F:
Something wrong with CI.

Gordon-F:
@kvark ready for merge.

kvark (approved):

PR #1553, opened 2021-11-26
Omit WGSL output types where they are inferred

kvark:
Also refactors (and greatly simplifies) our snapshot options.


kvark:
Those pesky `msl_custom` and `glsl_custom` parameters that very few understood.... BEGONE

PR #1571, opened 2021-12-04
glsl: reflect textures without samplers

kvark:
GLSL backend wasn't properly reflecting textures used without samplers, it's required for https://github.com/kvark/vange-rs/issues/161 and others.


kvark:
Published in 0.7.3

PR #1586, opened 2021-12-11
readme: list supported GLSL versions for glsl-in

markus-wa:


kvark:
@JCapucho is that the version we are expecting today?

JCapucho:
It's the versions we accept today, though 440 doesn't get as much attention as the other.

PR #1594, opened 2021-12-14
[msl-out]: Properly write default values for non-scalar types.

### LINKED ISSUE #1585: Metal incorrect bounds checking on the shadow shader
Blocking `wgpu` update right now:
```
Caused by:
      note: label = `main`
    Internal error in VERTEX shader: Metal: program_source:138:83: error: incompatible operand types ('const constant Light' and 'int')
        Light light = metal::uint(_e21) < 1 + (_buffer_sizes.size2 - 0 - 96) / 96 ? s_lights.data[_e21] : 0;
                                                                                  ^ ~~~~~~~~~~~~~~~~~~~   ~
program_source:185:48: error: incompatible operand types ('const constant Light' and 'int')
        Light light_1 = metal::uint(_e21) < 10 ? u_lights.data.inner[_e21] : 0;
                                               ^ ~~~~~~~~~~~~~~~~~~~~~~~~~   ~


', wgpu/src/backend/direct.rs:2233:5
thread 'shadow' panicked at 'UNEXPECTED TEST FAILURE', wgpu/examples/shadow/../../tests/common/mod.rs:269:9
```

Shader - https://github.com/gfx-rs/wgpu/blob/3b48ed28f37d41aaac53ce76501cd801bc41bdc1/wgpu/examples/shadow/shader.wgsl

jimblandy:
Fixes #1585.

Commits should be individually CI-clean.

### DIFF
--- a/src/proc/index.rs
+++ b/src/proc/index.rs
@@ -148,6 +148,11 @@ impl BoundsCheckPolicies {
             _ => self.index,
         }
     }
+
+    /// Return `true` if any of `self`'s policies are `policy`.
+    pub fn contains(&self, policy: BoundsCheckPolicy) -> bool {
+        self.index == policy || self.buffer == policy || self.image == policy
+    }
 }
 
 /// An index that may be statically known, or may need to be computed at runtime.
@@ -204,9 +209,7 @@ pub fn find_checked_indexes(
     let mut guarded_indices = BitSet::new();
 
     // Don't bother scanning if we never need `ReadZeroSkipWrite`.
-    if policies.index == BoundsCheckPolicy::ReadZeroSkipWrite
-        || policies.buffer == BoundsCheckPolicy::ReadZeroSkipWrite
-    {
+    if policies.contains(BoundsCheckPolicy::ReadZeroSkipWrite) {
         for (_handle, expr) in function.expressions.iter() {
             // There's no need to handle `AccessIndex` expressions, as their
             // indices never need to be cached.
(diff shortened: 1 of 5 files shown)


kvark:
As discussed on the matrix, let's investigate if it's possible to simplify the zero/default initialization without as much code complication. For example, https://godbolt.org/z/s1h3T6a4E

jimblandy:
Okay, this implements the suggestions from chat.

src/proc/index.rs
```diff
@@ -204,9 +209,7 @@ pub fn find_checked_indexes(
     let mut guarded_indices = BitSet::new();
 
     // Don't bother scanning if we never need `ReadZeroSkipWrite`.
-    if policies.index == BoundsCheckPolicy::ReadZeroSkipWrite
-        || policies.buffer == BoundsCheckPolicy::ReadZeroSkipWrite
-    {
+    if policies.contains(BoundsCheckPolicy::ReadZeroSkipWrite) {
```
kvark, line 212: this appears to change the semantics a bit, since it includes a case where image policy is `ReadZeroSkipWrite` but index/buffer policies are something different
but I think it's ok, given how niche this scenario is

kvark (approved):
Sweet!

src/proc/index.rs
jimblandy: Right, this is not quite needed. But it will be once I have texture bounds checks done.

PR #1597, opened 2021-12-15
Turn on panic = "abort" by default

jrmuizel:
This reduces the size of a release naga binary on macOS
from 3.5MB to 3.2MB. It's also the configuration we use
for naga in Firefox.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,12 @@ petgraph = { version ="0.6", optional = true }
 pp-rs = { version = "0.2.1", optional = true }
 hexf-parse = { version = "0.2.1", optional = true }
 
+[profile.release]
+panic = "abort"
+
+[profile.dev]
+panic = "abort"
+
 [features]
 default = []
 dot-out = []


Cargo.toml
```diff
@@ -30,6 +30,12 @@ petgraph = { version ="0.6", optional = true }
 pp-rs = { version = "0.2.1", optional = true }
 hexf-parse = { version = "0.2.1", optional = true }
 
+[profile.release]
+panic = "abort"
+
+[profile.dev]
```
JCapucho, line 36: I don't think this is a decision for the library to make, applying it to the `cli` makes more sense, but even then I find the backtraces helpful when debugging panics.

jrmuizel: Having this in the Cargo.toml doesn't affect users of this as a library. i.e. only the top level `[profile]` section applies.

I'm also pretty sure this doesn't impact your ability to get backtraces when debugging panics.

cwfitzgerald: Yeah, the backtrace/printing is done by the panic hook, which always fires.

JCapucho: Oh I didn't know only the top level profile counted.

Don't we lose stack unwinding if we change the abort strategy to abort?

JCapucho: Oh wait seems like `abort` does have backtraces what I was thinking about is an unstable feature called `panic_immediate_abort`

kvark: > Having this in the Cargo.toml doesn't affect users of this as a library

Who does it affect then? Are we improving the developer velocity here?

kvark: I guess we'll just have slightly better compile times and exec size for naga-cli

kvark (approved):

PR #1599, opened 2021-12-15
glsl-in: Flatten entry point arguments

### LINKED ISSUE #1278: [glsl-in] output interface block fails validation
The following vertex shader:
```glsl
#version 450

layout(location = 0) out FragmentData {
    vec2 position;
} frag;

void main() {
    frag.position = position;
}
```
causes the following validaiton error:
```
    0: failed to validate shader
    1: Type [4] '' is invalid
    2: The composite type contains a top-level structure
```

`Type [4]` in the above references the "result" type of the `main` entry point: it is a struct that consists of two members: 1) gl_Position and 2) the `FragmentData` struct. The `FragmentData` struct seems to be (incorrectly?) marked as being a top-level struct.

JCapucho:
The IR doesn't allow having structs has bindings for entry point
input/output, so the glsl frontend must flatten them.

Glsl defines that the locations are sequential from the block base location.

Closes #1278

### DIFF
--- a/src/front/glsl/functions.rs
+++ b/src/front/glsl/functions.rs
@@ -1058,6 +1058,82 @@ impl Parser {
         });
     }
 
+    /// Helper function for building the input/output interface of the entry point
+    ///
+    /// Calls `f` with the data of the entry point argument, flattening composite types
+    /// recursively
+    ///
+    /// The passed arguments to the callback are:
+    /// - The name
+    /// - The pointer expression to the global storage
+    /// - The handle to the type of the entry point argument
+    /// - The binding of the entry point argument
+    /// - The expression arena
+    fn arg_type_walker(
+        &self,
+        name: Option<String>,
+        binding: crate::Binding,
+        pointer: Handle<Expression>,
+        ty: Handle<Type>,
+        expressions: &mut Arena<Expression>,
+        f: &mut impl FnMut(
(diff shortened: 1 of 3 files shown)


src/front/glsl/functions.rs
```diff
@@ -1058,6 +1058,71 @@ impl Parser {
...
     }
 
+    fn arg_type_walker(
+        &self,
+        name: Option<String>,
+        binding: crate::Binding,
+        pointer: Handle<Expression>,
+        ty: Handle<Type>,
+        expressions: &mut Arena<Expression>,
+        f: &mut impl FnMut(
```
kvark, line 1068: could this be `mut f: impl FnMut`?

src/front/glsl/functions.rs
```diff
@@ -1058,6 +1058,71 @@ impl Parser {
         });
     }
 
+    fn arg_type_walker(
```
kvark, line 1061: could we drop a comment here explaining what it does? looks to be not simply walking but also populating expressions

src/front/glsl/functions.rs
JCapucho: That was my original idea but as the function is recursive it needs to pass the function to itself and passing it as a reference when recursing causes infinite type recursion.

PR #1600, opened 2021-12-15
spv-in: Allow branch weights in OpBranchConditional

### LINKED ISSUE #1589: [spirv-in] naga rejects shader that uses branch weights
Naga (c4b331acb44e8ffb6595030a16725ddb137c2dcb) rejects this:

```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 10
; Bound: 10
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main"
               OpExecutionMode %main OriginUpperLeft
               OpSource GLSL 450
               OpName %main "main"
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
       %bool = OpTypeBool
       %true = OpConstantTrue %bool
       %main = OpFunction %void None %3
          %5 = OpLabel
               OpSelectionMerge %9 None
               OpBranchConditional %true %8 %9 1564123493 301814045
          %8 = OpLabel
               OpBranch %9
          %9 = OpLabel
               OpReturn
               OpFunctionEnd
```

with:

```
unknown instruction
```

If I remove the branch weights from the OpBranchConditional instruction, the shader is accepted. If I change the branch weights to `1 2` I get "invalid word count" instead.

JCapucho:
Closes #1589

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -2814,6 +2814,11 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                         reject,
                     });
 
+                    // Consume branch weights
+                    for _ in 4..inst.wc {
+                        let _ = self.next()?;
+                    }
+
                     return Ok(());
                 }
                 Op::Switch => {


src/front/spv/mod.rs
```diff
@@ -2761,7 +2761,10 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                     break None;
                 }
                 Op::BranchConditional => {
-                    inst.expect_at_least(4)?;
+                    let weighted = inst.wc == 6;
```
kvark, line 2764: perhaps we could refactor it slightly differently?
I.e. it already used to check `expect_at_least(4)`. What it was missing though is eating those additional words.
So we could have just added something like this to the end:
```rust
for _ in 4 .. inst.wc {
  let _ = self.next()?;
}
```

JCapucho: The way I did it in this PR adds a little bit of validation, the spirv spec only allows 6 words (with weights) or 4 (without weights).

Your solution would allow for 5 words and 7+ words inputs. If you think this kinda of validation hurts more than it helps then I'm okay with your proposal.

kvark: We aren't in the business of validating input SPIR-V. That would blow up the complexity of spv-in rather significantly.
We talked previously about whether or not GLSL-in needs to be validated, and I argued against it. But a strong point for GLSL is that it's human-writable, therefore even limited validation is useful.
SPIR-V though is not human writable, so we expect it to be generated correctly in the first place.

kvark (approved):

PR #1601, opened 2021-12-15
spv-in: Remove if when both branches are equal

### LINKED ISSUE #1588: [spirv-in] Assertion failure when both targets of branch are the same
The following SPIR-V is valid according to the current SPIR-V spec:

```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 10
; Bound: 605
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %4 "main"
               OpExecutionMode %4 OriginUpperLeft
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
       %uint = OpTypeInt 32 0
       %bool = OpTypeBool
       %true = OpConstantTrue %bool
        %599 = OpConstantNull %uint
          %4 = OpFunction %void None %3
          %5 = OpLabel
               OpSelectionMerge %244 DontFlatten
               OpSwitch %599 %245
        %245 = OpLabel
               OpBranchConditional %true %92 %92
         %92 = OpLabel
               OpBranch %244
        %244 = OpLabel
               OpReturn
               OpFunctionEnd
```

Naga (c4b331acb44e8ffb6595030a16725ddb137c2dcb) fails this assertion:

```
thread 'main' panicked at 'assertion failed: prev.is_none()', src/front/spv/mod.rs:2808:25
```

The problematic code appears to be the `OpBranchConditional` with identical targets. That's legal in current SPIR-V.

(Do let me know if it's more useful for me to upload SPIR-V binary.)

JCapucho:
This needs to be handled this  way because when both branches are equal the spirv doesn't need a `OpSelectionMerge` so this could cause a huge amount of duplication in the two branches.

Closes #1588

### DIFF
--- a/src/front/spv/mod.rs
+++ b/src/front/spv/mod.rs
@@ -2769,8 +2769,15 @@ impl<I: Iterator<Item = u32>> Parser<I> {
                         get_expr_handle!(condition_id, lexp)
                     };
 
+                    block.extend(emitter.finish(ctx.expressions));
+                    ctx.blocks.insert(block_id, block);
+                    let body = &mut ctx.bodies[body_idx];
+                    body.data.push(BodyFragment::BlockId(block_id));
+
                     let true_id = self.next()?;
                     let false_id = self.next()?;
+
+                    let same_target = true_id == false_id;
 
                     // Start a body block for the `accept` branch.
                     let accept = ctx.bodies.len();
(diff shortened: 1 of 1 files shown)


kvark:
why on earth would anything generate a SPIR-V module with true==false branches?

kvark (approved):

Issue #1605, opened 2021-12-16
[spv-out] Double Declaration of OpTypeBool and Push Constant Bug


cwfitzgerald:
Given the following wgsl:

<details>
<summary>Wgsl</summary>

```
struct Plane {
    inner: vec4<f32>;
};

struct Frustum {
    left: Plane;
    right: Plane;
    top: Plane;
    bottom: Plane;
    near: Plane;
};

struct CullingUniforms {
    view: mat4x4<f32>;
    view_proj: mat4x4<f32>;
    frustum: Frustum;
    object_count: u32;
};

struct ObjectInputUniforms {
    uniforms: CullingUniforms;
};

struct ObjectOutputData {
    model_view: mat4x4<f32>;
    model_view_proj: mat4x4<f32>;
    material_idx: u32;
    inv_squared_scale: vec3<f32>;
};

struct ObjectOutputDataBuffer {
    object_output: [[stride(160)]] array<ObjectOutputData>;
};

struct IndirectCall {
    vertex_count: u32;
    instance_count: u32;
    base_index: u32;
    vertex_offset: i32;
    base_instance: u32;
};

struct IndirectBuffer {
    draw_call_count: u32;
    pad0_: u32;
    pad1_: u32;
    pad2_: u32;
    indirect_call: [[stride(20)]] array<IndirectCall>;
};

struct IntermediateBufferA {
    result_index_a: [[stride(4)]] array<u32>;
};

struct ObjectInputData {
    start_idx: u32;
    count: u32;
    vertex_offset: i32;
    material_idx: u32;
    transform: mat4x4<f32>;
    bounding_sphere: vec4<f32>;
};

struct ObjectInputDataBuffer {
    object_input: [[stride(96)]] array<ObjectInputData>;
};

[[group(0), binding(1)]]
var<uniform> unnamed: ObjectInputUniforms;
[[group(0), binding(4)]]
var<storage, read_write> unnamed_1: ObjectOutputDataBuffer;
[[group(0), binding(5)]]
var<storage, read_write> unnamed_2: IndirectBuffer;
var<private> gl_GlobalInvocationID_1: vec3<u32>;
[[group(0), binding(2)]]
var<storage, read_write> unnamed_3: IntermediateBufferA;
[[group(0), binding(0)]]
var<storage> unnamed_4: ObjectInputDataBuffer;

fn main_1() {
    switch(bitcast<i32>(0u)) {
        default: {
[...]

PR #1607, opened 2021-12-16
Enforce casting width better, fix IEqual/INotEqual in spv

kvark:
Fixes a regression from #1603
Addresses half of #1605
cc @JCapucho


JCapucho:
Nice catch, quite a dumb error on my part

kvark:
not dumb at all! binary operators are just not all equal, I constantly fail on them as well :)

PR #1580, opened 2021-12-07
Fixes for `fma` function

parasyte:
- Using the `mad` intrinsic should be enough for HLSL because we only support f32 for now.
- Adds support for `fma` function on GLSL, including trivial arithmetic transformation when necessary.
- Adds a new test for WGSL functions, in the spirit of operators.wgsl. And one for WebGL output.
- Closes #1579

PR #1610, opened 2021-12-17
Release v0.8

kvark:
Blocked on #1596 and #1611
Optionally, also on #1580 and #1473

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,42 @@
 # Change Log
 
-## v0.8 (TBD)
+## v0.8 (2021-12-18)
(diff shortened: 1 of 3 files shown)


Gordon-F:
Since the MSRV for wgpu 0.12 is 1.53, can we bump MSRV for naga too?

kvark:
> Since the MSRV for wgpu 0.12 is 1.53, can we bump MSRV for naga too?

That's certainly a possibility! But we need to take into account that Naga is used more widely than just in wgpu. So, generally speaking, Naga MSRV can be lower but not higher.

PR #1611, opened 2021-12-17
wgsl-in: correctly parse `else if`

### LINKED ISSUE #1545: The else if statement is not supported
Naga does not allow an if_statement inside an else_statement, but the [spec](https://www.w3.org/TR/WGSL/#if-statement) allows for it.

```
if (true) {} else if (false) {}
```

```
Parse Error: expected '{', found 'if'
```

nolanderc:
Fixes #1545. Removes support for `elseif` keyword.

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -3474,35 +3474,34 @@ impl Parser {
                         let mut elsif_stack = Vec::new();
                         let mut elseif_span_start = lexer.current_byte_offset();
                         let mut reject = loop {
-                            if lexer.skip(Token::Word("else")) {
-                                if lexer.skip(Token::Word("if")) {
-                                    // ... else if (...) { ... }
-                                    let mut sub_emitter = super::Emitter::default();
-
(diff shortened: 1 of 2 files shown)


src/front/wgsl/mod.rs
```diff
@@ -3470,33 +3470,41 @@ impl Parser {
                         lexer.expect(Token::Paren(')'))?;
 
                         let accept = self.parse_block(lexer, context.reborrow(), false)?;
+
                         let mut elsif_stack = Vec::new();
                         let mut elseif_span_start = lexer.current_byte_offset();
-                        while lexer.skip(Token::Word("elseif")) {
-                            let mut sub_emitter = super::Emitter::default();
+                        let mut reject = loop {
+                            if lexer.skip(Token::Word("else")) {
```
kvark, line 3477: can we rewrite this with early exists instead?
```rust
if !lexer.skip(...) {
  break ...;
}
```

nolanderc: Good point! Pushing the changes...

kvark (approved):
nice!

PR #1620, opened 2021-12-21
impl Clone for `WithSpan<E>`

jakobhellermann:
I wanted to create a `CreateShaderModuleError` from a `&WithSpan<ValidationError>` to reuse `wgpu`'s shader error formatting, but that isn't possible right now because `WithSpan` can't be cloned and you can't get an `&E` from `&WithSpan<E>`.

Solution: `#[derive(Clone)]`


kvark (approved):

kvark:
published in 0.8.1

PR #1629, opened 2021-12-27
hlsl: support arrays of matrices

### LINKED ISSUE #1628: error X3000: unrecognized identifier 'type_6' with DX12 Backend
**Description**
I have a shader that I will post below that works flawlessly on the VULKAN backend but doesn't work on the DX12 backend. The strang thing is that the exact same shader code has worked in previous wgpu versions (0.9) (on the DX12 backend)

**Repro steps**
Vertex Shader:
```glsl
#version 450

layout(location=0) in vec3 a_position;

layout(set=1, binding=0) // 1.
uniform Uniforms {
    mat4 projection; // 2.
};
layout(set=1, binding=1) // 1.
uniform Uniforms2 {
    mat4 view; // 2.
};
layout(set=2, binding=0) // 1.
uniform Uniforms3 {
    mat4[2] transform; // 2.
};

void main() {
    gl_Position = projection * view * transform[gl_VertexIndex] * vec4(a_position, 1.0);
}
```

This is a line shader and the mat4[2] is used to move the two points the line starts and ends.

It should draw a line like it does on VULKAN vs it crashes with this error message: error X3000: unrecognized identifier 'type_6'

**Extra materials**
Longer error:
https://pastebin.com/Grmeb9Td

**Platform**
OS: Windows 10
wgpu-version: 0.12
AdapterInfo { name: "NVIDIA GeForce GTX 1080 Ti", vendor: 4318, device: 6918, device_type: DiscreteGpu, backend: Dx12 }

kvark:
Fixes #1628

### DIFF
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -715,22 +715,11 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     stride: _,
                 } => {
                     // HLSL arrays are written as `type name[size]`
-                    let (ty_name, vec_size) = match module.types[base].inner {
-                        // Write scalar type by backend so as not to depend on the front-end implementation
-                        // Name returned from frontend can be generated (type1, float1, etc.)
-                        TypeInner::Scalar { kind, width } => (kind.to_hlsl_str(width)?, None),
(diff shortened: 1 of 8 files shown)


kvark:
published in 0.8.1

PR #1630, opened 2021-12-27
msl: fix packed float support

### LINKED ISSUE #1625: Error: unknown type name 'packed_float3'
I have a very similar issue to #999 with naga 0.8.0 on macOS 10.13 High Sierra (if that matters). The generated code includes:
```
// language: metal2.0
#include <metal_stdlib>
#include <simd/simd.h>

[...]

struct View {
    metal::float4x4 view_proj;
    metal::float4x4 inverse_view;
    metal::float4x4 projection;
    packed_float3 world_position;
    float near;
    float far;
    float width;
    float height;
};
```

Compilation fails on the `packed_float3`, same as @Lichtso from #999.

```
program_source:18:5: error: unknown type name 'packed_float3'
    packed_float3 world_position;
```

Is that really supposed to be fixed as stated in #999? If so, was it rather in a version later than 0.8.0, which is the one Bevy is using? Or is that a separate issue maybe?

Cargo.lock:
```
[[package]]
name = "naga"
version = "0.8.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "bc7216bec6be822a562dda435236539c7fb5aa5e5ceecc8c565e81139ed1d4b0"
```

kvark:
Fixes #1625

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -2357,8 +2357,9 @@ impl<W: Write> Writer<W> {
                             Some(kind) => {
                                 writeln!(
                                     self.out,
-                                    "{}packed_{}3 {};",
+                                    "{}{}::packed_{}3 {};",
                                     back::INDENT,
+                                    NAMESPACE,
                                     scalar_kind_string(kind),
                                     member_name
                                 )?;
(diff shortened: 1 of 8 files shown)


kvark:
published in 0.8.1

PR #1631, opened 2021-12-27
glsl-in: fix composite constructors

### LINKED ISSUE #1623: [GLSL] "Constructor: Too many arguments" when creating constant array
I have the following vertex shader.

```glsl
#version 450

layout(location=0) out vec3 clip_coords;

const vec3 positions[6] = vec3[6](
    // Bottom left triangle
    vec3(-1.0, 1.0, 0.0),
    vec3(-1.0, -1.0, 0.0),
    vec3(1.0, -1.0, 0.0),

    // Top right triangle
    vec3(-1.0, 1.0, 0.0),
    vec3(1.0, -1.0, 0.0),
    vec3(1.0, 1.0, 0.0)
);

void main() {
    gl_Position = vec4(positions[gl_VertexIndex], 1.0);
    clip_coords = gl_Position.xyz;
}
```

This compiles fine on 0.7, but on 0.8 it fails with the following error.

      Error {
          kind: SemanticError(
              "Constructor: Too many arguments",
          ),
          meta: Span {
              start: 82,
              end: 302,
          },
      },

I've tracked the error down to the `positions` array that I use in lieu of a vertex buffer. Are constant arrays no longer a thing? If so, their should probably be a more specific error for this.

JCapucho:
In the recent rework of the constructors it seems that the logic for
composite types (arrays and structs) was accidentally removed by me.

This PR reimplements it, adds implicit conversions that were missing on them and adds tests.

Closes #1623

### DIFF
--- a/src/front/glsl/context.rs
+++ b/src/front/glsl/context.rs
@@ -881,7 +881,7 @@ impl Context {
 
     pub fn expr_scalar_components(
         &mut self,
-        parser: &mut Parser,
+        parser: &Parser,
         expr: Handle<Expression>,
         meta: Span,
     ) -> Result<Option<(ScalarKind, crate::Bytes)>> {
@@ -891,7 +891,7 @@ impl Context {
 
     pub fn expr_power(
         &mut self,
-        parser: &mut Parser,
+        parser: &Parser,
         expr: Handle<Expression>,
         meta: Span,
     ) -> Result<Option<u32>> {
@@ -921,7 +921,7 @@ impl Context {
 
     pub fn implicit_conversion(
         &mut self,
-        parser: &mut Parser,
+        parser: &Parser,
         expr: &mut Handle<Expression>,
         meta: Span,
         kind: ScalarKind,
@@ -941,7 +941,7 @@ impl Context {
 
     pub fn binary_implicit_conversion(
         &mut self,
-        parser: &mut Parser,
+        parser: &Parser,
         left: &mut Handle<Expression>,
         left_meta: Span,
         right: &mut Handle<Expression>,
@@ -974,7 +974,7 @@ impl Context {
 
     pub fn implicit_splat(
         &mut self,
-        parser: &mut Parser,
+        parser: &Parser,
         expr: &mut Handle<Expression>,
         meta: Span,
         vector_size: Option<VectorSize>,
--- a/src/front/glsl/functions.rs
+++ b/src/front/glsl/functions.rs
@@ -195,14 +195,39 @@ impl Parser {
                 (value, expr_meta),
                 meta,
             )?,
-            TypeInner::Struct { .. } | TypeInner::Array { .. } => ctx.add_expression(
-                Expression::Compose {
-                    ty,
-                    components: vec![value],
-                },
-                meta,
-                body,
-            ),
+            TypeInner::Struct { ref members, .. } => {
+                let scalar_components = members
+                    .get(0)
+                    .and_then(|member| scalar_components(&self.module.types[member.ty].inner));
+                if let Some((kind, width)) = scalar_components {
+                    ctx.implicit_conversion(self, &mut value, expr_meta, kind, width)?;
+                }
+
+                ctx.add_expression(
+                    Expression::Compose {
+                        ty,
+                        components: vec![value],
+                    },
+                    meta,
+                    body,
+                )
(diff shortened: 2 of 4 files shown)


src/front/glsl/context.rs
```diff
@@ -891,7 +891,7 @@ impl Context {
 
     pub fn expr_power(
         &mut self,
-        parser: &mut Parser,
+        parser: &Parser,
```
kvark, line 894: nice!

src/front/glsl/functions.rs
```diff
@@ -195,14 +195,39 @@ impl Parser {
...
-                },
-                meta,
-                body,
-            ),
+            TypeInner::Struct { ref members, .. } => {
+                let scalar_components = members
+                    .get(0)
+                    .and_then(|member| scalar_components(&self.module.types[member.ty].inner));
+                if let Some((kind, width)) = scalar_components {
+                    ctx.implicit_conversion(self, &mut value, expr_meta, kind, width)?;
```
kvark, line 203: aren't you changing the second parameter to be `&value` now?

src/front/glsl/functions.rs
```diff
@@ -195,14 +195,39 @@ impl Parser {
...
+                    },
+                    meta,
+                    body,
+                )
+            }
+
+            TypeInner::Array { base, .. } => {
+                let scalar_components = scalar_components(&self.module.types[base].inner);
+                if let Some((kind, width)) = scalar_components {
+                    ctx.implicit_conversion(self, &mut value, expr_meta, kind, width)?;
```
kvark, line 219: same here

src/front/glsl/functions.rs
```diff
@@ -535,6 +560,26 @@ impl Parser {
             TypeInner::Vector { size, kind, width } => {
                 return self.vector_constructor(ctx, body, ty, size, kind, width, &args, meta)
             }
+            TypeInner::Array { base, .. } => {
+                for (mut arg, meta) in args.iter().copied() {
+                    let scalar_components = scalar_components(&self.module.types[base].inner);
+                    if let Some((kind, width)) = scalar_components {
+                        ctx.implicit_conversion(self, &mut arg, meta, kind, width)?;
```
kvark, line 567: and here

src/front/glsl/functions.rs
```diff
@@ -535,6 +560,26 @@ impl Parser {
             TypeInner::Vector { size, kind, width } => {
                 return self.vector_constructor(ctx, body, ty, size, kind, width, &args, meta)
             }
+            TypeInner::Array { base, .. } => {
+                for (mut arg, meta) in args.iter().copied() {
+                    let scalar_components = scalar_components(&self.module.types[base].inner);
```
kvark, line 565: would it be easier to do this via
```rust
impl crate::TypeInner {
  fn scalar_components(&self) -> ... {... }
}
```

src/front/glsl/functions.rs
```diff
@@ -535,6 +560,26 @@ impl Parser {
...
+                    }
+
+                    components.push(arg)
+                }
+            }
+            TypeInner::Struct { ref members, .. } => {
+                for ((mut arg, meta), member) in args.iter().copied().zip(members.iter()) {
+                    let scalar_components = scalar_components(&self.module.types[member.ty].inner);
+                    if let Some((kind, width)) = scalar_components {
+                        ctx.implicit_conversion(self, &mut arg, meta, kind, width)?;
```
kvark, line 577: same here

src/front/glsl/functions.rs
JCapucho: No, the update signature is
```diff
pub fn implicit_conversion(
        &mut self,
-       parser: &mut Parser,
+       parser: &Parser,
        expr: &mut Handle<Expression>,
        meta: Span,
        kind: ScalarKind,
        width: crate::Bytes,
    )
```

So only the mutability of the first argument is changed (`self` in this case), while the second argument (`&mut value`) is still a mutable reference.

Otherwise the code wouldn't compile because `implicit_conversion` mutates `expr`

src/front/glsl/functions.rs
JCapucho: It would be almost the same thing, I have no preference so if you prefer the method call instead of a function call that's fine by me.

PR #1639, opened 2022-01-03
Validate structure resource types

kvark:
Fixes https://github.com/gfx-rs/wgpu/issues/2351


kvark:
Published in naga-0.8.2

PR #1640, opened 2022-01-03
Add a way to emit errors with a path

laptou:
I am using `naga` in my build script to validate all of my shaders, and it would be nice to be able to display the file path inside of the diagnostic, so I added a method `emit_to_stderr_with_path` to `naga::front::wgsl::ParserError`.


kvark (approved):

kvark:
CI mad :)

kvark (approved):

PR #1653, opened 2022-01-07
spv-out: use OpFRem instead of OpFMod

### LINKED ISSUE #1652: WGSL float % translates into incorrect spirv op code
**Description**
WGSL spec says:
> Floating point remainder, where sign of non-zero result matches sign of e1. Component-wise when T is a vector.
> Result equal to: e1 - e2 * trunc(e1 / e2)
> (OpFRem)

While looking at the shader code in RenderDoc when using Vulkan backend I can see that the modulo operation was translated to:
> float _95 = FMod(_92, _94);

**Repro steps**
Use float % operator in WGSL shader where the first argument is a negative value. 

The result of the % operation should be a negative value, but it is a positive value, because OpFMod is used instead of OpFRem.

**Platform**
Using wgpu 0.12.0 on Linux (Arch linux if it matters).

JCapucho:
Closes #1652

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ bitflags = "1"
 bit-set = "0.5"
 codespan-reporting = { version = "0.11.0", optional = true }
 rustc-hash = "1.1.0"
-indexmap = "1.6" # 1.7 has MSRV 1.49
+indexmap = "~1.6" # 1.7 has MSRV 1.49
 log = "0.4"
 num-traits = "0.2"
 spirv = { version = "0.2", optional = true }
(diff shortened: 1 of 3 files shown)


kvark (approved):

kvark:
Looks like indexmap broke MSRV

kvark:
Would you mind specifying indexmap dependency as "1.7"?

JCapucho:
Sorry only noticed now.

But I'm confused we have in the `Cargo.toml`
```
indexmap = "1.6" # 1.7 has MSRV 1.49
```

why is it pulling `1.8` and how would changing it to `1.7` help?

kvark:
Damn, hmm, interesting. Technically 1.8 is compatible with 1.6 that we are requesting. We need to request it differently:
```toml
indexmap = "~1.6"
```

JCapucho:
It's fixed now

kvark (approved):

kvark:
Published in naga-0.8.2

PR #1658, opened 2022-01-10
[wgsl-in] Write a real type in type mismatch error

### LINKED ISSUE #1655: Confusing error messages on type mismatch
Consider the following wgsl code:
```
fn main() {
    var x: u32 = 0;
}
```
Attempting to evaluate it leads to a naga output of:
```
ERROR naga::front::wgsl] Given type Scalar { kind: Uint, width: 4 } doesn't match expected Scalar { kind: Sint, width: 4 }
error: the type of `x` is expected to be [1]
  ┌─ wgsl:3:9
  │
3 │     var x: u32 = 0;
  │         ^ definition of `x`

Could not parse WGSL
```

While the initial error message can clue an experienced user into what the problem is, the rest of the error message is, in my opinion, rather mystifying. What exactly does "[1]" refer to here, and how is it useful to someone reading the error? For this example, it may be worth adding a note explaining to the user that perhaps they may want to declare the right-hand side as unsigned.

Gordon-F:
Fix #1655

```shell
error: the type of `x` is expected to be `i32`
  ┌─ wgsl:3:9
  │
3 │     var x: u32 = 0;
  │         ^ definition of `x`
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ Cargo.lock
 /*.frag
 /*.comp
 /*.wgsl
+/*.hlsl
+/*.txt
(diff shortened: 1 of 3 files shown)


kvark:
Published in naga-0.8.2

PR #1660, opened 2022-01-11
Standardize some docs

NoelTautges:
This pull request standardizes some list items, links some references, fills in some missing docs, corrects capitalization, adds periods at the ends of items, and converts module-level block comments to line comments because docs.rs adds an extra exclamation point when those are used (due to a bug, I think).

It doesn't touch every doc item, but it's a start. Let me know how I can improve it (this is my first pull request to naga). Thank you!

### DIFF
--- a/src/back/dot/mod.rs
+++ b/src/back/dot/mod.rs
@@ -1,8 +1,9 @@
-/*! GraphViz (DOT) backend
- *
- * This backend writes a graph in the DOT format, for the ease
- * of IR inspection and debugging.
-!*/
+//! Backend for [DOT][dot] (Graphviz).
+//!
+//! This backend writes a graph in the DOT format, for the ease
+//! of IR inspection and debugging.
+//! 
+//! [dot]: https://graphviz.org/doc/info/lang.html
 
 use crate::{
     arena::Handle,
--- a/src/back/dot/mod.rs
+++ b/src/back/dot/mod.rs
@@ -1,9 +1,11 @@
-//! Backend for [DOT][dot] (Graphviz).
-//!
-//! This backend writes a graph in the DOT language, for the ease
-//! of IR inspection and debugging.
-//!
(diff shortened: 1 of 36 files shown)


kvark:
Thank you! Looks like Rustfmt isn't happy about some of the lines?

src/back/dot/mod.rs
```diff
@@ -1,8 +1,9 @@
-/*! GraphViz (DOT) backend
- *
- * This backend writes a graph in the DOT format, for the ease
- * of IR inspection and debugging.
-!*/
+//! Backend for [DOT][dot] (Graphviz).
```
kvark, line 1: Why do we need to convert the module documentation into per-line format?
I see that the current rendering is broken because of "*" at the start of lines. Just removing the "*" appears to be enough, and a single multi-line block is much easier to maintain than a bunch of individual prefixed lines.

NoelTautges: My reasoning was:
1. The formatting error you mentioned, which could be fixed the way you said,
2. The comment format is inconsistent between front/backends ([block comment for MSL](https://github.com/gfx-rs/naga/blob/f51f468a47312608e8f4fc59c8f1e65ce37012de/src/back/msl/mod.rs#L1) vs [line comments for GLSL](https://github.com/gfx-rs/naga/blob/f51f468a47312608e8f4fc59c8f1e65ce37012de/src/back/glsl/mod.rs)), and
3. [In-module comments use the line format.](https://github.com/gfx-rs/naga/blob/f51f468a47312608e8f4fc59c8f1e65ce37012de/src/lib.rs#L236)

If this isn't convincing, let me know and I can change them all to block comments!

NoelTautges: Thanks for writing all this out. I switched all of the module line comments to block comments. 😃

kvark (approved):
Thank you!

kvark:
Please rebase and make CI happy

PR #1668, opened 2022-01-13
Arbitrary IR fuzzing

kvark:
This change adds `arbitrary` feature and implements fuzzing of IR based on it, with cargo-fuzz.
We aren't enabling it on CI yet because it catches a ton of low-hanging fruits - let's fix those first.


src/valid/function.rs
```diff
@@ -843,7 +837,14 @@ impl super::Validator {
 
         #[cfg(feature = "validate")]
         for (index, argument) in fun.arguments.iter().enumerate() {
-            match module.types[argument.ty].inner.pointer_class() {
+            let ty = module.types.get_handle(argument.ty).ok_or(
+                FunctionError::InvalidArgumentType {
```
jimblandy, line 841: This isn't the right error for this situation. `InvalidArgumentType` is supposed to be for types that Naga forbids passing into functions, not wild type handles.

I think we should just have one error, `BadArenaHandle { arena: &'static str, holder: &'static str, index: usize }`, and not bother providing any more detail about it, since the only people who care about these errors are front end developers. I guess we can't put `BadArenaHandle` in `ValidationError`, since the context provided by wrapping `FunctionError`, `TypeError`, etc. in a `ValidationError` is valuable. Maybe it's good enough to just add identical `BadArenaHandle` variants to each of the more specific error enums.

src/valid/interface.rs
```diff
@@ -332,7 +332,10 @@ impl super::Validator {
         use super::TypeFlags;
 
         log::debug!("var {:?}", var);
-        let type_info = &self.types[var.ty.index()];
+        let type_info = self
+            .types
+            .get(var.ty.index())
+            .ok_or(GlobalVariableError::InvalidType)?;
```
jimblandy, line 338: Here, too, the error message for `InvalidType` is `"Type isn't compatible with the storage class"`, which is just bad.

src/valid/mod.rs
```diff
@@ -240,7 +240,11 @@ impl Validator {
                 }
             }
             crate::ConstantInner::Composite { ty, ref components } => {
-                match types[ty].inner {
+                match types
+                    .get_handle(ty)
+                    .ok_or(ConstantError::InvalidType)?
```
jimblandy, line 245: Again, bad message.

jimblandy (changes requested):
In addition to fixing CI, I think we need to change the errors being used here.

jimblandy (approved):
I think it's good to get these changes in, but I definitely think there are two loose ends we should tie up.

- We should have a single error type for bad handles, used everywhere. See my comments in the patch for some ideas. Another idea would be to have this error type defined in `arena.rs`, returned directly by some function that resembles `get` but returns a `Result`, and then have `#[transparent]` variants in the error types of functions that use those arena methods. But, it's important for us to establish a clear practice for this, because otherwise everyone who contributes patches to the validator or processors is going to improvise.
- It's not spelled out in any way where it's okay to unwrap `size`, and where one should pass an error. I know that the rationale right now is, "If validation should already have caught it, then unwrap," but I don't think people are going to pick up on this on their own. There should at least be comments. But it might? be better to have separate functions, designated for use only on validated modules, that panic on bad handles. I'm not sure it's much better - but either way, there needs to be a comment we can point people at that explains the policy.

kvark:
Thank you for the well-thought review!
I believe everything is addressed now. I'm going to merge it without squashing, since the commits are individual, and the only failures here are related to MSRV of arbitrary and the formatting stuff - we can live with that when bisecting.

kvark:
I had an abrupt connection loss and my last commit wasn't pushed here before merging. So following up with  #1669

PR #1670, opened 2022-01-15
Add .vscode to .gitignore

NoelTautges:
The .gitignore already has .idea in it and it can be removed later if there are ever official extensions/tasks/launches.


grovesNL (approved):
Thank you!

PR #1671, opened 2022-01-15
glsl-out: Write correct argument name when skipping samplers

JCapucho:
Previously the name of the sampler would be picked for the next argument, this is bad for two reasons:

- The debug data would be wrong
- If the argument was a pointer and written to, the expression would pick the right name but there wouldn't be a argument with that name


kvark (approved):

kvark:
published in 0.8.3

PR #1672, opened 2022-01-15
glsl-out: Add support for push constant emulation

JCapucho:
This PR allows for the glsl backend to output push constants as uniform buffers to allow for push constant emulation.

It only allows for one push constant per entry point, which is fine in my opinion since even vulkan glsl only allows one `push_constant` decoration per shader.

It also changes the `ReflectionInfo` to return a name of the uniform buffer used for the emulation if it was used and `Options` to take the binding to use for the buffer if explicit bindings are supported.

This interface is deeply tied with the changes for push constant emulation on wgpu, so it might be useful to also see it https://github.com/JCapucho/wgpu/commit/ee3887ebb53e04471e221f21d1126aeb73630532

**Note**: I don't have much experience with openGL so the code might not be the best.

### DIFF
--- a/tests/snapshots.rs
+++ b/tests/snapshots.rs
@@ -439,6 +439,7 @@ fn convert_wgsl() {
             Targets::SPIRV | Targets::METAL | Targets::HLSL | Targets::WGSL | Targets::GLSL,
         ),
         ("extra", Targets::SPIRV | Targets::METAL | Targets::WGSL),
+        ("push-constants", Targets::GLSL),
         (
             "operators",
             Targets::SPIRV | Targets::METAL | Targets::GLSL | Targets::HLSL | Targets::WGSL,
--- a/src/back/glsl/features.rs
+++ b/src/back/glsl/features.rs
@@ -343,6 +343,8 @@ impl<'a, W> Writer<'a, W> {
             }
         }
 
+        let mut push_constant_used = false;
+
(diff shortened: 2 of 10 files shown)


kvark (approved):
Looks great!
Can we have a small glsl test for this (or add it to an existing one)?

JCapucho:
@kvark I've added a test, it's basically a stripped down version of `extra.wgsl`, I tried to reuse it but it uses `f64`s which gles doesn't support, so I've copied the push constant parts and made it only output to glsl with version 320 to also check if the explicit bindings are working.

tests/snapshots.rs
```diff
@@ -439,6 +439,7 @@ fn convert_wgsl() {
             Targets::SPIRV | Targets::METAL | Targets::HLSL | Targets::WGSL | Targets::GLSL,
         ),
         ("extra", Targets::SPIRV | Targets::METAL | Targets::WGSL),
+        ("push-constants", Targets::GLSL),
```
kvark, line 442: Since `extra` enables push constants, could we just enable it instead of the new test?

JCapucho: I wrote about it in my comment, `extra.wgsl` uses `f64`s which goes doesn't support

kvark: oh sorry, didn't see this

kvark (approved):

PR #1675, opened 2022-01-19
[wgsl-in] implement block comments

### LINKED ISSUE #1582: error occurs in validating the block comment in WGSL
WGSL allows the block comment, and it works fine in WebGPU apps. Here is an example of using block (or multi-line) comment from the official WGSL site (https://www.w3.org/TR/WGSL/#comments):

EXAMPLE: COMMENTS
let f = 1.5;   // This is line-ending comment.
let g = 2.5;   /* This is a block comment
                  that spans lines.
                  /* Block comments can nest.
                   */
                  But all block comments must terminate.
                */

However, naga gives the following error for this block comment in wgpu apps:

Shader error:
error: expected global item ('struct', 'let', 'var', 'type', ';', 'fn') or the end of the file, found '/'

How to get around this error?

kocsis1david:
Fixes #1582

Implement block comments that support nesting.

### DIFF
--- a/src/front/wgsl/lexer.rs
+++ b/src/front/wgsl/lexer.rs
@@ -356,9 +356,47 @@ fn consume_token(mut input: &str, generic: bool) -> (Token<'_>, &str) {
                 (Token::UnterminatedString, quote_content)
             }
         }
-        '/' if chars.as_str().starts_with('/') => {
-            let _ = chars.position(|c| c == '\n' || c == '\r');
-            (Token::Trivia, chars.as_str())
+        '/' => {
+            input = chars.as_str();
+            match chars.next() {
+                Some('/') => {
+                    let _ = chars.position(|c| c == '\n' || c == '\r');
+                    (Token::Trivia, chars.as_str())
+                }
+                Some('*') => {
+                    input = chars.as_str();
+
+                    let mut depth = 1;
+                    let mut prev = '\0';
+
+                    for c in &mut chars {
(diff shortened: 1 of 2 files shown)


src/front/wgsl/lexer.rs
```diff
@@ -356,9 +356,41 @@ fn consume_token(mut input: &str, generic: bool) -> (Token<'_>, &str) {
...
+                    let _ = chars.position(|c| c == '\n' || c == '\r');
+                    (Token::Trivia, chars.as_str())
+                }
+                Some('*') => {
+                    input = chars.as_str();
+
+                    let mut depth = 1;
+                    while let Some(c) = chars.next() {
+                        match c {
+                            '*' if chars.next() == Some('/') => {
```
kvark, line 372: this will fail on a case like this: `/* blabla **/`

kvark (commented):
Very nice, thank you for tackling this!
We may need to fix it a little bit better

src/front/wgsl/lexer.rs
kocsis1david: hopefully it's fixed, there can be strange cases like `/*/**/****/`, which is a valid closed comment

kvark (approved):

kvark:
Some tests would be nice to have

kvark:
And clippy is unhappy

kvark (approved):
Thank you!

Issue #1677, opened 2022-01-20
GLSL barriers should come with barrier()


kvark:
See https://github.com/gfx-rs/wgpu/issues/2408
We need to issue both `groupMemoryBarrier()` and `barrier()` at the same time.

PR #1680, opened 2022-01-21
glsl-out: Call proper memory barrier functions

francesco-cattoglio:
This should fix https://github.com/gfx-rs/naga/issues/1677
OpenGL memory barriers and visibility are really confusing to me, but I spent some time researching to the best of my ability to make this correct: I started from https://www.khronos.org/opengl/wiki/Memory_Model#Ensuring_visibility and then read some extra questions/answers/blog posts around the net to confirm that I understood the khronos wiki correctly.

After the code change I did run `cargo test --all-features --workspace` and it reported no error, then I ran `make validate-glsl` and finally committed the new (modified) test output.

Edit: P.S: the `barrier()` GLSL function can only be called in a compute shader, but I am pretty sure this is fine because WGSL synchronization built-in functions can only be called in compute shaders anyway.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1776,8 +1776,6 @@ impl<'a, W: Write> Writer<'a, W> {
             Statement::Barrier(flags) => {
                 if flags.contains(crate::Barrier::STORAGE) {
                     writeln!(self.out, "{}memoryBarrierBuffer();", level)?;
-                    writeln!(self.out, "{}memoryBarrierImage();", level)?;
-                    writeln!(self.out, "{}memoryBarrierAtomicCounter();", level)?;
                 }
 
                 if flags.contains(crate::Barrier::WORK_GROUP) {
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1771,13 +1771,20 @@ impl<'a, W: Write> Writer<'a, W> {
             // keyword which ceases all further processing in a fragment shader, it's called OpKill
             // in spir-v that's why it's called `Statement::Kill`
             Statement::Kill => writeln!(self.out, "{}discard;", level)?,
-            // Issue an execution or a memory barrier.
+            // Issue a memory barrier. Please note that to ensure visibility,
+            // OpenGL always requires a call to the `barrier()` function after a `memoryBarrier*()`
             Statement::Barrier(flags) => {
-                if flags.is_empty() {
-                    writeln!(self.out, "{}barrier();", level)?;
-                } else {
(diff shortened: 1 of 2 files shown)


src/back/glsl/mod.rs
```diff
@@ -1771,13 +1771,20 @@ impl<'a, W: Write> Writer<'a, W> {
...
+            // OpenGL always requires a call to the `barrier()` function after a `memoryBarrier*()`
             Statement::Barrier(flags) => {
-                if flags.is_empty() {
-                    writeln!(self.out, "{}barrier();", level)?;
-                } else {
-                    writeln!(self.out, "{}groupMemoryBarrier();", level)?;
+                if flags.contains(crate::Barrier::STORAGE) {
+                    writeln!(self.out, "{}memoryBarrierBuffer();", level)?;
+                    writeln!(self.out, "{}memoryBarrierImage();", level)?;
+                    writeln!(self.out, "{}memoryBarrierAtomicCounter();", level)?;
```
kvark, line 1780: I don't think this is necessary, it seems to be affecting the stuff we aren't exposing in WebGPU - https://www.lighthouse3d.com/tutorials/opengl-atomic-counters/

src/back/glsl/mod.rs
```diff
@@ -1771,13 +1771,20 @@ impl<'a, W: Write> Writer<'a, W> {
...
+            // Issue a memory barrier. Please note that to ensure visibility,
+            // OpenGL always requires a call to the `barrier()` function after a `memoryBarrier*()`
             Statement::Barrier(flags) => {
-                if flags.is_empty() {
-                    writeln!(self.out, "{}barrier();", level)?;
-                } else {
-                    writeln!(self.out, "{}groupMemoryBarrier();", level)?;
+                if flags.contains(crate::Barrier::STORAGE) {
+                    writeln!(self.out, "{}memoryBarrierBuffer();", level)?;
+                    writeln!(self.out, "{}memoryBarrierImage();", level)?;
```
kvark, line 1779: I think the current semantics of `crate::Barrier::STORAGE` doesn't include images. Trying to clarify that with the group. So we should probably remove it for now.

francesco-cattoglio: I decided to add the `Image` barrier after taking a look at the naga HLSL-out writer, which produces a `DeviceMemoryBarrier()` function call. Then I looked for more information and found this: https://github.com/microsoft/DirectXShaderCompiler/blob/master/docs/SPIR-V.rst#synchronization-intrinsics
Anyway, I will remove this for now.

kvark (approved):

kvark:
published in 0.8.4

PR #1681, opened 2022-01-21
wgsl: remove stride

kvark:
Matches https://github.com/gpuweb/gpuweb/pull
I first thought we'd remove the stride from the IR but then realized the IR is a bit more detailed than WGSL. In particular, it already has struct member offsets and struct sizes. So keeping `stride` in the IR is more consistent.

### DIFF
--- a/src/front/glsl/types.rs
+++ b/src/front/glsl/types.rs
@@ -259,10 +259,13 @@ impl Parser {
         mut meta: Span,
         array_specifier: Option<(ArraySize, Span)>,
     ) -> Handle<Type> {
-        array_specifier
-            .map(|(size, size_meta)| {
+        match array_specifier {
+            Some((size, size_meta)) => {
                 meta.subsume(size_meta);
-                let stride = self.module.types[base].inner.size(&self.module.constants);
+                self.layouter
+                    .update(&self.module.types, &self.module.constants)
+                    .unwrap();
+                let stride = self.layouter[base].to_stride();
                 self.module.types.insert(
                     Type {
                         name: None,
@@ -270,7 +273,8 @@ impl Parser {
                     },
                     meta,
                 )
-            })
-            .unwrap_or(base)
+            }
+            None => base,
+        }
     }
 }
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -158,7 +158,6 @@ pub enum Error<'a> {
     UnknownType(Span),
     UnknownStorageFormat(Span),
     UnknownConservativeDepth(Span),
-    ZeroStride(Span),
     ZeroSizeOrAlign(Span),
     InconsistentBinding(Span),
     UnknownLocalFunction(Span),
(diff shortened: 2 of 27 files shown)


src/front/wgsl/mod.rs
```diff
@@ -1047,9 +1041,7 @@ impl Composition {
 }
 
 #[derive(Default)]
-struct TypeAttributes {
-    stride: Option<NonZeroU32>,
-}
+struct TypeAttributes {}
```
jimblandy, line 1044: This deserves a comment:

> Although WGSL has no type attributes at the moment, it has had them in the past and may well acquire some again in the future, so just in case, we're leaving the plumbing in.

src/front/glsl/types.rs
```diff
@@ -259,18 +259,22 @@ impl Parser {
...
     ) -> Handle<Type> {
-        array_specifier
-            .map(|(size, size_meta)| {
+        match array_specifier {
+            Some((size, size_meta)) => {
                 meta.subsume(size_meta);
-                let stride = self.module.types[base].inner.size(&self.module.constants);
+                self.layouter
+                    .update(&self.module.types, &self.module.constants)
+                    .unwrap();
```
jimblandy, line 267: nice catch

src/valid/type.rs
```diff
@@ -328,19 +328,21 @@ impl super::Validator {
...
-                let general_alignment = self.layouter[base].alignment;
+                let general_alignment_mask = base_layout.alignment.get() - 1;
                 let uniform_layout = match base_info.uniform_layout {
                     Ok(base_alignment) => {
                         // combine the alignment requirements
-                        let align = ((base_alignment.unwrap().get() - 1)
-                            | (general_alignment.get() - 1))
-                            + 1;
+                        let align =
+                            ((base_alignment.unwrap().get() - 1) | general_alignment_mask) + 1;
```
jimblandy, line 345: Wouldn't it be simpler and clearer just to use `max`?

src/valid/type.rs
```diff
@@ -357,9 +359,8 @@ impl super::Validator {
                 };
                 let storage_layout = match base_info.storage_layout {
                     Ok(base_alignment) => {
-                        let align = ((base_alignment.unwrap().get() - 1)
-                            | (general_alignment.get() - 1))
-                            + 1;
+                        let align =
+                            ((base_alignment.unwrap().get() - 1) | general_alignment_mask) + 1;
```
jimblandy, line 363: Same here.

tests/out/hlsl/access.hlsl
```diff
@@ -24,7 +24,7 @@ float4 foo(uint vi : SV_VertexID) : SV_Position
     float4x4 matrix_ = float4x4(asfloat(bar.Load4(0+0)), asfloat(bar.Load4(0+16)), asfloat(bar.Load4(0+32)), asfloat(bar.Load4(0+48)));
     uint2 arr[2] = {asuint(bar.Load2(104+0)), asuint(bar.Load2(104+8))};
     float b = asfloat(bar.Load(0+48+0));
-    int a = asint(bar.Load((((NagaBufferLengthRW(bar) - 120) / 8) - 2u)*8+120));
+    int a = asint(bar.Load((((NagaBufferLengthRW(bar) - 120) / 4) - 2u)*4+120));
```
jimblandy, line 27: Was this supposed to change? Should we have updated `tests/in/access.wgsl` to use a dummy struct?

jimblandy (approved):
Looks good in general - just need to fix a stray attribute, and I had a few questions.

src/valid/type.rs
kvark: whoa, nicely spotted!

tests/out/hlsl/access.hlsl
kvark: good idea!

PR #1683, opened 2022-01-21
glsl-out: push constants use anonymous uniforms

JCapucho:
Previously this was done with UBOs but this posed some problems when
integrating with wgpu (see https://github.com/gfx-rs/wgpu/pull/2400)

It basically reverses the previous PR retaining only the feature detection changes

### DIFF
--- a/tests/out/glsl/push-constants.main.Fragment.glsl
+++ b/tests/out/glsl/push-constants.main.Fragment.glsl
@@ -9,7 +9,7 @@ struct PushConstants {
 struct FragmentIn {
     vec4 color;
 };
-layout(std140, binding = 4) uniform PushConstants_block_0Fragment { PushConstants pc; };
+uniform PushConstants pc;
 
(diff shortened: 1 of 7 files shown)


tests/out/glsl/push-constants.main.Fragment.glsl
```diff
@@ -9,7 +9,7 @@ struct PushConstants {
 struct FragmentIn {
     vec4 color;
 };
-layout(std140, binding = 4) uniform PushConstants_block_0Fragment { PushConstants pc; };
+uniform PushConstants pc;
```
kvark, line 12: does that work? will it show up as separate constants for the fields?

kvark (approved):
Great

tests/out/glsl/push-constants.main.Fragment.glsl
JCapucho: Yes OpenGL will report each field of the struct separately

Issue #1667, opened 2022-01-13
[wgsl-in] dot product on integers isn't implemented


hasali19:
```
fn main() {
    dot(vec2<i32>(0), vec2<i32>(0));
}
```

```
error: 
  ┌─ y.wgsl:1:1
  │
1 │ ╭   fn main() {
  │ │ ╭───────────^
2 │ │ │     dot(vec2<i32>(0), vec2<i32>(0));
  │ │ ╰───────────────────────────────────^ naga::Expression [5]
3 │ │   }
  │ ╰───^ naga::Function [1]

Function [1] 'main' is invalid:
        Expression [5] is invalid
        Argument [0] to Dot as expression [2] has an invalid type.
```

kvark:
We haven't added this yet. Upstream - https://github.com/gpuweb/gpuweb/issues/2231

PR #1689, opened 2022-01-24
Add support for `vecN<i32>` and `vecN<u32>` to `dot()` function

francesco-cattoglio:
Work-in-progress pull request to fix https://github.com/gfx-rs/naga/issues/1667

Right now I have only implemented this for glsl-out to get an early review.
By reading the discussion at https://github.com/gpuweb/gpuweb/issues/2231 and the minutes in https://docs.google.com/document/d/1o3BrMEZnE3Yat95uLiY06opF2Hmcj5XGlQmBhTT-_dQ/edit# I understand that:
- HLSL requires no extra changes, the `dot()` function will just work 
- MSL and Spir-V will require changes to the code similar to the ones made for GLSL: if vectors of integers are involved we need to turn the `dot()` call into an arithmetic expression.


src/back/glsl/mod.rs
```diff
@@ -2532,7 +2532,63 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                                    ScalarKind::Float => "dot",
+                                    ScalarKind::Sint | ScalarKind::Uint => {
+                                        // No intrinsic function for integer dot product in GLSL,
+                                        // transform the function call into an arithmetic expression
+                                        let arg1 =
+                                            arg1.ok_or_else(|| Error::Custom("Missing dot arg1".to_owned()))?;
+
+                                        write!(self.out, "(")?;
+
+                                        self.write_expr(arg, ctx)?;
```
kvark, line 2550: Writing the same expression multiple times is a valid thing to do, but it can end up with a lot of unnecessary repetition.
We need to force the arguments of `dot(integer, integer)` to be "baked".
Currently the baking logic in GLSL is just checking `bake_ref_count()`. Here, it needs to change to something like:
```rust
if ref_count > 1 || is_argument_of_integer_dot_product(expr) {
 // bake
}
```

francesco-cattoglio: I tried to find a way to easily bake the arguments of a dot product, but to no avail. If I understand correctly, the "needs to be baked" information should live somewhere in the `FunctionCtx` that is created by each frontend (just like `named_expressions` are), so that when the time comes, the backend can know that the expression needs to be baked. This is a major change to Naga, so I don't think this is something I can achieve.

OTOH, I was wondering: would it be reasonable to do some "local" baking by creating a scope and using a temporary variable? I recon this is a lot of code that handles a very specific case, but perhaps it could be a good compromise.

kvark: The way I thought this could work is you making a `BitSet` to store indices of expressions (within the currently processed functions) that are integer arguments to `dot()` product. When starting each function, you'd clear that bitset, then walk through all expressions and add the relevant indices to the bitset. And then `is_argument_of_integer_dot_product` is basically a check if `expr.index()` is in this bitset.

francesco-cattoglio: I think I get it now, this would be a change that lives mostly only inside the `write_function` call of a backend. I will see if I can manage something reasonable!

src/back/glsl/mod.rs
```diff
@@ -1136,12 +1136,44 @@ impl<'a, W: Write> Writer<'a, W> {
         func: &crate::Function,
         info: &valid::FunctionInfo,
     ) -> BackendResult {
+
+        // Do a first pass on all statements and store the handle of the ones
+        // that will need baking
+        use crate::Expression;
+        let mut baking_set = crate::FastHashSet::default();
```
kvark, line 1143: could we make `baking_set` a part of `self` to re-use the allocation between functions?
This whole block should be moved into a function as well.

kvark (changes requested):
This is looking good!
Just one more thing :)

src/back/msl/writer.rs
```diff
@@ -1192,7 +1195,67 @@ impl<W: Write> Writer<W> {
...
+                    Mf::Dot => {
+                        use crate::TypeInner;
+                        let inner = context.resolve_type(arg);
+                        match *inner {
+                            TypeInner::Vector { kind, size, .. } => {
+                                use crate::ScalarKind;
+                                match kind {
+                                    ScalarKind::Float => "dot",
+                                    ScalarKind::Sint | ScalarKind::Uint => {
+                                        // No intrinsic function for integer dot product in MSL,
```
kvark, line 1207: let's move this into a function

src/back/msl/writer.rs
```diff
@@ -1192,7 +1195,67 @@ impl<W: Write> Writer<W> {
                     Mf::Log2 => "log2",
                     Mf::Pow => "pow",
                     // geometry
-                    Mf::Dot => "dot",
+                    Mf::Dot => {
```
kvark, line 1198: should just do `=> match *context.resolve_type(arg) {..}` from here

src/back/msl/writer.rs
```diff
@@ -1192,7 +1195,67 @@ impl<W: Write> Writer<W> {
...
+                        use crate::TypeInner;
+                        let inner = context.resolve_type(arg);
+                        match *inner {
+                            TypeInner::Vector { kind, size, .. } => {
```
kvark, line 1202: let's pattern match the kind here, i.e.
```rust
TypeInner::Vector { kind: ScalarKind::Float, .. } => "dot",
TypeInner::Vector { size, .. } => {
  self.write_dot_product(...);
}
```

src/back/msl/writer.rs
```diff
@@ -1192,7 +1195,67 @@ impl<W: Write> Writer<W> {
...
+                                        let arg1 = arg1.unwrap();
+
+                                        write!(self.out, "(")?;
+
+                                        self.put_expression(arg, context, false)?;
+                                        write!(self.out, ".x * ")?;
+
+                                        self.put_expression(arg1, context, false)?;
+                                        write!(self.out, ".x + ")?;
+
```
kvark, line 1220: should refactor this into a loop:
```rust
for component in COMPONENTS[1..size as usize] {
                                            write!(self.out, " + ")?;

                                            self.put_expression(arg, context, false)?;
                                            write!(self.out, ".{} * ", component)?;

                                            self.put_expression(arg1, context, false)?;
                                            write!(self.out, ".{}", component)?;
}
```

src/back/msl/writer.rs
```diff
@@ -1192,7 +1195,67 @@ impl<W: Write> Writer<W> {
...
+                                    ScalarKind::Sint | ScalarKind::Uint => {
+                                        // No intrinsic function for integer dot product in MSL,
+                                        // transform the function call into an arithmetic expression.
+
+                                        // arg1 should be safe to unwrap due to previous validation
+                                        let arg1 = arg1.unwrap();
+
+                                        write!(self.out, "(")?;
+
+                                        self.put_expression(arg, context, false)?;
```
kvark, line 1215: since the backend guarantees `arg` and `arg1` to be baked here, we can and should write this directly:
```rust
        let arg0_name = self.named_expressions[&arg];
        let arg1_name = self.named_expressions[&arg1];
        write!(self.out, "+ {}.{} * {}.{}", arg0_name, component, arg1_name, component)?;
```

src/back/spv/block.rs
```diff
@@ -552,13 +552,175 @@ impl<'w> BlockContext<'w> {
...
+
+                                        let a_z = self.gen_id();
+                                        block.body.push(Instruction::composite_extract(
+                                            result_type_id,
+                                            a_z,
+                                            arg0_id,
+                                            &[2],
+                                        ));
+                                        let b_z = self.gen_id();
+                                        block.body.push(Instruction::composite_extract(
```
kvark, line 647: similarly here, let's move it out into a function and try to write as a loop

src/back/spv/block.rs
```diff
@@ -552,13 +552,175 @@ impl<'w> BlockContext<'w> {
...
-                        spirv::Op::Dot,
-                        result_type_id,
-                        id,
-                        arg0_id,
-                        arg1_id,
-                    )),
+                    Mf::Dot => {
+                        use crate::TypeInner;
+                        let inner_ty = self.fun_info[arg].ty.inner_with(&self.ir_module.types);
+                        dbg!(inner_ty);
```
kvark, line 558: stray `dbg!`

src/back/spv/block.rs
```diff
@@ -552,13 +552,175 @@ impl<'w> BlockContext<'w> {
...
                     Mf::Ldexp => MathOp::Ext(spirv::GLOp::Ldexp),
                     // geometry
-                    Mf::Dot => MathOp::Custom(Instruction::binary(
-                        spirv::Op::Dot,
-                        result_type_id,
-                        id,
-                        arg0_id,
-                        arg1_id,
-                    )),
+                    Mf::Dot => {
```
kvark, line 555: let's also leave a comment to use `VK_KHR_shader_integer_dot_product` when available

kvark (changes requested):
very nice!

francesco-cattoglio:
I implemented the requested changes. On MSL and GLSL back I think I did a pretty good job so far. When refactoring the sum of components into a loop I incorporated the x as well, and therefore there will be an extra "+" sign at the beginning of the produced expression. Let me know if you like it or not. If not, I will move the product of x outside of the loop.

On spv back, I tried my best to reduce repetition, but perhaps I wrote some "clever code" that is just "confusing and obscure" code for something that could have been done much easier.

kvark (approved):
Nice!

src/back/spv/block.rs
```diff
@@ -1120,6 +1141,101 @@ impl<'w> BlockContext<'w> {
...
+        size: crate::VectorSize,
+        block: &mut Block,
+    ) -> Result<(), Error> {
+        // Create an iterator over the IDs that will store the partial sums of products.
+        // Last iterator element will always be result_id, while the others will be generated IDs.
+        let sum_2_ids: [Word; 1];
+        let sum_3_ids: [Word; 2];
+        let sum_4_ids: [Word; 3];
+        use crate::VectorSize;
+        let sum_ids = match size {
```
kvark, line 1160: I think we can simplify this by starting off `OpConstantNull` of type `result_type_id` and just adding to it in a loop

kvark (changes requested):
One last thing!

kvark (approved):
great!

PR #1693, opened 2022-01-24
Document `proc::layouter`.

jimblandy:
Rename `TypeLayoutError` to `LayoutErrorInner`, for consistency. It makes no sense to have 'Type' in the name of one of the error types but not the other. The 'Inner' convention is used by the module types.

Cite the layout rules in the WGSL spec, not a GitHub issue.

(I just wrote this while reviewing another PR.)

### DIFF
--- a/src/proc/layouter.rs
+++ b/src/proc/layouter.rs
@@ -3,7 +3,7 @@ use std::{num::NonZeroU32, ops};
 
 pub type Alignment = NonZeroU32;
 
-/// Alignment information for a type.
+/// Size and alignment information for a type.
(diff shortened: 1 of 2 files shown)


src/proc/layouter.rs
```diff
@@ -83,6 +101,19 @@ impl Layouter {
         (start..start + span, alignment)
     }
 
+    /// Extend this `Layouter` with layouts for any new entries in `types`.
```
kvark, line 104: oh sweet, this function needed some text on it for sure!

kvark (approved):
very well done

PR #1701, opened 2022-01-26
naga-cli: Support reading input from stdin

### LINKED ISSUE #1700: naga-cli: Support reading input from stdin
Current interface of `naga-cli` is as follows:

```
naga [OPTIONS] INPUT [OUTPUT...]
```

Since `INPUT` must be specified, the input file must exist in filesystem.

Since the input validation while editing shader program is great for those who are not familiar with WGSL like me. I'm considering to run `naga` command in a text editor on the fly.

However, while editing a source, the source text is not written to a file yet. It is only on memory of the text editor.

I suppose there are two possible solutions.

1. `naga` supports reading input from stdin (`INPUT` argument will be optional). A text editor can pass the current source via stdin
2. A text editor writes the current source to a temporary file and pass the path to `naga`

Cons of 1. is that it needs new implementation.
Cons of 2. is that it requires overhead of writing source to temporary file. The file path is different from actual file path being edited.

I think 1. is cleaner and faster.

rhysd:
Fixes #1700 

Three changes are added to support reading from stdin:

- ~With `argh` crate, the `INPUT` argument cannot be optional. So this PR replaces `argh` with `clap` crate.~
- The `INPUT` argument is now optional: `naga [OPTIONS] [FILES...]`
- Since a file path cannot be known when reading from stdin, so a new option `--stdin-file-path <PATH>` is added to notify the file path. This is the same way as `eslint` does.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,7 +1,7 @@
 #![allow(clippy::manual_strip)]
 #[allow(unused_imports)]
 use std::fs;
-use std::{error::Error, fmt, path::Path, str::FromStr};
+use std::{error::Error, fmt, io::Read, path::Path, str::FromStr};
 
 /// Translate shaders to different formats
 #[derive(argh::FromArgs, Debug, Clone)]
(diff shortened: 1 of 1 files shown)


kvark:
> With argh crate, the INPUT argument cannot be optional.

I don't think this reasoning is really sound. All it would take is to change this:
```rust
    /// the input file
    #[argh(positional)]
    input: String,

    /// the output file. If not specified, only validation will be performed
    #[argh(positional)]
    output: Vec<String>,
```
Into this:
```rust
    /// the input and output files
    #[argh(positional)]
    files: Vec<String>,
```

Would you be open to refactor the PR on top of argh?

rhysd:
> Would you be open to refactor the PR on top of argh?

That also makes sense to me. I will update this patch today.

rhysd:
@kvark I update this patch to keep `argh` dependency and force pushed the branch. Could you review the changes?

kvark (approved):
Nice!

Issue #1643, opened 2022-01-03
MSL bounds checks don't allow taking an address of an element of an array


kyren:
In WGSL we have code like this

```
let v = atomicAdd(&histogram_shared_group[bucket_index], 1u);
```

and when compiled by naga into MSL with bounds checks, that gets turned into this:

```
metal::uint _e26 = metal::atomic_fetch_add_explicit(&metal::uint(_e23) < 256 ? histogram_shared_group.inner[_e23] : DefaultConstructible(), 1u, metal::memory_order_relaxed);
```

which then generates this error from metal:

```
program_source:67:61: error: cannot take the address of an rvalue of type 'metal::uint' (aka 'unsigned int')
        metal::uint _e26 = metal::atomic_fetch_add_explicit(&metal::uint(_e23) < 256 ? histogram_shared_group.inner[_e23] : DefaultConstructible(), 1u, metal::memory_order_relaxed);
```

This doesn't compile obviously because of the badly placed `&`, but even if that was fixed, in order to make the entire thing work it would still require taking the address of an rvalue (namely `DefaultConstructible()`) right?  I'm not sure what the best way to fix this is.

This is a pretty rough bug because currently this is the only way of using atomics that I am aware of.

jimblandy:
@kyren, thanks for reporting this. As a workaround, would it suffice to use `BoundsCheckPolicy::Restrict` instead of `ReadZeroSkipWrite`?

kyren:
@jimblandy I'm using naga indirectly through wgpu, is there a way to pipe that setting through wgpu currently?

kyren:
Oh also, if it helps our particular use case, we don't actually need the return value from `atomicAdd`, that was actually added as a workaround for another bug which I'm not sure of right now, but it no longer exists in wgpu `0.12` and the version of naga that it uses so that part has been removed.

kvark:
> @kyren, thanks for reporting this. As a workaround, would it suffice to use `BoundsCheckPolicy::Restrict` instead of `ReadZeroSkipWrite`?

Since we don't have the full shader, we can just try to construct a case and see.

jimblandy:
@kyren, does bb604fd6 fix this bug?

zeh:
> @kyren, does [bb604fd](https://github.com/gfx-rs/naga/commit/bb604fd6a3ad9cb9dbe2ff579be55d008e624b18) fix this bug?

[...]

PR #1703, opened 2022-01-27
Mitigation for MSL atomic bounds check.

glalonde:
Will CI trigger automatically?

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -791,50 +791,34 @@ impl<W: Write> Writer<W> {
         value: Handle<crate::Expression>,
         context: &ExpressionContext,
     ) -> BackendResult {
-        // TODO Why not always put parentheses?
-        let is_scoped = true;
-        // Atomic ops are structured different from a normal load.
+        // Atomic ops are structured differently from a normal load.
         // Rewrite this for the bounds-check case so that the bounds check happens once
-        // when resolving the pointer, and then use an unchecked access within the atomic expression itself.
+        // when resolving the ptr, and then use an unchecked access within the atomic expression itself.
         let policy = context.choose_bounds_check_policy(expr_handle);
-        if policy == index::BoundsCheckPolicy::ReadZeroSkipWrite
-            && self.put_bounds_checks(
-                expr_handle,
-                context,
-                back::Level(0),
-                if is_scoped { "" } else { "(" },
-            )?
-        {
-            write!(self.out, " ? ")?;
 
-            {
-                // True branch of bounds check ternary.
-                write!(
-                    self.out,
-                    "{}::atomic_fetch_{}_explicit({}",
-                    NAMESPACE, key, ATOMIC_REFERENCE
-                )?;
-                self.put_access_chain(expr_handle, policy, context)?;
-                write!(self.out, ", ")?;
-                self.put_expression(value, context, true)?;
-                write!(self.out, ", {}::memory_order_relaxed)", NAMESPACE)?;
-            }
-
-            write!(self.out, " : DefaultConstructible()")?;
-
-            if !is_scoped {
-                write!(self.out, ")")?;
-            }
-        } else {
+        let put_unchecked_atomic_access = |writer: &mut Writer<W>| -> BackendResult {
             write!(
-                self.out,
+                writer.out,
                 "{}::atomic_fetch_{}_explicit({}",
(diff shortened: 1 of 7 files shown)


glalonde:
Mostly untested attempt for  #1643. This moves the bounds check outside of the atomic expression, but I'm unsure if it is correct to just immediately go to `self.put_access_chain(expr_handle, policy, context)?;` within or if there can be a more complicated expression than just direct member or array access that wouldn't work in this case.

Also not sure about the `is_scoped` logic. I figured it should be safe to always wrap in `()` but for some reason that's discouraged.

glalonde:
Looks like the CI failure is because I added the atomic tests which aren't implemented for spv. I guess I can remove those before submission, or maybe make a separate group of msl tests.

kvark:
@jimblandy please review this
@glalonde sorry about the delay! Please ping us if this gets lost again.

jimblandy:
@glalonde The PR title says "Mitigation", but are you aware of any cases where this doesn't address the problem?

glalonde:
I'm not aware of situations it doesn't handle, but I wouldn't be surprised if they exist.

For example, I'm not sure if there are more complicated access patterns than just array access or struct access ( in the tests) which might break.


I just tried this locally as a more convoluted access:
```
struct AtomicArray {
   a: array<atomic<u32>, 10>;
};

struct Globals {
...
    e: array<AtomicArray, 10>;
};

fn fetch_add_atomic_array2(i: i32, j: i32) -> u32 {
   return atomicAdd(&globals.e[i].a[j], 1u);
}
```

```
metal::uint fetch_add_atomic_array2_(
    int i_14,
    int j_2,
    device Globals& globals,
    constant _mslBufferSizes& _buffer_sizes
) {
    metal::uint _e8 = metal::uint(j_2) < 10 && metal::uint(i_14) < 10 ? metal::atomic_fetch_add_explicit(&globals.e.inner[i_14].a.inner[j_2], 1u, metal::memory_order_relaxed) : DefaultConstructible();
    return _e8;
}
```
And that looks like it would work...

jimblandy:
Yeah, I think as long as we do the same things that an ordinary load would
do, we should be fine. I'll finish the review in a few minutes.

src/back/msl/writer.rs
```diff
@@ -786,20 +786,57 @@ impl<W: Write> Writer<W> {
...
+            if !is_scoped {
+                write!(self.out, ")")?;
+            }
+        } else {
+            write!(
+                self.out,
+                "{}::atomic_fetch_{}_explicit({}",
+                NAMESPACE, key, ATOMIC_REFERENCE
+            )?;
+            self.put_expression(expr_handle, context, true)?;
```
jimblandy, line 834: This is a pre-existing problem, but, this should call `put_access_chain` instead of `put_expression`, for consistency with the atomic case in `put_unchecked_load`. That also won't check again for a `ReadZeroSkipWrite` policy, which at this point we already know does not apply. I would expect this to have no effect on the generated code.

jimblandy (changes requested):
This looks like just the right approach. I have a few small comments, and a suggestion on how to fix the tests.

src/back/msl/writer.rs
glalonde: Factored it into a closure.

glalonde:
Thanks for the review, ptal

src/back/msl/writer.rs
```diff
@@ -786,20 +786,41 @@ impl<W: Write> Writer<W> {
...
+                writer.out,
+                "{}::atomic_fetch_{}_explicit({}",
+                NAMESPACE, key, ATOMIC_REFERENCE
+            )?;
+            writer.put_access_chain(expr_handle, policy, context)?;
+            write!(writer.out, ", ")?;
+            writer.put_expression(value, context, true)?;
+            write!(writer.out, ", {}::memory_order_relaxed)", NAMESPACE)?;
+            Ok(())
+        };
```
jimblandy, line 810: Yes, I noticed the duplicated code as well. Naga's style is to not use closures in this way, to keep control flow simple. This is certainly subjective, but it's what we've settled on.

Instead, please declare a `bool`, and then have two `if` statements on either side of the common code.

jimblandy:
I sorted out the style points - this is a good fix and I want to get it in right away.

Thank you very much!

glalonde:
awesome, thanks! looking forward to the next release : )

PR #1713, opened 2022-02-03
glsl-in: Type qualifiers rework

JCapucho:
The previous implementation had many issues, most importantly it didn't
allow to implement (at least in a sane way) some future features like
default qualifiers and format qualifiers.

This implementations centralizes the qualifiers into a struct with a
hashmap for layout qualifiers, error reporting is also centralized which
means that `add_global_var` and `add_local_var` no longer need to
duplicate effort and the ugly loop + match combo is gone.

Finally some minor fixes are also done as part of this rework, like
a default location being provided for input/outputs variables, layout
qualifiers being allowed to be repeated (overwriting the previous one)
like the spec defines, more syntax validation is added to qualifiers and
more errors are made to not stop parsing when encountered.

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use std::{borrow::Cow, fmt};
 
 use super::{builtins::MacroCall, context::ExprPos, Span};
 use crate::{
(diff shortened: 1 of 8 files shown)


src/front/glsl/ast.rs
```diff
@@ -134,19 +134,117 @@ pub enum HirExprKind {
...
+    /// Removes the layout qualifier with `name`, if it exists and adds an error if it isn't
+    /// a [`QualifierValue::Uint`]
+    pub fn uint_layout_qualifier(
+        &mut self,
+        name: &str,
+        errors: &mut Vec<super::Error>,
+    ) -> Option<u32> {
+        match self
+            .layout_qualifiers
+            .remove(&QualifierKey::String(String::from(name)))
```
kvark, line 208: Can `QualifierKey` have `Cow<>` instead of `String` in it to avoid heap allocations in places like this?

src/front/glsl/ast.rs
```diff
@@ -134,19 +134,117 @@ pub enum HirExprKind {
     },
 }
 
+#[derive(Debug, Hash, PartialEq, Eq)]
+pub enum QualifierKey {
+    String(String),
+    Layout,
```
kvark, line 140: I'm curious why the layout is special here?

JCapucho: `Layout` applies both to `std140` and `std430`, so you can have
```glsl
layout(std140, std430, std140)
```
and you'll get only the final `std140` (this is horrible but that is how glsl defines it)

kvark: maybe adding a short comment about this would help :)

kvark (approved):
nice!

PR #1715, opened 2022-02-03
glsl-in: texelFetch accept multisampled textures

JCapucho:


src/front/glsl/builtins.rs
```diff
@@ -445,12 +445,19 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
             // bit 1 - array
-            // bit 2 - dim part 2 - 3D
+            // bit 2 - multisample
+            // bit 3 - dim part 2 - 3D
             //
-            // 0b100 is the latest since 3D arrayed images aren't allowed
-            for bits in 0..(0b101) {
-                let dim = bits & 0b1 | (bits & 0b100) >> 1;
+            // 0b1000 is the latest since 3D arrayed images aren't allowed
+            for bits in 0..(0b1001) {
```
kvark, line 452: nit: in order to align with the comment, we could do `0..=0b1000`

src/front/glsl/builtins.rs
```diff
@@ -445,12 +445,19 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
             // bit 0 - dim part 1 - 1D/2D
             // bit 1 - array
-            // bit 2 - dim part 2 - 3D
+            // bit 2 - multisample
+            // bit 3 - dim part 2 - 3D
             //
-            // 0b100 is the latest since 3D arrayed images aren't allowed
-            for bits in 0..(0b101) {
-                let dim = bits & 0b1 | (bits & 0b100) >> 1;
+            // 0b1000 is the latest since 3D arrayed images aren't allowed
```
kvark, line 451: now that we have multisampling, either the comment needs to be fixed, or the logic, e.g.
> 0b1000 is the latest since 3D images aren't allowed to be multisampled or arrayed

src/front/glsl/builtins.rs
```diff
@@ -445,12 +445,19 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
             // bits layout
             // bit 0 - dim part 1 - 1D/2D
             // bit 1 - array
-            // bit 2 - dim part 2 - 3D
+            // bit 2 - multisample
+            // bit 3 - dim part 2 - 3D
```
kvark, line 449: actually, why aren't we doing `bit 0-1 - dimension 1D/2D/3D` ?

src/front/glsl/builtins.rs
```diff
@@ -445,12 +445,19 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
             //
-            // 0b100 is the latest since 3D arrayed images aren't allowed
-            for bits in 0..(0b101) {
-                let dim = bits & 0b1 | (bits & 0b100) >> 1;
+            // 0b1000 is the latest since 3D arrayed images aren't allowed
+            for bits in 0..(0b1001) {
+                let dim = bits & 0b1 | (bits & 0b1000) >> 2;
                 let arrayed = bits & 0b10 == 0b10;
+                let multi = bits & 0b100 == 0b100;
```
kvark, line 455: it's easier to just write `bits & something != 0`

src/front/glsl/builtins.rs
JCapucho: Because you have to iterate less times and have less checks, but honestly this system doesn't work out as well as I initially thought and I'm tempted to just replace it with nested loops with proper types and free standing functions to reduce the right drift.

src/front/glsl/builtins.rs
JCapucho: it's was just keeping the same style as the other bit checks

src/front/glsl/builtins.rs
kvark: I don't think you need to iterate any less. In either case there are 2 bits for the dimension. I'm just saying let's put them together and treat as a number between 0 and 4 instead of 2 separate bits.

kvark (approved):

PR #1719, opened 2022-02-06
glsl-in: Texture builtins fixes

JCapucho:
This one grew out of hand quick. Initially it was just replacing the bit
loops with a function driving the declarations to make the code sharing
better and support different scalar kinds.

Now it includes a lot of fixes:
- `textureSize` now also returns the array layers as the last component
  on the return (only for arrayed textures)
- `textureSize` now supports multisampled textures
- the `texture` family of functions now consumes a `vec3` coordinate
  vector for all 1D shadow textures
- Shadow textures can use the bias version of `texture` functions
  (temporarily disabled since naga doesn't support it)
- 3D textures can be used in `textureProj`
- `sampler2DArrayShadow` can't be used in `textureLod` or in `texture` with bias
- Cube textures cannot be used with offsets

Updates the tests to cover all functions


src/front/glsl/builtins.rs
```diff
@@ -147,337 +147,219 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
-                        ..
-                    }
-                );
-                let lod = matches!(
-                    builtin,
-                    MacroCall::Texture {
-                        level_type: TextureLevelType::Lod,
-                        ..
+                    // Parse out the variant settings.
+                    let grad = matches!(level_type, TextureLevelType::Grad);
```
kvark, line 191: would it be easier to just compare for equality instead of using macros?

src/front/glsl/builtins.rs
```diff
@@ -2099,3 +2039,71 @@ pub fn sampled_to_depth(
...
+/// f(ScalarKind, ImageDimension, arrayed, multi, shadow)
+/// ```
+///
+/// The generation of the shadow component can be controlled with `generate_shadow`,
+/// if it's `true` then shadow will be passed as `true` and `false`, otherwise only\
+/// `false` is passed.
+///
+/// Likewise `generated_multi` controls the generation of the multi component.
+fn texture_args_generator(
+    generate_multi: bool,
```
kvark, line 2056: boolean arguments should generally not be used
did you consider bitflags instead?

tests/in/glsl/samplers.frag
```diff
@@ -89,21 +101,26 @@ void testTex1DArray(in vec2 coord) {
     c = textureLodOffset(sampler1DArray(tex1DArray, samp), coord, 3.0, 5);
     c = textureOffset(sampler1DArray(tex1DArray, samp), coord, 5);
     c = textureOffset(sampler1DArray(tex1DArray, samp), coord, 5, 2.0);
+    c = texelFetch(sampler1DArray(tex1DArray, samp), ivec2(coord), 3);
+    // c = texelFetchOffset(sampler1DArray(tex1DArray, samp), ivec2(coord), 3, 5);
```
kvark, line 105: what's happening here?

kvark (commented):
Very nice work!
A few minor comments from me here

src/front/glsl/builtins.rs
JCapucho: Oh yeah, I just modified the old code which tested on the builtin, but now it can be definitely a equality

PR #1723, opened 2022-02-10
glsl-in: support for storage image types

### LINKED ISSUE #1456: [glsl-in] Support image variables
Currently, attempting to compile this shader:

```glsl
#version 450 core

layout(r32ui, binding = 0) writeonly uniform uimage2D img;

void main() {}
```
produces these errors:
```
error: Unknown layout qualifier: r32ui
  ┌─ image.comp:3:8
  │
3 │ layout(r32ui, binding = 0) writeonly uniform uimage2D img;
  │        ^^^^^

error: Not implemented: variable qualifier
  ┌─ image.comp:3:46
  │
3 │ layout(r32ui, binding = 0) writeonly uniform uimage2D img;
  │                                              ^^^^^^^^
```

JCapucho:
Adds support for the `gimageDim` family of types and the `imageSize`, `imageLoad` and `imageStore` functions.

# Missing

- [x] ~~support for image types as function arguments~~ (impossible see https://github.com/KhronosGroup/GLSL/issues/57, `writeonly` images should work but see the following points)
- [x] tests

# Unresolved

- glsl supports multisampled storage images, naga doesn't, how do we handle it?
- glsl allows for `writeonly` storage images to not have their format specified, naga doesn't like it, once again how do we handle it?
- there are some texture formats to which I don't know the proper naga equivalent
- glsl allows for cube images which naga doesn't allow in `imageLoad/Store`, how do we handle it?

Closes #1456


kvark:
> glsl supports multisampled storage images, naga doesn't, how do we handle it

let's support it in the IR, but have a capability in the validator to check for it. Relevant SPIR-V capability - https://www.khronos.org/registry/vulkan/specs/1.2/html/chap31.html#features-shaderStorageImageMultisample

> glsl allows for writeonly storage images to not have their format specified, naga doesn't like it, once again how do we handle it?

Not sure what to do with it. I suggest just not supporting this in Naga, at least for now.
SPIR-V has capabilities for either [reads](https://www.khronos.org/registry/vulkan/specs/1.2/html/chap31.html#features-shaderStorageImageReadWithoutFormat) or [writes](https://www.khronos.org/registry/vulkan/specs/1.2/html/chap31.html#features-shaderStorageImageWriteWithoutFormat) to be unspecified with formats, but that would complicate our IR and backend logic possibly unnecessarily.

> there are some texture formats to which I don't know the proper naga equivalent

We'll probably just need to add them.

> glsl allows for cube images which naga doesn't allow in imageLoad/Store, how do we handle it?

This doesn't appear useful in modern APIs, since one would just create a 2D view instead. It looks like a GLSL artifact, so we shouldn't try supporting it.

src/front/glsl/builtins.rs
```diff
@@ -384,12 +366,131 @@ pub fn inject_builtin(declaration: &mut FunctionDeclaration, module: &mut Module
...
         }
+        "imageSize" => {
+            let f = |kind, dim, arrayed, _, _| {
+                let image = TypeInner::Image {
+                    dim,
+                    arrayed,
+                    class: ImageClass::Storage {
+                        // dummy value
+                        format: match kind {
+                            Sk::Sint => crate::StorageFormat::R16Sint,
```
kvark, line 383: lets do something like
```rust
impl crate::ScalarKind {
  fn dummy_storage_format(&self) -> crate::StorageFormat {..}
}
```

src/front/glsl/parser/types.rs
```diff
@@ -326,3 +330,54 @@ impl<'source> ParsingContext<'source> {
         })
     }
 }
+
+fn map_image_format(word: &str) -> Option<crate::StorageFormat> {
+    let format = match word {
+        // float-image-format-qualifier:
+        "rgba32f" => crate::StorageFormat::Rgba32Float,
```
kvark, line 337: perhaps doing `use crate::StorageFormat as Sf;` at the start of the function would help?

kvark (approved):

JCapucho:
I've applied your suggestions, I'll merge this as an initial implementation and work on the missing bits that require some IR changes later when I have more time.

PR #1727, opened 2022-02-16
[hlsl-out] Remove `MipLevelCoordinate` type.

jimblandy:
Validation insures that `ImageLoad` expressions always have an `index` operand when the image's type requires it, so there is no need for `MipLevelCoordinate::Zero`. This means that `Option<Handle<Expression>>` serves adequately for this case.


kvark (approved):
Looks reasonable

PR #1728, opened 2022-02-16
Introduce TexelSelector type

jimblandy:
I'm curious whether folks see this as an improvement. It means more detailed matches in some places, but it also means we don't have to reach into the types to figure out what that `index` means, and I think it's a cleaner IR.

(Consider the last commit only; GitHub won't let me exclude the other PR's commit from this one, but it depends on it.)

### DIFF
--- a/src/valid/expression.rs
+++ b/src/valid/expression.rs
@@ -75,11 +75,11 @@ pub enum ExpressionError {
     InvalidDerivative,
     #[error("Image array index parameter is misplaced")]
     InvalidImageArrayIndex,
-    #[error("Image other index parameter is misplaced")]
+    #[error("Inappropriate sample or level-of-detail index for texel access")]
     InvalidImageOtherIndex,
     #[error("Image array index type of {0:?} is not an integer scalar")]
     InvalidImageArrayIndexType(Handle<crate::Expression>),
-    #[error("Image other index type of {0:?} is not an integer scalar")]
+    #[error("Image sample or level-of-detail index's type of {0:?} is not an integer scalar")]
     InvalidImageOtherIndexType(Handle<crate::Expression>),
     #[error("Image coordinate type of {1:?} does not match dimension {0:?}")]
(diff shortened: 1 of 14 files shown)


kvark:
I think I see where you are coming from. This is meant to improve the IR clarity.
There is one problem though. Mipmap levels and MSAA samples aren't directly related. Therefore, they shouldn't be switched in the same enum.
What we have right now isn't the best either, since the IR doesn't reflect clearly what this field is.

How do you feel about the following proposal: instead of changing `Option<A | B>` into `enum { A | B | () }`, how about we just split it into the clear `Option<A>` with `Option<B>`. This way, each of the options have clear semantics. The backends can still merge them together if needed, but the IR is clear.

jimblandy:
That was a great suggestion, things look nice now.

jimblandy:
Rebased on recent GLSL changes.

kvark:
I believe https://github.com/gfx-rs/naga/pull/1730#discussion_r808554896 applies here

jimblandy:
> I believe [#1730 (comment)](https://github.com/gfx-rs/naga/pull/1730#discussion_r808554896) applies here

I'm not sure exactly what you had in mind - there's no `is_multisampled` helper at present. But this version adds `is_multisampled` and `is_mipmapped` helpers to `ImageClass`, and tries to use them in a few cases. Is this better?

src/valid/expression.rs
```diff
@@ -560,16 +561,9 @@ impl super::Validator {
                                 ))
                             }
                         };
-                        let needs_index = match class {
-                            crate::ImageClass::Storage { .. } => false,
-                            _ => true,
```
kvark, line 565: we appear to be losing this check, aren't we?
i.e. a situation where either the sample index or the level is required but not provided.
The old code would detect this, but the new code doesn't care.

kvark (approved):

PR #1729, opened 2022-02-16
Benchmarking

kvark:
Integrates Criterion and also adds support for bincode-encoded IR form. The purpose here is to see the upper bound for shader loading performance. Today, bincode IR deserialization is roughly 7x faster than WGSL parsing.


kvark:
Results from my machine:
```
Gnuplot not found, using plotters backend
front/bin               time:   [143.70 us 146.65 us 149.74 us]                      
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
Benchmarking front/wgsl: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.7s, enable flat sampling, or reduce sample count to 60.
front/wgsl              time:   [1.1065 ms 1.1268 ms 1.1453 ms]                        
front/spv               time:   [56.855 us 58.214 us 59.836 us]                      
Found 14 outliers among 100 measurements (14.00%)
  5 (5.00%) high mild
  9 (9.00%) high severe
front/glsl              time:   [3.7642 ms 3.8215 ms 3.8864 ms]                        
Found 11 outliers among 100 measurements (11.00%)
  5 (5.00%) high mild
  6 (6.00%) high severe

valid/safe              time:   [122.50 us 123.12 us 123.97 us]                       
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high severe
valid/unsafe            time:   [86.239 us 86.530 us 86.842 us]                         
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

back/wgsl               time:   [637.90 us 640.21 us 642.65 us]                      
back/spv                time:   [402.15 us 403.86 us 405.86 us]                     
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe
back/spv-separate       time:   [323.96 us 324.86 us 325.73 us]                              
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
back/msl                time:   [895.71 us 900.71 us 906.08 us]                     
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
back/hlsl               time:   [774.44 us 778.98 us 784.89 us]                      
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
back/glsl-separate      time:   [930.78 us 932.80 us 934.88 us]                               
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe
```

benches/criterion.rs
```diff
@@ -0,0 +1,271 @@
...
+            }
+        });
+    });
+    #[cfg(feature = "wgsl-in")]
+    group.bench_function("wgsl", |b| {
+        let inputs = gather_inputs("tests/in", "wgsl");
+        let mut parser = naga::front::wgsl::Parser::new();
+        b.iter(move || {
+            for input in inputs.iter() {
+                let string = std::str::from_utf8(input).unwrap();
```
jimblandy, line 66: These `from_utf8` calls need to scan the input string to validate the UTF-8. Could we do this outside the `b.iter` call?

benches/criterion.rs
```diff
@@ -0,0 +1,271 @@
...
+        .map(|input| {
+            let string = std::str::from_utf8(input).unwrap();
+            parser.parse(string).unwrap()
+        })
+        .collect()
+}
+#[cfg(not(feature = "wgsl-in"))]
+fn gather_modules() -> Vec<naga::Module> {
+    Vec::new()
+}
```
jimblandy, line 111: It's a little confusing that when necessary features aren't available, we do still run benchmarks and get meaningless numbers. The loop over the modules is inside the `b.iter` call, not outside it. It would be clearer to somehow just not run benchmarks that require features we don't have.

benches/criterion.rs
```diff
@@ -0,0 +1,271 @@
...
+            naga::valid::Capabilities::empty(),
+        );
+        let input_modules = gather_modules();
+        input_modules
+            .into_iter()
+            .flat_map(|module| validator.validate(&module).ok().map(|info| (module, info)))
+            .collect::<Vec<_>>()
+    };
+    #[cfg(not(feature = "validate"))]
+    let inputs = Vec::<(naga::Module, naga::valid::ModuleInfo)>::new();
```
jimblandy, line 156: Again, we shouldn't run benchmarks on empty input lists because a feature was turned off.

src/front/wgsl/mod.rs
```diff
@@ -4343,10 +4343,15 @@ impl Parser {
         }
     }
 
-    pub fn parse(&mut self, source: &str) -> Result<crate::Module, ParseError> {
+    pub fn reset(&mut self) {
         self.scopes.clear();
+        self.module_scope_identifiers.clear();
```
jimblandy, line 4348: Whoa, nice catch

jimblandy (approved):
Looks good!

I think we should do better with feature handling; see comments.

benches/criterion.rs
```diff
@@ -0,0 +1,271 @@
...
+
+fn gather_inputs(folder: &str, extension: &str) -> Vec<Box<[u8]>> {
+    let mut list = Vec::new();
+    let read_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join(folder)
+        .read_dir()
+        .unwrap();
+    for file_entry in read_dir {
+        match file_entry {
+            Ok(entry) => match entry.path().extension() {
```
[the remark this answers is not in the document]
kvark: That would require some shenanigans on going from OSStr to str to be able to compare them

benches/criterion.rs
kvark: that's a bit of an edge case. We don't really care about the numbers when features aren't enabled. What I did was a minimal thing to make it compile for these cases

benches/criterion.rs
kvark: similar reasoning to the last one - this is the minimal change to compile

kvark:
Thanks for the review! Some of the notes are addressed, other left for follow-ups

PR #1736, opened 2022-02-18
glsl-in: inject samplerCubeArray builtins on use

### LINKED ISSUE #1622: [glsl-in] inject_builtin injects dependency on CUBE_TEXTURES_ARRAY feature
The `inject_builtin` function adds all the texture access functions in one batch, which means that any texture usage will cause all the cube map array function versions to be exposed as well. This is a problem because WebGL doesn't support them, but later feature detection code will see those function signatures and think the shader requires them.
 
The specific error produced will look like:
> Shader translation error for stage VERTEX: The selected version doesn't support CUBE_TEXTURES_ARRAY

JCapucho:
Arrayed cube images require a special capabilities in some backends, so
like how we already do with doubles, we now only inject them if a call
uses one as an argument.

This required some refractoring on the builtins handling of variations
but now the interface is cleaner and adding new variations (if needed)
should be easier.

Closes #1622

### DIFF
--- a/src/front/glsl/ast.rs
+++ b/src/front/glsl/ast.rs
@@ -62,14 +62,26 @@ pub struct Overload {
     pub void: bool,
 }
 
+bitflags::bitflags! {
+    /// Tracks the variations of the builtin already generated, this is needed because some
+    /// builtins overloads can't be generated unless explicitly used, since they might cause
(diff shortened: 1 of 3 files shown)


kvark (approved):
Would be good to have it tested too

PR #1740, opened 2022-02-20
[msl-out] Write 'using metal::uint' at the top, to work around bug.

jimblandy:
This lets us remove some special cases where `uint` must be fully qualified, even though other similar types are not.

This will conflict with the texture access bounds check branch (#1730), but it affects every single file in `tests/out/msl`, so combining it with that would flood the patch with noise that has nothing to do with bounds checks. So we can land the two separately, and I'll take care of whatever merging needs to be done. Whichever order you prefer is fine, but it's a lot less work to land this one second.

### DIFF
--- a/src/back/msl/keywords.rs
+++ b/src/back/msl/keywords.rs
@@ -18,6 +18,7 @@ pub const RESERVED: &[&str] = &[
     "bool",
     "char",
     "int",
+    "uint",
     "long",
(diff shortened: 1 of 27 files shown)


cwfitzgerald (approved):
Love it, LGTM

jimblandy:
Thanks very much!

PR #1742, opened 2022-02-22
glsl-out: texture function fixes

JCapucho:
- samplerCubeShadow requires vec3 gradients
- Bias must be written after offset
- sampler1D hack on GLES requires vec2 gradients
- textureSize doesn't accept a lod argument for MS samplers
- sampler1DArray hack on GLES requires a y component on imageLoad

Also adds some tests in wgsl

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2192,7 +2192,11 @@ impl<'a, W: Write> Writer<'a, W> {
                     // Zero needs level set to 0
                     crate::SampleLevel::Zero => {
                         if workaround_lod_array_shadow_as_grad {
-                            write!(self.out, ", vec2(0,0), vec2(0,0)")?;
+                            let vec_dim = match dim {
+                                crate::ImageDimension::Cube => 3,
+                                _ => 2,
+                            };
+                            write!(self.out, ", vec{}(0.0), vec{}(0.0)", vec_dim, vec_dim)?;
                         } else if gather.is_none() {
                             write!(self.out, ", 0.0")?;
                         }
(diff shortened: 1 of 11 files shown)


JCapucho:
Add to also patch spv-out since they suffered the same problem of using lod for `ImageSize`s on multisampled images

src/back/glsl/mod.rs
```diff
@@ -2192,7 +2192,12 @@ impl<'a, W: Write> Writer<'a, W> {
                     // Zero needs level set to 0
                     crate::SampleLevel::Zero => {
                         if workaround_lod_array_shadow_as_grad {
-                            write!(self.out, ", vec2(0,0), vec2(0,0)")?;
+                            // samplerCubeShadow requires vec3 gradients
+                            if dim == crate::ImageDimension::Cube {
```
kvark, line 2196: nit: could move the condition out of the `write`:
```rust
let vec_dim = match dim {
  crate::ImageDimension::Cube => 3,
  _ => 2,
};
write!(self.out, ", vec{}(0), vec{}(0)", vec_dim, vec_dim)?;
```

src/back/glsl/mod.rs
```diff
@@ -2333,14 +2356,22 @@ impl<'a, W: Write> Writer<'a, W> {
                         write!(self.out, ")",)?;
                     }
                     crate::ImageQuery::NumLayers => {
-                        let fun_name = match class {
-                            ImageClass::Sampled { .. } | ImageClass::Depth { .. } => "textureSize",
-                            ImageClass::Storage { .. } => "imageSize",
+                        let (fun_name, lod) = match class {
```
kvark, line 2359: we should better use the existing helpers `class.is_multisampled()` for consistency

kvark (approved):

JCapucho:
@kvark thanks for the review, should I take your approval as a merge once fixed?

kvark:
Yes, that was the idea. If I'm not sure that merge is OK after fixes, I'm not making it "Approved" :)

PR #1746, opened 2022-02-24
glsl-in: Add support for texelFetchOffset

JCapucho:


src/front/glsl/builtins.rs
```diff
@@ -322,7 +323,7 @@ pub fn inject_builtin(
                 let dim_value = image_dims_to_coords_size(dim);
                 let coordinates = make_coords_arg(dim_value + arrayed as usize, Sk::Sint);
 
-                let args = vec![
+                let mut args = vec![
```
kvark, line 326: perhaps we should do `ArrayVec` instead?

kvark (commented):
Are these clippy warnings new?

kvark:
We should probably run Clippy on MSRV only, so that it's predictable.

src/front/glsl/builtins.rs
JCapucho: I would fine with doing this in another PR but I don't think mixing unrelated changes in the same PR would be a good idea.

JCapucho:
@kvark sorry for the late response, was out of town for a while, jim seems to have already fixed the clippy lint but it might still be a good idea to change the pipeline to use the MSRV or the latest stable (and maybe have another optional pipeline for checking on nightly), I can implement it if you want, just wanted to hear your thoughts on the matter.

PR #1749, opened 2022-02-26
[wgsl-in] Remove unused field Token::Number::width.

jimblandy:


jimblandy:
I landed the Clippy fix separately in #1752. This is rebased on master, and has no further excuses.

kvark (approved):
Great!

PR #1750, opened 2022-02-26
[wgsl-in] consume_number: Check uint_suffix flag at top of match.

jimblandy:
This makes it unnecessary to check that field in every other pattern in the match.


jimblandy:
The new Clippy complains about master as well. Added a commit to sort that out.

jimblandy:
Rebased on master, after merging #1752.

kvark (approved):

PR #1752, opened 2022-02-27
[wgsl-in] Use unwrap_or to placate Clippy.

jimblandy:
I had this minor fix as part of #1750, but since it's causing CI failures for everything, I'm going to land it separately.


kvark:
I'm surprised this makes Clippy happy, since previously it pretty much forced us to use `unwrap_or_else` on the basis of the None path requiring some actual logic (even if it's just a call to `.len()`).

PR #1756, opened 2022-03-02
Split logical and bitwise not

### LINKED ISSUE #1252: [msl-out] Bitwise negation is broken for Metal
```wgsl
fn main() {
    let foo = 42u;
    let bar = ~foo;
}
```

... produces:

```metal
// language: metal1.1
#include <metal_stdlib>
#include <simd/simd.h>


kernel void main1(
) {
    uint bar = !42u;
}
```

This is incorrect since it evaluates to boolean not 42 which then evaluates to 0.

vincentisambart:
Fixes #1252

Hi, I wanted to play with Naga and #1252 seems it might be easy to fix, and I ended up with something that looked fine so here it is.

My goal was to fix the Metal issue, but for that I had to change how WGSL unary operators were handled. I guess WGSL probably originally represented bitwise and logical not the same way (like in Rust where both are `!`).

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -315,10 +315,8 @@ impl<'w> BlockContext<'w> {
                             return Err(Error::FeatureNotImplemented("negation"));
                         }
                     },
-                    crate::UnaryOperator::Not => match expr_ty_inner.scalar_kind() {
-                        Some(crate::ScalarKind::Bool) => spirv::Op::LogicalNot,
-                        _ => spirv::Op::Not,
-                    },
+                    crate::UnaryOperator::LogicalNot => spirv::Op::LogicalNot,
+                    crate::UnaryOperator::BitwiseNot => spirv::Op::Not,
                 };
 
                 block
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -2500,25 +2500,19 @@ impl Parser {
         self.push_scope(Scope::UnaryExpr, lexer);
         //TODO: refactor this to avoid backing up
         let expr = match lexer.peek().0 {
-            Token::Operation('-') => {
+            token @ Token::Operation('-')
+            | token @ Token::Operation('!')
+            | token @ Token::Operation('~') => {
+                let op = match token {
+                    Token::Operation('-') => crate::UnaryOperator::Negate,
+                    Token::Operation('!') => crate::UnaryOperator::LogicalNot,
+                    Token::Operation('~') => crate::UnaryOperator::BitwiseNot,
+                    _ => unreachable!(),
+                };
                 let _ = lexer.next();
                 let unloaded_expr = self.parse_unary_expression(lexer, ctx.reborrow())?;
                 let expr = ctx.apply_load_rule(unloaded_expr);
-                let expr = crate::Expression::Unary {
-                    op: crate::UnaryOperator::Negate,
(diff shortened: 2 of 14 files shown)


src/back/spv/block.rs
```diff
@@ -315,10 +315,8 @@ impl<'w> BlockContext<'w> {
                             return Err(Error::FeatureNotImplemented("negation"));
                         }
                     },
-                    crate::UnaryOperator::Not => match expr_ty_inner.scalar_kind() {
-                        Some(crate::ScalarKind::Bool) => spirv::Op::LogicalNot,
-                        _ => spirv::Op::Not,
-                    },
+                    crate::UnaryOperator::LogicalNot => spirv::Op::LogicalNot,
+                    crate::UnaryOperator::BitwiseNot => spirv::Op::Not,
```
vincentisambart, line 319: For `Negate` types are checked so I was not sure if I should also do it here, but I went the easy route.

src/front/wgsl/mod.rs
```diff
@@ -2500,25 +2500,19 @@ impl Parser {
         self.push_scope(Scope::UnaryExpr, lexer);
         //TODO: refactor this to avoid backing up
         let expr = match lexer.peek().0 {
-            Token::Operation('-') => {
```
vincentisambart, line 2503: The code for `'-'` was exactly the same as the one for `'!'` and `'~'` except for `op` so I thought refactoring a bit would make code easier to read.

src/valid/expression.rs
```diff
@@ -633,11 +633,11 @@ impl super::Validator {
                 use crate::UnaryOperator as Uo;
                 let inner = resolver.resolve(expr)?;
                 match (op, inner.scalar_kind()) {
-                    (_, Some(Sk::Sint))
-                    | (_, Some(Sk::Bool))
-                    //TODO: restrict Negate for bools?
```
vincentisambart, line 638: I did not see any reason to properly check all types so that TODO should be fixed by this change.

kvark:
Thank you for the PR!
Why are we splitting the logical from bitwise in IR? The reason they are represented via one operator is because it's not ambiguous: if the arguments are boolean, then it's logical, otherwise it's bitwise.

vincentisambart:
> Why are we splitting the logical from bitwise in IR? The reason they are represented via one operator is because it's not ambiguous: if the arguments are boolean, then it's logical, otherwise it's bitwise.

It seemed that splitting made the code simpler and more easy to follow, not having in every backend to look at the type when you need to output the correct operator, especially as all seem to have a different operator for bitwise and logical.

But if you want I can go the other route, not splitting and looking at the type in the HLSL and MSL backends. (The issue was about MSL but it seems the HLSL had the same bug).

vincentisambart:
Another perk when splitting: if a user used the incorrect operator in their WGSL you can now easily detect that.

jimblandy:
There are a lot of cases where the type is needed to decide what the back end should emit for a given IR expression. You can always slide information back and forth between types and expressions, so there's a subjective aspect to the question. My feeling about this case is that depending on the type is the better choice for the IR.

kvark:
@vincentisambart please refactor the PR without IR changes.
The path we took can be discussed and argued freely, but fixing #1252 should come first.

vincentisambart:
As the whole premise changed, I created a new PR for it: #1760

PR #1757, opened 2022-03-02
glsl-in: matrix operations fixes

### LINKED ISSUE #1408: [glsl-in] matrix division
Glsl allows the following code
```glsl
mat3(1) /mat3(1);
mat3(1) / 1;
```
Currently this results in a validation error since matrix division isn't supported in the IR.

This will probably require decomposing the matrix applying the division to each vector and compose the matrix again but I'm waiting for @kvark's confirmation.

JCapucho:
Contains multiple fixes for matrix related binary operations:
- Equality and inequality operations
- Matrix by matrix division
- Matrix by scalar division/addition/subtraction

Closes #1408

All commits can be reviewed and tested individually


src/front/glsl/context.rs
```diff
@@ -579,8 +579,165 @@ impl Context {
...
+                            rows: right_rows,
+                            width: right_width,
+                        },
+                    ) => {
+                        // Check that the two arguments have the same dimensions
+                        if left_columns != right_columns
+                            || left_rows != right_rows
+                            || left_width != right_width
+                        {
+                            parser.errors.push(Error {
```
kvark, line 599: is this check necessary? general validation would produce a very similar error, I think

kvark (approved):

src/front/glsl/context.rs
JCapucho: Checking the width and the rows is not necessary, but the columns need to be checked since otherwise code that uses matrices with different columns would either fail on validation or truncate depending on the order.

PR #1760, opened 2022-03-04
Not operator depending on expression type

### LINKED ISSUE #1252: [msl-out] Bitwise negation is broken for Metal

vincentisambart:
Fixes #1252

In HLSL and MSL, when outputting a not operator, look at the expression type to choose the correct operator.

Mostly took code from the GLSL backend and adapted it for the other backends.

### DIFF
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -1751,10 +1751,21 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                 }
             }
             Expression::Unary { op, expr } => {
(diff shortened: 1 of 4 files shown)


kvark (approved):
Thank you!
Note that the backend doesn't need to validate. So it could just check for float in an "if/else" fashion, assuming it's correct.

PR #1763, opened 2022-03-06
msl: qualify read-only storage with const

### LINKED ISSUE #1761: read-only storage binding in VS results in scary warning in MSL
I have a binding like this in a WGSL shader

`@group(0) @binding(1) var<storage,read> particle_buffer: ParticleBufferReadonly;`

when translated to MSL, it becomes this parameter to vs_main:

`    , device ParticleBufferReadonly& particle_buffer [[buffer(1)]]`

which then results in this scary warning:

```
    program_source:277:22: warning: writable resources in non-void vertex function
    vertex vs_mainOutput vs_main(
                         ^
    program_source:282:34: note: writable buffer defined here
    , device ParticleBufferReadonly& particle_buffer [[buffer(1)]]
                                     ^
```

This seems related to this PR https://github.com/gfx-rs/naga/pull/1711

If I revert the part of the PR that changes the function `to_msl_name` back to:
```
            Self::Storage { access } if access.contains(crate::StorageAccess::STORE) => {
                Some("device")
            }
            Self::Storage { .. } => Some("constant"),
```

then the parameter to the vs_main function becomes:

``    , constant ParticleBufferReadonly& particle_buffer [[buffer(1)]]``

and the warning disappears.

I'm not sure if the warning has an actual negative effect or not, it doesn't seem to yet but I'm still fixing other problems.  I only noticed this because the shader was failing MSL compilation for a different reason.

kvark:
Fixes #1761

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -238,13 +238,12 @@ impl<'a> TypedGlobalVariable<'a> {
 
         let (space, access, reference) = match var.space.to_msl_name() {
             Some(space) if self.reference => {
-                let access = match var.space {
-                    crate::AddressSpace::Private | crate::AddressSpace::WorkGroup
-                        if !self.usage.contains(valid::GlobalUse::WRITE) =>
-                    {
-                        "const"
-                    }
-                    _ => "",
+                let access = if var.space.needs_access_qualifier()
+                    && !self.usage.contains(valid::GlobalUse::WRITE)
+                {
+                    "const"
+                } else {
+                    ""
                 };
                 (space, access, "&")
             }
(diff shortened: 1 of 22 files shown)


src/back/msl/writer.rs
```diff
@@ -401,13 +400,25 @@ impl crate::AddressSpace {
...
+            | Self::Storage { .. }
+            | Self::Private
+            | Self::WorkGroup
+            | Self::PushConstant
+            | Self::Handle => true,
+            Self::Function => false,
+        }
+    }
+
+    /// Returns true if the address space may need a "const" qualifier.
```
jimblandy, line 413: I'm in the dark as to why, say, `Uniform` and `PushConstant` don't need `const`. Is it because their MSL address space alone implies that they're not writable, so `const` is unnecessary, unlike `Storage`, `Private`, and `WorkGroup`, which could be writable?

jimblandy (commented):
Looks good to me.

It's helpful to separate out the changes to the tests into a separate commit, so that the effects of the changes to the code are completely apparent. I'm more likely to miss something when (say) the `const` qualifiers are mixed in with a lot of renames and code movement.

src/back/msl/writer.rs
kvark: Added a comment here, and made the MSL output actually contain push constants for the test, in #1767

PR #1768, opened 2022-03-09
[glsl-out] Write constant sized array type for uniform

### LINKED ISSUE #1747: Latest `wgpu` Shadow Example Causes GLSL backend to error
With naga and wgpu co-checked out, run 

```
cargo run --bin naga ..\wgpu\wgpu\examples\shadow\shader.wgsl --entry-point fs_main_without_storage ..\wgpu\wgpu\examples\shadow\shader.frag
```

and you get the error `Non-struct type of a buffer` from https://github.com/gfx-rs/naga/blob/master/src/back/glsl/mod.rs#L972-L974

The shadow example in naga tests has drifted a bit from the one in wgpu, it probably should be updated. In particular it is missing the fs_main_without_storage entry point which is failing, which is absent from naga tests.

hatoo:
resolves #1747

Writing `wgpu` Shadow Example as GLSL failed because `glsl-out` doesn't allow constant sized array type for uniform.

Hi, I'm interested in this project and I believe this PR resolve an issue. But I'm new to this project, sorry if the PR is nonsense.

This PR
* Fixes `glsl-out` to allow constant sized array type for uniform global.
* Update test case from `wgpu`'s shadow example.


kvark (approved):

tests/in/shadow.wgsl
```diff
@@ -1,61 +1,115 @@
...
+    world: mat4x4<f32>;
+    color: vec4<f32>;
+};
+
+@group(1)
+@binding(0)
+var<uniform> u_entity: Entity;
+
+@stage(vertex)
+fn vs_bake(@location(0) position: vec4<i32>) -> @builtin(position) vec4<f32> {
```
kvark, line 20: I don't think this entry point useful for testing. Perhaps, we could remove it?

kvark (commented):
Thank you for the fix!

tests/in/shadow.wgsl
hatoo: Make sense. I removed the entry point.

btw, I tried to remove `tests/out/glsl/*` and regenerate again. 
But `tests/out/glsl/quad.main.Fragment.glsl` and `tests/out/glsl/quad.main.Vertex.glsl` doesn't regenerated.
Maybe those are renamed to `vert_main` and `frag_main` but forgot to delete?

kvark: yes that sounds right

PR #1773, opened 2022-03-12
[wgsl] Uses commas to separate struct members instead of semicolons

Gordon-F:
Spec PR: https://github.com/gpuweb/gpuweb/pull/2656


kvark (approved):
Nice! One step closer to 🦀

PR #1779, opened 2022-03-17
glsl-in: Fix memory qualifiers being inverted

### LINKED ISSUE #1770: [glsl-in] writeonly images
This shader:
```glsl
#version 460 core

layout(rgba32f, binding = 0) writeonly uniform image2D tex;

void main() {
    imageStore(tex, ivec2(0, 0), vec4(0.0));
}
```
produces this error:
```
error: 'imageStore': image needs STORE access but only LOAD was provided
  ┌─ shader.comp:6:5
  │
6 │     imageStore(tex, ivec2(0, 0), vec4(0.0));
  │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

JCapucho:
Adds some documentation to better explain how the memory qualifier works
troughout the parser and some storage textures tests.

Closes #1770

### DIFF
--- a/src/front/glsl/parser/types.rs
+++ b/src/front/glsl/parser/types.rs
@@ -122,7 +122,7 @@ impl<'source> ParsingContext<'source> {
             | TokenValue::Shared
             | TokenValue::Buffer
             | TokenValue::Restrict
-            | TokenValue::StorageAccess(_)
+            | TokenValue::MemoryQualifier(_)
             | TokenValue::Layout => true,
             _ => false,
         })
@@ -219,11 +219,11 @@ impl<'source> ParsingContext<'source> {
 
                     qualifiers.precision = Some((p, token.meta));
                 }
-                TokenValue::StorageAccess(access) => {
+                TokenValue::MemoryQualifier(access) => {
                     let storage_access = qualifiers
                         .storage_acess
-                        .get_or_insert((crate::StorageAccess::empty(), Span::default()));
-                    if storage_access.0.contains(access) {
+                        .get_or_insert((crate::StorageAccess::all(), Span::default()));
+                    if !storage_access.0.contains(!access) {
                         parser.errors.push(Error {
                             kind: ErrorKind::SemanticError(
                                 "The same memory qualifier can only be used once".into(),
@@ -232,7 +232,7 @@ impl<'source> ParsingContext<'source> {
                         })
                     }
 
-                    storage_access.0 |= access;
+                    storage_access.0 &= access;
                     storage_access.1.subsume(token.meta);
                 }
                 TokenValue::Restrict => continue,
--- a/src/front/glsl/token.rs
+++ b/src/front/glsl/token.rs
@@ -38,7 +38,11 @@ pub enum TokenValue {
     Shared,
 
     Restrict,
-    StorageAccess(crate::StorageAccess),
+    /// A `glsl` memory qualifier such as `writeonly`
+    ///
+    /// The associated [`crate::StorageAccess`] is the access being allowed
+    /// (for example `writeonly` has an associated value of [`crate::StorageAccess::STORE`])
+    MemoryQualifier(crate::StorageAccess),
 
     Interpolation(Interpolation),
     Sampling(Sampling),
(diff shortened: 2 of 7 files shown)


src/front/glsl/token.rs
```diff
@@ -38,7 +38,11 @@ pub enum TokenValue {
     Shared,
 
     Restrict,
-    StorageAccess(crate::StorageAccess),
+    /// A `glsl` memory qualifier such as `writeonly`
+    ///
+    /// The associated [`crate::StorageAccess`] is the access being forbidden
+    /// (for example `writeonly` has an associated value of [`crate::StorageAccess::LOAD`])
+    MemoryQualifier(crate::StorageAccess),
```
kvark, line 45: let's use the naming instead of comments. This would look:
```rust
MemoryQualifier {
  forbid: crate::StorageAccess,
}
```
Another choice here, and I think a better one, is to consider `MemoryQualifier(allowed storage access mask)` semantics (i.e. "allow mask" instead of "forbid")

JCapucho: I do remember trying to use an allow mask and had some issues, but now I managed to implement it so I don't really remember why.

src/front/glsl/parser/types.rs
```diff
@@ -219,11 +219,11 @@ impl<'source> ParsingContext<'source> {
...
                     qualifiers.precision = Some((p, token.meta));
                 }
-                TokenValue::StorageAccess(access) => {
+                TokenValue::MemoryQualifier(access) => {
                     let storage_access = qualifiers
                         .storage_acess
-                        .get_or_insert((crate::StorageAccess::empty(), Span::default()));
-                    if storage_access.0.contains(access) {
+                        .get_or_insert((crate::StorageAccess::all(), Span::default()));
+                    if !storage_access.0.contains(!access) {
```
kvark, line 226: wow, that's a bit strange of a check we have here.
How about `if !access.contains(storage_access.0)` instead?

src/front/glsl/variables.rs
```diff
@@ -465,8 +465,8 @@ impl Parser {
             StorageQualifier::AddressSpace(mut space) => {
                 match space {
                     AddressSpace::Storage { ref mut access } => {
-                        if let Some((restricted_access, _)) = qualifiers.storage_acess.take() {
-                            access.remove(restricted_access);
+                        if let Some((allowed_access, _)) = qualifiers.storage_acess.take() {
+                            *access = allowed_access;
```
kvark, line 469: shouldn't this be `*access &= allowed_access;` instead?

src/front/glsl/variables.rs
```diff
@@ -480,10 +480,8 @@ impl Parser {
                                 mut format,
                             } = class
                             {
-                                if let Some((restricted_access, _)) =
-                                    qualifiers.storage_acess.take()
-                                {
-                                    access.remove(restricted_access);
+                                if let Some((allowed_access, _)) = qualifiers.storage_acess.take() {
+                                    access = allowed_access;
```
kvark, line 484: same here

src/front/glsl/parser/types.rs
JCapucho: Won't work because if you have for example a single `writeonly` qualifier, it will try to check if `STORE.contains(LOAD | STORE)` which will be false and the branch will be taken producing an error.

src/front/glsl/variables.rs
JCapucho: It's the same thing since the only way to change access which defaults to `all` is trough memory qualifiers

src/front/glsl/parser/types.rs
fintelia: The check should be equivalent to `storage_access.0 | access != StorageAccess::all()`, though I'll admit I don't fully understand how that condition is connected with a memory qualifier being repeated

JCapucho: This check is done to see if a memory qualifier was already used, as there are only two configurations `LOAD` or `STORE` and we `and` the storage access with the memory qualifier, the only way to check (without adding extra structures) is to check if the complement of the memory qualifier isn't present.

For example:
```
writeonly writeonly

Start:
- storage_access: all

First writeonly
- storage_access: all
- !STORE = LOAD
- storage_access.contains(LOAD) = true
- storage_access &= access <=> LOAD
 
Second writeonly
- storage_access: LOAD
- !STORE = LOAD
- storage_access.contains(LOAD) = false
- throw error
```

kvark (approved):

PR #1780, opened 2022-03-18
glsl-in: Allow multiple array specifiers

### LINKED ISSUE #1769: [glsl-in] 2D array support
This shader:

```glsl
#version 460 core

shared float h[2][2];

void main() {}
```

produces this error:
```
error: Expected Comma or Semicolon, found LeftBracket
  ┌─ shader.comp:3:18
  │
3 │ shared float h[2][2];
  │                  ^
```

JCapucho:
Closes #1769

### DIFF
--- a/src/front/glsl/parser/types.rs
+++ b/src/front/glsl/parser/types.rs
@@ -10,45 +10,76 @@ use crate::{
 };
 
 impl<'source> ParsingContext<'source> {
-    /// Parses an optional array_specifier returning `Ok(None)` if there is no
-    /// LeftBracket
+    /// Parses an optional array_specifier returning wether or not it's present
+    /// and modifying the type handle if it exists
     pub fn parse_array_specifier(
         &mut self,
         parser: &mut Parser,
-    ) -> Result<Option<(ArraySize, Span)>> {
-        if let Some(Token { mut meta, .. }) = self.bump_if(parser, TokenValue::LeftBracket) {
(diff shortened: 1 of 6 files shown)


src/front/glsl/parser/types.rs
```diff
@@ -10,45 +10,76 @@ use crate::{
...
                     },
                 },
-                span,
+                *span,
             );
-            let end_meta = self.expect(parser, TokenValue::RightBracket)?.meta;
-            meta.subsume(end_meta);
-            Ok(Some((ArraySize::Constant(constant), meta)))
+
+            Ok(true)
```
kvark, line 72: I don't understand why we are mutating `ty` and returning `Ok(bool)` if we could instead return something like `Result<Option<Handle<Type>>>`. Do you think this would be a feasible thing to change here?

JCapucho: We definitely could but I don't what the benefit would be since we would be mutating the type anyways.

kvark: Where would we be mutating the type? This would just return a new type.
Or do you mean that the caller would likely reassign the type anyway? That's OK, I think.

JCapucho: Yes the caller always reassigns the type

kvark: I think that would still be a better internal API.

PR #1781, opened 2022-03-19
[spv-out] fix wrong MatrixStride for matCx2 and mat2xR

teoxoy:
Matrices should have a `MatrixStride` of 8 for `matCx2` however the current implementation outputs a `MatrixStride` of 8 for `mat2xR`.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -871,12 +871,12 @@ impl Writer {
                             ref other => other,
                         };
                         if let crate::TypeInner::Matrix {
-                            columns,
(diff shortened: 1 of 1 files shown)


JCapucho (approved):
It does makes sense to me, but I would like either @kvark or @jimblandy  to also have a look since I'm not confident in my understanding of the `MatrixStride` decoration.

teoxoy:
There is some more relevant info here: https://github.com/gpuweb/gpuweb/issues/1258

kvark (approved):
Nicely spotted!

PR #1199, opened 2021-08-14
[hlsl-out] Fix reading from mat3

cwfitzgerald:
Fixes one part of the issues I'm having with rend3. This now matches the behavior of vulkan exactly.

PR #1784, opened 2022-03-20
[hlsl-out] fix matrix not being declared as transposed

teoxoy:
HLSL uses inverted col/row notation (relative to GLSL, WGSL, MSL).

i.e. `mat3x2` in WGSL = `float2x3` in HLSL

Reference: [HLSL Matrix Doc](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-matrix)

related to original change in #1199

### DIFF
--- a/src/back/hlsl/storage.rs
+++ b/src/back/hlsl/storage.rs
@@ -118,26 +118,23 @@ impl<W: fmt::Write> super::Writer<'_, W> {
                 rows,
                 width,
             } => {
-                // we declared the matrix as transposed in HLSL
-                let (rows, columns) = (columns, rows);
-
                 write!(
                     self.out,
                     "{}{}x{}(",
                     crate::ScalarKind::Float.to_hlsl_str(width)?,
-                    rows as u8,
                     columns as u8,
+                    rows as u8,
                 )?;
 
                 // Note: Matrices containing vec3s, due to padding, act like they contain vec4s.
-                let padded_columns = match columns {
+                let padded_rows = match rows {
                     crate::VectorSize::Tri => 4,
-                    columns => columns as u32,
+                    rows => rows as u32,
                 };
-                let row_stride = width as u32 * padded_columns;
-                let iter = (0..rows as u32).map(|i| {
+                let row_stride = width as u32 * padded_rows;
+                let iter = (0..columns as u32).map(|i| {
                     let ty_inner = crate::TypeInner::Vector {
-                        size: columns,
+                        size: rows,
                         kind: crate::ScalarKind::Float,
                         width,
                     };
@@ -256,9 +253,6 @@ impl<W: fmt::Write> super::Writer<'_, W> {
                 rows,
                 width,
             } => {
-                // we declared the matrix as transposed in HLSL
-                let (rows, columns) = (columns, rows);
-
                 // first, assign the value to a temporary
                 writeln!(self.out, "{}{{", level)?;
                 let depth = level.0 + 1;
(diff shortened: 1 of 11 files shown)


cwfitzgerald:
I'm confused, this PR appears to be making the hlsl declaration types be the same as wgsl, with column first and then row, but the comments and the PR say that it should be inverted.

teoxoy:
HLSL's matrix type is of the form `matRxC`

matrix type in GLSL/WGSL/MSL is of the form `matCxR`

So passing column first, row second, to HLSL will cause the matrix to be the transposed version of the original.

teoxoy:
Before this PR, trying to change the access test as in https://github.com/gfx-rs/naga/pull/1784/commits/0d920f508f83d0611e8fe501aeab64fc39e9815d and passing the resulting HLSL code trough DXC results in the error below

hlsl_fail

With the fix in this PR this no longer happens.

src/back/hlsl/storage.rs
```diff
@@ -118,6 +118,9 @@ impl<W: fmt::Write> super::Writer<'_, W> {
                 rows,
                 width,
             } => {
+                // we declared the matrix as transposed in HLSL
+                let (rows, columns) = (columns, rows);
```
cwfitzgerald, line 122: I think this is confusing, I'd much rather just flip the usages everywhere

cwfitzgerald (changes requested):
One comment, but otherwise looks okay

kvark:
Thank you for the PR!
We need to take a bit of extra thought to review this, since we specifically output all matrices in buffers as row-major.
Our HLSL matrix logic is a bit non-trivial. Maybe it's a good chance to try to document it thoroughly in the code (i.e. at the top of the module) and understand clearly.

teoxoy:
No problem.

Regarding documentation, there is already this nicely written note

https://github.com/gfx-rs/naga/blob/05f050fad485763bf7bf6e1ad4206007b52bc345/src/back/hlsl/mod.rs#L9-L21

is there anything else you'd like me to add?

kvark:
So to me this comment reads like the current code is already correct, isn't it?

teoxoy:
> Finally because all of our matrices are transposed, if you use `mat3x4`, it'll become `float4x3` in HLSL.

This line is not right and what this PR is solving. `mat3x4` will become a `mat4x3` (GLSL/WGSL/MSL notation) but will become a `float3x4` in HLSL (due to HLSL's matrix notation being inverted already).

I fixed this doc line in this PR as well.

This whole thing is quite confusing and I got confused as well while trying to fix it.

kvark:
> This line is not right and what this PR is solving. mat3x4 will become a mat4x3 (GLSL/WGSL/MSL notation)

Sorry, I'm not following this. Here is the GLSL code for matrix type:
```rust
            TypeInner::Matrix {
                columns,
                rows,
                width,
            } => write!(
                self.out,
                "{}mat{}x{}",
                glsl_scalar(crate::ScalarKind::Float, width)?.prefix,
                columns as u8,
                rows as u8
            )?,
```
Clearly it makes it so `mat3x4` in WGSL becomes `mat3x4` in GLSL. I.e. there is no difference in notation. The only difference is with HLSL.

teoxoy:
Sorry I think I made my explanation more confusing than it should have been.

See quote below with `()` added by me.

> Finally because all of our matrices are transposed, if you use `mat3x4` (3 columns, 4 rows), it'll become `float4x3` (3 columns, 4 rows -> issue is here, not transposed) in HLSL.

kvark (approved):
Thanks for your patience!
I launched a small sanity test in https://github.com/gfx-rs/naga/runs/5714948153?check_suite_focus=true to see if the current algorithm produces invalid HLSL, and indeed that's the case.

PR #1786, opened 2022-03-20
[hlsl-out] Insert padding between struct members

### LINKED ISSUE #1783: [hlsl-out] account for "vector-relaxed" std140 layout for constant buffers
HLSL uses "vector-relaxed" std140 layout for constant buffers, however the current implementation seems to assume std140 layout

The example below shows the issue.

Going from WGSL to HLSL trough naga translates the struct as if it had the same layout in HLSL.
When translating the same WGSL shader to SPIR-V via naga, the offsets are: 0, 8, 16.
However when translating the produced HLSL to SPIR-V via DXC, the offsets are: 0, 4, 12.

WGSL -> HLSL
wgsl_hlsl

WGSL -> SPIR-V
wgsl_spirv

HLSL -> SPIR-V
hlsl_spirv

### Solution

Insert padding (when necessary) for vector members that are in constant buffers.

### References

[HLSL Packing Rules for Constant Variables](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-packing-rules)
[DirectXShaderCompiler Buffer-Packing Wiki](https://github.com/microsoft/DirectXShaderCompiler/wiki/Buffer-Packing/4e59b6a713aa5b363ac57e87574bc1d13557bd0e)
[DirectXShaderCompiler Memory layout rules](https://github.com/Microsoft/DirectXShaderCompiler/blob/2322f53773fa6e1c449046e091fe18415ae65d57/docs/SPIR-V.rst#memory-layout-rules)

teoxoy:
fixes #1783

With this fix WGSL's `align` & `size` attributes and also spir-v's `Offset` decoration are respected since the offset in the IR is affected by those.

### DIFF
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -712,7 +712,20 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
         let struct_name = &self.names[&NameKey::Type(handle)];
         writeln!(self.out, "struct {} {{", struct_name)?;
 
+        let mut last_offset = 0;
         for (index, member) in members.iter().enumerate() {
+            if member.binding.is_none() && member.offset > last_offset {
+                // using int as padding should work as long as the backend
+                // doesn't support a type that's less than 4 bytes in size
+                // (Error::UnsupportedScalar catches this)
+                let padding = (member.offset - last_offset) / 4;
+                for i in 0..padding {
+                    writeln!(self.out, "{}int _pad{}_{};", back::INDENT, index, i)?;
+                }
+            }
+            let ty_inner = &module.types[member.ty].inner;
+            last_offset = member.offset + ty_inner.size(&module.constants);
+
             // The indentation is only for readability
             write!(self.out, "{}", back::INDENT)?;
 


teoxoy:
I noticed `half` is supported by the hlsl backend but the issue with using `half` as the padding type is that its size can be either 2 or 4 bytes. How should we deal with this?

kvark:
> but the issue with using half as the padding type is that its size can be either 2 or 4 bytes

Is that not defined in the spec? we shouldn't expose f16 (or half) until we figure out the details on how it maps to HLSL.
Note that fp16 extension has still not landed on WebGPU - https://github.com/gpuweb/gpuweb/pull/2647

src/back/hlsl/writer.rs
```diff
@@ -712,7 +712,20 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
 
+        let mut last_offset = 0;
         for (index, member) in members.iter().enumerate() {
+            if member.binding.is_none() && member.offset > last_offset {
+                // using int as padding should work as long as the backend
+                // doesn't support a type that's less than 4 bytes in size
+                // (Error::UnsupportedScalar catches this)
+                let padding = (member.offset - last_offset) / 4;
+                for i in 0..padding {
+                    writeln!(self.out, "{}int _pad{}_{};", back::INDENT, index, i)?;
```
kvark, line 723: If we add this padding, all of the accesses into the structure would be broken. I.e. we need to handle indexing, know where padding was inserted, and modify the indices accordingly.

kvark (changes requested):
Thank you for the PR!

src/back/hlsl/writer.rs
teoxoy: The [constructor function](https://github.com/gfx-rs/naga/blob/7d0644589755efefd7c6c7307ba59418f7af6f84/src/back/hlsl/help.rs#L355) doesn't require explicit initialization for all members.

I'm not sure what indexing means in this context since it's a struct.

teoxoy:
> half - 16-bit floating point value. This data type is provided only for language compatibility. Direct3D 10 shader targets map all half data types to float data types.

[reference](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-scalar)

DXC requires the `-enable-16bit-types` flag to be passed as described [here](https://github.com/microsoft/DirectXShaderCompiler/wiki/Buffer-Packing/4e59b6a713aa5b363ac57e87574bc1d13557bd0e) in order for `half` to be a true 16bit type otherwise it's 32bit

Maybe a better way to do this would be to check if the user is using a `half` type anywhere in the struct and only then use `half` for padding since at this point it would be safe to assume that the `half` type is 16bit.

What do you think?

src/back/hlsl/writer.rs
teoxoy: Ah, I think I know what you mean by adjusting the indices. Structs also use `Expression::AccessIndex` which uses an index for accessing struct members.

This is not an issue though since we are just adding the padding without any other part of the code knowing about it. As mentioned previously we can get away with it since we don't need to initialize the padding explicitly.

kvark: Sounds about right. This is more complicated in MSL, but fortunately not here.

kvark:
@teoxoy sounds like a reasonable proposal!

kvark (approved):

src/back/hlsl/writer.rs
teoxoy: Indeed, I first used the same approach as in the MSL backend only to discover that it wasn't necessary.

teoxoy:
> Maybe a better way to do this would be to check if the struct contains a member with a `half` type and only then use `half` for padding since at this point it would be safe to assume that the `half` type is 16bit. Also the padding only needs to be as small as the smallest struct member.

> sounds like a reasonable proposal!

so, should I open a new PR implementing that logic?

kvark:
If you feel that the FP16 spec has issues with `half`, please open an issue. Suggesting or writing a PR with a solution is greatly appreciated, obviously!
I suspect we are the only ones detecting it because Google virtualized uniform buffers on DX12, so they aren't actually using `half` in them. So your worries are likely right.

teoxoy:
I looked into this some more and it looks like the [WGSL FP16 extension](https://github.com/gpuweb/gpuweb/issues/2512) will require generation of DXIL ([ref](https://docs.google.com/presentation/d/1PyF0PnKW7zcFWQSsa5p9n8Sp6tRBP-7OQp6SljnYyhU/edit#slide=id.g111ff1b0a64_2_30)). So I guess when we start implementing the extension, we will just fail if we try to target the HLSL backend when the extension is required.

kvark:
we'll just not expose it on DX12 until we are able to generate DXIL

PR #1788, opened 2022-03-22
[wgsl-in] Implement increment and decrement statements

teoxoy:
Implements increment (++) and decrement (--) statements as per [spec](https://gpuweb.github.io/gpuweb/wgsl/#increment-decrement).

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -3312,12 +3312,13 @@ impl Parser {
                     .expressions
                     .append(crate::Expression::Binary { op, left, right }, span.into())
             }
-            (op @ Token::IncrementOperation | op @ Token::DecrementOperation, op_span) => {
-                let op = match op {
+            token @ (Token::IncrementOperation, _) | token @ (Token::DecrementOperation, _) => {
+                let op = match token.0 {
                     Token::IncrementOperation => Bo::Add,
                     Token::DecrementOperation => Bo::Subtract,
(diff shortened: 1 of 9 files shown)


src/front/wgsl/mod.rs
```diff
@@ -3295,13 +3305,69 @@ impl Parser {
...
+                    crate::Span::default(),
+                );
+
+                let left = context.expressions.append(
+                    crate::Expression::Load {
+                        pointer: reference.handle,
+                    },
+                    lhs_span.into(),
+                );
+                let right = context.interrupt_emitter(
```
kvark, line 3362: why are we interrupting the emitter here?

kvark (commented):
Thank you for the PR! Just one question

src/front/wgsl/mod.rs
teoxoy: I kept getting this validation error without it

```
thread 'convert_wgsl' panicked at 'Naga module validation failed: WithSpan { inner: EntryPoint { stage: Fragment, name: "fs_main", error: Function(ExpressionAlreadyInScope([22])) }, spans: [] }', tests/snapshots.rs:108:10
```

and I've seen it used here most likely for the same reason

https://github.com/gfx-rs/naga/blob/cf992a9a95026d8da2f4847e4884e6c0efbba1db/src/front/wgsl/mod.rs#L2301

kvark (approved):

PR #1789, opened 2022-03-23
Implement invariant attribute

### LINKED ISSUE #1659: Support invariant qualifier
See https://gpuweb.github.io/gpuweb/wgsl/#attribute-invariant

teoxoy:
closes #1659

Implements the invariant attribute in all frontends and backends.

References

[WGSL Invariant Attribute](https://gpuweb.github.io/gpuweb/wgsl/#attribute-invariant)
[SPIR-V Invariant Decoration (\#18)](https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#Decoration)
[MSL Invariant Attribute](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#page=87)
[GLSL Invariant Qualifier](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.60.pdf#page=105)
[HLSL Variable Syntax](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-variable-syntax)
[HLSL Function Declaration Syntax](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-function-syntax)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -748,7 +748,11 @@ pub enum ConstantInner {
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
 pub enum Binding {
     /// Built-in shader variable.
-    BuiltIn(BuiltIn),
+    BuiltIn {
+        built_in: BuiltIn,
+        /// can only be `true` for [`BuiltIn::Position`]
+        invariant: bool,
+    },
 
     /// Indexed location.
     ///
--- a/src/back/glsl/features.rs
+++ b/src/back/glsl/features.rs
@@ -404,7 +404,7 @@ impl<'a, W> Writer<'a, W> {
             _ => {
                 if let Some(binding) = binding {
                     match *binding {
-                        Binding::BuiltIn(builtin) => match builtin {
+                        Binding::BuiltIn { built_in, .. } => match built_in {
                             crate::BuiltIn::ClipDistance => {
                                 self.features.request(Features::CLIP_DISTANCE)
                             }
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -336,7 +336,7 @@ impl fmt::Display for VaryingName<'_> {
                 };
                 write!(f, "_{}_location{}", prefix, location,)
             }
-            crate::Binding::BuiltIn(built_in) => {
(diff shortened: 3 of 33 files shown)


teoxoy:
Also fixes HLSL's modifiers not being written in the vertex output and fragment input structs.

Before
no_mod

After
mod

src/lib.rs
```diff
@@ -748,7 +748,11 @@ pub enum ConstantInner {
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
 pub enum Binding {
     /// Built-in shader variable.
-    BuiltIn(BuiltIn),
+    BuiltIn {
+        built_in: BuiltIn,
+        /// can only be `true` for [`BuiltIn::Position`]
+        invariant: bool,
```
kvark, line 754: Perhaps there is a better place for this flag to be?
If user locations are to be supported in the future than the current API of this PR wouldn't work, anyway.
If it's only *ever* going to be supported on VS positions, then we could place it on the entry point itself (see `struct EntryPoint`). Would you be interested to check and see how this works out in terms of code complexity?

teoxoy: I checked the `EntryPoint` way but I think the code would end up a lot more complex than it is now since the frontends would need to first read the invariant attribute, then look for all entry points that use that position builtin (since it can be in a struct) and set it on the entry point. Backends would also have trouble with this since they would have to map the invariant attribute back.

If the spec changes to allow the attribute to be used on locations, we can change the binding to be a struct with an inner field being the enum we have now and add the invariant flag on the struct.

I agree that it might not be the best spot for it however I tried to put it in multiple places and decided at last to put it here.

kvark: That's a good insight, thank you!

kvark (approved):

kvark:
Going to ask @jimblandy to take a look as well - on the API change only.
We take extra care for any API related changes. Sorry about the delay this is causing!

teoxoy:
Thanks, no hurry!

teoxoy:
Rebased to fix conflicts

kvark:
Jim seems to have missed this one, unfortunately. Let's go ahead and merge then.

PR #1790, opened 2022-03-26
[wgsl-in] Constructor improvements

### LINKED ISSUE #1733: [wgsl-in] Zero value expressions are not implemented
WGSL spec section:
https://gpuweb.github.io/gpuweb/wgsl/#zero-value-expr

Shaders should be able to do things like this:
```
var f = f32();
var v = vec4<f32>();
var a = array<i32, 4>();
```

The Tint test suite makes extensive use of this feature, so supporting this would enable a large corpus of tests (~1900) to be pushed through Naga to help uncover other issues.

### LINKED ISSUE #1774: WGSL constructor doesn't check length
The following code produces an error message that was unhelpful 

```
@stage(vertex)
fn passthrough(@location(0) pos: vec2<f32>) -> @builtin(position) vec4<f32> {
   return vec4<f32>(pos);
}
```

The error is:

```
[2022-03-13T21:53:32Z ERROR naga::valid::function] Returning Some(Vector { size: Bi, kind: Float, width: 4 }) where Some(Vector { size: Quad, kind: Float, width: 4 }) is expected
error: 
  ┌─ triangle.wgsl:3:10
  │
3 │    return vec4<f32>(pos);
  │          ^^^^^^^^^^^^^^^ naga::Expression [2]

Entry point passthrough at Vertex is invalid: 
	The `return` value Some([2]) does not match the function return value
```

From the log output, it seems like the `4` in the type constructor expression `vec4<f32>(pos)` got ignored, and the expression is typed as a `vec2`, which doesn't match the return type.

teoxoy:
Adds support for zero value constructors and constructors that infer their type from their parameters.

Also adds 4 more types of errors to better indicate issues with constructors.

For instance, before you could construct atomic types

```
var _ = atomic<i32>(0);
```

now you will get this error:

```
error: type `atomic` is not constructible
  ┌─ wgsl:3:25
  │
3 │                 var _ = atomic<i32>(0);
  │                         ^^^^^^ type is not constructible
```

closes #1733
closes #1774

References
[Constructors](https://gpuweb.github.io/gpuweb/wgsl/#type-constructor-expr)
[Zero value constructors](https://gpuweb.github.io/gpuweb/wgsl/#zero-value-expr)


src/front/wgsl/mod.rs
```diff
@@ -466,6 +490,126 @@ impl<'a> Error<'a> {
     }
 }
 
+/// Represents the type of the constructor
+///
+/// Vectors, Matrices and Arrays can have partial type information
+/// which later gets inferred from the constructor parameters
+enum ConstructorType {
```
kvark, line 497: there is quite a bit of machinery now associated to constructors. Perhaps it could live in another module?

src/front/wgsl/mod.rs
```diff
@@ -466,6 +490,126 @@ impl<'a> Error<'a> {
...
+    PartialArray,
+    Array {
+        base: Handle<crate::Type>,
+        size: crate::ArraySize,
+        stride: u32,
+    },
+    Struct(Handle<crate::Type>),
+}
+
+impl TryFrom<ConstructorType> for TypeResolution {
```
kvark, line 528: is `TryFrom` as a trait really necessary here? Could it instead be just a method on `ConstructorType`?

src/front/wgsl/mod.rs
```diff
@@ -959,6 +1103,95 @@ impl<'a> ExpressionContext<'a, '_, '_> {
...
+                    crate::Type {
+                        name: None,
+                        inner: crate::TypeInner::Scalar { width, kind },
+                    },
+                    Default::default(),
+                );
+                crate::ConstantInner::Composite {
+                    ty,
+                    components: (0..size as u8)
+                        .map(|_| self.create_zero_value_constant(scalar_ty))
```
kvark, line 1135: can we call it once and just copy the value instead of calling it for each iteration?

src/front/wgsl/mod.rs
```diff
@@ -2034,6 +2267,106 @@ impl Parser {
...
+        word: &'a str,
+        type_arena: &mut UniqueArena<crate::Type>,
+        const_arena: &mut Arena<crate::Constant>,
+    ) -> Result<Option<ConstructorType>, Error<'a>> {
+        if let Some((kind, width)) = conv::get_scalar_type(word) {
+            return Ok(Some(ConstructorType::Scalar { kind, width }));
+        }
+
+        let partial = match word {
+            "vec2" => ConstructorType::PartialVector {
```
kvark, line 2282: it looks like we are not using the `enum ConstructorType` properly here. The `Partial*` variants are only appearing internally here. Perhaps, we could instead just have an "internal" enum for partial stuff and the main `enum ConstructorType` having no partial stuff? I suppose there wouldn't be needed `enum ConstructorType` at all, potentially, since we could just use `TypeInner`.

src/front/wgsl/mod.rs
```diff
@@ -2045,104 +2378,222 @@ impl Parser {
...
+                    None => {
+                        match self.parse_type_decl_impl(
+                            lexer,
+                            TypeAttributes::default(),
+                            type_name,
+                            ctx.types,
+                            ctx.constants,
+                        )? {
+                            Some(_) => {
+                                let span = self.pop_scope(lexer);
```
kvark, line 2395: why are we popping the scope here? I don't see where the corresponding push is

src/front/wgsl/mod.rs
```diff
@@ -2045,104 +2378,222 @@ impl Parser {
...
+                // type of the first component
+                &'a crate::TypeInner,
+            ),
+        }
 
-            while lexer.next_argument()? {
-                components.push(last_component);
-                last_component = self.parse_general_expression(lexer, ctx.reborrow())?;
+        impl<'a> Components<'a> {
+            fn components(self) -> Vec<Handle<crate::Expression>> {
```
kvark, line 2433: maybe `to_components_vec`? The `components()` method looks like an accessor, but it's not an accessor because it creates a vector and returns it.

src/front/wgsl/mod.rs
```diff
@@ -2045,104 +2378,222 @@ impl Parser {
...
+            }
+        }
+
+        enum Components<'a> {
+            None,
+            One(Handle<crate::Expression>, Span, &'a crate::TypeInner),
+            Many(
+                Vec<Handle<crate::Expression>>,
+                Vec<Span>,
+                // type of the first component
```
kvark, line 2427: let's just name those fields instead of putting comments

src/front/wgsl/mod.rs
```diff
@@ -2045,104 +2378,222 @@ impl Parser {
...
-                (
-                    &crate::TypeInner::Vector {
-                        size, kind, width, ..
+                if let Some(constant) = ctx.create_zero_value_constant(ty) {
+                    return Ok(Some(ctx.interrupt_emitter(
+                        crate::Expression::Constant(constant),
+                        span.into(),
+                    )));
+                } else {
+                    return Err(Error::TypeNotConstructible(span));
```
kvark, line 2477: nit: should we move `return` outside of `if/else`?

kvark (changes requested):
Very impressive work here!
We'll need a small pass to clean up some bits.

src/front/wgsl/mod.rs
teoxoy: I'll use `self.current_byte_offset()` to get the span instead. I was thinking since it will error out it wouldn't matter.

src/front/wgsl/mod.rs
teoxoy: Looking at it again I think a method returning an Option would be better.

src/front/wgsl/mod.rs
teoxoy: How about `into_components_vec` (since it takes ownership)?

src/front/wgsl/mod.rs
teoxoy: It will only parse the component type if present. If not, will still return the partial variants. See the [last arm](https://github.com/teoxoy/naga/blob/6ae0745cccc45204c4673166842067755066db13/src/front/wgsl/mod.rs#L2366) of the match.

The partial variants are later used in the `parse_construction` function.

src/front/wgsl/mod.rs
teoxoy: Actually, to use `current_byte_offset` there I would need the start of the span as well. The start of the span would have to be passed to the `parse_construction` function since it already gets the parsed type name from its caller.

Should I pass the span of the type into the `parse_construction` fn or leave it as is? Does `pop_scope` cause issues in this instance? (I'm not too familiar with that part of the code)

src/front/wgsl/mod.rs
kvark: oh yes, that sounds great!

src/front/wgsl/mod.rs
kvark: Just keep the push and pop corresponding to each other. Having this `pop` without the relevant `push` is the weird part. If you need to pass the span into `parse_construction` - so be it.

teoxoy: Ah, I think I get what your concern is now.

The `parse_construction` expects there to be a `PrimaryExpr` scope pushed already.

https://github.com/gfx-rs/naga/blob/012f2a6b2e9328e93939429705fdf9102d9f69b7/src/front/wgsl/mod.rs#L2047

Nonetheless, I changed the fn signature to pass the span trough it. It's a more accurate span actually (doesn't point to whitespace).

kvark (approved):
Glorious!

PR #1791, opened 2022-03-27
wgsl: Make semicolon after struct decl optional

stshine:
Spec PR: https://github.com/gpuweb/gpuweb/pull/2499


kvark (approved):
Thank you!

PR #1793, opened 2022-03-27
Add glsl barrier function

### LINKED ISSUE #1771: [glsl-in] Unknown function 'barrier'
This shader:
```glsl
#version 460 core

void main() {
    barrier();
}
```
produces this error:
```
error: Unknown function 'barrier'
  ┌─ shader.comp:4:5
  │
4 │     barrier();
  │     ^^^^^^^^^
```

fintelia:
This attempts to implement the GLSL `barrier` function based some of the surrounding builtin implementations. However, it panics later when emitting expressions, so it evidentially isn't correct yet. 

Fixes #1771

### DIFF
--- a/src/front/glsl/builtins.rs
+++ b/src/front/glsl/builtins.rs
@@ -1998,6 +1998,7 @@ impl MacroCall {
             ),
             MacroCall::Barrier => {
                 ctx.emit_flush(body);
+                ctx.emit_start();
                 body.push(crate::Statement::Barrier(crate::Barrier::all()), meta);
                 return Ok(None);
             }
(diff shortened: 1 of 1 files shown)


JCapucho:
Can you show an example that panics? I've tried to reproduce it on your branch with a very simple compute shader
```glsl
#version 460 core

void main() {
    barrier();
}
```
But that works

fintelia:
Two barriers in a row seems to be enough to cause a panic:
```glsl
#version 460 core

void main() {
    barrier();
    barrier();
}
```

src/front/glsl/builtins.rs
```diff
@@ -1992,6 +1996,11 @@ impl MacroCall {
                 Span::default(),
                 body,
             ),
+            MacroCall::Barrier => {
+                ctx.emit_flush(body);
```
JCapucho, line 2000: You also need to do `ctx.emit_start`

kvark (approved):
Would be nice to test it as well

PR #1795, opened 2022-03-27
glsl-in: Resume expression emit after imageStore

JCapucho:
This wasn't done before which caused the frontend to crash if anything
appeared after it.


fintelia:
Fixes #1797

kvark (approved):

PR #1802, opened 2022-04-05
[hlsl-out] fix matCx2 translation for uniform buffers

### LINKED ISSUE #1785: [glsl/hlsl-out] Matrix of the form matCx2 from WGSL not getting translated properly to GLSL/HLSL (only in uniform buffers)
WGSL expects a `matCx2` to have a matrix stride of 8 (inner vector components to be aligned to 8 bytes including itself).
GLSL/HLSL expect a `matCx2` to have a matrix stride of 16 (inner vector components to be aligned to 16 bytes including itself).

### Solution

Replace `matCx2` with a sequence of `C` `vec2`'s (since `vec2`s alignment is 8) and reconstruct the matrix in the shader.

### References

[WGSL Alignment and Size](https://gpuweb.github.io/gpuweb/wgsl/#alignment-and-size)
[GLSL Standard Uniform Block Layout](https://www.khronos.org/registry/OpenGL/specs/gl/glspec45.core.pdf#page=159)
[HLSL DirectXShaderCompiler Buffer-Packing Wiki](https://github.com/microsoft/DirectXShaderCompiler/wiki/Buffer-Packing/4e59b6a713aa5b363ac57e87574bc1d13557bd0e)

### Related issues

#1400 (https://github.com/gfx-rs/naga/issues/1400#issuecomment-1073222373) - inverse of this issue (concerning the frontends)

teoxoy:
Partially fixes #1785 (GLSL backend still needs to be changed)


src/back/hlsl/help.rs
```diff
@@ -393,13 +399,37 @@ impl<'a, W: Write> super::Writer<'a, W> {
...
+
             let field_name = &self.names[&NameKey::StructMember(constructor.ty, i)];
-            //TODO: handle arrays?
-            writeln!(
-                self.out,
-                "{}{}.{} = {}{};",
-                INDENT, RETURN_VARIABLE_NAME, field_name, ARGUMENT_VARIABLE_NAME, i,
-            )?;
+
+            match &module.types[member.ty].inner {
```
kvark, line 406: let's not match on a reference here, just match on a thing itself instead

src/back/hlsl/help.rs
```diff
@@ -393,13 +399,37 @@ impl<'a, W: Write> super::Writer<'a, W> {
...
+                &crate::TypeInner::Matrix { columns, rows, .. }
```
kvark, line 407: nit: should just match as `rows: crate::VectorSize::Bi`

src/back/hlsl/help.rs
```diff
@@ -413,6 +443,335 @@ impl<'a, W: Write> super::Writer<'a, W> {
...
+        let field_name = &self.names[&NameKey::StructMember(access.ty, access.index)];
+        match module.types[member.ty].inner {
+            crate::TypeInner::Matrix { columns, .. } => {
+                for i in 0..columns as u8 {
+                    write!(
+                        self.out,
+                        "{}.{}_{}",
+                        STRUCT_ARGUMENT_VARIABLE_NAME, field_name, i
+                    )?;
+                    if i < columns as u8 - 1 {
```
kvark, line 501: small trick: if you move this condition to the beginning of the loop body (instead of the end), you can just compare `i != 0`

src/back/hlsl/help.rs
```diff
@@ -413,6 +443,335 @@ impl<'a, W: Write> super::Writer<'a, W> {
...
+
+        writeln!(
+            self.out,
+            "{}switch({}) {{",
+            INDENT, MATRIX_INDEX_ARGUMENT_VARIABLE_NAME
+        )?;
+
+        let field_name = &self.names[&NameKey::StructMember(access.ty, access.index)];
+
+        match &module.types[member.ty].inner {
```
kvark, line 652: similarly, let's not match on a reference

src/back/hlsl/help.rs
```diff
@@ -413,6 +443,335 @@ impl<'a, W: Write> super::Writer<'a, W> {
...
```
kvark, line 746: and here

src/back/hlsl/help.rs
```diff
@@ -478,6 +837,46 @@ impl<'a, W: Write> super::Writer<'a, W> {
...
+                            resolved = &module.types[base].inner;
+                            Some(base)
+                        }
+                        _ => base_ty_res.handle(),
+                    };
+                    if let crate::TypeInner::Struct { ref members, .. } = *resolved {
+                        let member = &members[index as usize];
+
+                        match module.types[member.ty].inner {
+                            crate::TypeInner::Matrix { rows, .. }
```
kvark, line 858: nit: let's encode the condition into the pattern here: `rows: crate::VectorSize::Bi`

src/back/hlsl/writer.rs
```diff
@@ -749,6 +749,28 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                        size: rows,
+                        kind: crate::ScalarKind::Float,
+                        width,
+                    };
+                    let field_name_key = NameKey::StructMember(handle, index as u32);
+
+                    for i in 0..columns as u8 {
+                        self.write_value_type(module, &vec_ty)?;
+                        write!(self.out, " {}_{}", &self.names[&field_name_key], i)?;
+                        if i < columns as u8 - 1 {
```
kvark, line 769: same little trick could be done here

src/back/hlsl/writer.rs
```diff
@@ -1215,11 +1237,175 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     writeln!(self.out, "[_i] = _result[_i];")?;
                     writeln!(self.out, "{}}}", level)?;
                 } else {
+                    // We treat matrices of the form matCx2 as a sequence of C vec2's due to
```
kvark, line 1240: please make sure to write something down at the top doc comment of `mod.rs` about this

src/back/hlsl/help.rs
```diff
@@ -478,6 +837,46 @@ impl<'a, W: Write> super::Writer<'a, W> {
...
+                                if member.binding.is_none() && rows == crate::VectorSize::Bi =>
+                            {
+                                let ty = base_ty_handle.unwrap();
+                                let access = WrappedStructMatrixAccess { ty, index };
+
+                                if !self.wrapped.struct_matrix_access.contains(&access) {
+                                    self.write_wrapped_struct_matrix_get_function(module, access)?;
+                                    self.write_wrapped_struct_matrix_set_function(module, access)?;
```
kvark, line 866: I suspect this is right but still would like to hear your thoughts on one question. Since the transformation this PR is doing is really only needed for matCx2 in uniform buffers, why do we care about `set_xxx` functions? Uniform buffers are read-only after all.

kvark (commented):
This is very serious work, and it's done impressively well.
Thank you!
Small nits are noted.
Medium size question is present as well.

Finally, the biggest question I have to you is the following. Do you think it would be simpler to just give up on structured uniform buffers and represent them as arrays of `vec4` internally? That's what Tint does, last time I checked. I quite like our approach because it's more debuggable, but the amount of code for the "exceptions" like matCx2 is a bit worrying me.

teoxoy:
> Do you think it would be simpler to just give up on structured uniform buffers and represent them as arrays of vec4 internally? That's what Tint does, last time I checked. I quite like our approach because it's more debuggable, but the amount of code for the "exceptions" like matCx2 is a bit worrying me.

I think with this PR we will be at a stage where we are covering all the discrepancies (between the buffer layout of WGSL and HLSL) since to get full compatibility with WGSL's uniform buffers we need to:

- sometimes insert padding (due to the vector-relaxed layout of HLSL) which we already do
- translate matCx2's to a sequence of C vec2's which is what this PR is solving

I've put together a [resource](https://gist.github.com/teoxoy/936891c16c2a3d1c3c5e7204ac6cd76c) covering all the differences between layouts that we can reference.

What we might not be able to cover (but could with an array of vec4s) in the future is SPIR-V (with all its buffer layout flexibility) -> HLSL; I'm unsure if this in scope for naga or not (if it is maybe generating DXIL would be a better fit for this?).

PR #1805, opened 2022-04-08
[spv-in] Error on unsupported `MatrixStride`

teoxoy:
More or less reverts #908


kvark:
This would probably mean our lazy jobs for Dota2 shaders will start failing again. But it looks like you have a plan in place to address this properly.

teoxoy:
Seems like it's still working https://github.com/gfx-rs/naga/runs/5966771736?check_suite_focus=true

I think what might have fixed the issue with the dota shaders in #908 is that the check was fixed (i.e. instead of rows * width -> rounded_rows * width). Previously it was most likely failing only on matCx3's.

PR #1806, opened 2022-04-08
[glsl-in] Error on a `matCx2` used with the std140 layout

teoxoy:
Our IR expects matrices of the form `matCx2` to have a stride of 8 however matrices in the std140 layout have a stride of at least 16.


kvark (approved):
Nice!

PR #1809, opened 2022-04-08
[hlsl-out] Don't output interpolation modifier if it's the default

NoelTautges:
Currently, hlsl-out writes the interpolation modifier before most structs, e.g.

```hlsl
struct input_0_ {
    linear float4 SV_Position : LOC0;
};
```

However, [the linear intepolation modifier is the default](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-struct#interpolation-modifiers-introduced-in-shader-model-4) in HLSL.

This pull request makes hlsl-output not output `linear` to make outputted code look more human-like.

### DIFF
--- a/src/back/hlsl/conv.rs
+++ b/src/back/hlsl/conv.rs
@@ -111,12 +111,14 @@ impl crate::BuiltIn {
 }
 
 impl crate::Interpolation {
-    /// Helper function that returns the string corresponding to the HLSL interpolation qualifier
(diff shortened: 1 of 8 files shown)


NoelTautges:
I forgot to test with `--all-features`. Let me change the output tests. 😅

NoelTautges:
I ran the changed tests through FXC and the output was the same.

kvark (approved):

PR #1814, opened 2022-04-09
[hlsl-out] Add more padding when necessary

### LINKED ISSUE #1772: Shader working on VULKAN but not on DX12 FRAGMENT
I have a problem that my shader is only producing the right result on VULKAN but not DX12.

Expected result (works on VULKAN):
https://imgur.com/xIHAQdv
dragonVULKAN

Result on DX12:
https://imgur.com/1EbPQOf
dragonDX12

My fragment shader code looks like this:
```
void main() {

    float ambientStrength = 0.2;
    vec4 texel = texture(sampler2D(t_diffuse, s_diffuse), tex_coords_out);

    vec3 resultA = vec3(0.0,0.0,0.0);

    for(int i = 0;i<numLights;i++) {
        vec3 ambient = material.ambientColor * ambientStrength * light[i].lightColor;

        vec3 norm = normalize(normal);
        vec3 lightDir = normalize(light[i].lightPos - frag_pos);

        float diff = max(dot(norm,lightDir),0.0);
        vec3 diffuse = light[i].lightColor * (diff * material.diffuseColor);

        resultA += ambient + diffuse;
    }

    vec4 result = texel * vec4(resultA,1.0);

    f_color = result;
}
```

The Light struct looks like this:

```
struct Light {
    vec3 lightPos;
    vec3 lightColor;
};

layout(set = 3, binding = 0)
uniform LightUniform {
    Light light[4];
    int numLights;
};

```
If the numLights variable is 1 the lighting is working on DX12 (but only for one light of course).
There seems to be a problem with "resultA += ambient + diffuse;" because if two lights are present than the addition is not correct (everything black).

naga-version = "0.8"

teoxoy:
fixes #1772

Add padding at the end of structs and after struct members of type matrix and array since apparently the size of HLSL's matrices, arrays and structs doesn't get rounded up to a multiple of their alignment implicitly.

see also https://github.com/gfx-rs/naga/issues/1772#issuecomment-1093995973

You can now see that the HLSL output of the newly added test has the expected offsets.
https://shader-playground.timjones.io/86acbc3f9bec2bad5d3668e5ac707eeb

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -727,7 +727,11 @@ impl<'a, W: Write> Writer<'a, W> {
         self.collect_reflection_info()
     }
 
-    fn write_array_size(&mut self, size: crate::ArraySize) -> BackendResult {
+    fn write_array_size(
+        &mut self,
+        base: Handle<crate::Type>,
+        size: crate::ArraySize,
+    ) -> BackendResult {
         write!(self.out, "[")?;
 
         // Write the array size
@@ -751,6 +755,11 @@ impl<'a, W: Write> Writer<'a, W> {
         }
 
         write!(self.out, "]")?;
+
+        if let TypeInner::Array { base, size, .. } = self.module.types[base].inner {
+            self.write_array_size(base, size)?;
+        }
+
         Ok(())
     }
 
@@ -807,7 +816,7 @@ impl<'a, W: Write> Writer<'a, W> {
             // GLSL arrays are written as `type name[size]`
             // Current code is written arrays only as `[size]`
             // Base `type` and `name` should be written outside
-            TypeInner::Array { size, .. } => self.write_array_size(size)?,
+            TypeInner::Array { base, size, .. } => self.write_array_size(base, size)?,
             // Panic if either Image, Sampler, Pointer, or a Struct is being written
             //
             // Write all variants instead of `_` so that if new variants are added a
@@ -994,8 +1003,8 @@ impl<'a, W: Write> Writer<'a, W> {
         write!(self.out, " ")?;
         self.write_global_name(handle, global)?;
 
-        if let TypeInner::Array { size, .. } = self.module.types[global.ty].inner {
-            self.write_array_size(size)?;
+        if let TypeInner::Array { base, size, .. } = self.module.types[global.ty].inner {
+            self.write_array_size(base, size)?;
         }
 
         if global.space.initializable() && is_value_init_supported(self.module, global.ty) {
@@ -1299,8 +1308,8 @@ impl<'a, W: Write> Writer<'a, W> {
             write!(this.out, " {}", &this.names[&ctx.argument_key(i as u32)])?;
 
             // Write array size
-            if let TypeInner::Array { size, .. } = this.module.types[arg.ty].inner {
-                this.write_array_size(size)?;
+            if let TypeInner::Array { base, size, .. } = this.module.types[arg.ty].inner {
+                this.write_array_size(base, size)?;
             }
 
             Ok(())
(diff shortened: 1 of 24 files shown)


teoxoy:
Looks like the HLSL backend doesn't handle array constructors that well.

This WGSL

```wgsl
fn test(a: array<f32, 1>) -> f32 {
    return a[0];
}

@stage(vertex)
fn vertex() -> @builtin(position) vec4<f32> {
    return vec4<f32>(1.0) * test(array<f32, 1>(4.0));
}
```

produces this HLSL

```hlsl
float test(float a[1])
{
    return a[0];
}

float4 vertex() : SV_Position
{
    const float _e4 = test({ 4.0 });
    return (float4(1.0.xxxx) * _e4);
}
```

which DXC errors on with the same [error in the CI](https://github.com/gfx-rs/naga/runs/5955901929?check_suite_focus=true)

```
D:\local\Temp\340e7002-2bc1-4ebc-a460-75337a3d949f.hlsl:9:28: error: expected expression
    const float _e4 = test({ 4.0 });
                           ^
```

https://shader-playground.timjones.io/d3480954041d4f0515e1fada1d464256

teoxoy:
Got it working by adding wrapped constructors for arrays.

Also added support for multidimensional arrays for the HLSL and GLSL backends and fixed arrays not working as function arguments.

src/back/glsl/mod.rs
```diff
@@ -751,6 +755,11 @@ impl<'a, W: Write> Writer<'a, W> {
         }
 
         write!(self.out, "]")?;
+
+        if let TypeInner::Array { base, size, .. } = self.module.types[base].inner {
```
kvark, line 759: Let's avoid shadowing the same name as the same type. It makes code confusing wrt what actual values are used.
I filed https://github.com/rust-lang/rust-clippy/issues/3433#issuecomment-1094555963 upstream to get clippy helping with this guideline.

src/back/hlsl/conv.rs
```diff
@@ -35,6 +35,83 @@ impl crate::TypeInner {
...
+    }
+
+    /// Used to generate the name of the wrapped type constructor
+    pub(super) fn hlsl_type_id(
+        &self,
+        base: crate::Handle<crate::Type>,
+        types: &crate::UniqueArena<crate::Type>,
+        constants: &crate::Arena<crate::Constant>,
+        names: &crate::FastHashMap<crate::proc::NameKey, String>,
+    ) -> Result<String, Error> {
```
kvark, line 78: Can it return `Cow<str>` instead of `String`?

kvark (approved):

kvark:
This looks quite reasonable, I don't have serious concerns.
Other than a bigger question of whether we should just give up and make it a `float4 raw_data[]` representation for uniform buffers in HLSL, which I think I asked in another PR. What do you think?

teoxoy:
> Other than a bigger question of whether we should just give up and make it a `float4 raw_data[]` representation for uniform buffers in HLSL, which I think I asked in another PR. What do you think?

Well, hopefully this should be the last edge case.

Tbh I like the addition of this "transparent" padding more. DXC shows offsets for types in buffers which makes it a lot easier to check if something is wrong. I would imagine it's easier to optimize accesses when DXC does the translation to DXIL too since it would just see the padding is not used and therefore remove it. I would also think that reconstructing the types from a vec4 array is a lot more involved than inserting padding in the right spots. Overall I don't see clear advantages to using a vec4 array.

What's your take on vec4 array vs padding?

kvark:
> What's your take on vec4 array vs padding?

I think having proper structures is more user (read: debugging) friendly, and I'd love for it to stay as long as it doesn't involve too much work for workarounds.
Thank you for implementing this!

Issue #1811, opened 2022-04-08
GLSL backend can't handle matrix-typed globals


jimblandy:
When asked to produce GLSL from the following WGSL input, Naga declares the IR valid, but the GLSL back end produces the error "Non-struct type of a buffer":

```
@group(0)
@binding(0)
var<uniform> transform: mat4x4<f32>;

@stage(vertex)
fn main(
    @location(0) position: vec4<f32>,
) -> @builtin(position) vec4<f32> {
    return transform * position;
}
```

jimblandy:
I'm working on this, but it turns out there's a lot about GLSL I don't know, so it's taking a bit longer than expected.

PR #1815, opened 2022-04-10
[spv-out] Emit required decorations on wrapper struct types, too.

### LINKED ISSUE #1810: [spv-out] wrapping struct types for globals lack required decorations
Naga IR permits globals to have non-struct types, but Vulkan [requires][rq] that globals in the `Uniform` and `Storage` storage classes be struct types. Our SPIR-V backend accommodates this by synthesizing SPIR-V struct types wrapping the contents of Naga globals as needed, but it fails to apply decorations required on members that are matrices or arrays of matrices.

For example, the following WGSL input generates SPIR-V that fails to validate:

```
@group(0)
@binding(0)
var<uniform> transform: mat4x4<f32>;

@stage(vertex)
fn vs_main(
    @location(0) position: vec4<f32>,
    return transform * position;
}
```

This fails validation in the `vulkan1.0` target environment:

```
error: line 19: Structure id 7 decorated as Block must be explicitly laid out with MatrixStride decorations.
  %_struct_7 = OpTypeStruct %mat4v4float
```

[rq]:  https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#interfaces-resources-descset

jimblandy:
Fixes #1810. Draft PR because it contains no tests. Contains no tests because #1811.

[spv-out] Decorate array and struct type layouts unconditionally.

Remove the `decorate_layout` variable, which is unconditionally set to `true` anyway.

### DIFF
--- a/src/back/spv/helpers.rs
+++ b/src/back/spv/helpers.rs
@@ -62,7 +62,21 @@ impl crate::AddressSpace {
     }
 }
 
-/// Return true if the global requires a type decorated with "Block".
+/// Return true if the global requires a type decorated with `Block`.
+///
+/// Vulkan spec v1.3 §15.6.2, "Descriptor Set Interface", says:
+///
+/// > Variables identified with the `Uniform` storage class are used to
+/// > access transparent buffer backed resources. Such variables must
+/// > be:
+/// >
+/// > -   typed as `OpTypeStruct`, or an array of this type,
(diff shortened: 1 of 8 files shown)


kvark (approved):

jimblandy:
Once #1823 lands, this can land.

Issue #1561, opened 2021-11-26
[msl-out] Generated metal uses packed_float3 in an expression


scoopr:
With this wgsl
```wgsl
struct ObjectOutputData {
    model_view_proj: mat4x4<f32>;
    inv_squared_scale: vec3<f32>;
    material_idx: u32;
};

var<private> global : ObjectOutputData;

fn foo()  {
    let zero3 = vec3<f32>(0.0, 0.0, 0.0);
    let zero4 = vec4<f32>(0.0, 0.0, 0.0, 0.0);
    let data : ObjectOutputData = global;
    let zero_mat3: mat3x3<f32> = mat3x3<f32>(zero3, zero3, zero3);
    let foo = zero_mat3 * data.inv_squared_scale;
}
```

the converted struct ends up as
```c++
struct ObjectOutputData {
    metal::float4x4 model_view_proj;
    packed_float3 inv_squared_scale;
    metal::uint material_idx;
};
```
and the expression uses it directly
```
metal::float3 foo_1 = zero_mat3_ * data.inv_squared_scale;
```
But I believe it should be cast first, like `float3(data.inv_squared_scale)`

A workaround can be to not have something with align(4) (like the u32 here) next to vec3. Just moving the u32 to first member works, or having some vec4 or mat4x4 in between also works. Then it won't be generated as packed_float3.

Originally found with rend3 trunk version of `opaque.vert.cpu.wgsl`, cc @cwfitzgerald

teoxoy:
I'm not sure when this got fixed but it's no longer an issue as can be seen here: https://github.com/gfx-rs/naga/pull/1816/commits/a33518818a44db4808199d74455cc2491b9e739a

scoopr:
Given the example at the top (fixed for the struct semicolon change), with naga master as of a3d968e7
```sh
$ cargo run naga -- test.wgsl test.metal
$ xcrun metal test.metal
```
still yields exactly the same error, the struct field is of a packed type, but it is used in an expression that expects non-packed.

I believe in the example you gave, it is merely doing an assignment with the field, which I suppose is fine, but the arithmetic is not. In fact in the example above, assigning the field to a local variable first is a valid workaround.

teoxoy:
Ah, I see what's going on. The test I added takes advantage of baking the load expression.

https://github.com/gfx-rs/naga/blob/a3d968e79527b805a7639a4c4db4fa024c16a8c7/src/back/mod.rs#L139-L142

[...]

PR #1816, opened 2022-04-10
[msl-out] Packed vec3 improvements

teoxoy:


kvark (approved):

teoxoy:
Added more tests for packed vec3's and fixed issue in #1561 by casting them if they are used in expressions that MSL doesn't support on packed vec3's.

@kvark let me know if this is good to go or have any concerns with the approach.

PR #1819, opened 2022-04-12
[glsl-out] Document and flatten `back::glsl::Writer::write_varying`.

jimblandy:


JCapucho (approved):
lgtm, definitely an improvement in comparison to the big match

PR #1820, opened 2022-04-12
Operator improvements

### LINKED ISSUE #1527: [spv-out] `Matrix + Matrix` operations not implemented
WGSL test case:

```wgsl
[[stage(vertex)]]
fn vertex() {
    let model = mat4x4<f32>(
        vec4<f32>(1.0, 0.0, 0.0, 0.0),
        vec4<f32>(0.0, 1.0, 0.0, 0.0),
        vec4<f32>(0.0, 0.0, 1.0, 0.0),
        vec4<f32>(0.0, 0.0, 0.0, 1.0),
    );

    let foo = model + model;
}
```

Panics at: https://github.com/gfx-rs/naga/blob/e69a70bfb7b3dff1391d5189d5c68fb798c02040/src/back/spv/block.rs#L321-L328

Workaround by decomposing the matrices into their column vectors and constructing a new matrix after adding the columns separately:

```wgsl
[[stage(vertex)]]
fn vertex() {
    let model = mat4x4<f32>(
        vec4<f32>(1.0, 0.0, 0.0, 0.0),
        vec4<f32>(0.0, 1.0, 0.0, 0.0),
        vec4<f32>(0.0, 0.0, 1.0, 0.0),
        vec4<f32>(0.0, 0.0, 0.0, 1.0),
    );

    let foo = mat4x4<f32>(
        model.x + model.x,
        model.y + model.y,
        model.z + model.z,
        model.w + model.w,
    );
}
```

I believe this is the transformation required for SpirV.

teoxoy:
- add more tests for operators (covering the latest version of the WGSL spec)
- [glsl/hlsl/msl-out] support vectors for unary operators
- [spv-out] add support for matrix add and sub (fixes #1527)
- [hlsl-out] fix bool splat (see [failing CI](https://github.com/gfx-rs/naga/runs/5991330322?check_suite_focus=true))
- [spv-out] add support for integer vector - scalar multiplication
- [glsl-out] add support for boolean vector `~`, `|` and `&` ops


teoxoy:
Sorry for assigning so many reviewers - no sure who was best to review since the PR is a bit broad (github was suggesting you guys).

src/back/glsl/mod.rs
```diff
@@ -2428,23 +2430,30 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                                return Err(Error::Custom(format!(
-                                    "Cannot apply not to type {:?}",
-                                    other
-                                ))),
-                        },
+                let ty = ctx.info[expr].ty.inner_with(&self.module.types);
+
+                match *ty {
+                    TypeInner::Vector { kind: Sk::Bool, .. } => {
+                        write!(self.out, "not(")?;
```
JCapucho, line 2437: What about the other operations like `Negate` and `Not` for non booleans?

src/back/glsl/mod.rs
```diff
@@ -2467,27 +2476,22 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                                Bo::Modulo => BinaryOperation::Modulo,
-                                _ => BinaryOperation::Other,
-                            },
-                            _ => BinaryOperation::Other,
-                        },
+                        Bo::Modulo if kind == Sk::Float => BinaryOperation::Modulo,
+                        Bo::And if kind == Sk::Bool => {
+                            op = crate::BinaryOperator::LogicalAnd;
+                            BinaryOperation::VectorComponentWise
+                        }
```
JCapucho, line 2490: You can just do `vec & vec` why is this needed?

src/back/glsl/mod.rs
```diff
@@ -2467,27 +2476,22 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                        Bo::InclusiveOr if kind == Sk::Bool => {
+                            op = crate::BinaryOperator::LogicalOr;
+                            BinaryOperation::VectorComponentWise
```
JCapucho, line 2493: Same thing here you can do `vec | vec`

src/back/hlsl/writer.rs
```diff
@@ -2289,11 +2292,9 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     crate::VectorSize::Tri => "xxx",
                     crate::VectorSize::Quad => "xxxx",
                 };
-                let resolved = func_ctx.info[expr].ty.inner_with(&module.types);
-                self.write_value_type(module, resolved)?;
                 write!(self.out, "(")?;
                 self.write_expr(module, value, func_ctx)?;
-                write!(self.out, ".{})", number_of_components)?
+                write!(self.out, ").{}", number_of_components)?
```
JCapucho, line 2297: This is definitely wrong, the parentheses is there to stop problems with operator precedence so it needs to be outside.

JCapucho (changes requested):
I gave a review to the backends except `msl` (since I know next to nothing about it), there are problems that need to be fixed.

Also this shouldn't have been a monolith PR but a couple of PRs , it makes the life of everyone worse, you don't get as good nor a timely review, fixing the problems in the right places requires rebasing and other shenanigans, the reviewers need to look at a lot of unrelated code, it's more probable of causing/having merging problems with other  PRs and CI will only run on the last commit which will either require the reviewers to test every commit or it to be squashed which is bad because we have lot of changes and losing history on it would be bad.

So in the future if the changes aren't related please open separate PRs, you should also do this for commits, for example you have `add support for unary vector operators` which could have a been a commit for each backend, which would reduce the diff to only the code and tests changes for that specific backend. 

I only mean this as constructive criticism

src/back/glsl/mod.rs
teoxoy: They are handled below.

src/back/glsl/mod.rs
teoxoy: It doesn't work for boolean vectors.

src/back/glsl/mod.rs
teoxoy: It doesn't work for boolean vectors.

src/back/hlsl/writer.rs
teoxoy: This is for `Expression::Splat`; the outside parentheses were only needed because the splat was previously wrapped in a constructor which I don't think is needed. Validation failed here for boolean splat ([see CI](https://github.com/gfx-rs/naga/runs/5991330322?check_suite_focus=true)); surrounding the boolean and then writing the splat components fixes it.

src/back/glsl/mod.rs
JCapucho: Oh right this is for booleans my bad

src/back/hlsl/writer.rs
JCapucho: wow ok, I don't work much with hlsl so I wasn't aware of this stuff, but still if you don't use the constructor anymore why is the parentheses there? Isn't the hlsl backend generating the code having in attention the expression precedence? It seems to me like it especially when you look at `Expression::Swizzle` which is basically what your are outputting but without parentheses.

teoxoy:
> Also this shouldn't have been a monolith PR but a couple of PRs , it makes the life of everyone worse, you don't get as good nor a timely review, fixing the problems in the right places requires rebasing and other shenanigans, the reviewers need to look at a lot of unrelated code, it's more probable of causing/having merging problems with other PRs and CI will only run on the last commit which will either require the reviewers to test every commit or it to be squashed which is bad because we have lot of changes and losing history on it would be bad.
> 
> So in the future if the changes aren't related please open separate PRs, you should also do this for commits, for example you have add support for unary vector operators which could have a been a commit for each backend, which would reduce the diff to only the code and tests changes for that specific backend.
> 
> I only mean this as constructive criticism

While I generally try to keep PRs as focused as possible, this one is in somewhat of a grey area form me. I started by adding a bunch of tests and then with each commit slowly working out the problems. I understand the concerns and can split it up but at least in my opinion it didn't reach a threshold where this is necessary (I find it easier to see what changed and what the fixes/improvements are with PRs of this kind; i.e. where tests are added and fixes are implemented incrementally).

src/back/hlsl/writer.rs
teoxoy: Filed https://github.com/microsoft/DirectXShaderCompiler/issues/4393

teoxoy:
@JCapucho tbh, your first comment left a sour taste in my mouth. It comes off as aggressive even after rereading it a few times and I'm not sure why you felt the need to phrase it like that since this is our first interaction ("I only mean this as constructive criticism" is just a cop-out; if it weren't it wouldn't have been necessary in the first place). Regarding the content; if you feel so strongly about splitting commits and PRs as much as possible I would recommend adding a `CONTRIBUTING.md` document to the root of the repo highlighting this.

JCapucho:
> @JCapucho tbh, your first comment left a sour taste in my mouth. It comes off as aggressive even after rereading it a few times and I'm not sure why you felt the need to phrase it like that since this is our first interaction

I'm very sorry if it does come out as aggressive, English isn't my native language so sometimes I do have trouble expressing myself, I have no ill intent towards you I only wish to make the best shader translation library in the world :)

> ("I only mean this as constructive criticism" is just a cop-out; if it weren't it wouldn't have been necessary in the first place).

I think this is a problem of different cultures, in my mother tongue it's very common to say that after a longer comment on someone's work (and it comes out as aggressive if you don't say it).

> Regarding the content; if you feel so strongly about splitting commits and PRs as much as possible I would recommend adding a `CONTRIBUTING.md` document to the root of the repo highlighting this.

I do feel strongly about it and I have already expressed why, I think it makes everyone's life easier.

Once again I'm very sorry @teoxoy if I sounded aggressive, I truly mean it when I say it wasn't my intention and thanks for pointing it out, I'll try to improve how I communicate.

tests/out/glsl/operators.main.Compute.glsl
```diff
@@ -64,24 +56,156 @@ float constructors() {
     return _e70;
 }
 
-void modulo() {
-    int a_2 = (1 % 1);
-    float b_1 = (1.0 - 1.0 * trunc(1.0 / 1.0));
-    ivec3 c = (ivec3(1) % ivec3(1));
-    vec3 d = (vec3(1.0) - vec3(1.0) * trunc(vec3(1.0) / vec3(1.0)));
+void logical() {
+    bool unnamed_11 = (! true);
```
Gordon-F, line 60: Out of the PR scope, but I think we should remove a redundant extra space after `!` and `-`.

teoxoy:
I appreciate the reply. Communication is hard 😅 - especially between different cultures. All good!

> I only wish to make the best shader translation library in the world :)

Big props! :)

src/back/hlsl/writer.rs
Gordon-F: I think we should leave a comment here about filled issue. Just to make sure that we didn't confuse other contributors.

JCapucho: Ok makes sense to me

Gordon-F (approved):
@teoxoy Looked at glsl, hlsl, and msl changes. SPIR-V scares me, and I don't know much about it :) More tests are great! Thank you!

tests/out/glsl/operators.main.Compute.glsl
teoxoy: Can remove it, the output would be more consistent with the other backends.

JCapucho (approved):
As noted above I didn't review `msl`, but the other parts seem good to me.

PR #1821, opened 2022-04-12
[msl-out] Permit `invariant` qualifier on vertex shader outputs.

jimblandy:
Split into two commits for easier review, but it's pretty short either way.


teoxoy (approved):
Nicely spotted!

PR #1823, opened 2022-04-13
[glsl-out] Emit globals of any type, as required by Naga IR.

### LINKED ISSUE #1811: GLSL backend can't handle matrix-typed globals

jimblandy:
Fixes #1811.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -98,11 +98,12 @@ impl crate::AddressSpace {
     /// Whether a variable with this address space can be initialized
     fn initializable(&self) -> bool {
         match *self {
+            crate::AddressSpace::Function | crate::AddressSpace::Private => true,
             crate::AddressSpace::WorkGroup
             | crate::AddressSpace::Uniform
-            | crate::AddressSpace::PushConstant
-            | crate::AddressSpace::Storage { .. } => false,
-            _ => true,
+            | crate::AddressSpace::Storage { .. }
+            | crate::AddressSpace::Handle
+            | crate::AddressSpace::PushConstant => false,
         }
     }
 }
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -570,27 +570,19 @@ impl<'a, W: Write> Writer<'a, W> {
 
         let ep_info = self.info.get_entry_point(self.entry_point_idx as usize);
 
-        // Write all structs
+        // Write struct types.
         //
-        // This are always ordered because of the IR is structured in a way that you can't make a
-        // struct without adding all of it's members first
+        // This are always ordered because the IR is structured in a way that
+        // you can't make a struct without adding all of its members first.
         for (handle, ty) in self.module.types.iter() {
             if let TypeInner::Struct { ref members, .. } = ty.inner {
-                let generate_struct = match GlobalTypeKind::new(self.module, handle) {
-                    GlobalTypeKind::WrappedStruct => true,
-                    // If it's a global non-wrapped struct, it will be printed
-                    // with the corresponding global variable.
-                    GlobalTypeKind::Unsized(_) => false,
-                    GlobalTypeKind::Other => {
-                        let used_by_global =
-                            self.module.global_variables.iter().any(|(vh, var)| {
-                                !ep_info[vh].is_empty() && var.space.is_buffer() && var.ty == handle
-                            });
-                        // If not used by a global, it's safe to just spew it here
-                        !used_by_global
-                    }
-                };
-                if generate_struct {
+                // Structures ending with runtime-sized arrays can only be
+                // rendered as shader storage blocks in GLSL, not stand-alone
+                // struct types.
+                if !self.module.types[members.last().unwrap().ty]
+                    .inner
+                    .is_dynamically_sized(&self.module.types)
+                {
                     let name = &self.names[&NameKey::Type(handle)];
                     write!(self.out, "struct {} ", name)?;
                     self.write_struct_body(handle, members)?;
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -108,45 +108,6 @@ impl crate::AddressSpace {
     }
 }
 
-#[derive(PartialEq)]
-enum GlobalTypeKind<'a> {
-    WrappedStruct,
-    Unsized(&'a [crate::StructMember]),
-    Other,
-}
-
-impl<'a> GlobalTypeKind<'a> {
-    //Note: similar to `back/spv/helpers.rs`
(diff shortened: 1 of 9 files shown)


jimblandy:
Draft because:

- [ ] needs tests
- [ ] should be split up into cleaner commits

src/back/hlsl/writer.rs
```diff
@@ -153,15 +153,13 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
-                if let Some(member) = members.last() {
-                    if let TypeInner::Array {
-                        size: crate::ArraySize::Dynamic,
-                        ..
-                    } = module.types[member.ty].inner
-                    {
-                        // unsized arrays can only be in storage buffers, for which we use `ByteAddressBuffer` anyway.
-                        continue;
-                    }
+                if module.types[members.last().unwrap().ty]
```
Gordon-F, line 156: An empty struct in input shader will break this code? I think it would be better to use `if let Some(member) = members.last() ....`

src/back/glsl/mod.rs
```diff
@@ -570,27 +532,19 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                            });
-                        // If not used by a global, it's safe to just spew it here
-                        !used_by_global
-                    }
-                };
-                if generate_struct {
+                // Structures ending with runtime-sized arrays can only be
+                // rendered as shader storage blocks in GLSL, not stand-alone
+                // struct types.
+                if !self.module.types[members.last().unwrap().ty]
```
Gordon-F, line 544: One more empty struct problem?

src/back/glsl/mod.rs
```diff
@@ -1021,11 +957,68 @@ impl<'a, W: Write> Writer<'a, W> {
...
+            ty_name,
+            self.block_id.generate(),
+            self.entry_point.stage,
+        );
+        write!(self.out, "{} ", block_name)?;
+        self.reflection_names_globals.insert(handle, block_name);
+
+        match self.module.types[global.ty].inner {
+            crate::TypeInner::Struct { ref members, .. }
+                if self.module.types[members.last().unwrap().ty]
```
Gordon-F, line 994: And more :)

src/back/glsl/mod.rs
```diff
@@ -955,56 +909,38 @@ impl<'a, W: Write> Writer<'a, W> {
...
-            let global_name = self.get_global_name(handle, global);
-            self.reflection_names_globals.insert(handle, global_name);
+            crate::AddressSpace::WorkGroup => {
+                self.write_simple_global(handle, global)?;
+            }
+            crate::AddressSpace::PushConstant => {
+                self.write_simple_global(handle, global)?;
+            }
+            crate::AddressSpace::Uniform | crate::AddressSpace::Handle => {
+                self.write_interface_block(handle, global)?;
```
JCapucho, line 927: Isn't `AddressSpace::Handle` supposed to be used for textures and samplers? In that case using `write_interface_block` doesn't make much sense

src/back/glsl/mod.rs
```diff
@@ -570,27 +532,19 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                    // with the corresponding global variable.
-                    GlobalTypeKind::Unsized(_) => false,
-                    GlobalTypeKind::Other => {
-                        let used_by_global =
-                            self.module.global_variables.iter().any(|(vh, var)| {
-                                !ep_info[vh].is_empty() && var.space.is_buffer() && var.ty == handle
-                            });
-                        // If not used by a global, it's safe to just spew it here
-                        !used_by_global
-                    }
```
JCapucho, line 591: Where did this code go? FIY it's needed because some systems might not support things like dynamic arrays and this code prevents it from being generated if it isn't requested.

JCapucho (changes requested):
Besides what Gordon already pointed out I have some more worries about this changes.

src/back/hlsl/writer.rs
jimblandy: Empty structures are forbidden in WGSL, and I thought the validator enforced this - but it does not. GLSL doesn't allow empty structures either, but SPIR-V does (and Vulkan doesn't seem to add any further restrictions about that).

But for the time being, since we do have empty structs passing through Naga, I'll change this and the other sites.

src/back/glsl/mod.rs
jimblandy: That's not what this code says. It says, if the struct type is *not* used by any global, then the backend should write its definition. So it will cause *unused* structs to be written.

In any case, this code isn't actually reachable, because `GlobalTypeKind::new` will never return `Other` for a `Struct` type.

JCapucho: 🤦🏻 please ignore me

src/back/glsl/mod.rs
jimblandy: You're right - I'm not sure what's going on here. It didn't change any of our snapshots, but I don't understand why not. I'll figure this out and clarify the code.

src/back/hlsl/writer.rs
jimblandy: @kvark suggested just banning empty structs in validation; that's implemented in #1826.

src/back/glsl/mod.rs
jimblandy: Indeed, textures and samplers are handled directly by `Writer::write_global`'s caller, `Writer::write`. I've changed `write_global` to refuse to, uh, handle handles.

src/back/hlsl/writer.rs
Gordon-F: Can we avoid unwrap here and return `EmptyStruct` error instead? In case where validation is disabled (it disabled by default).

JCapucho (approved):
everything looks good on my part now

src/back/hlsl/writer.rs
JCapucho: The general stance backends have over panicking is that it's fine if the IR is illegal there are multiple cases where we panic if the IR isn't good, just grep `unwrap`,  if validation isn't enabled it's the responsibility of the user to ensure it's legal, otherwise the behavior may include panics.

jimblandy: That principle makes sense; I've submitted #1828 to write it down.

Gordon-F (approved):

PR #1826, opened 2022-04-13
Reject empty struct types.

jimblandy:


src/valid/type.rs
```diff
@@ -466,6 +468,10 @@ impl super::Validator {
 
                 let mut prev_struct_data: Option<(u32, u32)> = None;
 
+                if members.is_empty() {
```
Gordon-F, line 471: We can check that `members` is empty before doing extra work :)

Gordon-F (approved):

PR #1827, opened 2022-04-14
add methods on the error types for reusability in wgsl-analyzer

jakobhellermann:
I integrated naga error messages into [wgsl-analyzer](https://github.com/wgsl-analyzer/wgsl-analyzer) and in doing so made a few improvements to the error types allowing me to extract information I need.

Namely:
- expose `message` and `labels` from `frong::wgsl::ParseError`
- add `as_inner(&self) -> &E` to `WithSpan` (`into_inner(self) -> E` already exists)
- add `+ ExactSizeIterator` to the label/span iterators. This is not required but nice to have when checking whether any spans exist at all with `iter.len() == 0`

### DIFF
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -1287,6 +1287,16 @@ pub struct ParseError {
 }
 
 impl ParseError {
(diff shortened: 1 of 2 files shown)


teoxoy (approved):
Looks good, looking forward to see this being used in the next release of `wgsl-analyzer`!

PR #1828, opened 2022-04-14
Document Naga's promises around validation and panics.

jimblandy:


JCapucho (approved):
Not sure if this is the best spot to put them but the contents seems about right to me

jimblandy:
> Not sure if this is the best spot to put them but the contents seems about right to me

Yeah, it's a good point. Would `Validator::new` be better? Or the `valid` module documentation?

JCapucho:
> > Not sure if this is the best spot to put them but the contents seems about right to me
> 
> Yeah, it's a good point. Would `Validator::new` be better? Or the `valid` module documentation?

I think that in terms of api discoverability this would make more sense in the `back` module documentation since this refers to the operation of the backends, like have this part
```rust
// If you do not perform full validation, invalid shaders may
/// cause Naga to panic. If you do perform full validation and
/// [`Validator::validate`] returns `Ok`, then Naga promises that
/// code generation will either succeed or return an error; it
/// should never panic.
```

In the `back` module and the rest where it's (and maybe add a link to the `back` documentation)

PR #1838, opened 2022-04-16
Update MSRV to 1.56

teoxoy:
As mentioned on matrix, `wgpu`'s MSRV has been recently bumped up to 1.59 and I proposed to bump the MSRV of `naga` to 1.56.

Besides API changes, I think thees are the most relevant features for us since 1.43 (our current MSRV):

- [1.56](https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html); rust 2021 (can upgrade in the future), `rust-version` in Cargo.toml, multiple bindings in match arms
- [1.55](https://blog.rust-lang.org/2021/09/09/Rust-1.55.0.html#faster-more-correct-float-parsing); faster, more correct float parsing
- [1.53](https://blog.rust-lang.org/2021/06/17/Rust-1.53.0.html#or-patterns); ability to nest OR patterns in match arms
- [1.51](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#const-generics-mvp); Const Generics MVP
- [1.50](https://blog.rust-lang.org/2021/02/11/Rust-1.50.0.html); const-generic array indexing, `const` value repetition for arrays
- [1.46](https://blog.rust-lang.org/2020/08/27/Rust-1.46.0.html#const-fn-improvements); `const fn` improvements

Tested changes locally via `cargo msrv verify -- cargo test --all-features --workspace` ([cargo-msrv](https://github.com/foresterre/cargo-msrv)).

PR #1841, opened 2022-04-16
Make use of new language features

teoxoy:
Depends on #1838 (will rebase once it lands)


kvark (approved):
nice!
Although, the proliferation of `const` makes me worry a bit if this is a step in the direction of `public static void main()` verbosity level of Java :)

kvark:
@cwfitzgerald should decide if this lands now or after #1845

cwfitzgerald:
Let's land this first, I should touch up that one function anyway and I won't do that until tomorrow night.

PR #1845, opened 2022-04-17
Implement Binding Arrays

### LINKED ISSUE #751: Support arrays of descriptors (textures, samplers, buffers).
Descriptor array, especially textures and samplers are crucial for many rendering approaches.

For example automatic batching is possible to implement by binding textures into array and drawing instances with different materials by indexing into texture array.

With descriptor indexing extension it becomes possible to render whole 2d scene with single drawcall.

cwfitzgerald:
Closes #751

Sorry that this PR is larger than I'd like, I couldn't really land any part of it without other parts, so in an attempt to make things easier to review, I have segmented the PR into sections based on the section of code. While the commits are _not_ standalone and this PR needs to be squashed, **you can review this pr commit by commit**.

This PR implements Binding Arrays and their indexing in the IR, Validator, wgsl-in, and {msl,hlsl,spv}-out.


src/valid/analyzer.rs
```diff
@@ -169,10 +169,18 @@ enum GlobalOrArgument {
 }
 
 impl crate::Expression {
-    fn to_global_or_argument(&self) -> Result<GlobalOrArgument, ExpressionError> {
+    fn to_global_or_argument(
+        &self,
+        expression_arena: &Arena<crate::Expression>,
```
kvark, line 174: this becomes bad with this argument. If the function input is not one expression but multiple ones, it shouldn't be a method on the `Expression` any more. Let's turn it to a freestanding function (accepting an arena and a handle)? Or put it on the arena?

kvark (approved):
I looked at the core parts, and it all seems very reasonable.
Amazing (and titanic) work on your part, excited to see this landing!

teoxoy:
Should we also implement this for the SPIR-V frontend? (related to #1734)

cwfitzgerald:
Ideally both spirv and glsl, but those can be future prs

cwfitzgerald:
I'm going to merge this, and if you want a different representation of that function, I'll do a follow up PR

PR #1847, opened 2022-04-17
[wgsl-in] Update reserved keywords

teoxoy:
https://gpuweb.github.io/gpuweb/wgsl/#keyword-summary


tests/in/image.wgsl
```diff
@@ -97,7 +97,7 @@ fn levels_queries() -> @builtin(position) vec4<f32> {
 var sampler_reg: sampler;
 
 @fragment
-fn sample() -> @location(0) vec4<f32> {
+fn texture_sample() -> @location(0) vec4<f32> {
```
Gordon-F, line 100: Since you rename entry point function, we also should remove unused output snapshot for `glsl-out`. New file for glsl-out generates for each shader stage.

Gordon-F (changes requested):

Gordon-F (approved):

PR #1850, opened 2022-04-17
[hlsl-out] Flesh out matrix handling documentation.

jimblandy:


src/back/hlsl/mod.rs
```diff
@@ -6,29 +6,99 @@ Backend for [HLSL][hlsl] (High-Level Shading Language).
...
 
-# Matrix struct member of the form `matCx2` Note
+However, with a bit of trickery, it is possible to use HLSL's `m[i]`
+as the translation of WGSL's `m[i]`:
 
-Struct member matrices of the form `matCx2` are translated to a sequence of C `vec2`s due to
-differences in alignment between WGSL and HLSL for uniform buffers.
+- We declare all matrices in uniform buffers in HLSL with the
+  `row_major` qualifier, and transpose the row and column counts: a
+  WGSL `mat3x4<f32>`, say, becomes an HLSL `row_major float4x3`. Since
```
teoxoy, line 37: A WGSL `mat3x4<f32>` becomes an HLSL `row_major float3x4` (not `float4x3`) due to HLSL's matrix notation being floatRxC (i.e. inverted relative to WGSL's).

src/back/hlsl/mod.rs
```diff
@@ -6,29 +6,99 @@ Backend for [HLSL][hlsl] (High-Level Shading Language).
...
+
+- For vector-matrix multiplication, since `mul(transpose(m), v)` is
+  equivalent to `mul(v, m)` (note the reversal of the arguments), and
+  `mul(v, transpose(m))` is equivalent to `mul(m, v)`, we can
+  translate WGSL `m * v` and `v * m` to HLSL by simply reversing the
+  arguments to `mul`.
+
+## Padding in two-row matrices
+
+An HLSL `row_major floatRx2` matrix has padding between its rows that
```
teoxoy, line 52: Should be `row_major float2xC`.

teoxoy (approved):
Besides the 2 comments, this is really well written!

src/back/hlsl/mod.rs
jimblandy: Good catch, thank you!

src/back/hlsl/mod.rs
jimblandy: Are you sure? It's stored in row major order, so it's a series of rows. There is padding, so the padding must be between one row and the next. In order for there to be padding not present in WGSL, each row must be two values long. Hence, the matrix must have two columns.

jimblandy: I think it's the WGSL type on the next line that's wrong: an HLSL `row_major floatRx2` matrix would represent a WGSL `mat2xC` matrix, not a `matCx2`.

jimblandy: I was confused - the original `matCx2` was correct. If what's stored in memory is some series of two-element vectors (and HLSL and WGSL disagree about the padding between them), then since WGSL uses a column-major storage order, those two-element vectors must each be a WGSL `vec2<f32>`, of which there are `C`. That's `matCx2`.

jimblandy:
[[hlsl-out] Flesh out matrix handling documentation. naga#1850](https://trello.com/c/RYr1Cjoq/1-hlsl-out-flesh-out-matrix-handling-documentation-naga1850)

src/back/hlsl/mod.rs
teoxoy: Ah yes it should be right in that order (`row_major floatRx2`). I got confused about the naming since I was expecting `C` to appear somewhere (akin to `matCx2`).

I think a shared value between the two notations would be helpful here (something like `row_major floatKx2`, `matKx2`).

What do you think?

jimblandy: excellent suggestion

PR #1853, opened 2022-04-18
Replace `unicode-id` with `unicode-xid`

teoxoy:
In 743c8aa387cbfbd3c47e6097c8c2694b3f6ad0a0 I mistakenly used `unicode-id` instead of `unicode-xid`.

Unicode defines both `ID_{Start,Continue}` and `XID_{Start,Continue}` in [Table_Lexical_Classes_for_Identifiers](https://www.unicode.org/reports/tr31/tr31-35.html#Table_Lexical_Classes_for_Identifiers).


jimblandy (approved):
Looks good to me.

PR #1855, opened 2022-04-19
Binding arrays play nice with bounds checks

cwfitzgerald:
Found an issue with binding indexing in the MSL backend when bounds checks were on.

This fixes the issue and adds a binding array bounds check policy that, for now, must be unchecked.

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -1270,13 +1270,14 @@ impl<W: Write> Writer<W> {
         let expression = &context.function.expressions[expr_handle];
         log::trace!("expression {:?} = {:?}", expr_handle, expression);
         match *expression {
-            crate::Expression::Access { .. } | crate::Expression::AccessIndex { .. } => {
+            crate::Expression::Access { base, .. }
+            | crate::Expression::AccessIndex { base, .. } => {
                 // This is an acceptable place to generate a `ReadZeroSkipWrite` check.
                 // Since `put_bounds_checks` and `put_access_chain` handle an entire
                 // access chain at a time, recursing back through `put_expression` only
                 // for index expressions and the base object, we will never see intermediate
                 // `Access` or `AccessIndex` expressions here.
-                let policy = context.choose_bounds_check_policy(expr_handle);
+                let policy = context.choose_bounds_check_policy(base);
                 if policy == index::BoundsCheckPolicy::ReadZeroSkipWrite
                     && self.put_bounds_checks(
                         expr_handle,
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -3340,11 +3340,19 @@ impl<W: Write> Writer<W> {
                     }
                     if let Some(ref br) = var.binding {
                         let good = match options.per_stage_map[ep.stage].resources.get(br) {
-                            Some(target) => match module.types[var.ty].inner {
-                                crate::TypeInner::Image { .. } => target.texture.is_some(),
-                                crate::TypeInner::Sampler { .. } => target.sampler.is_some(),
-                                _ => target.buffer.is_some(),
-                            },
+                            Some(target) => {
+                                let binding_ty = match module.types[var.ty].inner {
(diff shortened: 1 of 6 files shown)


kvark:
@jimblandy please review

src/proc/index.rs
```diff
@@ -140,7 +144,13 @@ impl BoundsCheckPolicies {
...
         info: &valid::FunctionInfo,
     ) -> BoundsCheckPolicy {
-        match info[access].ty.inner_with(types).pointer_space() {
+        let ty = info[access].ty.inner_with(types);
+
+        if let crate::TypeInner::BindingArray { .. } = *ty {
+            return self.binding_array;
+        }
+
+        match ty.pointer_space() {
```
jimblandy, line 153: This is fine (especially since the SPIR-V back end was already passing the base, not the access itself), but please rename `access` to `base`, and change the docs to say:

```
    /// Determine which policy applies to an indexed access of `base`.
    ///
    /// `base` is the base of an `Access` or `AccessIndex` expression,
    /// either a pointer to a value, or a value being indexed directly.
    ///
    /// See the documentation for [`BoundsCheckPolicy`] for details about
    /// when each policy applies.
```

tests/out/msl/binding-arrays.msl
```diff
@@ -146,11 +152,17 @@ fragment main_Output main_(
...
     metal::float4 _e250 = v4_;
-    texture_array_storage[uniform_index].write(_e250, metal::uint2(pix));
+    if (metal::all(metal::uint2(pix) < metal::uint2(texture_array_storage[uniform_index].get_width(), texture_array_storage[uniform_index].get_height()))) {
+        texture_array_storage[uniform_index].write(_e250, metal::uint2(pix));
+    }
     metal::float4 _e252 = v4_;
-    texture_array_storage[non_uniform_index].write(_e252, metal::uint2(pix));
+    if (metal::all(metal::uint2(pix) < metal::uint2(texture_array_storage[non_uniform_index].get_width(), texture_array_storage[non_uniform_index].get_height()))) {
+        texture_array_storage[non_uniform_index].write(_e252, metal::uint2(pix));
+    }
```
jimblandy, line 165: This doesn't seem to be checking the value of `non_uniform_index` at all - isn't that what `BoundsCheckPolicies::binding_array` is supposed to apply to?

jimblandy (changes requested):
It looks to me like the generated MSL code isn't right.

tests/out/msl/binding-arrays.msl
cwfitzgerald: Correct. I haven't actually implemented bounds checking for the index into the array yet. This was merely to get binding indexing to play nice with existing bounds checks, and part of that was adding a new policy (that only works on unrestricted right now) so that the choose policy works correctly.

src/back/msl/writer.rs
```diff
@@ -3339,11 +3340,19 @@ impl<W: Write> Writer<W> {
...
+                                        &module.types[base].inner
+                                    }
+                                    ref ty => ty,
+                                };
+                                match *binding_ty {
+                                    crate::TypeInner::Image { .. } => target.texture.is_some(),
+                                    crate::TypeInner::Sampler { .. } => target.sampler.is_some(),
+                                    _ => target.buffer.is_some(),
+                                }
+                            }
```
jimblandy, line 3355: This seems like an unrelated fix. It's fine to include it in this PR, but if it has something to do with bounds checks, let me know, because I'm not understanding the patch.

jimblandy (approved):
One question to check over, but otherwise, looks good to me! Thank you very much.

src/back/msl/writer.rs
cwfitzgerald: Ah yeah, sorry this should have been titled "fixes needed to get the binding indexing wgpu PR working" :)

tl;dr: Because of the change above this, it is now possible to get the type of binding array, and we need to "punch through" the binding array to find the actual type of binding we're looking at.

PR #1862, opened 2022-04-25
Clarify accepted types for `Expression::AccessIndex`

NoelTautges:
The docs for `AccessIndex` only note it can be used with arrays ("Array access with a known index.") and don't mention [it can be used with structs or other types](https://github.com/gfx-rs/naga/blob/1aa9154964238af8c692cf521ff90e1f2395e147/src/valid/expression.rs#L227).

This pull request documents the full list of types relative to those for `Expression::Access`, which already have a more extensive blurb written about them.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1108,14 +1108,19 @@ pub enum Expression {
         base: Handle<Expression>,
         index: Handle<Expression>,
     },
-    /// Array access with a known index.
+    /// Access the same types as [`Access`] plus [`Struct`] with a known index.
+    /// 
+    /// [`Array`]s must have a [`Constant`] size.
+    /// 
+    /// [`Access`]: Expression::Access
+    /// [`Struct`]: TypeInner::Struct
+    /// [`Array`]: TypeInner::Array
+    /// [`Constant`]: ArraySize::Constant
     AccessIndex {
         base: Handle<Expression>,
         index: u32,
     },
     /// Constant value.
-    ///
-    /// Every `Constant` expression
     Constant(Handle<Constant>),
     /// Splat scalar into a vector.
     Splat {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1108,14 +1108,19 @@ pub enum Expression {
         base: Handle<Expression>,
         index: Handle<Expression>,
     },
-    /// Array access with a known index.
+    /// Access the same types as [`Access`] plus [`Struct`] with a known index.
+    ///
+    /// [`Array`]s must have a [`Constant`] size.
```
jimblandy, line 1113: Is this true? I couldn't find code in the validator that would reject dynamically-sized arrays.

src/lib.rs
```diff
@@ -1108,14 +1108,19 @@ pub enum Expression {
...
+    /// [`Struct`]: TypeInner::Struct
+    /// [`Array`]: TypeInner::Array
+    /// [`Constant`]: ArraySize::Constant
     AccessIndex {
         base: Handle<Expression>,
         index: u32,
     },
     /// Constant value.
-    ///
-    /// Every `Constant` expression
```
jimblandy, line 1118: agreed

jimblandy (changes requested):
Thank you for this!

I have one concern.

src/lib.rs
NoelTautges: The validator [sets a bound of the size of the array](https://github.com/gfx-rs/naga/blob/1aa9154964238af8c692cf521ff90e1f2395e147/src/valid/expression.rs#L223) if its size is constant, but sets it to `!0` if it's not, which somehow evaluates to -1, making the bounds check always fail. At least in theory, I haven't explicitly tested this.

cwfitzgerald: It makes the compile time bounds check always _succeed_ because `!0` as a u32 is `u32::MAX` so no index is ever out of bounds. The code could be made easier to grok by using `u32::MAX` explicitly.

jimblandy: (Agreed that the code is not as clear as it might be.)

jimblandy (approved):

Issue #1843, opened 2022-04-16
[wgsl-in] Numeric suffixes + unsanitised related error


hanawatson:
Hi - I noticed that naga doesn't seem to correctly parse numeric suffixes (except for `u` for unsigned integers u32). According to the spec, WGSL theoretically also allows `i` (i32), `f` (f32) and `h` (f16, although f16 support/extension seems to be very new to the spec).

code example (analogous behavior for the `i` suffix and the `f` without exponent):
```
@stage(compute) @workgroup_size(1)
fn compute_main(
) {
	var var0: f32 = -2.9007446E38f;
}
```

resultant error:
```
error: expected ';', found 'f'
  ┌─ test.wgsl:6:31
  │
6 │     var var0: f32 = -2.9007446E38f;
  │                                  ^ expected ';'

Could not parse WGSL
```

Also (sorry to bombard you with issues!) - while checking the special `0x1f` case just above this section of the spec https://gpuweb.github.io/gpuweb/wgsl/#keywords, I got an error that I suspect might not have been meant to be user-exposed.

code:
```
@stage(compute) @workgroup_size(1)
fn compute_main(
) {
	var var0: f32 = 0x1f;
}
```

error:
```
[2022-04-16T19:02:13Z ERROR naga::front::wgsl] Given type Scalar { kind: Float, width: 4 } doesn't match expected Scalar { kind: Sint, width: 4 }
error: the type of `var0` is expected to be `i32`
  ┌─ test1.wgsl:4:6
  │
4 │     var var0: f32 = 0x1f;
  │         ^^^^ definition of `var0`

Could not parse WGSL
```

PR #1863, opened 2022-04-25
[wgsl-in] Overhaul number lexing / parsing

teoxoy:
Updated number literal format and behavior to latest spec version.

Reduced the number errors to the following errors:

```
error: `-0x1f.p500f` numeric literal not representable by target type
  ┌─ test.wgsl:7:18
  │
7 │     var var0: f32 = -0x1f.p500f;
  │                     ^^^^^^^^^^^ numeric literal not representable by target type
```

```
error: `1h` unimplemented f16 type
  ┌─ test.wgsl:7:18
  │
7 │     var var0: f32 = 1h;
  │                     ^^ unimplemented f16 type
```

```
error: `` invalid numeric literal format
  ┌─ test.wgsl:7:18
  │
7 │     var var0: f32 = 0xx;
  │                     ^ invalid numeric literal format
```
(the span of the invalid error can't be more accurate than the first character since we don't know where it ends)

fixes https://github.com/gfx-rs/naga/issues/1843
gets us closer to https://github.com/gpuweb/gpuweb/pull/2227

Related
https://github.com/gpuweb/gpuweb/pull/2762
https://github.com/gpuweb/gpuweb/issues/2769


jimblandy:
I wish `hexf_parse` provided an API that just let you hand over the segments of text you'd identified.

jimblandy:
@teoxoy Have you profiled the effect of using regular expressions to parse numeric literals? You can get some big shaders from the sources given in `.github/workflows/lazy.yml`.

src/front/wgsl/lexer.rs
```diff
@@ -1,267 +1,237 @@
...
+            // 0[xX] and [pP] is required by hexf
+            let hexf_input = &format!(
+                "{}0x{}{}",
+                sign,
+                significand,
+                if exponent.is_empty() { "p0" } else { exponent }
+            );
+
+            (parse_hex_float(hexf_input, kind), rest)
+        } else if let Some((_, sign, significand, exponent, kind, rest)) = lazy_regex::regex_captures!(r"
```
kvark, line 158: Is using "lazy" semantics here really warranted? This is effectively an implicit global state with initialization at the first run. Since we know all of the regex patterns in advance, alternatively we could just build them all at the creation of WGSL parser and store with it. This would make our first shader compilation predictable, and as a small bonus avoid an extra dependency on `lazy_regex`.

kvark (commented):
Thank you for improving the number parsing!

We are generally trying hard to keep the code simple and dependencies to the minimum. So involving `regex` should come with a good justification. I suppose in case of your PR the justification looks like this:
  1. it fixes a bunch of cases
  2. WGSL spec uses regular expressions already, so we can encode it more directly

If that sounds right to you, please consider building regexes in a non-lazy way.

src/front/wgsl/lexer.rs
teoxoy: I've updated the code in this direction. The fact that we now need to pass the `NumberRegexes` struct around isn't as nice as `lazy_regex`'s API but I agree it's not really warranted to add another dep. (Also `lazy_regex` doesn't support `RegexSet`s)

jimblandy:
@teoxoy One comment in my review (I'll finish it soon) was, I wonder if the regexps ending in `(.*)` to capture the `rest` might be scanning all the way to the end of the input, each time we parse a literal. But I see your most recent version has taken those out. Was the 60% impact measured with, or without, the `(.*)` rest matchers?

jimblandy:
> Do you guys know of other ways that we could optimize the regex approach?

tl;dr: I don't think so, unfortunately.

I think it would have to entail using some sort of build-time code generator. We've generally shied away from those because of the build-time impact (`serde` and `thiserror` being exceptions), since that affects developers constantly.

I've also noticed a tendency for projects to move away from code generators in favor of hand-written parsers. This is pretty disappointing to me. The reason seems to be that generators have to get a whole range of things Just Right to be competitive with hand-written code:
- flexibility (real-world grammars are not well-behaved; remember having to write `> >` in C++ all the time?)
- error messages (i.e. statically detectable problems)
- debuggability (i.e. dynamically detectable problems)
- runtime performance
- speed of the generator itself
- generated code size (rustc is slow)

Generators let you write your code at a higher-level, but coders have this "that's hard, but I could do it" mindset that undervalues that. Generators generate more correct code, but the use of safe Rust makes this less crucial than it used to be.

teoxoy:
ping @jimblandy (since it seems I can't re-request a review)

src/front/wgsl/number.rs
```diff
@@ -0,0 +1,506 @@
...
+        }
+        (output, input.len() - output.len())
+    }
+
+    const fn to_str(bytes: &[u8]) -> &str {
+        // SAFETY: the bytes are valid UTF-8 since the buffer
+        // was created from a &str and we only advance
+        // it by one byte codepoints
+        unsafe { core::str::from_utf8_unchecked(bytes) }
+    }
```
jimblandy, line 169: Instead of the unsafe code, please use the length of `bytes` versus the length of `input` in bytes to do a fresh slice of `input`.

src/front/wgsl/number.rs
```diff
@@ -0,0 +1,506 @@
...
+                let (bytes, consumed) = consume_dec_digits(bytes);
+
+                if consumed == 0 {
+                    return None;
+                }
+
+                bytes
+            } else {
+                bytes
+            };
```
jimblandy, line 309: My thought was that letting `consume!` update bytes as a side effect would make odd plumbing like this go away.

src/front/wgsl/number.rs
```diff
@@ -0,0 +1,506 @@
...
+    };
+    (Token::Number(num), res.1)
+}
+
+struct ExtractInfo<'a> {
+    info: Info<'a>,
+    rest: &'a str,
+}
+
+enum Info<'a> {
```
jimblandy, line 54: I don't think this enum is earning its keep. It looks to me as if you could:
1) replace each variant with a function (i.e. `parse_hex_float_missing_exponent`) whose arguments are the variant's fields
2) call those functions from `extract_info` instead of constructing the variants, and
3) eliminate the `map` function altogether.

The comments on the variants here would make sense as doc comments on the functions.

jimblandy (changes requested):
This is pretty intricate. Thanks for taking this on!

I have a few changes I'd like to see and then re-review.

teoxoy:
@jimblandy thanks for the review and suggestions - they were great! I addressed the comments, simplified some things further and added a few more docs (I know you like those 😉).

src/front/wgsl/number.rs
```diff
@@ -0,0 +1,438 @@
...
+
+// hex_float_literal:
+// | / 0[xX][0-9a-fA-F]*    \.[0-9a-fA-F]+      ([pP][+-]?[0-9]+        [fh]?)? /
+// | / 0[xX][0-9a-fA-F]+    \.[0-9a-fA-F]*      ([pP][+-]?[0-9]+        [fh]?)? /
+// | / 0[xX][0-9a-fA-F]+                         [pP][+-]?[0-9]+        [fh]?   /
+
+// You could visualize the regex below via https://debuggex.com to get a rough idea what `extract_info` is doing
+// -?(?:0[xX](?:([0-9a-fA-F]+\.[0-9a-fA-F]*|[0-9a-fA-F]*\.[0-9a-fA-F]+)(?:([pP][+-]?[0-9]+)([fh]?))?|([0-9a-fA-F]+)([pP][+-]?[0-9]+)([fh]?)|([0-9a-fA-F]+)([iu]?))|((?:[0-9]+[eE][+-]?[0-9]+|(?:[0-9]+\.[0-9]*|[0-9]*\.[0-9]+)(?:[eE][+-]?[0-9]+)?))([fh]?)|((?:[0-9]|[1-9][0-9]+))([iufh]?))
+
+fn parse(input: &str) -> Option<(Result<Number, NumberError>, &str)> {
```
jimblandy, line 86: It seems like this could just return `Result<Number, NumberError>`, and replace the `None` returns with `Err(NumberError::Invalid)`, and then `consume_number` could just use the `?` operator. Does that work?

src/front/wgsl/mod.rs
```diff
@@ -210,25 +192,12 @@ impl<'a> Error<'a> {
...
-                                format!(
-                                    "{} literal of arbitrary width",
-                                    literal_ty_str,
-                                )
-                            }
+                        ExpectedToken::Number(ty) => {
+                            match ty {
+                                NumberType::I32 => "32bit signed integer literal",
+                                NumberType::U32 => "32bit unsigned integer literal",
+                                NumberType::F32 => "32bit floating-point literal",
```
jimblandy, line 199: Let's spell this "32-bit".

src/front/wgsl/mod.rs
```diff
@@ -258,21 +227,13 @@ impl<'a> Error<'a> {
...
-                ),
-                labels: vec![(bad_span.clone(), "expected unsigned integer".into())],
-                notes: vec![err.to_string()],
-            },
-            Error::BadI32(ref bad_span, ref err) => ParseError {
+            Error::BadNumber(ref bad_span, ref err) => ParseError {
                 message: format!(
-                    "expected integer literal, found `{}`",
-                    &source[bad_span.clone()],
+                    "`{}` {}",
```
jimblandy, line 232: I think this is going to produce error messages like:

> \`10.0h\` unimplemented f16 type

which is a little odd. It seems like

> unimplemented f16 type: \`10.0h\`

would be better.

src/front/wgsl/mod.rs
```diff
@@ -2305,10 +2233,14 @@ impl Parser {
...
-                let const_handle =
-                    self.parse_const_expression_impl(token, lexer, None, ctx.types, ctx.constants)?;
+            (Token::Word("true" | "false") | Token::Number(..), _) => {
+                let const_handle = self.parse_const_expression_impl(
+                    lexer.next(),
+                    lexer,
+                    None,
+                    ctx.types,
+                    ctx.constants,
+                )?;
```
jimblandy, line 2243: I think this is just `self.parse_const_expression(lexer, None, ctx.types, ctx.constants)`.

src/front/wgsl/mod.rs
```diff
@@ -4269,11 +4200,6 @@ impl Parser {
                             }
                         }
                     }
-                    for size in workgroup_size.iter_mut() {
-                        if *size == 0 {
-                            *size = 1;
-                        }
-                    }
```
jimblandy, line 4276: So the difference in behavior here is that if `parse_generic_non_negative_int_literal` returns `0`, the old code will correct it to a `1`, whereas the new code leaves it zero. Is that change intentional?

src/front/wgsl/mod.rs
```diff
@@ -2853,15 +2785,13 @@ impl Parser {
                 match lexer.next_ident_with_span()? {
                     ("size", _) => {
                         lexer.expect(Token::Paren('('))?;
-                        let (value, span) = lexer
-                            .capture_span(|lexer| parse_non_negative_sint_literal(lexer, 4))?;
+                        let (value, span) = lexer.capture_span(parse_non_negative_i32_literal)?;
```
jimblandy, line 2788: I love all these changes

src/front/wgsl/number_literals.rs
```diff
@@ -1,133 +1,18 @@
...
-        other => Err(Error::Unexpected(
-            other,
-            ExpectedToken::Number {
-                ty: Some(NumberType::Uint),
-                width: Some(width),
-            },
-        )),
-    }
-}
+use super::{lexer::Lexer, number::Number, Error, ExpectedToken, NumberType, Token};
```
jimblandy, line 3: The `number_literals.rs` module is pretty small now. Could we fold it into `mod.rs` or `number.rs`?

src/front/wgsl/lexer.rs
```diff
@@ -641,41 +385,195 @@ fn sub_test(source: &str, expected_tokens: &[Token]) {
...
 
+#[test]
+fn test_numbers() {
+    // WGSL spec examples //
+
+    // decimal integer
+    sub_test(
+        "0x123 0X123u 1u 123 0 0i 0x3f",
+        &[
+            Token::Number(Ok(Number::I32(291))),
```
jimblandy, line 396: Unless these tests now need access to something private to `lexer.rs`, I'd rather they stayed in `tests.rs`. But I haven't tried this - if there's some reason it's difficult to have them in `tests.rs`, then that's fine. (Ideally, the tests in `lexer.rs` should move to `tests.rs`, too.)

jimblandy (changes requested):
This looks really nice. Just a few more points I'd like to see addressed.

src/front/wgsl/number.rs
teoxoy: Since the rest is also returned I wasn't sure if it was the responsibility of `parse` to give back the full input string. Another option would be: `(Result<Number, NumberError>, Option<&str>)`. What do you think?

src/front/wgsl/number.rs
```diff
@@ -0,0 +1,438 @@
...
+    /// returns `true` and consumes `X` bytes from the given byte buffer
+    /// if the given `X` nr of patterns are found at the start of the buffer
+    macro_rules! consume {
+        ($bytes:ident, $($($pattern:pat)|*),*) => {
```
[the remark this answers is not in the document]
teoxoy: Seems so https://doc.rust-lang.org/edition-guide/rust-2021/or-patterns-macro-rules.html

src/front/wgsl/lexer.rs
teoxoy: I mainly moved them here so that we can test the output and how the tokens are generated.

i.e.
`0x123` should be a concrete number literal with value `291` and not a number `0` followed by an `x` followed by a 123 number literal

src/front/wgsl/number.rs
jimblandy: I tried making this change myself and realized I'd missed a bunch of details. There are really two levels of error here:

1) We may fail to recognize a numeric literal at all. In this case, the token returned by `consume_number` is `Token::Number(Err(NumberError::Invalid))`, and we don't advance the input.

2) We may recognize a numeric literal, but it may not be representable, or it may be an unimplemented type like `f16`. In this case, the token returned by `consume_number` is `Token::Number(Err(NumberError:: something else))`, and we do advance the input to the next token.

In the long term, we would like Naga to be able to return more than one error message at a time. I think that means that, since we're here, `parse` should always return a rest string, even when we fail to recognize any literal at all, and try to advance us across the initial run of characters that looked okay. (There should always be such an initial run, or else the lexer wouldn't have called `consume_number`.)

So I think `parse` should return `(Result<Number, NumberError>, &str)`, and then `consume_number` should look like this:

```
pub(super) fn consume_number(input: &str) -> (Token<'_>, &str) {
    let (result, rest) = parse(input);
    (Token::Number(result.and_then(Number::abstract_to_concrete)), rest)
}
```

where:

```
impl Number {
    fn abstract_to_concrete(self) -> Result<Number, NumberError> {
        match self {
            Number::AbstractInt(num) => {
                use std::convert::TryFrom;
                i32::try_from(num)
                    .map(Number::I32)
                    .map_err(|_| NumberError::NotRepresentable)
            }
            Number::AbstractFloat(num) => {
                let num = num as f32;
                if num.is_finite() {
                    Ok(Number::F32(num))
                } else {
                    Err(NumberError::NotRepresentable)
                }
            }
            num => Ok(num),
        }
    }
}
```

And then instead of `return None`, `parse` would say something like:

```
return (Err(NumberError::Invalid), rest_to_str!(bytes));
```

src/front/wgsl/mod.rs
jimblandy: Okay, I see - nice catch.

jimblandy:
@teoxoy Since I had misunderstood some bits about the behavior of `consume_number` in my previous comments, this time I tried to put my money where my mouth is and actually implement my suggested change. It seems to work fine. If you'd like to look it over, I can push it to your branch (thanks, GitHub!) and if it seems all right, we can squash, rebase, and merge. If you don't like it, then we can just merge the PR as-is.

teoxoy:
@jimblandy feel free to push the change; I think the reasoning for it makes sense 👍

jimblandy (approved):

Issue #1842, opened 2022-04-16
[wgsl-in] Phony assignment


hanawatson:
Hi, phony assignments in WGSL input seem to not be recognised by naga currently

code example (slightly altered copy of an example in the spec https://gpuweb.github.io/gpuweb/wgsl/#phony-assignment-section, since as this is an example use it should presumably definitely work):
```
var <private> counter: i32;

fn increment_and_yield_previous() -> i32 {
	let previous = counter;
	counter = counter + 1;
	return previous;
}

@stage(compute) @workgroup_size(1)
fn compute_main(
) {
	_ = increment_and_yield_previous();
}
```

resultant error:
```
error: unknown local function `_`
   ┌─ phonytest.wgsl:12:2
   │
12 │     _ = increment_and_yield_previous();
   │     ^ unknown local function

Could not parse WGSL
```

teoxoy:
We currently allow the `_` identifier for let and var assignments and this works like phony assignment but according to the spec (and this issue) we shouldn't require let/var since the `_` is not supposed to be an identifier and therefore not able to be declared via let/var.

stshine:
It is strange for phony assignment not requiring `let`.

teoxoy:
@stshine see https://github.com/gpuweb/gpuweb/pull/2127 for the reasons behind it

PR #1866, opened 2022-04-26
[wgsl-in] Implement phony assignment

teoxoy:
fixes https://github.com/gfx-rs/naga/issues/1842


stshine:
It feels really unfortunate to introduce so many changes for a single syntax ...
What if the spec changed to keep up with how rust handle phony assignment?

tests/out/wgsl/operators.wgsl
```diff
@@ -32,175 +32,164 @@ fn bool_cast(x: vec3<f32>) -> vec3<f32> {
...
+    let unnamed_104 = (2 == 1);
+    let unnamed_105 = (2u == 1u);
+    let unnamed_106 = (2.0 == 1.0);
+    let unnamed_107 = (vec2<i32>(2) == vec2<i32>(1));
+    let unnamed_108 = (vec3<u32>(2u) == vec3<u32>(1u));
+    let unnamed_109 = (vec4<f32>(2.0) == vec4<f32>(1.0));
+    let unnamed_110 = (2 != 1);
+    let unnamed_111 = (2u != 1u);
+    let unnamed_112 = (2.0 != 1.0);
+    let unnamed_113 = (vec2<i32>(2) != vec2<i32>(1));
```
kvark, line 166: I think we know how many times a particular expression is used. So we could just generate `_ = xxx` if it's not used, and the name isn't specified

kvark (approved):

teoxoy:
@stshine the first time I used search and replace to replace `let _ =` with `_ =` I forgot to only filter wgsl files and was surprised to see that the code still compiled. As far as I can see rust also supports `_ =`.

tests/out/wgsl/operators.wgsl
teoxoy: I'm now thinking a new kind of statement for phony assignment might be better (i.e. `Phony(Handle<Expression>)`) since using named_expressions with _ feels like a hack. What do you think?

stshine:
@teoxoy Oh that's something TIL

tests/out/wgsl/operators.wgsl
kvark: It's not supposed to be a named expression at all. It's just an expression that is emitted at a particular point.
I don't think this warrants an IR change, we shouldn't be introducing a new statement type.

Issue #802, opened 2021-04-29
Handle differences between the glsl `mod` function and the `metal::fmod` function.


expenses:
Utgh.

When using `spirv-cross` on a spir-v file that uses `OpFMod`, it emits the following metal function:
```metal
// Implementation of the GLSL mod() function, which is slightly different than Metal fmod()
template<typename Tx, typename Ty>
inline Tx mod(Tx x, Ty y)
{
    return x - y * floor(x / y);
}
```
Currently we emit `metal::fmod` for `OpFMod` functions. This results in some glitches in shaders. For example, the shadertoy shader https://www.shadertoy.com/view/XsXXDn looks like this with `spirv-cross` (using https://github.com/repi/shadertoy-browser/pull/63):

<img width="1025" alt="Screen Shot 2021-04-29 at 13 27 05" src="https://user-images.githubusercontent.com/13566135/116544164-105b4880-a8ef-11eb-8053-e7af602245e3.png">
But like this with naga:

<img width="1025" alt="Screen Shot 2021-04-29 at 13 27 47" src="https://user-images.githubusercontent.com/13566135/116544314-43054100-a8ef-11eb-88d7-fa366aceb09c.png">

expenses:
From https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf table 6.4, it looks like metal uses `x - y * trunc(x / y)`, and `trunc` has slightly different semantics. We should probable have a comment saying what the semantics of the IR is meant to be.

kvark:
Oh I see, so this is about the sign. Filed https://github.com/gpuweb/gpuweb/issues/1696
Let's change our SPIR-V frontend/backend to use `OpFRem` instead of `OpFMod` here.

PR #1867, opened 2022-04-27
Modulo improvements

teoxoy:
With these changes we should be on par with the WGSL spec (not including handling undefined behavior in the back-ends; added TODOs for those).

fixes https://github.com/gfx-rs/naga/issues/802

Commits are self contained - no need to squash.


kvark (approved):
top work!

PR #1869, opened 2022-04-28
Remove named expression entry for phony assignment

teoxoy:
Follow up to https://github.com/gfx-rs/naga/pull/1866#discussion_r859374751


teoxoy:
The GLSL frontend is generating some extra expressions that end up being emitted as phony assignments.
I tried to look into it without much success; @JCapucho do you have any pointers to what might be the cause?

JCapucho:
The glsl frontend generates a lot of expressions that aren't needed because it's generally easier to do it that way than to add them as needed later, but I do see some weird ones there that I'll need to check out.

kvark (approved):

PR #1872, opened 2022-04-29
Rename type flags to better align with the current WGSL spec.

jimblandy:
TypeFlags::INTERFACE -> TypeFlags::IO_SHARED
(WGSL §4.4.4's "IO-shareable types")

TypeFLags::HOST_SHARED -> TypeFlags::HOST_SHAREABLE
(WGSL §4.4.5's "Host-shareable types")

### DIFF
--- a/src/valid/interface.rs
+++ b/src/valid/interface.rs
@@ -364,7 +364,7 @@ impl super::Validator {
                         ));
                     }
                 }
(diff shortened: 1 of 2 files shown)


jimblandy:
These are nowhere used in wgpu, so renames shouldn't cause trouble.

teoxoy (approved):
Small but great change, lowers the barrier of entry/understanding!

PR #1874, opened 2022-04-30
Simplify type handling in VaryingContext.

jimblandy:
When `VaryingContext::validate` discovers that it needs to examine each member of a struct, pass the type as an argument to `validate` and `validate_impl`, rather than mutating `self.ty` on each iteration of the loop. This matches the way we're handling the binding, and avoids stashing per-iteration state in a surprising place.

This should have no effect on execution.


teoxoy (approved):
LGTM

PR #1875, opened 2022-04-30
Validate `ValuePointer` exactly like a `Pointer` to a `Scalar`.

jimblandy:
A `ValuePointer` type should always be handled exactly like the equivalent `Pointer` to `Scalar` type. To this end:

- Adjust the handling of `TypeFlags::ARGUMENT`.

- Neither `Pointer` nor `ValuePointer` types are ever `DATA`.

### DIFF
--- a/src/valid/type.rs
+++ b/src/valid/type.rs
@@ -323,12 +323,28 @@ impl super::Validator {
                 size: _,
                 kind,
                 width,
-                space: _,
+                space,
             } => {
+                // ValuePointer should be treated the same way as the equivalent
+                // Pointer / Scalar / Vector combination, so each step in those
+                // variants' match arms should have a counterpart here.
+                //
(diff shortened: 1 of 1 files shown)


src/valid/type.rs
```diff
@@ -323,12 +323,28 @@ impl super::Validator {
                 size: _,
                 kind,
                 width,
-                space: _,
+                space,
             } => {
+                // ValuePointer should be treated the same way as the equivalent
+                // Pointer / Scalar combination, so each step in the `Scalar`
+                // and `Pointer` match arms should have a counterpart here.
```
teoxoy, line 330: nit: ValuePointer is a Pointer to a Scalar or Vector.

teoxoy (approved):
Besides the comment, looks good!

PR #1876, opened 2022-04-30
Change `!0` to equivalent results for readability

NoelTautges:
[What can I say, it screwed me up!](https://github.com/gfx-rs/naga/pull/1862#discussion_r857831242) It didn't register that `!` in Rust performs both boolean and bitwise [negation](https://doc.rust-lang.org/std/ops/trait.Not.html), unlike other languages where it's `!`/`not` vs `~`.

This PR changes all instances of `!0` to equivalent results (`<type>::MAX` or `-1`, depending on signedness) to improve readability and prevent such errors in the future.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -340,7 +340,7 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
             println!("Validation successful");
             return Ok(());
         } else {
-            std::process::exit(!0);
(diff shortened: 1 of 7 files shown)


teoxoy (approved):
Good stuff, I was also not familiar with the magic of `!0`.

jimblandy (approved):
Looks good to me - thanks!

PR #1877, opened 2022-04-30
glsl-in: Fix the ternary operator to behave like an if

### LINKED ISSUE #1854: [glsl-in] True and false branch of conditional expression operands are always evaluated
The expression:
```glsl
localSums[left] = globalLeft < globalSumBuffer.count ? globalSumBuffer.sums[globalLeft] : 0;
```
compiles to:
```wgsl
let _e36 = globalLeft;
let _e38 = globalSumBuffer.count;
let _e40 = globalLeft;
let _e43 = globalSumBuffer.sums[_e40];
localSums[_e34] = select(u32(0), _e43, (_e36 < _e38));
```
which evaluates both the true and false branches before evaluation the condition. This is wrong according to the GLSL spec:
> The ternary selection operator (?:). It operates on three expressions (exp1 ? exp2 : exp3). This
operator evaluates the first expression, which must result in a scalar Boolean. If the result is true, it
selects to evaluate the second expression, otherwise it selects to evaluate the third expression. Only
one of the second and third expressions is evaluated.
[GLSL 4.4 Spec page 109](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.40.pdf)

In the example above, if the condition is false, then the true branch is an out-of-bound access, which cannot be evaluated.

JCapucho:
Fixes #1854

As pointed out in the issue, the glsl spec defines that only one of the expressions should be evaluated making the ternary operator syntactic sugar to an `if`.

There are two commits that can be reviewed and tested separately, the first contains the actual fix and is the majority of the work. The seconds introduces a new method to the `Context` called `emit_restart` that performs a very common operation on the frontend of flushing the emitter and restarting it.


src/front/glsl/context.rs
```diff
@@ -1069,35 +1069,138 @@ impl Context {
...
-                    parser,
-                    &mut accept,
+                // Add the local that will hold the result of our conditional
+                let local = self.locals.append(
+                    LocalVariable {
+                        name: None,
+                        ty,
+                        init: None,
+                    },
+                    Span::default(),
```
jimblandy, line 1175: Would it make sense to use `meta` (i.e. the whole expression) for this `LocalVariable`'s span? I have no idea whether any errors will try to point to it.

src/front/glsl/context.rs
```diff
@@ -1069,35 +1069,138 @@ impl Context {
...
+            // `HirExprKind::Conditional` represents the ternary operator in glsl (`:?`)
+            //
+            // The ternary operator is defined to only evaluate one of the two possible
+            // expressions which means that it's behavior is that of an `if` statement,
+            // and it's merely syntatic sugar for it.
             HirExprKind::Conditional {
                 condition,
                 accept,
                 reject,
             } if ExprPos::Lhs != pos => {
```
jimblandy, line 1101: This match arm is 122 lines long. Would it make sense to move it into its own function or `Context` method?

src/front/glsl/context.rs
```diff
@@ -1069,35 +1069,138 @@ impl Context {
...
+                        .and_then(|(kind, width)| Some((type_power(kind, width)?, width, kind))),
+                ) {
+                    match accept_power.cmp(&reject_power) {
+                        std::cmp::Ordering::Less => {
+                            self.conversion(&mut accept, accept_meta, reject_kind, reject_width)?;
+                            // The expression belongs to the `true` branch so we need to flush to
+                            // the respective body
+                            self.emit_flush(&mut accept_body);
+                        }
+                        // Technically there's nothing to flush but we need to add some
```
jimblandy, line 1149: "but later we will need" would have helped me follow this.

src/front/glsl/context.rs
```diff
@@ -1069,35 +1069,138 @@ impl Context {
...
                     reject_meta,
-                )?;
+                );
 
-                self.add_expression(
-                    Expression::Select {
+                // Finally add the `If` to the main body with the `condition` we lowered
+                // earlier and the branches we prepared.
+                body.push(
+                    Statement::If {
```
jimblandy, line 1203: noice

src/front/glsl/types.rs
```diff
@@ -225,6 +225,18 @@ pub const fn type_power(kind: ScalarKind, width: Bytes) -> Option<u32> {
 }
 
 impl Parser {
+    /// Resolves the types of the expressions until `handle` (inclusive)
+    ///
+    /// This needs to be done before the [`typifier`][typifier] can be queried for
```
jimblandy, line 230: You can just say 

    [`typifier`]

and then at the bottom say:

    [`typifier`]: Context::typifier

There are similar cases elsewhere in this file.

src/front/glsl/types.rs
```diff
@@ -258,6 +276,42 @@ impl Parser {
...
+    /// Consider using [`resolve_type`][Self::resolve_type] whenever possible
+    /// since it doesn't require adding each type to the [`types`][types] arena
+    /// and it doesn't need to mutably borrow the [`Parser`][Self]
+    ///
+    /// [typifier]: Context::typifier
+    /// [types]: crate::Module::types
+    pub(crate) fn resolve_type_handle(
+        &mut self,
+        ctx: &mut Context,
+        handle: Handle<Expression>,
```
jimblandy, line 295: Since we have a lot of handles flying around here, could we call this `expr`?

src/front/glsl/types.rs
```diff
@@ -258,6 +276,42 @@ impl Parser {
...
+        meta: Span,
+    ) -> Result<Handle<Type>> {
+        self.typifier_grow(ctx, handle, meta)?;
+        let resolution = &ctx.typifier[handle];
+        Ok(match *resolution {
+            // If the resolution is already a handle return early
+            crate::proc::TypeResolution::Handle(handle) => handle,
```
jimblandy, line 302: ... and this `ty`?

tests/in/glsl/expressions.frag
```diff
@@ -121,6 +121,13 @@ void testNonImplicitCastVectorCast() {
 float global;
 void privatePointer(inout float a) {}
 
+void ternary(bool a) {
+    uint b = a ? 0 : 1u;
+    uint c = a ? 0u : 1;
+
+    uint nested = a ? (a ? (a ? 0u : 1) : 1u) : 1;
```
jimblandy, line 128: It'd be a little easier to follow the progress of this expression in the output if you used different values for each constant:

    a ? (a ? (a ? 2u : 3) : 4u) : 5

jimblandy (approved):
Looks beautiful! Nothing major, just suggestions and nits.

src/front/glsl/context.rs
```diff
@@ -1069,35 +1069,138 @@ impl Context {
...
+            //
+            // The ternary operator is defined to only evaluate one of the two possible
+            // expressions which means that it's behavior is that of an `if` statement,
+            // and it's merely syntatic sugar for it.
             HirExprKind::Conditional {
                 condition,
                 accept,
                 reject,
             } if ExprPos::Lhs != pos => {
+                // Lower the condition first to the current bodyy
```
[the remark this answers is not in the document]
JCapucho: Great suggestion

src/front/glsl/context.rs
JCapucho: I specifically was trying to avoid it to prevent the errors to highlight the same span twice but now that I look at it it shouldn't be a problem since the span isn't used in expression analysis

src/front/glsl/context.rs
JCapucho: Yes this part needs to be refractored a bit but I thimk that should be done in another PR.

src/front/glsl/types.rs
JCapucho: Thanks didn't know about this

jimblandy (approved):

PR #1880, opened 2022-04-30
[glsl-in] Simplify control flow in search for `main`.

jimblandy:
Purely aesthetic. If it's not to your taste, feel free to turn it down.


JCapucho:
Looks good to me, I'll admit I'm big fan of `match` so I tend to overuse it a bit like in this situation.

PR #1887, opened 2022-05-02
glsl-out: Fix feature search in expressions

JCapucho:
The fma wasn't being searched in case it wasn't supported which is the
opposite of what we want since if it's used we want to error if it isn't
supported.

It was also searching in all entry points instead of only in the current
one.

All samples queries need the ARB_shader_texture_image_samples extension
so we need to check if any samples queries are made and if so request
the extension.

### DIFF
--- a/src/back/glsl/features.rs
+++ b/src/back/glsl/features.rs
@@ -36,6 +36,8 @@ bitflags::bitflags! {
         const MULTI_VIEW = 1 << 17;
         /// Fused multiply-add.
         const FMA = 1 << 18;
+        /// Texture samples query
+        const TEXTURE_SAMPLES = 1 << 19;
     }
 }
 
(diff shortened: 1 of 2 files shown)


tests/in/functions-webgl.param.ron
```diff
@@ -1,6 +1,6 @@
 (
 	glsl: (
-		version: Embedded(300),
+		version: Embedded(310),
```
Gordon-F, line 3: This doesn't feel right. Test name is `functions-webgl`, but output glsl version is 3.1.

tests/out/glsl/functions-webgl.main.Vertex.glsl
```diff
@@ -1,4 +1,5 @@
 #version 300 es
+#extension GL_EXT_gpu_shader5 : require
```
Gordon-F, line 2: Based on the [ext docs](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_gpu_shader5.txt), it requires OpenGL ES 3.1.

src/back/glsl/features.rs
```diff
@@ -101,7 +103,10 @@ impl FeaturesManager {
         check_feature!(SAMPLE_VARIABLES, 400, 300);
         check_feature!(DYNAMIC_ARRAY_SIZE, 430, 310);
         check_feature!(MULTI_VIEW, 140, 310);
-        check_feature!(FMA, 400, 310);
```
Gordon-F, line 104: I guess this line was accidentally removed ?

Gordon-F (changes requested):

src/back/glsl/features.rs
JCapucho: No this line is correct, `check_feature!` is only used for features that cannot be workarounded, fma can.

tests/out/glsl/functions-webgl.main.Vertex.glsl
JCapucho: You're right forgot to add the check on the version

Gordon-F (approved):

PR #1888, opened 2022-05-03
Show naga-cli instead of cargo run on README

yutannihilation:
I think installing `naga-cli` is more convenient than using `cargo run`. How about showing it on README rather than `cargo run`?


teoxoy (approved):
Thanks! I think this is a change in the right direction (keeping the readme for the users audience).

For potential contributors, we could add a contributing document (`cargo run` would be more useful for that audience).

Let's wait for a few more reviewers to chime in.

yutannihilation:
> For potential contributors, we could add a contributing document (`cargo run` would be more useful for that audience).

Ah, thanks for the suggestion. I agree with you. I'll try adding some words on this, but feel free to take over!

teoxoy:
Looks good! I think we can merge this.

yutannihilation:
Thanks!

PR #1897, opened 2022-05-09
Implement reverseBits and countOneBits for SPIR-V

### LINKED ISSUE #1895: [spv-out] reverseBits and countOneBits unimplemented
I noticed that the `reverseBits` and `countOneBits` functions don't seem to be implemented for the SPIR-V backend, though they seem to be implemented for all the other backends. Is this just an oversight?

hasali19:
Closes #1895


teoxoy:
Looks good!

Just to make sure everything works as expected, could you add those 2 functions to the test file at `tests/in/bits.wgsl`? (remember to run `cargo test` so that the snapshots get generated)

hasali19:
Looks like the glsl implementation is incorrect. I think `bitCount` always returns a signed integer so needs a conversion.

teoxoy (approved):
Good stuff!

I think the GLSL change is small enough to not warrant a separate PR.

PR #1903, opened 2022-05-10
Fix generated hlsl for writes to scalar/vector storage buffer

### LINKED ISSUE #1902: [hlsl-out] Invalid hlsl for write to scalar/vector storage buffer
WGSL:

```
@group(0)
@binding(0)
var<storage, read_write> buf: vec2<i32>;

@compute
@workgroup_size(1)
fn main() {
    buf = vec2<i32>(1);
}
```

HLSL:

```hlsl
RWByteAddressBuffer buf : register(u0);

[numthreads(1, 1, 1)]
void main()
{
    buf.Store2(, asuint((1).xx));
    return;
}
```

The `Store2` call seems to have a missing argument.

hasali19:
Closes #1902

### DIFF
--- a/src/back/hlsl/storage.rs
+++ b/src/back/hlsl/storage.rs
@@ -44,6 +44,9 @@ impl<W: fmt::Write> super::Writer<'_, W> {
         chain: &[SubAccess],
         func_ctx: &FunctionCtx,
     ) -> BackendResult {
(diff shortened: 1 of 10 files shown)


teoxoy (approved):
Looks good, thanks!

Issue #1913, opened 2022-05-12
[hlsl-out] hlsl compilation error for global array initialiser


hasali19:
WGSL:

```
@group(0)
@binding(1)
var<storage, read_write> s_output: u32;

var<private> global1: array<u32, 2> = array<u32, 2>(1u, 2u);

@compute
@workgroup_size(1)
fn main() {
    s_output = global1[0];
}
```

HLSL:

```hlsl

uint Constructarray2_uint_(uint arg0, uint arg1)[2] {
    uint ret[2] = { arg0, arg1 };
    return ret;
}

RWByteAddressBuffer s_output : register(u1);
static uint global1_[2] = Constructarray2_uint_(1u, 2u);

[numthreads(1, 1, 1)]
void main()
{
    uint _expr4 = global1_[0];
    s_output.Store(0, asuint(_expr4));
    return;
}

```

HLSL compiler error:

```
D:\local\Temp\98929c1d-4490-4b2e-a958-cccfa42228a4.hlsl(2,38): error X3000: syntax error: unexpected token '['
```

teoxoy:
I've noticed this a few hours ago and am already fixing it. The issue was that the array return syntax only works with DXC (not FXC).

PR #1914, opened 2022-05-12
[hlsl-out] Fix array constructor return type (for FXC)

teoxoy:
According to https://github.com/microsoft/DirectXShaderCompiler/issues/1808 the syntax used for returning arrays (which was added in https://github.com/gfx-rs/naga/pull/1814) is specific to DXC. This PR changes it to use a typedef which is supported by both FXC and DXC.

fixes https://github.com/gfx-rs/naga/issues/1913


teoxoy:
We should probably add HLSL validation via FXC to CI as well.

kvark (approved):

PR #1917, opened 2022-05-14
Add error for matrices without `f32` scalar type

### LINKED ISSUE #1896: [wgsl-in] Error on matrices with non float scalar kinds
The following works (on naga master cf32c2b):

````wgsl
@group(0) @binding(0)
var<storage, write> output_0: vec3<f32>;

@compute @workgroup_size(1)
fn main(@builtin(global_invocation_id) global_id: vec3<u32>) {
	let zero: f32 = 0.0;
	let zero_vec = vec3<f32>(zero,zero,zero);
	let m = mat3x3<f32>(zero_vec, zero_vec, zero_vec);
}
````

````bash
% cargo run --  ./test.wgsl
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/naga ./test.wgsl`
Validation successful
````

The following shows an error (`f32` is replaced with `i32`):

````wgsl
@group(0) @binding(0)
var<storage, write> output_0: vec3<i32>;

@compute @workgroup_size(1)
fn main(@builtin(global_invocation_id) global_id: vec3<u32>) {
	let zero: i32 = 0;
	let zero_vec = vec3<i32>(zero,zero,zero);
	let m = mat3x3<i32>(zero_vec, zero_vec, zero_vec);
}
````

````bash
% cargo run --  ./test.wgsl
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/naga ./test.wgsl`
[2022-05-09T19:06:29Z ERROR naga::valid::compose] Matrix component[0] type Handle([1])
error: 
  ┌─ test.wgsl:8:9
  │
8 │     let m = mat3x3<i32>(zero_vec, zero_vec, zero_vec);
  │            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ naga::Expression [5]

Entry point main at Compute is invalid: 
	Expression [5] is invalid
	Composing 0's component type is not expected
````

This seems... a bug, or is this just a counterintuitive aspect of the spec? 

Use case: in [wonnx](https://github.com/webonnx/wonnx) we generate shaders using templates (e.g. [this one for generalized matrix multiplication](https://github.com/webonnx/wonnx/blob/master/wonnx/templates/matrix/gemm.wgsl)) to implement machine learning operators, which may use different data types (typically f32, i32 or i64) and would like to make the shaders work on all types. We often need to construct zero values (or other statically initialized values).

### PR DESCRIPTION
**Pull request by grovesNL**
Fixes #1896

### DIFF
--- a/src/front/wgsl/construction.rs
+++ b/src/front/wgsl/construction.rs
@@ -184,12 +184,15 @@ fn parse_constructor_type<'a>(
             Ok(Some(ConstructorType::Vector { size, kind, width }))
         }
         (Token::Paren('<'), ConstructorType::PartialMatrix { columns, rows }) => {
-            let (_, width) = lexer.next_scalar_generic()?;
-            Ok(Some(ConstructorType::Matrix {
-                columns,
-                rows,
-                width,
-            }))
+            let (kind, width, span) = lexer.next_scalar_generic_with_span()?;
+            match kind {
+                ScalarKind::Float => Ok(Some(ConstructorType::Matrix {
+                    columns,
+                    rows,
+                    width,
+                })),
+                _ => Err(Error::BadMatrixScalarKind(span, kind, width)),
+            }
         }
         (Token::Paren('<'), ConstructorType::PartialArray) => {
             lexer.expect_generic_paren('<')?;
--- a/src/front/wgsl/mod.rs
+++ b/src/front/wgsl/mod.rs
@@ -130,6 +130,7 @@ pub enum Error<'a> {
     BadFloat(Span, BadFloatError),
     BadU32Constant(Span),
     BadScalarWidth(Span, Bytes),
+    BadMatrixScalarKind(Span, crate::ScalarKind, u8),
     BadAccessor(Span),
     BadTexture(Span),
     BadTypeCast {
@@ -296,11 +297,19 @@ impl<'a> Error<'a> {
                 labels: vec![(bad_span.clone(), "expected unsigned integer".into())],
                 notes: vec![],
             },
-
             Error::BadScalarWidth(ref bad_span, width) => ParseError {
                 message: format!("invalid width of `{}` bits for literal", width as u32 * 8,),
                 labels: vec![(bad_span.clone(), "invalid width".into())],
                 notes: vec!["the only valid width is 32 for now".to_string()],
+            },
+            Error::BadMatrixScalarKind(
+                ref span,
+                kind,
+                width,
+             ) => ParseError {
+                message: format!("matrix scalar type must be floating-point, but found `{}`", kind.to_wgsl(width)),
+                labels: vec![(span.clone(), "must be floating-point (e.g. `f32`)".into())],
+                notes: vec![],
             },
             Error::BadAccessor(ref accessor_span) => ParseError {
                 message: format!(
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/front/wgsl/mod.rs [hunk 1]
```diff
@@ -2901,6 +2910,23 @@ impl Parser {
...
 
+    fn parse_matrix_scalar_type<'a>(
+        &mut self,
+        lexer: &mut Lexer<'a>,
+        columns: crate::VectorSize,
+        rows: crate::VectorSize,
+    ) -> Result<crate::TypeInner, Error<'a>> {
+        let (kind, width, span) = lexer.next_scalar_generic_with_span()?;
+        match (kind, width) {
+            (crate::ScalarKind::Float, 4) => Ok(crate::TypeInner::Matrix {
```
teoxoy, line 2921: We shouldn't check the width here since it will be checked by the validation code where a width of 8 might be allowed.

https://github.com/gfx-rs/naga/blob/ab2806e05fbd69a502b4b25a85f99ae4d3e82278/src/valid/type.rs#L205-L208

src/front/wgsl/construction.rs
```diff
@@ -184,12 +184,15 @@ fn parse_constructor_type<'a>(
...
-            }))
+            let (kind, width, span) = lexer.next_scalar_generic_with_span()?;
+            match (kind, width) {
+                (ScalarKind::Float, 4) => Ok(Some(ConstructorType::Matrix {
+                    columns,
+                    rows,
+                    width,
+                })),
+                _ => Err(Error::BadMatrixScalarKind(span, kind, width)),
+            }
```
teoxoy, line 195: Can we use the new `parse_matrix_scalar_type` fn here as well?

src/front/wgsl/mod.rs
```diff
@@ -296,12 +297,20 @@ impl<'a> Error<'a> {
...
                 labels: vec![(bad_span.clone(), "invalid width".into())],
                 notes: vec!["the only valid width is 32 for now".to_string()],
             },
+            Error::BadMatrixScalarKind(
+                ref span,
+                kind,
+                width,
+             ) => ParseError {
+                message: format!("matrix scalar type must be `f32`, but found `{}`", kind.to_wgsl(width)),
+                labels: vec![(span.clone(), "must be `f32`".into())],
```
teoxoy, line 311: Let's only mention the kind here (i.e. floating-point) since f64 might be allowed as well and once we support WGSL's f16 extension that would as well.

teoxoy (commented):
Looks good overall, left a few comments.

src/front/wgsl/construction.rs
grovesNL: We don't have a great way to do this currently, because this returns `ConstructorType::Matrix` but the `parse_matrix_scalar_type ` returns `TypeInner::Matrix`.

We could consider making those share a type, e.g. change the enum variants to be `ConstructorType::Matrix(Matrix)` and `TypeInner::Matrix(Matrix)`. That change would affect a bunch of areas though.

grovesNL:
Thank you! Updated the PR

src/front/wgsl/construction.rs
teoxoy: Ah right, it's ok - can leave it like this for now.

src/front/wgsl/mod.rs
teoxoy: @grovesNL I think you missed this one

grovesNL: Oops, thanks for catching this! Should be fixed now

teoxoy (approved):
LGTM, thanks!

PR #1918, opened 2022-05-14
Add HLSL validation via FXC to CI

teoxoy:
Added HLSL validation via FXC to CI since `wgpu` currently uses FXC for the D3D backends.

There are 4 issues with our current tests/backend that have to be resolved for FXC validation to succeed.

- #1919
- #1920
- #1921
- #1922

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all clean validate-spv validate-msl validate-glsl validate-dot validate-wgsl validate-hlsl
+.PHONY: all clean validate-spv validate-msl validate-glsl validate-dot validate-wgsl validate-hlsl-dxc validate-hlsl-fxc
 .SECONDARY: boids.metal quad.metal
 SNAPSHOTS_BASE_IN=tests/in
 SNAPSHOTS_BASE_OUT=tests/out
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -93,3 +93,37 @@ validate-hlsl: $(SNAPSHOTS_BASE_OUT)/hlsl/*.hlsl
...
+	@set -e && for file in $^ ; do \
+		FXC_PARAMS="-Zi -Od"; \
+		echo "Validating" $${file#"$(SNAPSHOTS_BASE_OUT)/"}; \
+		config="$$(dirname $${file})/$$(basename $${file}).config"; \
+		. $${config}; \
+		for (( i=0; i<$${#vertex[@]}; i++ )); do \
+			name=`echo $${vertex[i]} | cut -d \: -f 1`; \
+			profile=`echo $${vertex[i]} | cut -d \: -f 2`; \
+			sm=`echo $${profile} | cut -d \_ -f 2`; \
+			if (( sm < 6 )); then \
```
Gordon-F, line 108: Is it necessary to add shader model version check?

teoxoy: FXC doesn't support any SM >= 6. When we add f16 support which requires SM6.2 the validation would fail but shouldn't.

Gordon-F (approved):

jimblandy:
Just noting: we merged this CI test before merging the PRs that would allow it to pass.

PR #1921, opened 2022-05-14
[hlsl-out] Avoid error X3694: race condition writing to shared resource detected (for FXC)

teoxoy:
Avoid FXC `error X3694: race condition writing to shared resource detected, consider making this write conditional` triggered by the `empty-global-name.spv` example.

Minimum reproducible example:

src WGSL

```
@group(0) @binding(0)
var<storage, read_write> buf: i32;

@compute @workgroup_size(2)
fn compute_main() {
	buf += 1;
}
```

generated HLSL

```hlsl
RWByteAddressBuffer buf : register(u0);

[numthreads(2, 1, 1)]
void compute_main()
{
    int _expr1 = asint(buf.Load(0));
    buf.Store(0, asuint((_expr1 + 1)));
    return;
}
```

I'm not sure what else we can do in this case other than trying to not trigger it.


Gordon-F:
@teoxoy I think we can remove HLSL from this test and add comment why we did this:
https://github.com/gfx-rs/naga/blob/master/tests/snapshots.rs#L565

teoxoy:
I changed the test instead of removing HLSL from the test since the test itself is just making sure we can handle globals with no name from SPIR-V's side. Changing LocalSize from 64 to 1 shouldn't impact the test's purpose.

jimblandy (approved):

PR #1922, opened 2022-05-14
[hlsl-out] Avoid error X3504: array index out of bounds (for FXC)

teoxoy:
Avoid FXC `error X3504: array index out of bounds` triggered by the `access.wgsl` example.

### DIFF
--- a/tests/in/access.wgsl
+++ b/tests/in/access.wgsl
@@ -26,7 +26,7 @@ var<uniform> baz: Baz;
 var<storage,read_write> qux: vec2<i32>;
 
 fn test_matrix_within_struct_accesses() {
-	var idx = 9;
+	var idx = 1;
 
     idx--;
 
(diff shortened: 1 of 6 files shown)


teoxoy:
Opened #1923 to track a proper fix for this.

Gordon-F:
Same as https://github.com/gfx-rs/naga/pull/1921. I think it would be better to remove hlsl from test case instead of change test case.

teoxoy:
Similarly to my response in #1921, I changed the test instead of removing HLSL from the test since the test itself is just making sure the extra HLSL generated code for accessing matrices of the form `matCx2` works properly. 

https://github.com/gfx-rs/naga/blob/60ae549fe1bec8df40518a6164d723d41a2b76b5/tests/out/hlsl/access.hlsl#L32-L52

Changing the index shouldn't impact the test's purpose and we could also revert the change once #1923 gets resolved.

jimblandy (approved):

PR #1928, opened 2022-05-16
[hlsl-out] Fix countOneBits and reverseBits for signed integers

### LINKED ISSUE #1904: [hlsl-out] Incorrect result for min(reverseBits(-1), 0)
WGSL:

```
struct f {
    a: i32,
}

@group(0)
@binding(1)
var<storage, read_write> h: f;

@stage(compute)
@workgroup_size(1)
fn main() {
    h = f(min(reverseBits(-1), 0));
}
```

HLSL:

```hlsl
struct f {
    int a;
};

RWByteAddressBuffer h : register(u1);

f Constructf(int arg0) {
    f ret;
    ret.a = arg0;
    return ret;
}

[numthreads(1, 1, 1)]
void main()
{
    {
        f _value2 = Constructf(min(reversebits(-1), 0));
        h.Store(0, asuint(_value2.a));
    }
    return;
}
```

This incorrectly produces 0, though `reverseBits(-1)` should in fact be a no-op. I think hlsl's `reversebits` operates on `uint`, so needs to be transformed to `asint(reversebits(asuint(-1)))`.

hasali19:
This PR fixes usage of the HLSL `countbits` and `reversebits` functions with signed integers, by adding the necessary `asuint` and `asint` conversions.

Fixes #1904

### DIFF
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -6,7 +6,7 @@ use super::{
 use crate::{
     back,
     proc::{self, NameKey},
-    valid, Handle, Module, ShaderStage, TypeInner,
+    valid, Handle, Module, ScalarKind, ShaderStage, TypeInner,
 };
 use std::{fmt, mem};
 
@@ -2190,6 +2190,7 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     Asincosh { is_sin: bool },
                     Atanh,
                     Regular(&'static str),
+                    MissingIntOverload(&'static str),
                 }
 
                 let fun = match fun {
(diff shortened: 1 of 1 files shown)


src/back/hlsl/writer.rs
```diff
@@ -2190,6 +2190,7 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     Asincosh { is_sin: bool },
                     Atanh,
                     Regular(&'static str),
+                    UintWrapped(&'static str),
```
teoxoy, line 2193: Could we name this `NoIntOverload` or `MissingIntOverload`? What do you think?

teoxoy (commented):
Looks good! Could you add HLSL to this list and generate the snapshot for it? 

https://github.com/gfx-rs/naga/blob/60ae549fe1bec8df40518a6164d723d41a2b76b5/tests/snapshots.rs#L418-L419

src/back/hlsl/writer.rs
hasali19: I was trying to think of a good name, that sounds much better!

hasali19:
> Looks good! Could you add HLSL to this list and generate the snapshot for it?
> 
> https://github.com/gfx-rs/naga/blob/60ae549fe1bec8df40518a6164d723d41a2b76b5/tests/snapshots.rs#L418-L419

I noticed that previously and attempted to add it, but it looks like there are some other functions which aren't implemented yet causing it to fail:

```
thread 'convert_wgsl' panicked at 'HLSL write failed: Unimplemented("write_expr_math Pack4x8snorm")', tests/snapshots.rs:313:54
```

teoxoy:
Alright, we can enable the test later then.

teoxoy (approved):

PR #1933, opened 2022-05-20
Support for the `OVR_multiview2` WebGL extension

expenses:
The WebGL `OVR_multiview2` extension is implemented in:

* Chrome on Windows, Linux and Android
* Firefox on Android

but NOT on Firefox on Linux afaik.

This is the main (there's also an Occulus extension: https://developer.mozilla.org/en-US/docs/Web/API/OVR_multiview2) extension used for multiview rendering in WebGL.

Registry link: https://www.khronos.org/registry/webgl/extensions/OVR_multiview2/
MDN link: https://developer.mozilla.org/en-US/docs/Web/API/OVR_multiview2

The number of views needs to be passed in as a pipeline option and then set in the vertex shader with `layout(num_views = X)`. I have a working (but messy) integration with wgpu here: https://github.com/expenses/wgpu/commit/7da24f3782a606cb708d68bbe236773a3dbab8d3. Note that this relies on https://github.com/rustwasm/wasm-bindgen/pull/2903 and https://github.com/grovesNL/glow/compare/main...expenses:multiview.

We need a good way to check whether the target API is WebGL so that we can determine whether to use this extension or the regular `GL_EXT_multiview` one.


expenses:
> We need a good way to check whether the target API is WebGL so that we can determine whether to use this extension or the regular `GL_EXT_multiview` one.

I've solved this by using `Embedded { version: u16, is_webgl: bool }` in the version. Not sure if this is the best way.

jimblandy:
@JCapucho, if this isn't something you can review feel free to turn down the review request - I just wanted to put this on someone's plate to see if we could make some progress.

src/back/glsl/mod.rs
```diff
@@ -3580,6 +3605,7 @@ const fn glsl_built_in(built_in: crate::BuiltIn, output: bool) -> &'static str {
                 "gl_FragCoord"
             }
         }
+        Bi::ViewIndex if multiview_ovr => "int(gl_ViewID_OVR)",
```
JCapucho, line 3608: What's the purpose of having two branches if we always take the first one?

src/back/glsl/features.rs
```diff
@@ -106,7 +106,10 @@ impl FeaturesManager {
         check_feature!(CULL_DISTANCE, 450, 300);
         check_feature!(SAMPLE_VARIABLES, 400, 300);
         check_feature!(DYNAMIC_ARRAY_SIZE, 430, 310);
-        check_feature!(MULTI_VIEW, 140, 310);
+        match version {
+            Version::Embedded { is_webgl: true, .. } => check_feature!(MULTI_VIEW, 140, 300),
```
JCapucho, line 110: Looking at the extension you linked below this should only be supported from 300 and up

JCapucho (changes requested):
I'll be honest, I don't have a lot of experience with webgl and have almost 0 experience with multiview, but I've looked at the extensions and overall everything looks mostly fine.

But I think there are some problems.

src/back/glsl/mod.rs
expenses: Oops, forgot to have this take `is_webgl` as input.,

expenses:
@JCapucho I've re-done this a little bit to make selecting which multiview extension to use explicit. I've put the multiview options on the glsl options for convenience, but wgpu creates these when creating a pipeline layout so we need to set it to `None` there:

https://github.com/expenses/wgpu/blob/dbab152093dd9ca1ca255a57df4048a92b857325/wgpu-hal/src/gles/device.rs#L881-L888

Should I be putting the multiview options on the pipeline options instead?

Also the nightly test seems to be broken by something outside of this PR.

expenses:
> What's the benefit of doing it this way? It seems to add a lot of code with no real benefit

I think it's nicer to only specify which extension to use if the multiview options are set, as opposed to always having `is_webgl` in the version. It could be streamlined though but I don't think that it adds that much code.

JCapucho:
> I think it's nicer to only specify which extension to use if the multiview options are set, as opposed to always having `is_webgl` in the version. It could be streamlined though but I don't think that it adds that much code.

My worry is that we are opening ourselves up to have the backend mismatch extensions and version, with this approach it's possible to set the version to core and produce a webgl extension, and if we end up doing this for other extensions the problem will quickly get worse.

expenses:
> My worry is that we are opening ourselves up to have the backend mismatch extensions and version, with this approach it's possible to set the version to core and produce a webgl extension, and if we end up doing this for other extensions the problem will quickly get worse.

So my thinking is that we'd only need to check for WebGL with this extension, so doing this would be okay. Unfortunately https://www.khronos.org/registry/webgl/extensions/WEBGL_multi_draw/ also needs a specific WebGL extension: `GL_ANGLE_multi_draw`, so this would cause problems if we wanted to support that.

So I think I'll go and revert these changes to have `is_webgl` on the version again. Sorry about that.

src/back/glsl/mod.rs
```diff
@@ -119,15 +119,31 @@ pub enum Version {
     /// `core` GLSL.
     Desktop(u16),
     /// `es` GLSL.
-    Embedded(u16),
+    Embedded { version: u16, is_webgl: bool },
 }
 
 impl Version {
+    /// Create a new non-WebGL embedded version
+    pub const fn new_embedded(version: u16) -> Self {
```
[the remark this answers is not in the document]
expenses: Done

tests/snapshots.rs
```diff
@@ -531,6 +537,11 @@ fn convert_wgsl() {
             "binding-arrays",
             Targets::WGSL | Targets::HLSL | Targets::METAL | Targets::SPIRV,
         ),
+        ("multiview", Targets::SPIRV | Targets::GLSL | Targets::WGSL),
+        (
+            "multiview_webgl",
+            Targets::SPIRV | Targets::GLSL | Targets::WGSL,
```
JCapucho, line 543: Running this for spirv and even wgsl seems unnecessary since we only want to observe how it affects webgl and we the other multiview test already tests them

JCapucho (changes requested):
Just one more small thing and we are good to go

tests/snapshots.rs
expenses: Done!

JCapucho:
Brilliant work, just needs the auto formatter to run and it's good to merge

PR #1934, opened 2022-05-20
[spv-in] Add MultiView to SUPPORTED_CAPABILITIES

expenses:
Naga's support for multiview shaders seems pretty good so I think we can enable this SPIR-V capability, especially because `SPV_KHR_multiview` is already in `SUPPORTED_EXTENSIONS`.


expenses:
Also fixes some cargo check and clippy issues

kvark (approved):

PR #1937, opened 2022-05-23
Add a helper for getting the location of a span in some textual source

nical:
The idea is to get an easily consumable location (line/column/etc.) in the shader source for errors, where "easily consumable" really just means what [the WebGPU spec](https://www.w3.org/TR/webgpu/#shader-module-compilation-information) expects for errors.

r? @jimblandy


jimblandy (approved):
Looks good - just tweaked the docs a bit.

PR #1885, opened 2022-05-02
Use Unicode 14 for WGSL

mehmetoguzderin:
The unicode-xid crate just had a release with Unicode 14 data, which the WGSL spec requires.

PR #1940, opened 2022-05-25
Require unicode-xid 0.2.2, not 0.2.3, for Firefox.

jimblandy:
Firefox tries to avoid having multiple copies of crates in the tree, and currently it is using `unicode-xid` 0.2.2. Since Naga does not specifically require 0.2.3, we can relax this requirement to make Firefox `wgpu` updates easier.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ serde = { version = "1.0.103", features = ["derive"], optional = true }
 petgraph = { version ="0.6", optional = true }
 pp-rs = { version = "0.2.1", optional = true }
 hexf-parse = { version = "0.2.1", optional = true }
-unicode-xid = { version = "0.2.3", optional = true }
+unicode-xid = { version = "0.2.2", optional = true }
 
 [dev-dependencies]
 bincode = "1"


mehmetoguzderin:
I am not sure how this goes, but the spec requires Unicode 14, which is present only in the 0.2.3 version. Some kind of TODO note needs to be added somewhere to remind, as that was the case before my PR.

jimblandy:
@mehmetoguzderin Oh, thank you for checking this out! I assumed that the differences were minor.

Let me look into upgrading Firefox's version instead.

mehmetoguzderin:
@jimblandy Thank you! Just a tiny note: there is a very competitive Rust solution built directly by ICU named ICU4X for Unicode utilities that produces super compact & fast runtime versions of such tables. I am not super actively looking into it (as unicode-xid does 14 well), but if I ever get around to trying plugging it, I'll notify somewhere.

jimblandy:
@mehmetoguzderin I think my friend Erik works on ICU4X, actually. That would be great if that was a better solution for us.

mehmetoguzderin:
@jimblandy Wow, that's cool!

teoxoy:
Just for linking purposes, it has been bumped in https://github.com/gfx-rs/naga/pull/1885 specifically for unicode 14 support.

jimblandy:
Yeah, I'll do my homework next time. :(

teoxoy:
Ah, no worries :)

PR #1950, opened 2022-05-27
 glsl-out: Bake Constant expressions 

### LINKED ISSUE #1944: [glsl-out] Thread panic when generating GLSL of dot function with zero-valued vector(s)
Attempting to translate the following WGSL code to GLSL 440 or 450 with naga causes a main thread panic.

WGSL:
```
@compute @workgroup_size(1)
fn main() {
    var var0 = dot(vec2<i32>(), vec2<i32>());
}
```

Error:
```
        thread 'main' panicked at 'no entry found for key', (MY_NAGA_DIR)/naga/src/back/glsl/mod.rs:1524:26
```

This doesn't happen if both vectors are provided arguments, seemingly only if they've been initialized via zero-value.

JCapucho:
These expressions don't appear in Emit statements so they need to be
baked manually.

closes #1944

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1425,6 +1425,17 @@ impl<'a, W: Write> Writer<'a, W> {
             writeln!(self.out, ";")?
         }
 
+        // Loop trough all Constant expressions and bake them if needed.
+        for (handle, expr) in ctx.expressions.iter() {
+            if let crate::Expression::Constant(_) = *expr {
+                if self.need_bake_expressions.contains(&handle) {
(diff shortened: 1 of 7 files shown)


jimblandy (changes requested):
This makes the output harder to follow, for no good reason. Wouldn't it be better to just never put `Constant` expressions in `Writer::need_bake_expressions` in the first place?

JCapucho:
> This makes the output harder to follow, for no good reason. Wouldn't it be better to just never put `Constant` expressions in `Writer::need_bake_expressions` in the first place?

Definitely feasible, we could just use `write_expr` in the `write_dot_product` function and let the backend cache as it sees fit

PR #1952, opened 2022-05-27
glsl-in: Stop emitter in conditional

### LINKED ISSUE #1942: [glsl] Unreachable code error causing main thread crash during GLSL validation
Translating the following piece of WGSL code to GLSL 440 or 450 using naga returns no errors, but validating the translated `.comp` shader with naga results in a main thread panic/illegal instruction(?). I'm unsure if this is an issue with glsl-in or glsl-out, given that I am giving naga code that it has generated that I'm not sure on whether is legal or not.

WGSL (substituting any of the `true` literals below for `false` or a Boolean expression seems to make no difference for those I have tried):
```
@compute @workgroup_size(1)
fn main() {
	let var0 = select(true, true, true);
}
```

Error:
```
thread 'main' panicked at 'internal error: entered unreachable code: Emitting has already started!', src/front/mod.rs:31:13
zsh: IOT instruction  cargo run test.comp
```

The last line is different depending on my shell - for bash/sh it is `Aborted`.

Translated GLSL from glsl-out:
```
#version 440 core
#extension GL_ARB_compute_shader : require


void main() {
    bool var0_ = (true ? true : true);
}
```

Running with `RUST_LOG=trace` returns several trace and debug level logs, but no errors/warnings before the thread panic.

JCapucho:
This wasn't being done when there were no implicit conversions, causing
the emitter to panic.

closes #1942

### DIFF
--- a/src/front/glsl/context.rs
+++ b/src/front/glsl/context.rs
@@ -1174,6 +1174,10 @@ impl Context {
                             self.emit_end(&mut reject_body);
                         }
                     }
+                } else {
+                    // Technically there's nothing to flush but later we will need to
+                    // add some expressions that must not be emitted.
+                    self.emit_end(body)
                 }
 
                 // We need to get the type of the resulting expression to create the local,


jimblandy (commented):
Ah, sorry I didn't catch this in the first review. Thanks for the fix.

jimblandy (approved):

PR #1962, opened 2022-06-01
Let validation check for more unsupported builtins.

### LINKED ISSUE #1961: Naga generates invalid WGSL for struct members with unsupported builtin attributes
WGSL does not support all the builtins that Naga IR does; for example, [there is no WGSL analogue](https://github.com/gpuweb/gpuweb/issues/390) of `BuiltIn::ClipDistance`. When emitting a struct type, the WGSL back end simply drops members that are builtins it can't handle, but composition expressions for that type still include all members, resulting in WGSL that doesn't compile.

Rather than generating invalid code, Naga should error out in validation.

jimblandy:
Add `valid::Capabilities` flags for the `ClipDistance` and `CullDistance` builtins, which are not supported by all back ends.

Have the CLI perform valation with only those capabilities that the requested back ends support.

Fixes #1961.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -317,12 +317,23 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
         _ => return Err(CliError("Unknown input file extension").into()),
     };
 
+    // Decide which capabilities our output formats can support.
+    let validation_caps =
+        output_paths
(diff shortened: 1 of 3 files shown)


teoxoy (approved):
LGTM

PR #1965, opened 2022-06-02
[glsl-in] Fix issues with shadow parameters

magcius:
The argument was getting converted to a Depth image,
but the parameter type was not... see provided test.

### DIFF
--- a/tests/in/glsl/sampler-functions.frag
+++ b/tests/in/glsl/sampler-functions.frag
@@ -1,0 +1,17 @@
+
+#version 440
+precision mediump float;
+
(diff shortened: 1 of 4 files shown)


tests/in/glsl/sampler-functions.frag
```diff
@@ -0,0 +1,17 @@
+
```
Gordon-F, line 1: Could you rerun tests with the `wgsl-out` feature? It looks like your PR missing wgsl output snapshots.

JCapucho:
The test case you provided is still good so we can merge it

magcius:
Feel free to cherry-pick my test to your PR then. Closing.

PR #1966, opened 2022-06-02
[glsl-out] Handle vector bitcasts

expenses:
This fixes an issue where an `OpBitcast` on a vector didn't generate the correct glsl:

```diff
diff --git a/before.frag b/after.frag
index 8b2968b8..eb1074a2 100644
--- a/before.frag
+++ b/after.frag
@@ -126,8 +126,8 @@ void main_1() {
             }
             ivec3 _e2636 = phi_22745_;
             ivec3 _e2638 = phi_22742_;
-            ivec3 _e2643 = int(bitfieldExtract(uint(_e2638), int(uint(0)), int(uint(7))));
-            ivec3 _e2657 = int(bitfieldExtract(uint(_e2636), int(uint(0)), int(uint(7))));
+            ivec3 _e2643 = ivec3(bitfieldExtract(uvec3(_e2638), int(uint(0)), int(uint(7))));
+            ivec3 _e2657 = ivec3(bitfieldExtract(uvec3(_e2636), int(uint(0)), int(uint(7))));
             int _e2672 = max(((81 + (_e162 * 3)) - ((_e162 > _e180) ? 1 : 0)), 82);
             int _e2676 = (((_e162 == 0) || (_e162 == _e180)) ? 2 : 3);
             local = _e163;
```


JCapucho:
I don't think this is the right approach, I think the problem is the backend generating the bad casts not the frontend generating the bad IR because it sets the width, but I need either @kvark or @jimblandy to confirm

jimblandy:
SPIR-V `OpBitcast` instructions can change the number of components of composite types, but Naga IR has nothing like that. The other `OpBitcast` meanings all require the bit widths to match. So I would expect these to always become `Expression::As` expressions with `convert: None`.

So I think @JCapucho is right here: if the back end is looking at the absence of a bit width and generating an `int` cast instead of an `ivec3` cast, that's the bug. It should be using the type of the operand and the `kind` in the `As` expression to decide what to output.

Here's the code in [typifier.rs](https://github.com/gfx-rs/naga/blob/89bed99bcc995bc5068c9c112fd9b7d7896bb148/src/proc/typifier.rs#L849-L882) that says what the output of an `As` expression should be:

```
            crate::Expression::As {
                expr,
                kind,
                convert,
            } => match *past(expr)?.inner_with(types) {
                Ti::Scalar { kind: _, width } => TypeResolution::Value(Ti::Scalar {
                    kind,
                    width: convert.unwrap_or(width),
                }),
                Ti::Vector {
                    kind: _,
                    size,
                    width,
                } => TypeResolution::Value(Ti::Vector {
                    kind,
                    size,
                    width: convert.unwrap_or(width),
                }),
                Ti::Matrix {
                    columns,
                    rows,
                    width,
                } => TypeResolution::Value(Ti::Matrix {
                    columns,
                    rows,
                    width: convert.unwrap_or(width),
                }),
                ref other => {
                    return Err(ResolveError::IncompatibleOperands(format!(
                        "{:?} as {:?}",
                        other, kind
                    )))
                }
            },
```

Every path chooses the resulting `kind` from the `As` expression, not the operand.

src/back/glsl/mod.rs
```diff
@@ -2892,29 +2892,54 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                                crate::VectorSize::Quad => "ivec4",
+                            },
+                            (Sk::Uint | Sk::Sint | Sk::Float, Sk::Bool, Some(size)) => match size {
+                                crate::VectorSize::Bi => "bvec2",
+                                crate::VectorSize::Tri => "bvec3",
+                                crate::VectorSize::Quad => "bvec4",
+                            },
+                            (Sk::Float, Sk::Sint, _) => "floatBitsToInt",
+                            (Sk::Float, Sk::Uint, _) => "floatBitsToUint",
+                            (Sk::Sint, Sk::Float, _) => "intBitsToFloat",
```
expenses, line 2924: It seems like these can take vectors as input?

src/back/glsl/mod.rs
```diff
@@ -2892,29 +2892,54 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                            (Sk::Float, Sk::Sint) => "floatBitsToInt",
-                            (Sk::Float, Sk::Uint) => "floatBitsToUint",
-                            (Sk::Sint, Sk::Float) => "intBitsToFloat",
-                            (Sk::Uint, Sk::Float) => "uintBitsToFloat",
+                        let conv_op = match (source_kind, target_kind, vector_size) {
+                            (Sk::Bool, Sk::Float, Some(size)) => match size {
+                                crate::VectorSize::Bi => "vec2",
+                                crate::VectorSize::Tri => "vec3",
+                                crate::VectorSize::Quad => "vec4",
+                            },
```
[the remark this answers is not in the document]
jimblandy: The linker should optimize out repeated strings, as long as they're read-only (which Rust `&'static str` always is).

jimblandy (changes requested):
This needs some test coverage. Snapshot tests that include a generated use of each of the GLSL functions in the output via this code should suffice, and that'll get the validator to check that it's valid.

src/back/glsl/mod.rs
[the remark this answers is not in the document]
expenses: Ah, no, I mean that we should (at some point) pull these out to a function so that we don't have to do this matching here.

src/back/glsl/mod.rs
```diff
@@ -2930,29 +2930,54 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                            _ => None,
+                        };
+
                         let source_kind = inner.scalar_kind().unwrap();
-                        let conv_op = match (source_kind, target_kind) {
-                            (Sk::Float, Sk::Sint) => "floatBitsToInt",
-                            (Sk::Float, Sk::Uint) => "floatBitsToUint",
-                            (Sk::Sint, Sk::Float) => "intBitsToFloat",
-                            (Sk::Uint, Sk::Float) => "uintBitsToFloat",
+                        let conv_op = match (source_kind, target_kind, vector_size) {
```
jimblandy, line 2939: It seems like almost all of these match arms could be handled with a single default call to `self.write_value_type(inner)`. I'd like to see if that works before we duplicate all those variations here.

jimblandy (changes requested):
Thanks very much for the tests - they look great.

I'm still concerned about code duplication. Could we go around one more time on this?

src/back/glsl/mod.rs
expenses: Oh nice, that makes things super easy!

jimblandy (approved):
Looks good - thank you!

PR #1975, opened 2022-06-07
glsl-out: don't rely on cached expressions

JCapucho:
Expressions marked for caching might not be cached, this can happen for
example when these expressions are pre emitted (for example Contants).

closes https://github.com/gfx-rs/naga/issues/1944


teoxoy:
Should we add a test for this?

JCapucho:
> Should we add a test for this?

@teoxoy added tests and found out that msl suffers from the same problem, thx for reminding to add tests.

teoxoy (approved):
I remember looking at that code a while back feeling something is off there but wasn't completely certain. Glad to see it resolved!

PR #1976, opened 2022-06-07
glsl-in: Splat inputs for smoothstep if needed

### LINKED ISSUE #1971: [glsl-in] naga does not accept all glsl smoothstep overloads
It appears naga does not support certain overloads of the glsl smoothstep function.
The following compute shader compiles successfully on other shader compilers.

```glsl
#version 450
void main() {
    smoothstep(0., 0., vec4(0.));
}
```
naga output:
```
	Expression [9] is invalid
	Argument [2] to SmoothStep as expression [8] has an invalid type.
Generating SPIR-V output requires validation to succeed, and it failed in a previous step
```

here are all the valid overloads copied from the glsl 4.60 spec
```glsl
genFType smoothstep(genFType edge0, genFType edge1, genFType x)
genFType smoothstep(float edge0, float edge1, genFType x)
genDType smoothstep(genDType edge0, genDType edge1, genDType x)
genDType smoothstep(double edge0, double edge1, genDType x)
```
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/smoothstep.xhtml
https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.60.html#common-functions

JCapucho:
Glsl defines two overloads for smoothstep that accept `min` and `max` as
scalars and the value as a vector, naga's IR is stricter and only allows
operators with the same dimensions, so this inputs must be splatted.

closes #1971

### DIFF
--- a/src/front/glsl/builtins.rs
+++ b/src/front/glsl/builtins.rs
@@ -1556,7 +1556,7 @@ fn inject_common_builtin(
                 };
                 declaration.overloads.push(module.add_builtin(
                     vec![ty(), ty(), base_ty()],
-                    MacroCall::MathFunction(MathFunction::SmoothStep),
+                    MacroCall::SmoothStep { splatted: size },
                 ))
             }
         }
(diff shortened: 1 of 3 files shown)


teoxoy (approved):
LGTM

PR #1978, opened 2022-06-07
glsl-out: Perform casts in int only math functions

### LINKED ISSUE #1968: [glsl-out] Invalid GLSL emission when using firstLeadingBit and others
Apologies for the lack of clarity in https://github.com/gfx-rs/naga/issues/1946, but I didn't mention specifically that the GLSL in that issue was emitted by naga after being provided valid WGSL. I assumed it was a glsl-in problem, but it seems I was incorrect - sorry for the confusion! I'm not sure that the GLSL mentioned in the linked issue should be being output by naga, as it's invalid (as answered in the issue).

WGSL:
```
@stage(compute) @workgroup_size(1)
fn main() {
    var var0 = firstLeadingBit(abs(0u));
}
```

GLSL output by naga (included in linked issue, also applies for 450 core):
```
#version 440 core
#extension GL_ARB_compute_shader : require

void main() {
    uint var0_ = findMSB(abs(0u));
}
```

Error when validating above GLSL with naga (included in linked issue):
```
error: Unknown function 'findMSB'
  ┌─ temptest.comp:6:18
  │
6 │     uint var0_ = findMSB(abs(0u));
  │                  ^^^^^^^^^^^^^^^^
```

JCapucho:
Some functions like abs only accept signed integers while naga's IR
accepts both signed and unsigned integers, this wasn't accounted for in
the glsl backend causing it to produce code that wouldn't type check.

This commit addresses it by performing casts from uint to int of the
function argument and then back from int to uint for the function
return.

closes #1968

~~Edit: found some problems that need to be fixed first~~

Edit2: Good to go now

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2810,12 +2810,12 @@ impl<'a, W: Write> Writer<'a, W> {
 
                 // Some GLSL functions always return signed integers (like findMSB),
                 // so they need to be cast to uint if the argument is also an uint.
-                let needs_int_to_uint =
+                let ret_might_need_int_to_uint =
                     matches!(fun, Mf::FindLsb | Mf::FindMsb | Mf::CountOneBits | Mf::Abs);
 
                 // Some GLSL functions only accept signed integers (like abs),
                 // so they need their argument cast from uint to int.
-                let needs_uint_to_int = matches!(fun, Mf::Abs);
+                let arg_might_need_uint_to_int = matches!(fun, Mf::Abs);
 
                 // Check if the argument is an unsigned integer and return the vector size
                 // in case it's a vector
@@ -2832,9 +2832,9 @@ impl<'a, W: Write> Writer<'a, W> {
                     _ => None,
                 };
 
-                if let Some(maybe_size) = maybe_uint_size {
-                    // Cast to uint if the function needs it
-                    if needs_int_to_uint {
+                // Cast to uint if the function needs it
+                if ret_might_need_int_to_uint {
+                    if let Some(maybe_size) = maybe_uint_size {
                         match maybe_size {
                             Some(size) => write!(self.out, "uvec{}(", size as u8)?,
                             None => write!(self.out, "uint(")?,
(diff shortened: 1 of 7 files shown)


teoxoy:
Can we add a test for this?

JCapucho:
> Can we add a test for this?
@teoxoy done

src/back/glsl/mod.rs
```diff
@@ -2798,32 +2798,59 @@ impl<'a, W: Write> Writer<'a, W> {
...
-                                ..
-                            } => {
-                                write!(self.out, "uvec{}(", size as u8)?;
-                                true
-                            }
-                            _ => false,
-                        }
+                // Some GLSL functions always return signed integers (like findMSB),
+                // so they need to be cast to uint if the argument is also an uint.
+                let needs_int_to_uint =
```
teoxoy, line 2803: Can we rename this var to something like `ret_needs_int_to_uint`?

src/back/glsl/mod.rs
```diff
@@ -2798,32 +2798,59 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                        kind: crate::ScalarKind::Uint,
+                        size,
+                        ..
+                    } => Some(Some(size)),
+                    _ => None,
                 };
 
+                if let Some(maybe_size) = maybe_uint_size {
+                    // Cast to uint if the function needs it
+                    if needs_int_to_uint {
```
teoxoy, line 2827: nit: It might be more self explanatory if we first check `if ret_might_need_int_to_uint` and then do the deconstruction of the option via `if let Some`

src/back/glsl/mod.rs
```diff
@@ -2798,32 +2798,59 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                            None => write!(self.out, "uint(")?,
+                        }
+                    }
+                }
+
                 write!(self.out, "{}(", fun_name)?;
+
+                if let Some(maybe_size) = maybe_uint_size {
+                    // Cast to int if the function needs it
+                    if needs_uint_to_int {
```
teoxoy, line 2839: nit: Same as above

teoxoy:
Behavior looks good; just a few things that I think would make the code easier to read.

jimblandy (changes requested):
Just marking this as "changes requested" per @teoxoy's comments, to make it easier to track status.

JCapucho:
Thanks @teoxoy, could you please re-review?

teoxoy (approved):

teoxoy:
Thanks, @jimblandy; not sure what happened there; I didn't mean to review 4 times 😅

jimblandy (approved):

PR #1983, opened 2022-06-13
Update changelog for v0.9

teoxoy:
Updates the changelog for v0.9 (also minor releases of v0.8.x from the 0.8 branch) and should include all user facing changes.

@cwfitzgerald and I discussed a bit about an updated format and this is what we came up with (`wgpu` will also adopt this format).


jimblandy (approved):
So much work - thank you!

PR #1987, opened 2022-06-15
dot-out: Improvements

JCapucho:
Had to debug some issues in the spirv frontend and decided to use the dot backend to help me debug it, this quickly snowballed into me hacking in the dot backend.

Improves the dot backend output by:
- Linking new nodes to the end of other blocks, instead of the beginning
- Generating merge nodes for conditional statements

<details>
  <summary>Comparison</summary>

  ```rust
  fn test(a: i32) {
      if (a == a) {
          var b = a;
          var c = b;
      }
  
      return;
  }
  ```
  
  Old             |  New
  :--------------:|:--------------:
  test | test

</details>

- Generating connections from break/continue nodes to their target

<details>
  <summary>Comparison</summary>

  ```rust
  fn test(a: i32) {
      switch(a) {
          default {
              break;
          }
      }
  }
  ```
  
  Old             |  New
  :--------------:|:--------------:
  old1 | new1
</details>

- Introducing a "cfg only" mode that only generates statements,
  useful for spirv
  
<details>
  <summary>Showcase</summary>

  ```rust
  fn test(a: i32) {
      switch(a) {
          default {
              break;
          }
      }
  }
  ```

cfg-only
</details>


cli/src/main.rs
```diff
@@ -60,6 +60,10 @@ struct Args {
     #[argh(switch)]
     keep_coordinate_space: bool,
 
+    /// enable the cfg only mode in the dot backend
```
jimblandy, line 63: This comment gets turned into the `--help` text from the CLI, so let's try to make it a bit more explicit. Suggest:

> In dot output, include only the control flow graph

jimblandy (approved):
This is an improvement (especially the comments), so I think we should merge it, but it doesn't do too well with code like this:

```
fn f(a: bool, b: bool) -> i32 {
   loop {
      if a {
         break;
      } else {
         break;
      }
      return 5;
   }
   return 0;
}
```

I think `StatementGraph::add` needs to return `(NodeId, Option<NodeId>)`, since some blocks never have control flow off the end. For example, a block that ends in a return, or a break within a loop, shouldn't get an outgoing edge from the end.

JCapucho:
> I think `StatementGraph::add` needs to return `(NodeId, Option<NodeId>)`, since some blocks never have control flow off the end. For example, a block that ends in a return, or a break within a loop, shouldn't get an outgoing edge from the end.

That would be a nice improvement but it isn't so easy because in case a block doesn't have an outgoing edge the graph might be split in two, like this one

test

jimblandy:
Doesn't that mean that the other fragment is unreachable? That seems like the right outcome, no?

JCapucho:
> Doesn't that mean that the other fragment is unreachable? That seems like the right outcome, no?

I don't think it's a good idea to leave it disconnected from the rest of the control flow graph, Imagine if you had 2 or more random fragments floating around

PR #1993, opened 2022-06-17
Add support for break if

### LINKED ISSUE #1831: [wgsl-in] Add break-if as optional at end of continuing
Spec PR: https://github.com/gpuweb/gpuweb/pull/2618

JCapucho:
Adds support for the `break if` statement that may appear in the `continuing` block of a loop.

Relevant part of the spec: https://gpuweb.github.io/gpuweb/wgsl/#break-if-statement

Implements parsing in the wgsl frontend, emission on all the backends and validation for it.

Uniformity analysis isn't implemented, following the same behavior as regular breaks (this will be fixed in the uniformity analysis v5 rewrite).

All commits can be reviewed individually and are fully tested

Closes #1831

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -37,6 +37,28 @@ enum ExpressionPointer {
     },
 }
 
+/// The termination statement to be added to the end of the block
+pub enum BlockExit {
+    /// Generates an OpReturn (void return)
+    Return,
+    /// Generates an OpBranch to the specified block
+    Branch {
+        /// The branch target block
+        target: Word,
+    },
+    /// Translates a loop `break if` into an `OpBranchConditional` to the
+    /// merge block if true (the merge block is passed through [`LoopContext::break_id`]
+    /// or else to the loop header (passed through [`preamble_id`])
+    ///
+    /// [`preamble_id`]: Self::BreakIf::preamble_id
+    BreakIf {
+        /// The condition of the `break if`
+        condition: Handle<crate::Expression>,
+        /// The loop header block id
+        preamble_id: Word,
+    },
+}
+
 impl Writer {
     // Flip Y coordinate to adjust for coordinate space difference
     // between SPIR-V and our IR.
@@ -1491,7 +1513,7 @@ impl<'w> BlockContext<'w> {
         &mut self,
         label_id: Word,
         statements: &[crate::Statement],
-        exit_id: Option<Word>,
+        exit: BlockExit,
         loop_context: LoopContext,
     ) -> Result<(), Error> {
         let mut block = Block::new(label_id);
@@ -1508,7 +1530,12 @@ impl<'w> BlockContext<'w> {
                     self.function.consume(block, Instruction::branch(scope_id));
 
                     let merge_id = self.gen_id();
-                    self.write_block(scope_id, block_statements, Some(merge_id), loop_context)?;
+                    self.write_block(
+                        scope_id,
+                        block_statements,
+                        BlockExit::Branch { target: merge_id },
+                        loop_context,
+                    )?;
 
                     block = Block::new(merge_id);
                 }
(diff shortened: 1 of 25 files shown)


JCapucho:
Nightly is changing the formatting of some parts, and if those changes are formatted with stable they are reverted

src/back/spv/block.rs
```diff
@@ -37,6 +37,28 @@ enum ExpressionPointer {
...
+pub enum BlockExit {
+    /// Generates an OpReturn (void return)
+    Return,
+    /// Generates an OpBranch to the specified block
+    Branch {
+        /// The branch target block
+        block_id: Word,
+    },
+    /// Translates a loop `break if` into an `OpBranchConditional` to the
+    /// merge block if true (the merge block is passed trough [`LoopContext::break_id`]
```
jimblandy, line 50: "through"

src/back/spv/block.rs
```diff
@@ -37,6 +37,28 @@ enum ExpressionPointer {
...
+    /// or else to the loop header (passed trough [`preamble_id`])
```
jimblandy, line 51: "through"

src/back/spv/block.rs
```diff
@@ -37,6 +37,28 @@ enum ExpressionPointer {
...
+    ///
+    /// [`preamble_id`]: Self::BreakIf::preamble_id
```
jimblandy, line 53: Whoa, I didn't know you could use `Self` like that.

src/back/spv/block.rs
```diff
@@ -37,6 +37,28 @@ enum ExpressionPointer {
...
 }
 
+/// The termination statement to be added to the end of the block
+pub enum BlockExit {
+    /// Generates an OpReturn (void return)
+    Return,
+    /// Generates an OpBranch to the specified block
+    Branch {
+        /// The branch target block
+        block_id: Word,
```
jimblandy, line 47: Suggestion: I would call this `target`. It makes the various construction sites in `write_block` read better.

src/back/spv/block.rs
```diff
@@ -1955,19 +2012,31 @@ impl<'w> BlockContext<'w> {
...
+            BlockExit::Branch { block_id } => Instruction::branch(block_id),
+            BlockExit::BreakIf {
+                condition,
+                preamble_id,
+            } => {
+                let condition_id = self.cached[condition];
+
+                Instruction::branch_conditional(
+                    condition_id,
+                    loop_context.break_id.unwrap(),
```
jimblandy, line 2036: This `unwrap` makes me wonder whether `LoopContext` and `BlockExit` couldn't be fused into a single enum. But that can be another PR.

jimblandy (changes requested):
Thank you so much for 1) this PR and 2) leaving the SPIR-V front end change to a separate PR so my brain doesn't break.

This also needs some failure case coverage in `tests/wgsl-errors.rs`:
- misplaced `break if` statements
- non-`bool` expressions

A few suggestions, too.

src/back/spv/block.rs
JCapucho: They could be merged but it would make it more painful to handle them in the `Break` and `Continue` statements

src/back/spv/block.rs
JCapucho: Item paths in doc comments have the same scope as the item were they are defined, it's very useful.

Shame there's no shorthand for addressing the variant the doc comment is on :(

src/back/wgsl/writer.rs
```diff
@@ -917,11 +918,26 @@ impl<W: Write> Writer<W> {
                     self.write_stmt(module, sta, func_ctx, l2)?;
                 }
 
-                if !continuing.is_empty() {
+                // The continuing is optional so we don't need to write it if
+                // it is empty, but the `break if` counts as a continuing statement
+                // so even if `continuing` is empty we must generate it if a
+                // `break if` exists
```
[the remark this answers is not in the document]
JCapucho: I have been implementing your suggestion and honestly I think it's more effort than it's worth it, because now everytime you encounter a continuing block you need to match on it, access the body and continue your work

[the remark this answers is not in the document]
jimblandy: Okay, thanks for trying it out. I'm fine with this as originally posted.

jimblandy (approved):

PR #1995, opened 2022-06-28
spv-in: Fix incorrect translation of SMod

JCapucho:
Since spirv's SMod doesn't map to naga's IR modulo operator the
instruction is mapped into it's direct formula, this formula requires
some casts from int to float and back.

These casts were present but their `convert` field was left to `None`
which implied that they were bitcasts, causing some wildly incorrect
results.

This commit fixes it by setting the `convert` field to the same size as
the result type.


cwfitzgerald (approved):
Thanks!

PR #2002, opened 2022-07-08
Handle Unpack2x16float in hlsl

expenses:
[`f16tof32`](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/f16tof32) handles unpacking of the lower half of a `uint` to a 16-bit float, so we need to call that twice, with a bitshift the second time.

### DIFF
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -2517,9 +2517,9 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     Function::Unpack2x16float => {
                         write!(self.out, "float2(f16tof32(")?;
                         self.write_expr(module, arg, func_ctx)?;
-                        write!(self.out, "), f16tof32(")?;
+                        write!(self.out, "), f16tof32((")?;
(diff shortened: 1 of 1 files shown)


src/back/hlsl/writer.rs
```diff
@@ -2512,6 +2514,13 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                         self.write_expr(module, arg, func_ctx)?;
                         write!(self.out, "))")?;
                     }
+                    Function::Unpack2x16float => {
+                        write!(self.out, "float2(f16tof32(")?;
+                        self.write_expr(module, arg, func_ctx)?;
+                        write!(self.out, "), f16tof32(")?;
+                        self.write_expr(module, arg, func_ctx)?;
+                        write!(self.out, " >> 16))")?;
```
expenses, line 2522: Do we need a second pair of parenthesis here?

cwfitzgerald (approved):
Cool stuff!

cwfitzgerald (approved):

PR #2005, opened 2022-07-10
hlsl-out: Add support for push constants

### LINKED ISSUE #1782: [hlsl-out] Implement push constants
Tracking issue for implementing push constants for the HLSL backend

https://github.com/gfx-rs/naga/blob/f90e563c281cfc71c794e0426ebcced9e3999202/src/back/hlsl/writer.rs#L583

JCapucho:
Push constants need to be configured by the consumer which must pass the
bind target of the constant buffer used for the push constants.

Closes #1782

### DIFF
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -633,27 +633,7 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
         // If the global is a push constant write the type now because it will be a
         // generic argument to `ConstantBuffer`
         if global.space == crate::AddressSpace::PushConstant {
-            let matrix_data = get_inner_matrix_data(module, global.ty);
-
-            // We treat matrices of the form `matCx2` as a sequence of C `vec2`s.
-            // See the module-level block comment in mod.rs for details.
-            if let Some(MatrixType {
-                columns,
-                rows: crate::VectorSize::Bi,
-                width: 4,
-            }) = matrix_data
-            {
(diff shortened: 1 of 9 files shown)


src/back/hlsl/writer.rs
```diff
@@ -622,12 +622,65 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
         };
 
+        // If the global is a push constant write the type now because it will be a
+        // generic argument to `ConstantBuffer`
+        if global.space == crate::AddressSpace::PushConstant {
+            let matrix_data = get_inner_matrix_data(module, global.ty);
+
+            // We treat matrices of the form `matCx2` as a sequence of C `vec2`s.
+            // See the module-level block comment in mod.rs for details.
+            if let Some(MatrixType {
```
cwfitzgerald, line 639: This feels like it's duplicated with some other place, if it is, can we deduplicate it?

tests/out/hlsl/push-constants.hlsl
```diff
@@ -0,0 +1,21 @@
+
+struct PushConstants {
```
cwfitzgerald, line 2: How does this interact with vertex shaders' builtin constants? It would be useful to put that into a test.

cwfitzgerald (changes requested):

cwfitzgerald (approved):
This seems well tested and generally g2g.

cwfitzgerald:
Seems to just need a rebase.

PR #2009, opened 2022-07-16
Update reserved WGSL keywords

### LINKED ISSUE #2007: `sample` is not a reserved keyword according to WebGPU spec
`sample` is listed among the reserved keywords here:

https://github.com/gfx-rs/naga/blob/17dcf39a1b2e4282943580302f510955704650f5/src/keywords/wgsl.rs#L301

However, the WebGPU spec does not consider it a reserved keyword: https://www.w3.org/TR/WGSL/#reserved-words.

There may be other reserved keywords that need updating; `sample` is just the one I ran into.

norepimorphism:
Fixes #2007. This brings us up-to-speed with https://www.w3.org/TR/WGSL/#reserved-words.


norepimorphism:
Oops. Disregard that.

cwfitzgerald:
Thanks!

PR #2012, opened 2022-07-17
add `emit_to_{stderr,string}` helpers to validation error

nolanderc:
Mirrors those already found for parse errors. Also took the time to remove the `StringErrorBuffer` writer in favor of the existing `NoColor` writer and added the `emit_to_string_with_path` method to complete the set.


JCapucho (approved):
LGTM

PR #2013, opened 2022-07-21
Implement `Clone` for `Module`

daxpedda:
Allows `Module` to be cloned.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ panic = "abort"
 
 [features]
 default = []
+clone = []
 dot-out = []
 glsl-in = ["pp-rs"]
 glsl-validate = []
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -161,7 +161,7 @@ impl<T> Iterator for Range<T> {
 /// Adding new items to the arena produces a strongly-typed [`Handle`].
 /// The arena can be indexed using the given handle to obtain
 /// a reference to the stored item.
-#[derive(Clone)]
+#[cfg_attr(feature = "clone", derive(Clone))]
 #[cfg_attr(feature = "serialize", derive(serde::Serialize))]
 #[cfg_attr(feature = "serialize", serde(transparent))]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
@@ -413,7 +413,7 @@ mod tests {
 ///
 /// `UniqueArena` is similar to [`Arena`]: If `Arena` is vector-like,
 /// `UniqueArena` is `HashSet`-like.
-#[derive(Clone)]
+#[cfg_attr(feature = "clone", derive(Clone))]
 pub struct UniqueArena<T> {
     set: IndexSet<T>,
 
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -161,6 +161,7 @@ impl<T> Iterator for Range<T> {
 /// Adding new items to the arena produces a strongly-typed [`Handle`].
 /// The arena can be indexed using the given handle to obtain
 /// a reference to the stored item.
+#[derive(Clone)]
 #[cfg_attr(feature = "serialize", derive(serde::Serialize))]
 #[cfg_attr(feature = "serialize", serde(transparent))]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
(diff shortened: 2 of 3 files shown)


teoxoy:
I think we should implement this via a feature flag that's on by default (and run CI with default features off) since @kvark previously mentioned that having these `Clone` implementations makes it too easy for the contributors of the project to clone potentially expensive structures unnecessarily.

daxpedda:
I used `cfg(dev)` instead, I hope that's appropriate.

teoxoy:
Oh, I think that's even better!

teoxoy (approved):
LGTM but let's wait for one more review

JCapucho:
I think we should do the opposite and add a feature flag to enable it that's not the default, because I think most people that would want to dabble in naga will try to run
```bash
$ cargo run
```
And this will cause them to be able to `Clone` even though they shouldn't and then later when they try to open a PR it will fail on CI and debugging it would be a little bit counter intuitive.

But either way we should document this feature flag

daxpedda:
I don't particularly understand the problem, calling `clone()` is explicit, why are we afraid of people misusing it or not be able to catch it in code review? I mean `Vec`s and `HashMap`s are also cloneable and they can be pretty large too, any `clone()` call should be paid attention to.

> I think we should do the opposite and add a feature flag to enable it that's not the default, ...

This PR is specifically made for https://github.com/gfx-rs/wgpu/pull/2903, so not having it on by default doesn't really work.

Alternatively, instead of `Clone`, I could also implement [`ToOwned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html), this should stand out even more.

Otherwise I could change https://github.com/gfx-rs/wgpu/pull/2903 to not use `Cow`, but some custom type not requiring `ToOwned` or `Clone`.

> But either way we should document this feature flag

I could add a note under [README#development-workflow](https://github.com/gfx-rs/naga/blob/cfbc91d6cf34983d25418a201f00d015dc23e6fc/README.md#development-workflow)?

JCapucho:
> I don't particularly understand the problem, calling `clone()` is explicit, why are we afraid of people misusing it or not be able to catch it in code review? I mean `Vec`s and `HashMap`s are also cloneable and they can be pretty large too, any `clone()` call should be paid attention to.

While `Clone`s are explicit they are a tool that people can easily reach to and will do so, and naga doesn't clone IR internally for memory and performance considerations.

And while we can review them, doing so in every PR and pointing it out every time is an added task to the reviewer and some are voluntary so easing their work is always better and it's also unexpected to the contributor because they saw nowhere that the codebase doesn't accept it and suddenly are being told to not do it.

> This PR is specifically made for [gfx-rs/wgpu#2903](https://github.com/gfx-rs/wgpu/pull/2903), so not having it on by default doesn't really work.
> 
> Alternatively, instead of `Clone`, I could also implement [`ToOwned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html), this should stand out even more.
> 
> Otherwise I could change [gfx-rs/wgpu#2903](https://github.com/gfx-rs/wgpu/pull/2903) to not use `Cow`, but some custom type not requiring `ToOwned` or `Clone`.

I don't understand the problem. wgpu is a user of naga so it can add to it's `Cargo.toml` the feature explicitly and I assume it already does so for other features.

No need to implement other traits.

> > But either way we should document this feature flag
> 
> I could add a note under [README#development-workflow](https://github.com/gfx-rs/naga/blob/cfbc91d6cf34983d25418a201f00d015dc23e6fc/README.md#development-workflow)?

If we do end up following the negative feature flag then yes this would be the preferred place

daxpedda:
> I don't understand the problem. wgpu is a user of naga so it can add to it's `Cargo.toml` the feature explicitly and I assume it already does so for other features.

Yeah, I don't know what I was thinking, will do that.

daxpedda:
Alright, I changed it to a non-default crate feature and documented it in the crate documentation!

JCapucho (approved):
I'll just wait for @teoxoy to re-review since he originally approved it but otherwise it seems good to me, thanks for your contribution.

daxpedda:
Sorry, I accidentally requested a review from you again.

teoxoy (approved):
Looks good; I think the previous `cfg(dev)` could have clashed with downstream users possibly setting the cmd line flag `--cfg dev` themselves.

PR #2014, opened 2022-07-23
Make termcolor dependency optional

AldaronLau:
Termcolor only needs to be pulled in to the dependencies when codespan-reporting is pulled in.


cwfitzgerald (approved):
Very cool beans

PR #2015, opened 2022-07-25
spv-out: Don't emit cast when types are the same

guusw:
While working with some shaders with wgpu & naga on android I noticed random driver crashes with the following logs:
```plain
07-25 13:17:37.578 15860 15900 I ____  : [2022-07-25 13:17:37.578] [GFX.Context] [info] Skip function Some("entryPoint_vertex_7")
07-25 13:17:37.578 15860 15900 I ____  : [2022-07-25 13:17:37.578] [GFX.Context] [info] Skip function Some("entryPoint_vertex_8")
07-25 13:17:37.587 15860 15900 I AdrenoVK-0: Shader compilation failed for shaderType: 1
07-25 13:17:37.587 15860 15900 W ____  : [2022-07-25 13:17:37.587] [GFX.Context] [warning] Unrecognized device error ERROR_UNKNOWN
07-25 13:17:37.587 15860 15900 E ____  : [2022-07-25 13:17:37.587] [GFX.Context] [error] [context.cpp:415] Device(Lost) (3)
--------- beginning of crash
```

After digging around I was able to isolate it to the usage of OpBitcast on a value that is already the correct type (in my case casting float to float)

```wgsl
struct Fragment_globals_t {
    n_dot_l: f32,
};
var<private> p_Fragment_globals: Fragment_globals_t;

fn entryPoint_fragment_1() {
    // %float_0 = OpConstant %float 0
    // ...

    // This breaks
    // %179 = OpBitcast %float %float_0
    // %180 = OpAccessChain %_ptr_Private_float %p_Fragment_globals %uint_5
    //        OpStore %180 %179
    p_Fragment_globals.n_dot_l = f32(0.000000);

    // This doesn't
    // %179 = OpAccessChain %_ptr_Private_float %p_Fragment_globals %uint_5
    //        OpStore %179 %float_0
    p_Fragment_globals.n_dot_l = 0.000000;
}
```

I tried to make a change that doesn't emit the casting instruction when the input and output types are the same, which In my case fixes this issue.

Would it be possible to merge something like this to avoid this driver issue?

I'm not too familiar with SPIR-V or this codebase so let me know If I missed something.


src/back/spv/block.rs
```diff
@@ -1087,18 +1087,22 @@ impl<'w> BlockContext<'w> {
...
-                    Cast::Binary(op, operand) => {
-                        Instruction::binary(op, result_type_id, id, expr_id, operand)
-                    }
-                    Cast::Ternary(op, op1, op2) => {
-                        Instruction::ternary(op, result_type_id, id, expr_id, op1, op2)
-                    }
-                };
-                block.body.push(instruction);
-                id
+                if src_kind == kind {
```
JCapucho, line 1090: This check also needs to account for the bit width of the type since an `f32` isn't equal to a `f64`.

JCapucho (changes requested):
I'm not familiar with the way driver issues are handled in wgpu and naga, so I need either @jimblandy or @kvark input here.

But there's a problem with your code that needs to be fixed either way

src/back/spv/block.rs
guusw: It should check the width now too.

jimblandy:
> I'm not familiar with the way driver issues are handled in wgpu and naga, so I need either @jimblandy or @kvark input here.

tl;dr: In broad terms, this seems fine. If this change were more involved, we'd have to make a harder call.

I think it's totally fine to make non-disruptive adjustments to accommodate driver bugs. A localized change that drops a no-op instruction seems harmless.

The underlying pragmatics are that we want `wgpu` to Just Work for the vast majority of users, and that driver quality is uneven, so a strict principled stand of never working around driver bugs just isn't in the cards for us. Titanic presences like Unity can pressure people to fix their drivers, so a principled stand might work for them. But `wgpu`'s not there... yet (j/k).

Also, it's my experience is that, without being able to read the source, it's easy to misunderstand the range of behaviors affected by a bug, so a workaround that helps today may not help tomorrow. That makes it riskier to invest time and complexity in the workaround. So a more complex workaround would require stronger justification.

jimblandy:
@guusw What kind of GPU are you using? Can you somehow run `wgpu-info` on it and post the output?

jimblandy:
I'd like to handle this change a bit differently. I've filed #2043 with my changes.

PR #2017, opened 2022-07-27
[glsl-in] add support for .length()

### LINKED ISSUE #2016: [glsl-in] naga rejects .length() unconditionally
minimal example:
```glsl
#version 440 core

layout(std430, binding = 0) buffer something_buf {
    uint something[];
};

void main() {
    uint test = uint(something.length());
}
```

naga-cli rejects this with the following error
```
error: Expected Comma or RightParen, found LeftParen
  ┌─ minimal.comp:8:38
  │
8 │     uint test = uint(something.length());
  │                                      ^
```

SpaceCat-Chan:
this fixes #2016

i see two reasons you may not want to merge this
1. this assumes that .length() is only valid on arrays and never has any other meaning
2. this assumes that .length() has the same semantics as `Expression::ArrayLength`, which it may not have

other than that there should be no problems, other than that this was done in a hasty way

### DIFF
--- a/src/front/glsl/context.rs
+++ b/src/front/glsl/context.rs
@@ -1412,13 +1412,7 @@ impl Context {
                                     meta,
                                     body,
                                 );
-                                self.forced_conversion(
-                                    parser,
-                                    &mut array_length,
-                                    meta,
-                                    ScalarKind::Sint,
-                                    4,
-                                )?;
+                                self.conversion(&mut array_length, meta, ScalarKind::Sint, 4)?;
                                 array_length
                             }
                         }
--- a/src/front/glsl/context.rs
+++ b/src/front/glsl/context.rs
@@ -1387,20 +1387,32 @@ impl Context {
                         }
                         let lowered_array =
                             self.lower_expect_inner(stmt, parser, object, pos, body)?.0;
-                        let array_length = self.add_expression(
-                            Expression::ArrayLength(lowered_array),
-                            stmt.hir_exprs[expr].meta,
-                            body,
-                        );
-                        self.add_expression(
-                            Expression::As {
-                                expr: array_length,
-                                kind: ScalarKind::Sint,
-                                convert: Some(4),
-                            },
-                            stmt.hir_exprs[expr].meta,
-                            body,
-                        )
+                        let array_type = parser.resolve_type(self, lowered_array, meta)?;
+                        match array_type {
+                            &TypeInner::Array {
+                                base: _,
+                                size: crate::ArraySize::Constant(size),
+                                stride: _,
+                            } => {
+                                let mut array_length = self.add_expression(
+                                    Expression::Constant(size),
+                                    stmt.hir_exprs[expr].meta,
+                                    body,
+                                );
+                                self.conversion(&mut array_length, meta, ScalarKind::Sint, 4)?;
+                                array_length
+                            }
+                            // let the error be handeled in type checking if it's not a dynamic array
+                            _ => {
+                                let mut array_length = self.add_expression(
+                                    Expression::ArrayLength(lowered_array),
+                                    stmt.hir_exprs[expr].meta,
+                                    body,
+                                );
+                                self.conversion(&mut array_length, meta, ScalarKind::Sint, 4)?;
+                                array_length
+                            }
+                        }
                     }
                     _ => {
                         return Err(Error {
--- a/src/front/glsl/context.rs
+++ b/src/front/glsl/context.rs
@@ -1366,13 +1366,41 @@ impl Context {
                     value
                 }
             }
-            HirExprKind::GetLength { array } if ExprPos::Lhs != pos => {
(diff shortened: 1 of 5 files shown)


JCapucho:
> this fixes #2016

thank you for working on this

> i see two reasons you may not want to merge this
> 
>     1. this assumes that .length() is only valid on arrays and never has any other meaning

According to the glsl specification this should be the case

>     2. this assumes that .length() has the same semantics as `Expression::ArrayLength`, which it may not have

There's a slight difference, `Expression::ArrayLength` follows wgsl semantics as almost all of naga's IR follows, in wgsl `arrayLength` returns a `u32` which is glsl's `uint`, meanwhile `.length()` should return a `int` so you need to add a cast.

> other than that there should be no problems, other than that this was done in a hasty way

I do have another issue with the current implementation, it specializes the parsing on `.length()` which is the only method on the base glsl specification, but I don't know if there any extensions that add other methods or might appear new extensions that do so, so I think we should parse a `HirExprKind::Method` and then handle this in lowering, this also has the added benefit that error messages can be better instead of `no such field`

SpaceCat-Chan:
latest commit adds a cast to Sint, and makes the parsing parse any method and then only specializes on .length in lowering instead

JCapucho (changes requested):
Amazing work, I wasn't expecting you to include argument handling but nice that you added it.

There are some issues that need to be addressed and I also added some nits.

Also please add a test (you can add it to `expressions.frag`)

JCapucho (commented):
Amazing work, but I noticed while testing your changes that `ArrayLength` only works on pointer to dynamic arrays, while glsl expects `.length` to work on all arrays.

To fix this we would need to check the type of the expression and if it's an array of fixed size or a pointer to it pass it as a constant.

src/front/glsl/context.rs
```diff
@@ -1366,6 +1366,64 @@ impl Context {
...
+                                    ".length() doesn't take any arguments".into(),
+                                ),
+                                meta,
+                            });
+                        }
+                        let lowered_array =
+                            self.lower_expect_inner(stmt, parser, object, pos, body)?.0;
+                        let array_type = parser.resolve_type(self, lowered_array, meta)?;
+                        match array_type {
+                            &TypeInner::Array {
```
JCapucho: Prefer de referencing the match expression 
```suggestion
                        match *array_type {
                            TypeInner::Array {
```

src/front/glsl/context.rs
```diff
@@ -1366,6 +1366,64 @@ impl Context {
...
+                                base: _,
+                                size: crate::ArraySize::Constant(size),
+                                stride: _,
```
JCapucho: Prefer using ellipsis `..`
```suggestion
                                size: crate::ArraySize::Constant(size),
                                ..
```

src/front/glsl/context.rs
```diff
@@ -1366,6 +1366,64 @@ impl Context {
...
+                            } => {
+                                let mut array_length = self.add_expression(
+                                    Expression::Constant(size),
+                                    stmt.hir_exprs[expr].meta,
+                                    body,
+                                );
+                                self.conversion(&mut array_length, meta, ScalarKind::Sint, 4)?;
```
JCapucho: Use `forced_conversion` in this case because constant array sizes might already be `Sint`
```suggestion
                                self.forced_conversion(&mut array_length, meta, ScalarKind::Sint, 4)?;
```

tests/in/glsl/expressions.frag
```diff
@@ -132,6 +132,14 @@ void testMatrixMultiplication(mat4x3 a, mat4x4 b) {
...
 }
 
+layout(std430, binding = 0) buffer a_buf {
+    float a[];
+};
+
+void testLength() {
+    int len = a.length();
+}
+
```
JCapucho, line 142: Let's add another test for arrays with constant size
```suggestion

void testConstantLength(float a[4u]) {
    int len = a.length();
}

```

JCapucho (changes requested):
Almost perfect just left some nits and ask that you add a test for constant size arrays.

JCapucho:
Sorry I might have misled you but since we are already lowering with a non `Lhs` context we don't need to check if the array is wrapped in a pointer.

src/front/glsl/context.rs
```diff
@@ -1366,6 +1366,73 @@ impl Context {
...
+                                    meta,
+                                    body,
+                                );
+                                self.forced_conversion(
+                                    parser,
+                                    &mut array_length,
+                                    meta,
+                                    ScalarKind::Sint,
+                                    4,
+                                )?;
```
JCapucho, line 1421: I only meant the previous branch, in here we can be sure it will be an `Uint` and a cast will be neeeded

JCapucho (changes requested):

JCapucho (approved):
Perfect, thanks for your contribution.

PR #2024, opened 2022-08-06
wgsl-in: Implement lexical scopes

### LINKED ISSUE #2021: [wgsl-in] let declarations from different scopes merged?
Shadowed let declarations can get incorrectly hoisted out of conditional scopes.
See example WGSL:
```
@vertex
fn main(@location(0) cond: f32) -> @location(0) f32 {
    let value = 1.;
    if cond > 0. {
        let value = 2.;
        return 3.;
    }
    return value;
}
```
This shader always returns 2 when the if condition evaluates to false.
Same shader translated to GLSL by Naga:
```
#version 310 es

precision highp float;
precision highp int;

layout(location = 0) in float _p2vs_location0;
layout(location = 0) smooth out float _vs2fs_location0;

void main() {
    float cond = _p2vs_location0;
    if ((cond > 0.0)) {
        _vs2fs_location0 = 3.0;
        gl_Position.yz = vec2(-gl_Position.y, gl_Position.z * 2.0 - gl_Position.w);
        return;
    }
    _vs2fs_location0 = 2.0;
    gl_Position.yz = vec2(-gl_Position.y, gl_Position.z * 2.0 - gl_Position.w);
    return;
}
```
Tested this happens on 0.9 and HEAD.

JCapucho:
Previously the wgsl frontend wasn't aware of lexical scopes causing all variables and named expressions to share a single function scope, this meant that if a variable was defined in a block with the same name as a variable in the function body, the variable in the function body would be lost and exiting the block all references to the variable in the function body would be replaced with the variable of the block.

This PR fixes that in two commits, the first commit introduces a new frontend agnostic type `SymbolTable` that will be responsible for managing lexical scopes and performing variable lookups, this type is only used in wgsl right now but I intend to change glsl's internal symbol table for it, the second commit wires it up to the wgsl frontend and inserts the proper push/pop of lexical scopes according to the wgsl spec.

A snapshot test is also included to ensure all lexical scopes are working.

Fixes #2021

### DIFF
--- a/src/front/mod.rs
+++ b/src/front/mod.rs
@@ -14,6 +14,7 @@ pub mod wgsl;
 use crate::{
     arena::{Arena, Handle, UniqueArena},
     proc::{ResolveContext, ResolveError, TypeResolution},
+    FastHashMap,
 };
 use std::ops;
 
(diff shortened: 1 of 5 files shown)


jimblandy:
This isn't really a problem with this PR, but one concern I have is that the WGSL `Parser` type already has functions named `push_scope` and `pop_scope`. Those do something entirely different from `SymbolTable`'s methods - they're purely for error checking - but it's confusing to have both of these flying around.

But I think this PR's use of the term 'scope' is more apt. Once it's merged, we should rename `naga::front::wgsl::Scope` to `Rule`, with corresponding changes to the related functions and variables.

jimblandy (approved):
This looks great - thank you!

I've put a bunch of copy-edits here: https://github.com/jimblandy/naga/tree/copy-edit
If they look good, then smoosh them into your commit before landing.

JCapucho:
> This looks great - thank you!
> 
> I've put a bunch of copy-edits here: https://github.com/jimblandy/naga/tree/copy-edit If they look good, then smoosh them into your commit before landing. EDIT: I wanted to merge immediately, so I took care of this.

@jimblandy if it's just some typos you don't even need to ask, thanks for fixing them.

PR #2025, opened 2022-08-11
Saturate

evahop:
Adds [saturate](https://www.w3.org/TR/WGSL/#saturate-float-builtin) and converts it to `clamp(e, 0.0, 1.0)` where necessary.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2755,20 +2755,12 @@ impl<'a, W: Write> Writer<'a, W> {
                         self.write_expr(arg, ctx)?;
 
                         match *ctx.info[arg].ty.inner_with(&self.module.types) {
-                            crate::TypeInner::Vector {
-                                size,
-                                kind: crate::ScalarKind::Float,
-                                ..
-                            } => write!(
+                            crate::TypeInner::Vector { size, .. } => write!(
                                 self.out,
                                 ", vec{}(0.0), vec{0}(1.0)",
                                 back::vector_size_str(size)
                             )?,
-                            crate::TypeInner::Scalar {
-                                kind: crate::ScalarKind::Float,
-                                ..
-                            } => write!(self.out, ", 0.0, 1.0")?,
-                            ref other => unreachable!("Unexpected type {:?}", other),
+                            _ => write!(self.out, ", 0.0, 1.0")?,
                         }
 
                         write!(self.out, ")")?;
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2749,6 +2749,7 @@ impl<'a, W: Write> Writer<'a, W> {
                     Mf::Min => "min",
                     Mf::Max => "max",
                     Mf::Clamp => "clamp",
+                    Mf::Saturate => todo!(),
                     // trigonometry
                     Mf::Cos => "cos",
                     Mf::Cosh => "cosh",
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2754,7 +2754,7 @@ impl<'a, W: Write> Writer<'a, W> {
 
                         self.write_expr(arg, ctx)?;
 
-                        match ctx.info[arg].ty.inner_with(&self.module.types) {
+                        match *ctx.info[arg].ty.inner_with(&self.module.types) {
                             crate::TypeInner::Vector {
                                 size,
                                 kind: crate::ScalarKind::Float,
@@ -2762,13 +2762,13 @@ impl<'a, W: Write> Writer<'a, W> {
                             } => write!(
                                 self.out,
                                 ", vec{}(0.0), vec{0}(1.0)",
-                                back::vector_size_str(*size)
+                                back::vector_size_str(size)
                             )?,
                             crate::TypeInner::Scalar {
                                 kind: crate::ScalarKind::Float,
                                 ..
                             } => write!(self.out, ", 0.0, 1.0")?,
-                            other => unreachable!("Unexpected type {:?}", other),
+                            ref other => unreachable!("Unexpected type {:?}", other),
                         }
 
                         write!(self.out, ")")?;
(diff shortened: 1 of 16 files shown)


src/proc/typifier.rs
```diff
@@ -704,6 +704,7 @@ impl<'a> ResolveContext<'a> {
                     Mf::Min |
                     Mf::Max |
                     Mf::Clamp |
+                    Mf::Saturate |
```
JCapucho, line 707: This seems like the wrong place to put it, if we look at the code for this match arm we have:

```rust
match *res_arg.inner_with(types) {
  Ti::Vector {
      kind,
      size: _,
      width,
  } => TypeResolution::Value(Ti::Scalar { kind, width }),
  ref other =>
      return Err(ResolveError::IncompatibleOperands(
          format!("{:?}({:?}, _)", fun, other)
      )),
},
```

but saturate is defined as

```
@const fn saturate(e: T) -> T
where
S is AbstractFloat, f32, or f16
T is S or vecN<S>

File: src/valid/expression.rs [hunk 2]
```diff
@@ -1040,6 +1040,20 @@ impl super::Validator {
                             ));
                         }
                     }
+                    Mf::Saturate => {
```
JCapucho, line 1043: You can merge this with the match below, they have the exact same body

src/back/glsl/mod.rs
```diff
@@ -2749,6 +2749,32 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                            } => write!(
+                                self.out,
+                                ", vec{}(0.0), vec{0}(1.0)",
+                                back::vector_size_str(size)
+                            )?,
+                            crate::TypeInner::Scalar {
+                                kind: crate::ScalarKind::Float,
+                                ..
+                            } => write!(self.out, ", 0.0, 1.0")?,
+                            ref other => unreachable!("Unexpected type {:?}", other),
```
JCapucho, line 2771: This is already handled in validation we can just use a catch all:
```suggestion
                            _ => write!(self.out, ", 0.0, 1.0")?,
```

src/back/glsl/mod.rs
```diff
@@ -2749,6 +2749,32 @@ impl<'a, W: Write> Writer<'a, W> {
...
                     Mf::Clamp => "clamp",
+                    Mf::Saturate => {
+                        write!(self.out, "clamp(")?;
+
+                        self.write_expr(arg, ctx)?;
+
+                        match *ctx.info[arg].ty.inner_with(&self.module.types) {
+                            crate::TypeInner::Vector {
+                                size,
+                                kind: crate::ScalarKind::Float,
```
JCapucho, line 2760: Same here
```suggestion
```

src/back/spv/block.rs
```diff
@@ -696,6 +696,88 @@ impl<'w> BlockContext<'w> {
...
+                            MathOp::Custom(Instruction::ext_inst(
+                                self.writer.gl450_ext_inst_id,
+                                spirv::GLOp::FClamp,
+                                result_type_id,
+                                id,
+                                &[arg0_id, arg1_id, arg2_id],
+                            ))
+                        }
+                        ref other => unimplemented!("Unexpected saturate({:?})", other),
+                    },
```
JCapucho, line 780: This could could be greatly simplified, we could do something like this
```suggestion
                    Mf::Saturate => {
                      let (maybe_size, width) = match *arg_ty {
                          crate::TypeInner::Vector {
                              size,
                              width,
                              ..
                          } => (Some(size), width),
                          crate::TypeInner::Scalar { width, .. } => (None, width),
                          ref other => unimplemented!("Unexpected saturate({:?})", other),
                      };
                      
                      let mut arg1_id = self
                          .writer
                          .get_constant_scalar(crate::ScalarValue::Float(0.0), width);
                      let mut arg2_id = self
                          .writer
                          .get_constant_scalar(crate::ScalarValue::Float(1.0), width);
                          
                      if let Some(size) = maybe_size {
                        let value = LocalType::Value {
                            vector_size: Some(size),
                            kind: crate::ScalarKind::Float,
                            width,
                            pointer_space: None,
                        };

                        let result_type_id = self.get_type_id(LookupType::Local(value));
  
                        self.temp_list.clear();
                        self.temp_list.resize(size as _, arg1_id);

                        let id = self.gen_id();
                        block.body.push(Instruction::composite_construct(
                            result_type_id,
                            id,
                            &self.temp_list,
                        ));
                        arg1_id = id;
  
                        self.temp_list.clear();
                        self.temp_list.resize(size as _, arg2_id);

                        let id = self.gen_id();
                        block.body.push(Instruction::composite_construct(
                            result_type_id,
                            id,
                            &self.temp_list,
                        ));
                        arg2_id = id;
                      }

                      MathOp::Custom(Instruction::ext_inst(
                          self.writer.gl450_ext_inst_id,
                          spirv::GLOp::FClamp,
                          result_type_id,
                          id,
                          &[arg0_id, arg1_id, arg2_id],
                      ))
                   },
```

JCapucho (changes requested):
Thanks for your contribution, there are some problems with some parts of it and some code could be improved but overall it's on the right track.

You should add a test (`math-functions.wgsl` seems like a good place).

Also the spec you point to is outdated, the newer version is at https://gpuweb.github.io/gpuweb/wgsl/#saturate-float-builtin

src/proc/typifier.rs
evahop: Thanks, making the changes now, this one appears to end on `Mf::Pow => res_arg.clone(),`

JCapucho: Oh yeah sorry confused with the match body for `Dot`

JCapucho (approved):
Thanks, everything looks great now, there seems to be problems with clippy not related to your changes, I'll fix them and then we can merge your PR (you might need to rebase).

evahop:
> Thanks, everything looks great now, there seems to be problems with clippy not related to your changes, I'll fix them and then we can merge your PR (you might need to rebase).
Cool, I’ll be on the lookout 👀

JCapucho:
Thanks for you work, I've gone ahead and rebased it so now it's ready to be merged

PR #2026, opened 2022-08-12
Fix clippy lints for 1.63

JCapucho:
1.63 Introduced some new clippy lints, this PR fixes them and disables the `derive_partial_eq_without_eq` lint because the structures it flags don't need to implement `Eq` (they are for internal use only) and deriving it might affect compilation times.

### DIFF
--- a/src/front/glsl/builtins.rs
+++ b/src/front/glsl/builtins.rs
@@ -537,7 +537,7 @@ fn inject_standard_builtins(
         | "dFdyFine" | "dFdyCoarse" | "fwidth" | "fwidthFine" | "fwidthCoarse" => {
             // bits layout
             // bit 0 trough 1 - dims
-            for bits in 0..(0b100) {
+            for bits in 0..0b100 {
                 let size = match bits {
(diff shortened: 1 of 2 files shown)


src/front/glsl/builtins.rs
```diff
@@ -537,7 +537,7 @@ fn inject_standard_builtins(
         | "dFdyFine" | "dFdyCoarse" | "fwidth" | "fwidthFine" | "fwidthCoarse" => {
             // bits layout
             // bit 0 trough 1 - dims
-            for bits in 0..(0b100) {
+            for bits in 0..0b100 {
```
cwfitzgerald, line 540: What's up with all these changes?

cwfitzgerald (approved):
One comment from code that looks like it was accidentally included - otherwise LGTM

src/front/glsl/builtins.rs
JCapucho: Clippy also started flagging these as unused parentheses so I just applied the suggestion

PR #2028, opened 2022-08-15
wgsl-in, glsl-in: Short circuit || and &&

### LINKED ISSUE #1931: [spv-out] Naga does not short-circuit `&&` and `||`
The WGSL specification says that `a && b` only evaluates `b` if `a` is true, but Naga's SPIR-V back end will translate that to a `OpLogicalAnd`, which does not short-circuit.

Naga IR's `BinaryOperator::LogicalAnd` and `BinaryOperator::LogicalOr` are translated to `&&` and `||` in the HLSL, GLSL, and MSL back ends, which I think means that those IR operators are indeed short-circuiting, although it's not documented.

(I thought we had an issue filed for this, but I can't find it.)

### LINKED ISSUE #4394: Naga front ends do not properly short-circuit `&&` and `||`
NOTE: description edited 2024-4-16, so much of the following discussion may look irrelevant

The WGSL specification says that `a && b` only evaluates `b` if `a` is true, but Naga generally will always evaluate both.

The following input:

```wgsl
fn h() -> bool {
  return f() || g();
}
```

produces the following output WGSL:

```wgsl
fn h() -> bool {
    let _e0 = f();
    let _e1 = g();
    return (_e0 || _e1);
}
```

The call to `g` is hoisted out to a statement and made unconditional, which is incorrect.

This means that even though most backends turn Naga IR's `BinaryOperator::LogicalAnd` and `BinaryOperator::LogicalOr` into short-circuiting operations in the target language, it doesn't help because the front end has already hoisted the right-hand side, which ought to be conditional, out into its own unconditional statement.

adeline-sparks:
Fixes gfx-rs/wgpu#4394.


JCapucho (changes requested):
Disclaimer: only read the glsl code yet, so this might or might not apply to the wgsl frontend

There's an issue with your implementation, you're always branching on the left expression, this is true for `LogicalAnd` but not for `LogicalOr`, the latter must only execute the right hand side if it evaluates to `false`.

Also you seem to be creating the following pseudocode:
```
if (left) {
  local = right
} else {
  local = constant
}
```
you could instead do:
```
local = left
if (left) {
  local = right
}
```

adeline-sparks:
Looking at this again, I think it is translating things correctly. The tests show that for the `&&` case, the RHS is evaluated in the accept branch of the if statement, while for the `||` it is evaluated in the reject branch. 

I'll have a go at generating multiple assignments instead of single assignment. Would it be better to create a logical not and always leave the reject branch empty so that the generated code doesn't use an `else` in the `||` case?

teoxoy:
> Looking at this again, I think it is translating things correctly. The tests show that for the `&&` case, the RHS is evaluated in the accept branch of the if statement, while for the `||` it is evaluated in the reject branch.

Tests seem correct indeed, but quite verbose (see below).

> Would it be better to create a logical not and always leave the reject branch empty so that the generated code doesn't use an `else` in the `||` case?

I think so (as per @JCapucho's suggestion) because the extra assignments are redundant, we already know the var is true/false.

cwfitzgerald:

PR #2029, opened 2022-08-16
glsl-in: Fix missing stores for local declarations

### LINKED ISSUE #1935: [glsl-in][wgsl-out] Nested loop miscompilation
The following glsl code with a simple nested loop:

<details>
<summary> main.frag </summary>

```glsl
void main() {
    float total = 0.0;

    for (int y = -1 ; y <= 1 ; y++) {
        for (int x = -1; x <= 1; x++) {
            total += x + y;
        }
    }
}
```

</details>

when translated to wgsl gives

<details>
<summary> main.wgsl </summary>

```glsl
fn main_1() {
    var total: f32 = 0.0;
    var y: i32 = -1;
    var x: i32 = -1;

    loop {
        let _e5 = y;
        if (!((_e5 <= 1))) {
            break;
        }
        {
            loop {
                let _e15 = x;
                if (!((_e15 <= 1))) {
                    break;
                }
                {
                    let _e22 = total;
                    let _e23 = x;
                    let _e24 = y;
                    total = (_e22 + f32((_e23 + _e24)));
                }
                continuing {
                    let _e19 = x;
                    x = (_e19 + 1);
                }
            }
        }
        continuing {
            let _e9 = y;
            y = (_e9 + 1);
        }
    }
    return;
}

[[stage(fragment)]]
fn main() {
    main_1();
    return;
}
```

</details>

Which is obviously wrong as `x` is not declared within the scope of the first nested loop, thus reusing it and summing total to -3 instead of 0.

I got this bug while implementing PCF, nested loops do exist in shaders! :D

Note that trying to declare the x out of the for like so doesn't work:

<details>
<summary> main.frag </summary>

```glsl
void main() {
    float total = 0.0;

    for (int y = -1 ; y <= 1 ; y++) {
        int x = -1;
        for (; x <= 1; x++) {
            total += x + y;
        }
    }
}
```

</details>

### Workaround

Declare main.frag like so:

<details>
<summary> main.frag </summary>

```glsl
void main() {
    float total = 0.0;
    int x;
    for (int y = -1 ; y <= 1 ; y++) {
        x = -1;
        for (; x <= 1; x++) {
            total += x + y;
        }
    }
}
```

</details>

Tested with naga-cli 8.0. I've searched through naga's issue but didn't find anything like it.

adeline-sparks:
Previously, if a local variable was declared with a constant value, we
would elide the store and instead give the variable an initial value (as
if it was a global variable). This caused variables to not be
re-initialized each time through a loop.

Fixes #1935, but the issue seems broad and might fix others.


src/lib.rs
```diff
@@ -188,7 +188,10 @@ tree.
     clippy::new_without_default,
     clippy::unneeded_field_pattern,
     clippy::match_like_matches_macro,
-    clippy::if_same_then_else
+    clippy::if_same_then_else,
+    clippy::derive_partial_eq_without_eq,
+    clippy::needless_parens_on_range_literals,
+    clippy::needless_borrow
```
JCapucho, line 194: Remove these and rebase on master

src/front/glsl/parser/declarations.rs
```diff
@@ -198,7 +198,9 @@ impl<'source> ParsingContext<'source> {
                 let is_const = ctx.qualifiers.storage.0 == StorageQualifier::Const;
 
                 match parser.solve_constant(ctx.ctx, root, meta) {
-                    Ok(res) => Some(res),
+                    // Note that even if we solve the constant, we only use the solution for global constants.
+                    // Locals have to be re-assigned every time they are encountered.
+                    Ok(res) if ctx.external => Some(res),
```
JCapucho, line 203: This check should be moved before the constant is solved, no point in wasting time processing it if it isn't going to be used

JCapucho (changes requested):
There are some problems with how the problem is handled but the overall approach is good, you should also change `add_local_var` to make the `init` field always `None`

```diff
# src/front/glsl/variables.rs:646
-                init: decl.init,
+                init: None,
```

src/front/glsl/parser/declarations.rs
```diff
@@ -190,22 +190,14 @@ impl<'source> ParsingContext<'source> {
...
-                match parser.solve_constant(ctx.ctx, root, meta) {
-                    Ok(res) => Some(res),
-                    // If the declaration is external (global scope) and is constant qualified
-                    // then the initializer must be a constant expression
-                    Err(err) if ctx.external && is_const => return Err(err),
-                    _ => None,
+            // If the variable is global and const qualified, solve the initializer for a constant
+            // and use that as the variable's initial value.
+            let is_const = ctx.qualifiers.storage.0 == StorageQualifier::Const;
+            let maybe_constant = match (init, ctx.external, is_const) {
```
JCapucho, line 196: This will prevent non constant globals from being evaluated, try this instead:
```rust
let maybe_constant = if ctx.external {
  match parser.solve_constant(ctx.ctx, root, meta) {
      Ok(res) => Some(res),
      // If the declaration is external (global scope) and is constant qualified
      // then the initializer must be a constant expression
      Err(err) is_const => return Err(err),
      _ => None,
  }
} else {
  None
};
```

adeline-sparks: Ah, thanks! There's a lot of cases here and I still didn't have it clear in my head.

adeline-sparks: Actually, if we have a non-constant qualified global with an initializer that can't be solved, should that also be an error?

JCapucho: No glsl allows globals to refer to other globals which can't be constified

adeline-sparks: Fascinating, ok! I will go with this. Thanks again!

JCapucho (approved):
Thanks for your contribution

PR #2035, opened 2022-08-31
spv-out: Decorate integer builtins as Flat in the spirv writer

### LINKED ISSUE #2032: SPIR-V output fails VUID-StandaloneSpirv-Flat-06202, VUID-StandaloneSpirv-Flat-04744
Using naga e7ddd356 to compile `wgpu` test and example shaders to SPIR-V produces code that fails the standalone validation rules `VUID-StandaloneSpirv-Flat-06202` and `VUID-StandaloneSpirv-Flat-04744`.

For example:
```
$ cd wgpu/wgpu/tests/shader_primitive_index
$ naga primitive_index.wgsl primitive_index.spv
$ spirv-val --target-env vulkan1.0 primitive_index.spv 
error: line 29: [VUID-StandaloneSpirv-Flat-04744] Fragment OpEntryPoint operand 27 with Input interfaces with integer or float type must have a Flat decoration for Entry Point id 31.
  %gl_PrimitiveID = OpVariable %_ptr_Input_uint Input

$ cd ../../examples/water/
$ naga water.wgsl water.spv
$ spirv-val --target-env vulkan1.0 water.spv 
error: line 139: [VUID-StandaloneSpirv-Flat-06202] OpEntryPoint interfaces variable must not be vertex execution model with an input storage class for Entry Point id 479.
  %465 = OpVariable %_ptr_Input_v2int Input

$ git rev-parse --short HEAD
2cd08a1c
```

nical:
Fixes #2032

Take what I write in this PR with a whole handful off salt, I'm new enough to naga and spirv that there's a decent chance I am jumping to conclusions.

See the investigation in #2032. My uneducated guess is that spirv wants the integer input builtins such as `primitive_index` to have the flat decoration. So I added that in fairly naive way.

This PR fixes the `primitive_index.wgsl` test case and doesn't appear to cause regressions as far as `cargo test --all` can tell.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1225,6 +1225,26 @@ impl Writer {
                 };
 
                 self.decorate(id, Decoration::BuiltIn, &[built_in as u32]);
+
+                use crate::ScalarKind as Sk;
+
+                // Per the Vulkan spec, `VUID-StandaloneSpirv-Flat-04744`:
+                //
+                // > Any variable with integer or double-precision floating-
+                // > point type and with Input storage class in a fragment
+                // > shader, must be decorated Flat
+                let is_flat = match ir_module.types[ty].inner {
+                    crate::TypeInner::Scalar { kind, .. }
+                    | crate::TypeInner::Vector { kind, .. } => match kind {
+                        Sk::Uint | Sk::Sint | Sk::Bool => true,
+                        Sk::Float => false,
+                    },
+                    _ => false,
+                };
+
+                if is_flat {
+                    self.decorate(id, Decoration::Flat, &[]);
+                }
             }
         }
 
--- a/tests/out/spv/access.spvasm
+++ b/tests/out/spv/access.spvasm
@@ -98,6 +98,7 @@ OpDecorate %78 Binding 3
 OpDecorate %79 Block
 OpMemberDecorate %79 0 Offset 0
 OpDecorate %228 BuiltIn VertexIndex
+OpDecorate %228 Flat
 OpDecorate %231 BuiltIn Position
 OpDecorate %273 Location 0
 %2 = OpTypeVoid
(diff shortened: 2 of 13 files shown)


nical:
It does not fix the `water.wgsl` test case, though.

src/back/spv/writer.rs
```diff
@@ -1172,6 +1172,11 @@ impl Writer {
             }
             crate::Binding::BuiltIn(built_in) => {
                 use crate::BuiltIn as Bi;
+                let is_flat = match built_in {
+                    Bi::Position { .. } => false,
+                    _ => true,
+                };
```
jimblandy, line 1178: The actual rule from the Vulkan spec is:

> - VUID-StandaloneSpirv-Flat-04744
>    Any variable with integer or double-precision floating-point type and with Input storage class in a fragment shader, must be decorated Flat

It's true that `spirv-val` doesn't complain about (say):

```
@fragment
fn fs_main(@location(0) index: u32) -> @location(0) vec4<f32> {
    return vec4<f32>(0.0, 0.0, 0.0, 0.0);
}
```

but since the background story is that these validations are just being implemented piecemeal as somebody's interest leads them, I think we should assume the above might well stop passing validation at some point in the future.

Anyway, that is too many words to say: Let's implement the rule as expressed in the Vulkan spec and attach `Decoration::Flat` to everything whose type is `Sint`, `Uint`, or 64-bit `Float`.

jimblandy (changes requested):

src/back/spv/writer.rs
```diff
@@ -1225,6 +1225,23 @@ impl Writer {
                 };
 
                 self.decorate(id, Decoration::BuiltIn, &[built_in as u32]);
+
+                use crate::ScalarKind as Sk;
+                use crate::TypeInner as Ty;
+                // Spirv spec: Any variable with integer or double-precision floating-point
```
jimblandy, line 1231: Do cite the specific VUID here - that's meant to be a stable identifier for this requirement. Also, it's not part of the SPIR-V spec per se; rather, it's a requirement that Vulkan imposes on the SPIR-V it accepts. (The SPIR-V spec refers to this as the "client API". SPIR-V is also used as the target for OpenCL, which has to run arbitrary C++, so its client API has much looser requirements.)

jimblandy (approved):
Looks fantastic. Two small changes, but then this can land.

jimblandy:
(I made the minor changes I requested.)

PR #2042, opened 2022-09-02
wgsl-in: Rename `Scope` to `Rule`, since we now have lexical scope.

jimblandy:
Rename related functions and variables appropriately.

Also, make `Rule` private to the WGSL front end, and remove the unused `ImportDecl` variant.


JCapucho (approved):
`Rule` doesn't sound right to me, but it makes sense when considered as grammar rule and I don't have any better suggestions, so fine by me.

PR #2043, opened 2022-09-02
[spv-out] Don't emit no-op `OpBitCast` instructions.

jimblandy:
As reported in #2015, some drivers trip over these.


JCapucho (approved):
LGTM

PR #2044, opened 2022-09-03
glsl-in: Migrate to `SymbolTable`

JCapucho:
Moves the glsl frontend from it's custom solution for lexical scopes to the new `SymbolTable`.

Adds an `add_root` method to `SymbolTable` that's needed for glsl builtins and implement `Debug`.

### DIFF
--- a/src/front/mod.rs
+++ b/src/front/mod.rs
@@ -254,7 +254,7 @@ where
     /// the current scope to the root scope, returning `Some` when a variable is
     /// found or `None` if there doesn't exist a variable with `name` in any
     /// scope.
-    pub fn lookup<Q: ?Sized>(&mut self, name: &Q) -> Option<&Var>
+    pub fn lookup<Q: ?Sized>(&self, name: &Q) -> Option<&Var>
     where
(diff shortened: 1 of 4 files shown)


src/front/mod.rs
```diff
@@ -277,6 +277,15 @@ where
     pub fn add(&mut self, name: Name, var: Var) -> Option<Var> {
         self.scopes[self.cursor - 1].insert(name, var)
     }
+
+    /// Adds a new variable to the root scope.
+    ///
+    /// Returns the previous variable with the same name in the root scope if it
+    /// exists, so that the frontend might handle it in case variable shadowing
+    /// is disallowed.
```
jimblandy, line 285: It might be nice to briefly say something like:

> This is used for certain kinds of declarations in GLSL, like yadda yadda, which introduce definitions into the global scope, not the innermost scope enclosing their declaration.

or something like that.

jimblandy (approved):
Looks good to me.

jimblandy (approved):
Great!

PR #2045, opened 2022-09-05
Remove the glsl-validate feature

JCapucho:
When it was introduced it was supposed to allow for fast compiles by
skipping glsl specific validation, but as it turns the subset of glsl that's
compilable code is already pretty close to the subset of valid glsl code.

So the current code gated behind glsl-validate amounts to a single branch that
isn't even performance sensitive, and most of the validation is not specific to
glsl and is made by naga's validator which can be turned off, so the original
goal of fast compile times by disabling validation can still be accomplished.


jimblandy (approved):
Looks great!

PR #2050, opened 2022-09-07
valid: Check regular functions don't have bindings

### LINKED ISSUE #2037: Reject `@location` in non-entry-point function argments
Naga currently accepts:

```rust
// Not an entry point, but takes an argument with a location.
fn foo(@location(1) bar: vec3<u32>) -> vec3<u32> {
    return bar;
}

@vertex
fn vs_main(
    @location(0) position: vec2<i32>,
    var v = foo();

    // ...
}
```

However https://gpuweb.github.io/gpuweb/wgsl/#attribute-location says: "[Must](https://gpuweb.github.io/gpuweb/wgsl/#shader-creation-error) only be applied to an entry point function parameter, entry point return type, or member of a [structure](https://gpuweb.github.io/gpuweb/wgsl/#structure) type."

JCapucho:
Checks that all arguments and return types have no bindings in regular functions.

Closes #2037

### DIFF
--- a/src/valid/function.rs
+++ b/src/valid/function.rs
@@ -143,6 +143,10 @@ pub enum FunctionError {
         Handle<crate::Expression>,
         UniformityDisruptor,
     ),
+    #[error("Functions that are not entry points cannot have `@location` or `@builtin` attributes on their arguments: \"{name}\" has attributes")]
+    PipelineInputRegularFunction { name: String },
+    #[error("Functions that are not entry points cannot have `@location` or `@builtin` attributes on their return value types")]
+    PipelineOutputRegularFunction,
 }
 
 bitflags::bitflags! {
@@ -859,6 +863,7 @@ impl super::Validator {
         fun: &crate::Function,
         module: &crate::Module,
         mod_info: &ModuleInfo,
+        entry_point: bool,
     ) -> Result<FunctionInfo, WithSpan<FunctionError>> {
         #[cfg_attr(not(feature = "validate"), allow(unused_mut))]
         let mut info = mod_info.process_function(fun, module, self.flags, self.capabilities)?;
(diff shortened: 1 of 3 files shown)


src/valid/function.rs
```diff
@@ -143,6 +143,10 @@ pub enum FunctionError {
         Handle<crate::Expression>,
         UniformityDisruptor,
     ),
+    #[error("Argument '{name}' at index {index} defines a pipeline input for a regular function")]
```
jimblandy, line 146: This is a statement of fact, but it doesn't explain what's wrong. Suggest phrasing this as:

> Functions that are not entry points cannot have `@location` or `@builtin` attributes on their arguments

src/valid/function.rs
```diff
@@ -143,6 +143,10 @@ pub enum FunctionError {
...
+    PipelineInputRegularFunction { index: usize, name: String },
+    #[error("Return type defines a pipeline output for a regular function")]
```
jimblandy, line 148: Similarly:

> Functions that are not entry points cannot have `@location` or `@builtin` attributes on their return value types

jimblandy:
Interesting. I thought this program would be forbidden by WGSL, but the spec doesn't seem to say it is:

```
struct Foo {
   @location(0) x: f32,
};

fn f(arg: Foo) -> f32 {
   return arg.x;
}
```

Tint accepts the program.

jimblandy (changes requested):
Thank you! This looks good, but the error messages need to be clarified:

JCapucho:
> Interesting. I thought this program would be forbidden by WGSL, but the spec doesn't seem to say it is:
> 
> ```
> struct Foo {
>    @location(0) x: f32,
> };
> 
> fn f(arg: Foo) -> f32 {
>    return arg.x;
> }
> ```
> 
> Tint accepts the program.

I guess this is to allow for programs like:
```
struct Foo {
   @location(0) x: f32,
};

@vertex
fn vs_main(foo: Foo) -> @location(0) vec4<f32> {
  let a = f(foo);
  return vec4<f32>(a);
}

fn f(arg: Foo) -> f32 {
    return arg.x;
}
```

src/valid/function.rs
JCapucho: I was having issues with the phrasing, thanks for the suggestion

jimblandy (approved):
Fantastic - thank you very much!

Issue #2053, opened 2022-09-10
Spans on expressions are bad


jimblandy:
Given the input:

```
fn f() {
   var v = vec2(0);
   v.xy = vec2(1);
}
```

Naga produces the error message:

```
error: the left-hand side of an assignment must be a reference
  ┌─ assign-to-swizzle.wgsl:2:20
  │  
2 │      var v = vec2(0);
  │ ╭───────────────────^
3 │ │    v.xy = vec2(1);
  │ ╰───────^ expression is not a reference

Could not parse WGSL
```

That span is terrible. It should simply refer to the `v.xy`, or ideally just the `.xy`.

The quality of the error message itself is addressed in #2052.

PR #2054, opened 2022-09-10
wgsl-in: Fix bad span in assignment lhs error

JCapucho:
The wgsl frontend was emiting errors for lhs expressions on assignments that weren't references using a span that didn't skip blankspaces causing the span to look weird (like starting at the end of the previous line)

This is fixed by consuming the blankspace before constructing the span

Related to #2053

### DIFF
--- a/src/front/wgsl/lexer.rs
+++ b/src/front/wgsl/lexer.rs
@@ -237,6 +237,22 @@ impl<'a> Lexer<'a> {
         }
     }
 
+    /// Consumes [`Trivia`] tokens until another token is encountered, returns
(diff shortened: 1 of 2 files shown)


jimblandy (approved):
This is fine for now. I'll put my concerns in a separate comment.

jimblandy:
I hope we can find a more general solution to this, because these spans are misplaced everywhere. Basically, everyplace that calls `Lexer::current_byte_offset` is either trying to find the start or the end of something, and whenever it's looking for the start of something it's probably wrong.

PR #2055, opened 2022-09-13
wgsl-in: Fix inclusion of trivia in spans

### LINKED ISSUE #2053: Spans on expressions are bad

SparkyPotato:
This fixes #2053 for all cases, ensuring that no spans have leading or trailing trivia in them.

The lexer API is changed to have `.start_byte_offset()` and `.end_byte_offset()`, to be used when trying to get the byte offset of the start and end of a node, respectively.

### DIFF
--- a/src/front/wgsl/lexer.rs
+++ b/src/front/wgsl/lexer.rs
@@ -196,6 +196,22 @@ impl<'a> Lexer<'a> {
         Ok((res, start..end))
     }
 
+    pub(super) fn start_byte_offset(&mut self) -> usize {
+        loop {
+            // Eat all trivia becuase `next` doesn't eat trailing trivia.
+            let (token, rest) = consume_token(self.input, false);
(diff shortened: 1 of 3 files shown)


jimblandy (approved):
This looks fantastic. Thank you very much!

PR #2057, opened 2022-09-15
[wgsl-in] Break up long string, reformat rest of file.

jimblandy:
Whitespace and formatting changes only.

It turns out that if `cargo fmt` comes across a single line it can't make fit within the margins, then it sort of gives up on the surrounding construct. So breaking up the error message for `Error::UnknownScalarType` in `Error::as_parse_error` and then re-running `cargo fmt` cleans up lots of other stuff in the file.

cargo fmt issue:
https://github.com/rust-lang/rustfmt/issues/3863


jimblandy:
merging without review, since it's only whitespace/formatting changes

PR #2059, opened 2022-09-17
[wgsl] Fix line endings on wgsl reserved words list.

jimblandy:


jimblandy:
Enabling auto-merge without review, since this is a whitespace-only change.

PR #2061, opened 2022-09-17
Fix for incorrect GLSL validation on witeonly image uniforms

spencerkohan:
Related to this issue: https://github.com/gfx-rs/wgpu/issues/4530

I noticed the validation on layout qualifiers for writeonly image uniforms was not working according to the GLSL spec.

I saw there was already a TODO in the code, and the fix seemed straightforward so I made a PR for this.

### DIFF
--- a/src/front/glsl/parser_tests.rs
+++ b/src/front/glsl/parser_tests.rs
@@ -320,6 +320,20 @@ fn declarations() {
         "#,
         )
         .unwrap();
+
+    parser
+        .parse(
+            &Options::from(ShaderStage::Compute),
+            r#"
+        #version 450
+        precision highp float;
+        
+        layout(binding = 0) uniform writeonly image2D image;
+
+        void main() {}
+        "#,
+        )
+        .unwrap();
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


JCapucho:
Thanks for trying to fix the issue, but unfortunately this doesn't solve the underlying problem it just side steps the error gating it.

The problem is that naga doesn't have a way to express a storage image without a format, so we initialize it by default `r8uint` and expect the user to change the format.

https://github.com/gfx-rs/naga/blob/7e3750a7f2eca0ab357dfa72d90bdc60b77c3202/src/front/glsl/types.rs#L170-L173

So your solution is just making all `writeonly` images without storage format default to `r8uint`.

Also if try this on the original issue reproduction case, it won't work, because the storage format kind and the data vector kind aren't equal.

spencerkohan:
Ok I see the problem.  Would this be somehow solvable by making `format` optional on `ImageClass::Storage`, or is there a hard requirement to have the format later on?

I am wondering how spriv-cross solves this problem.

JCapucho:
That would be a possibility, but the real problem is figuring out how to handle it in the backends.

So I recommend you first research how this would map to the different backends and then we can discuss how this would map to the IR.

JCapucho (changes requested):
Forgot to mark it as requesting changes

cwfitzgerald:

PR #2062, opened 2022-09-17
Explicitely enable std for indexmap

maxammann:
I noticed that depending on the environment the std check by the autocfg crate may fail. This enables std like recommended here: https://github.com/bluss/indexmap/issues/184


jimblandy (approved):
Okay - I see that indexmap on master actually makes `std` default, but the 1.8.1 release that Naga is using right now still uses autocfg to check the `CARGO_FEATURE_STD` environment variable. So once Naga is able to switch to indexmap 2.0.0, we can remove this.

Thanks!

Issue #2031, opened 2022-08-21
[wgsl-in] Remove fallthrough statement


Gordon-F:
Spec PR: https://github.com/gpuweb/gpuweb/pull/3315

teoxoy:
We also need to allow `default` to be used with other case selectors (which was changed by that same gpuweb PR).

Issue #2020, opened 2022-08-02
is "- 1" can be considered as const Expressions? 


caxieyou:
so i was testing the **textureSample** function, https://www.w3.org/TR/WGSL/#texturesample

I notice that the 

**fn textureSample(t: texture_2d<f32>,
                 s: sampler,
                 coords: vec2<f32>,
                 offset: vec2<i32>) -> vec4<f32>** usage,  

the last offset: vec2<i32> is consider to be the **const expression**, according to here https://www.w3.org/TR/WGSL/#const-expr

so if i want to use like textureSample(_, _, _, off) is gonna be wrong (off is a vec2<i32> parameter), make sense.

but if i use like textureSample(_, _, _, vec2<i32>(- 1, - 1))  (there is a space between '-' and '1') it's still wrong, I am not sure why the space can cause an issue.

The reason why this "space" is a problem to me is:  using Rust, most of the time you need to use macro.

the token stream will automatically separate "-1" to "- 1", same as “()” to “( )”, “!a” to "! a" 

so I am wondering if anyone wants to combine the wgls with rust macro, this textureSample's offset will be a problem?

any ideas?

thanks!!!!

mikialex:
The current implementation uses the core::num::i32::from_str_radix to parse an int in the tokenization stage, which doesn't allow any space between the preceding '-' and number literals. I think for const int literals, this limitation is ok. The real question is wgsl seems not to support the const operator (like this minus unary case) as a const expression at all.

cwfitzgerald:
Note this is expected, we haven't implemented any of the const stuff yet

caxieyou:
> Note this is expected, we haven't implemented any of the const stuff yet

so this could be somthing which will implement later?

cwfitzgerald:
Yeah, we'll need to eventually to be conformant.

Issue #1996, opened 2022-06-29
[wgsl-in] Confusing error: "expected unsigned/signed integer literal, found '1'"


kpreid:
I'm not familiar with the WGSL specification and couldn't figure it out quickly, so I'm not sure if this is a validation error or _merely_ a poor error message. Given that it used to be accepted, I'm guessing the latter.

The following passes validation with naga 0.8.5 (current crates.io release) but fails on git (as of commit 27d38aae33fdbfa72197847038cb470720594cb1)

```
fn switcher(input: u32) {
    switch (input) {
        case 1: {}
        default: {}
    }
}
```

```
error: expected unsigned/signed integer literal, found '1'
  ┌─ test.wgsl:3:14
  │
3 │         case 1: {}
  │              ^ expected unsigned/signed integer literal
```

The error goes away if I change `1` to `1u`, making it explicitly unsigned. If this program should pass validation, then the message is quite confusing: it doesn't specify that I need an unsigned integer, and it doesn't make it clear that this is a type mismatch and not a syntax error.

teoxoy:
This seems to be a side-effect of #1863.

The tokenizer now differentiates between concrete types (those with suffixes) and abstract ones.

The plan is to infer the type of abstract numbers as per spec.

teoxoy:
Track #2127 for automatic type conversions.

Issue #1957, opened 2022-05-29
[wgsl-in] Rejection of composite consts without explicit nested types


hanawatson:
Even though the types are specified in the declaration of the const, naga rejects consts that are composite if they omit type information from their constructors.

WGSL:
```
let const0: vec3<f32> = vec3(0.0, 0.0, 0.0);

@stage(compute) @workgroup_size(1)
fn main() {
    var var0 = 0;
}
```

Error:
```
        error: expected '<', found '('
          ┌─ test.wgsl:1:29
          │
        1 │ let const0: vec3<f32> = vec3(0.0, 0.0, 0.0);
          │                             ^ expected '<'
        
        Could not parse WGSL
```

The above works fine if given e.g. `vec3<f32>(0.0, 0.0, 0.0)`. I believe it should be acceptable to omit the type in the constructor.

PR #2065, opened 2022-09-18
[wgsl-in] Transition to a pass-based architecture

SparkyPotato:
The current WGSL frontend does everything in a single pass, with parsing depending on input from the type-checker, etc. This creates problems like #1745, const eval, type inference, and also has generally poor diagnostics due to the lack of context.

This PR converts it into a more traditional pass-based architecture with four passes:
1. Parsing.
2. Name resolution.
3. Declaration cycle detection, and topological ordering.
4. Lowering to Naga IR, while doing type inference and const evaluation.

### In its current state, it fixes:
1. #1745 
2. #1044 according to the current WGSL spec.
3. #1754 (I believe this has already been fixed?)
4. #1741: It makes all of Naga's native WGPU features WGSL `enable` directives, so you have to explicitly opt into using them while writing shaders.
5. #2031 
6. #989 by using only the dynamic `Access` expression for array indexing.
7. #1883 
8. #2071 
9. #2072 
10. #2020 
11. #1906 
12. #1775 
13. #791

### Planned features
1. #733 
2. #471 
3. #1583 
4. #1829 
5. #1777 
6. #1426 
7. #1957 
8. #1996 
9. #2068 
10. #1956 
11. #1931 
12. #1884 
13. #1154 

### Breaking language changes
There are some WGSL programs that were accepted by Naga before, but will no longer be accepted:
1. ~~All of Naga's extra native features will be gated behind WGSL features, and must be `enable`d manually, so the shader author explicitly opts out of supporting browser WebGPU.
For example: `binding_array` can be enabled by putting a `enable binding_array;` at the top of the source.~~
This is currently disabled.
2. `let`s are no longer allowed in global declaration position, `const` must be used instead.

### Breaking API changes
The frontend returns a `Result<Module, Vec<WgslError>>`, where `WgslError` is a type containing a diagnostic that can be trivially converted to a codespan `Diagnostic`.

### Other changes
The frontend's diagnostics have been improved, and now multiple of them can be emitted at once.
A little example:
Given:
```wgsl
const a: i32 = b;
const b = c;
const c = S();

struct S {
    field: array<S, a>,
}
```
Since cyclic declarations are not allowed anywhere, Naga errors with:
```
error: cyclic declarations are not allowed
  ┌─ wgsl:1:7
  │
1 │ const a: i32 = b;
  │       ^        ^ uses
  │       │
  │       this declaration
2 │ const b = c;
  │       ^   ^ uses
3 │ const c = S();
  │       ^   ^ uses
4 │
5 │ struct S {
  │        ^
6 │     field: array<S, a>,
  │                     ^ ending the cycle

error: recursive declarations are not allowed
  ┌─ wgsl:5:8
  │
5 │ struct S {
  │        ^
6 │     field: array<S, a>,
  │                  ^ uses itself here
```

### Performance regression
I have used a parser combinator library, `chumsky` for parsing. This causes a severe regression in performance (the `front/wgsl` benchmark takes 60 ms instead of 4ms).

However, the rest of the frontend has virtually no impact on performance, so regaining Naga's excellent performance should be possible by hand writing the parser.


src/front/wgsl/resolve/mod.rs
```diff
@@ -0,0 +1,1852 @@
...
+pub mod features;
+pub mod inbuilt;
+pub mod inbuilt_functions;
+mod index;
+pub mod ir;
+
+pub struct ResolveContext {
+    dependency_context: DependencyContext,
+    index: Index,
+    reserved_matcher: AhoCorasick,
```
jimblandy, line 38: It seems like an `FxHashSet<Text>` would be better here than an `AhoCorasick` matcher. You've already hashed the identifier once; it doesn't make sense to make another pass over it.

src/front/wgsl/parse/ast.rs
```diff
@@ -0,0 +1,284 @@
+use crate::{front::wgsl::text::Text, Span};
+
+#[derive(Clone, Debug, Default)]
+pub struct TranslationUnit {
```
jimblandy, line 4: This AST is very close to the grammar, but I don't think it's worth building all this intermediate structure. We should just touch the data once, lower it as far as we can immediately, and be done with it. The architecture here leads to odd bits like the check for generic parameters on the identifier in `Resolver::expr`; it seems like the code could be structured so that situations like that (and others) simply never arise.

Naga tries hard to gather everything into arrays (the `Arena` types, for example) and avoid heap allocation because such code runs very well on modern processors. As I say, I haven't profiled, but my guess is that you won't be able to get within 2x or 3x of the current Naga front end's run time without addressing this.

src/front/wgsl/parse/ast.rs
```diff
@@ -0,0 +1,284 @@
...
+#[derive(Clone, Debug)]
+pub struct Expr {
+    pub kind: ExprKind,
+    pub span: Span,
+}
+
+#[derive(Clone, Debug)]
+pub enum ExprKind {
+    Underscore,
+    VarDecl(Box<VarDecl>),
```
jimblandy, line 196: It is very surprising to see variable declarations as an expression type. Overall, I had a hard time tracking how identifiers were resolved. I think this could be simplified a lot, while still implementing WGSL's semantics.

src/front/wgsl/resolve/mod.rs
```diff
@@ -0,0 +1,1852 @@
...
+    access_mode: Matcher<AccessMode>,
+    address_space: Matcher<AddressSpace>,
+    builtin: Matcher<Builtin>,
+    interpolation_sample: Matcher<InterpolationSample>,
```
jimblandy, line 42: I think I understand why you want a separate `Matcher` for each role identifiers play in WGSL, but I think there are much nicer approaches to this. I outlined one possible idea in #1391.

jimblandy (commented):
This PR isn't suitable for us to merge. I think the best way to
proceed is for you to host it as your own crate, where you can
experiment as you please, and show people how they can use your
front end in their own `wgpu` code using
[`wgpu::ShaderSource::Naga`][ssn]. That way everyone can choose the
tradeoffs that make sense for them.

There are a number of reasons we don't want to take this PR.

First of all, as you noted, this front end is 15x slower than the
current Naga WGSL front end. I used the following command on
`master` and the `multipass` branch:

    cargo bench --bench criterion front/wgsl -- --measurement-time 30

Here's the 'before' and 'after' plot (red = master, blue = multipass):

Screenshot 2022-10-02 at 16-20-21 front_wgsl - Criterion rs

Any fully compliant front end is almost certainly going to be a bit
slower than our current front end, but I'm confident we can do a lot
better than this.

Second, this PR takes on a bunch of new dependencies, which we would
like to avoid. I haven't profiled, but I suspect that some of the
slowdown is due to `chumsky` and `logos`; in particular, `chumsky`
states in its `README.md` that it is much more concerned with the
quality of the diagnostics and error recovery than with speed. That
is a perfectly reasonable tradeoff, but again, I think we can do
better.

The new dependencies also increase Naga compile time by around a
third, for both naga + naga-cli - only builds and builds from
scratch. We've tried hard to keep build times under control, since
they impact developer experience pretty directly.

The new dependencies also create a review burden for embedders that
want to use `wgpu` in security-sensitive code, like Firefox.
Sometimes that's the right burden to choose - it's certainly not the
case that nobody should ever add a new dependency. But in this case,
I don't think the tradeoff is right.

Finally, the architecture just uses a lot of intermediate steps and
intermediate representations where I think a more direct path would
be both faster and more legible. This front end has three phases:
parsing, resolution, and lowering to Naga IR. I'm pretty sure we can
get by with just two, and use a representation for the unresolved
source code that is a minor variant on Naga IR, which will make the
code easier for people familiar with the rest of Naga to follow.

I haven't profiled, but I suspect the intermediate steps are also
responsible for some of the loss of performance. That is, I suspect
that rewriting the parser is not going to get you back as much speed
as you expect.

I've put some other comments at specific places in the code, in case
they're useful to you. But overall, it's really not a matter of
fixing a few problems here and there. When you first brought up the
idea of rewriting the front end in Naga chat, I tried to encourage
you to take a more incremental approach, which would have made it
easier for us to give feedback you could incorporate as you go, and
preserve the crucial characteristics of the original front end, like
performance and a light touch on dependencies. By jumping in with a
full rewrite, those characteristics were lost.

I didn't look into this PR's resolution algorithm in detail, nor the
abstract type handling, but I take your word that they're working;
that's certainly an advantage over Naga's present state. And I
definitely expect your front end to produce better diagnostics than
ours (at least for now), which many users will find valuable. So I
encourage you to continue development of this front end in your own
repository, and perhaps publish it as a separate crate.

If that approach doesn't appeal to you, the naga/wgpu license was
chosen specifically to ensure that you have the right to fork. This
is a feature, not a bug, intended for dealing with situations like
this, where different teams have different priorities.

But we expect to be able to bring Naga into compliance with the spec
over the coming months while preserving most (probably not all) of
its current performance.

[ssn]: https://github.com/gfx-rs/wgpu/blob/26f223964e7bf57b7f207cf09efa82b732c470cf/wgpu/src/lib.rs#L865

PR #2066, opened 2022-09-18
glsl-in: Update initializer list type when parsing

JCapucho:
Previously when parsing an initializer list, the type passed to `parse_initializer` was not updated to reflect the child type, this caused implicit conversions to not work and nested initializer lists to not be allowed.

### DIFF
--- a/src/front/glsl/parser/declarations.rs
+++ b/src/front/glsl/parser/declarations.rs
@@ -19,6 +19,55 @@ use crate::{
 
 use super::{DeclarationContext, ParsingContext, Result};
 
+/// Helper method used to retrieve the child type of `ty` at
+/// index `i`.
+///
+/// # Note
+///
+/// Does not check if the index is valid and returns the same type
+/// when indexing out-of-bounds a struct or indexing a non indexable
+/// type.
+fn element_or_member_type(
+    ty: Handle<Type>,
+    i: usize,
+    types: &mut crate::UniqueArena<Type>,
+) -> Handle<Type> {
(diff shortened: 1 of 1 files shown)


jimblandy (approved):
Looks good to me.

I wonder whether it would be helpful to put the bulk of this in a helper function, `element_or_member_type(i: usize) -> Type` or something.

JCapucho:
@jimblandy thanks for the suggestion, I've also took some time to improve the documentation, could you give it another look.

src/front/glsl/parser/declarations.rs
```diff
@@ -19,6 +19,55 @@ use crate::{
...
 use super::{DeclarationContext, ParsingContext, Result};
 
+/// Helper method used to retrieve the child type of `ty` at
+/// index `i`.
+///
+/// # Note
+///
+/// Does not check if the index is valid and returns the same type
+/// when indexing out-of-bounds a struct or indexing a non indexable
+/// type.
```
jimblandy, line 29: You're counting on validation to complain if there are too many components - correct? (It does, in `naga::valid::compose::validate_compose`.)

This PR is still an improvement over the status quo, but tolerating bad indices here does mean that if someone supplies too many values in the initializer, you're going to be passing meaningless types to `implicit_conversion`, and the error messages will be pretty confusing. That is, you'll get weird GLSL front end errors, instead of the nice "too many components" error from the validator, because you'll never make it that far.

jimblandy (approved):

Issue #2071, opened 2022-09-30
Global item does not support 'const'


haoyunfeix:
Shader:
```javascript
      const workGroupSizeX = 256u;
      const workGroupSizeY = 1u;
      const workGroupSizeZ = 1u;
```
The compiler doesn't like it:
$ naga simple.wgsl
```
error: expected global item ('struct', 'let', 'var', 'type', ';', 'fn') or the end of the file, found 'const'
  鈹屸攢 simple.wgsl:1:7
  鈹?[0m
1 鈹?[0m       const workGroupSizeX = 256u;
  鈹?[0m       ^^^^^ expected global item ('struct', 'let', 'var', 'type', ';', 'fn') or the end of the file
```

SparkyPotato:
Naga currently uses `let` for global constants.
It will be fixed in #2045.

SparkyPotato:
I'm a bit busy right now, but hope to get it done soon.

vicary:
@SparkyPotato no worries, take your time!

Issue #1775, opened 2022-03-15
Referencing a function without a return type yields an unknown identifier error.


OptimisticPeach:
I had code which resembled the following:
```rs
fn foo() {
    return vec2<f32>(0.0, 0.0);
}

fn bar() {
    let qux = foo();
}
```

Note that I forgot a return type on `foo`. (This was much more difficult to spot at a glance in my actual code).

Validating the code with Naga fails with:
```
❯ cargo run example.wgsl
    Finished dev [unoptimized + debuginfo] target(s) in 0.23s
     Running `target\debug\naga.exe example.wgsl`
error: no definition in scope for identifier: 'foo'
  ┌─ example.wgsl:6:15
  │
6 │     let qux = foo();
  │               ^^^ unknown identifier

Could not parse WGSL
error: process didn't exit successfully: `target\debug\naga.exe example.wgsl` (exit code: 1)
```
Whereas I'd expect it to either:

- Error on `foo` containing a `return` statement while it returns nothing, or
- Tell me that `foo` returns nothing and assigning to a variable is foolish.

Issue #1745, opened 2022-02-23
Support out-of-order module scope declarations in WGSL


jimblandy:
As described in [§3.8 Declaration and Scope](https://gpuweb.github.io/gpuweb/wgsl/#declaration-and-scope):

> A declaration is at module scope if the declaration appears outside the text of any other declaration. Module scope declarations are [in scope](https://gpuweb.github.io/gpuweb/wgsl/#in-scope) for the entire program. That is, a declaration at module scope may be referenced by source text that follows or precedes that declaration.

Our front end depends pretty seriously on being able to find definitions when they are referenced.

Issue #1044, opened 2021-06-28
Forbid local variable shadowing in WGSL


zesterer:
This is actually generated by wgpu, but the error seems to be a mistranslation by naga.

```
[2021-06-28T19:36:06Z ERROR wgpu::backend::direct] wgpu error: Validation Error

    Caused by:
        In Device::create_render_pipeline
          note: label = `Render pipeline`
        Internal error in VERTEX shader: Error compiling the shader "\"Compilation failed: 

program_source:72:103: error: cannot initialize an array element of type 'char' with an lvalue of type 'metal::float3' (aka 'float3')
    VertexOut _e57 = VertexOut {_e52 * metal::float4(metal::float3(_e44.pos.xy, _e49), 1.0), _e44.uv, _e44.pos};
                                                                                                      ^~~~~~~~\
program_source:72:103: warning: suggest braces around initialization of subobject\
    VertexOut _e57 = VertexOut {_e52 * metal::float4(metal::float3(_e44.pos.xy, _e49), 1.0), _e44.uv, _e44.pos};
                                                                                                      ^~~~~~~~
                                                                                                      {       }
program_source:99:45: error: cannot initialize an array element of type 'char' with an lvalue of type 'metal::float3' (aka 'float3')
    VertexOut _e44 = VertexOut {_e41, _e42, _e43};
                                            ^~~~
program_source:99:45: warning: suggest braces around initialization of subobject
    VertexOut _e44 = VertexOut {_e41, _e42, _e43};
                                            ^~~~
                                            {   }

thread 'main' panicked at 'Handling wgpu errors as fatal by default', /Users/foo/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.9.0/src/backend/direct.rs:1978:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

(I've made an attempt at formatting the error such that it's a little more intelligible)

The original offending shader (written in wgsl, translated to spir-v also with naga, although possibly with a slightly different version) is as follows:

```rust
[[block]]
struct Globals {
    vp: mat4x4<f32>;
};
[[group(0), binding(0)]] var<uniform> globals: Globals;

[[group(1), binding(0)]] var heightmap_texture: texture_2d<f32>;
[[group(1), binding(1)]] var heightmap_sampler: sampler;
[[group(1), binding(2)]] var noise_texture: texture_2d<f32>;
[[group(1), binding(3)]] var noise_sampler: sampler;
[...]

PR #2075, opened 2022-10-04
[wgsl-in] Implement module-level scoping

### LINKED ISSUE #2072: [wgsl-in] allow functions to be declared literally after entry point
This is discovered in Deno when using tensorflow.js with its WebGPU backend, when tfjs declares a function `fn main` after the entry point `fn _start`. The generated WGSL works in Chrome Canary 108.0.5329.0 (V8 10.8.79) but fails in Deno, error details at https://github.com/tensorflow/tfjs/issues/6842#issuecomment-1262708937.

Experiments in https://github.com/tensorflow/tfjs/issues/6842#issuecomment-1263196772 showed that in naga, invoked functions have to be declared literally before the entry point and the `@compute` directive.

The current WGSL specification (20220922) seems to have [function scoping](https://www.w3.org/TR/2022/WD-WGSL-20220922/#function-declaration-sec), replacing the [lexical scoping](https://github.com/gpuweb/gpuweb/issues/559) back in 2020.

> A function declaration [must](https://www.w3.org/TR/2022/WD-WGSL-20220922/#shader-creation-error) only occur at [module scope](https://www.w3.org/TR/2022/WD-WGSL-20220922/#module-scope). A function name is [in scope](https://www.w3.org/TR/2022/WD-WGSL-20220922/#in-scope) for the entire program.

Given the generated script [shader.wgsl](https://github.com/gfx-rs/naga/files/9682291/shader.wgsl.txt), should the function `main` be available in scope when `_start` is invoked?

SparkyPotato:
Try 2 :)

- fixes https://github.com/gfx-rs/naga/issues/1745
- fixes https://github.com/gfx-rs/naga/issues/1044
- fixes https://github.com/gfx-rs/naga/issues/2076
- fixes https://github.com/gfx-rs/naga/issues/2071
- fixes https://github.com/gfx-rs/naga/issues/2105
- fixes https://github.com/gfx-rs/naga/issues/1775
- fixes https://github.com/gfx-rs/naga/issues/2089
- fixes https://github.com/gfx-rs/naga/issues/1996

This PR separates parsing from lowering, by generating an AST, which desugars as much as possible down to something like Naga IR. 

The AST is then used to resolve identifiers while lowering to Naga IR.


SparkyPotato:
@jimblandy Would be great if you could a look at the AST definition!

Also, you were talking about how 2 passes would be all that was necessary.
However, lowering to Naga IR requires that the dependencies of each module-level declaration are known beforehand, as we need to (reverse) topologically sort the declaration list so that all dependencies are processed before their dependents.

The only way I can think of doing this in two passes is to maintain a set of the raw identifiers referenced by each declaration during parsing, and then using those to sort the declarations, but I'm not quite sure if this will actually be faster than resolving on the same IR, and then sorting.

jimblandy:
Here's how I was thinking it could work (I haven't actually tried this yet):

1) For the first pass, we parse mostly as we do now, except that identifiers in expressions, types, statements, etc. are represented by placeholders that just store the name. Except for the placeholders, we try to produce something as close to the `crate` IR as possible immediately. 

    The placeholders mean that we will have types that look very much like `crate::Expression`, etc. except that certain variants that assume knowledge of definitions (`LocalVariable`, obviously, but also things like `AccessIndex`, since we don't know struct type definitions) hold identifiers instead of handles. Some cases will need to change in more profound ways; for example, we can't tell whether `S(x)` is a `Call` or a `Compose` until we know the definition of `S`. But as much as possible, the unresolved types should resemble the `crate` types, so that people familiar with the rest of Naga can make assumptions.

2) For the resolution pass, starting with the entry points, start building true `crate::{Function, EntryPoint, Type, ...}` values. When we hit a placeholder, check to see if we already have a previously-resolved definition for that name. If so, use it. Otherwise, immediately recurse on resolving that name (which adds the resulting fully-resolved definition to the `Module` under construction), and then resume processing the placeholder, whose definition we now know. When we're done, add our completed definition to the `Module`.

This amounts to a depth-first traversal of the DAG of definitions, where we "mark nodes as visited" by adding resolved definitions to the `Module`. This ensures that definitions appear in the `Module` in dependency order, in the process of replacing placeholders with `Handle`s.

After we've processed the entry points, we may want to just walk the top-level unresolved definitions and finish off anything unvisited. Such definitions could be safely dropped from the module, since they're unreachable from any entry point (and we should probably produce warnings about this), but it's very useful for testing to have everything in the input appear in the `Module`. Maybe this step would be optional.

Naturally, we will need to check for cyclic references at the point of recursion.

As I say, I haven't tried this, so there may be aspects of the problem I've missed - but it seems like something along these lines ought to be possible somehow.

jimblandy:
You should use `Arena`, `UniqueArena`, and `Handle` for the unresolved representation. That is one of the tricks Naga uses for performance: broadly speaking, caches and prefetchers are much happier with arrays than heap allocation, especially if you can phrase some computation as an iteration over the array.

Otherwise, try to make the change as undisruptive and limited as you can while still preserving legibility.

SparkyPotato:
Recursion was something I thought of, but processing each declaration does use quite a large amount of stack, so we might end up stack overflowing quite quick, especially on Windows, where the default stack size is 2 MB.

We can probably use [stacker](https://github.com/rust-lang/stacker), which is what rustc uses.

About the warnings, I do plan to add error recovery to the frontend so we can output more than one error, and will add warnings for unused items there (I did have this implemented in my other frontend, it isn't too big of a change).

jimblandy:
Using arenas instead of trees for expressions should reduce the stack depth, though. And with a little care we could avoid recursing on statements at all - I think only `Call` and `Atomic` need resolution. So the depth would only be determined by the number of references we're in the midst of resolving, and not so much by the interior structure of the functions themselves.

jimblandy:
`stacker` sounds 1) amazing and 2) wildly unsafe and definitely not something we want to ship in a browser. So that's one easy solution out the window. I think stack depth is a legitimate concern here.

SparkyPotato:
I think keeping a set of (textual) dependencies used by each declaration during parsing, and then using that to post-order depth-first sort would be the easiest solution, and it shouldn't really impact performance too much.

The post-order sorting with cycle detection does require recursion, however it should require much less stack space than resolving the function.

I have also implemented parsing and finalized the AST structures, would be great if you could take a look, and tell me if any changes are needed :)

SparkyPotato:
I did have to keep track of locals during the initial parse phase, to fix an issue with code like this:
```rs
fn a() {
    let a = b();
}

fn b() -> i32 {
    let a = 20;
    return a;
}
```
If we just kept a set of textual dependencies, `fn a` being shadowed by `let a` would create a cyclic dependency error.

i509VCB:
I'm curious if the ast generated here is publicly accessible. I do have a case where I need to look at the shader's ast to figure out how to setup bind groups

SparkyPotato:
> I'm curious if the ast generated here is publicly accessible. I do have a case where I need to look at the shader's ast to figure out how to setup bind groups

It's not public at the moment, but I can make it so.

However, can you not just use the `Module` that the front end produces instead? It will probably have more information than the semi-AST IR used here.

SparkyPotato:
@jimblandy This has reached feature parity with the existing frontend, and passes all the functionality-based tests. The diagnostics snapshots don't pass yet, but fixing those shouldn't change too much. You can start reviewing now.

Surprisingly, performance has actually *improved*. The current frontend consistently takes around 3.6 ms for the `front/wgsl` benchmark, while this does it in around 2.9 ms (a 20% decrease).

The only difference in output is that I don't initialize `LocalVariable`s with constants since I'm not too sure if that added complexity is actually worth it. I can add it back if it's something wanted, though.

SparkyPotato:
@teoxoy I've moved as much as I think possible back (parsing + errors).

`construction.rs` hasn't been moved because I have made quite a few changes to it, and it also needs to be in the `lower` module for privacy reasons.

Do tell me if there's anything else I need to do to make reviewing easier!

teoxoy:
Thanks for reducing the diff, but it still looks like quite a lot of changes which will make browsing the history of the files challenging.

Let's move the code in `lower/mod.rs` to `mod.rs` and `lower/construction.rs` to `construction.rs` (with existing code where it previously was).

After this, the changes should be reviewable and after the review we can see how to proceed with new files/folders (and moving code around).

SparkyPotato:
Got it down to +4495 -2811, not sure if anything more is possible.

src/lib.rs
```diff
@@ -573,8 +573,7 @@ pub struct Type {
 }
 
 /// Enum with additional information, depending on the kind of type.
-#[derive(Debug, Eq, Hash, PartialEq)]
-#[cfg_attr(feature = "clone", derive(Clone))]
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
```
teoxoy, line 576: I think we can get by without having `TypeInner` implement `Clone`.

It looks like this was only needed because of `ConcreteConstructor`.

teoxoy (changes requested):
Thanks for reducing the diff!
I took a brief look around the code and left a few comments.

**General comment**
Only user defined structs should have a `crate::Type.name` (some of the backends now generate extra type aliases for built-in types).

tests/wgsl-errors.rs
```diff
@@ -64,12 +64,12 @@ fn invalid_integer() {
...
-        r###"error: expected ';', found '.'
-  ┌─ wgsl:1:21
+        "const scale: f32 = 1.1.;",
+        r###"error: expected identifier, found ';'
+  ┌─ wgsl:1:24
   │
-1 │ let scale: f32 = 1.1.;
-  │                     ^ expected ';'
+1 │ const scale: f32 = 1.1.;
+  │                        ^ expected identifier
```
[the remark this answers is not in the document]
SparkyPotato: Not sure anything can be done about this.

The previous parser parsed only a single literal or construction, but we now parse a full expression because it is valid in the spec (we don't actually support most expressions, but we should) - so this becomes a field access on `1.1`.

src/front/wgsl/construction.rs
```diff
@@ -1,679 +1,609 @@
...
-        width: Bytes,
-    },
+use crate::front::wgsl;
+use crate::front::wgsl::ast;
+use crate::front::wgsl::ast::{ArraySize, ConstructorType};
+use crate::front::wgsl::Error;
+use crate::front::wgsl::{ExpressionContext, Lowerer, OutputContext};
+use crate::{Handle, Span};
+
+enum ConcreteConstructor {
```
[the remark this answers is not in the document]
SparkyPotato: That would lead to #2105, so we do have to fully resolve the type.

benches/criterion.rs
```diff
@@ -42,12 +42,11 @@ fn frontends(c: &mut Criterion) {
     #[cfg(all(feature = "wgsl-in", feature = "serialize", feature = "deserialize"))]
     group.bench_function("bin", |b| {
         let inputs_wgsl = gather_inputs("tests/in", "wgsl");
-        let mut parser = naga::front::wgsl::Parser::new();
         let inputs_bin = inputs_wgsl
             .iter()
             .map(|input| {
                 let string = std::str::from_utf8(input).unwrap();
-                let module = parser.parse(string).unwrap();
+                let module = naga::front::wgsl::parse_str(string).unwrap();
```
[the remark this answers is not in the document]
SparkyPotato: `parse_str` does both parsing and lowering. Do you want me to split each benchmark into just parsing, and both parsing and lowering?

SparkyPotato:
> Only user defined structs should have a crate::Type.name (some of the backends now generate extra type aliases for built-in types).

Wouldn't this lead to worse validation errors, where something like `Vec { kind: Float, width: 4 }` would be used instead of just `vec4<f32>`?

tests/in/lexical-scopes.wgsl
```diff
@@ -29,9 +29,7 @@ fn loopLexicalScope(a: bool) {
 
 fn forLexicalScope(a: f32) {
     let a = false;
-    for (var a = 0; a < 1; a++) {
-        let a = 3.0;
-    }
+    for (var a = 0; a < 1; a++) {}
```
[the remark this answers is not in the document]
teoxoy: In the section about the [For Statement](https://gpuweb.github.io/gpuweb/wgsl/#for-statement) they say:

image

There is also an example in that section showing the additional scope around the loop (if you were to translate it to a loop).

tests/wgsl-errors.rs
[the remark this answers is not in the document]
teoxoy: ah, I see - it's fine then

benches/criterion.rs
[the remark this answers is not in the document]
teoxoy: Got tripped up by the lowering missing in the `front/wgsl` bench.

Considering the `Parser` was reused in the benchmarks before, can we try to reuse the `Parser` and maybe the `Index` as well? We might see some additional benefits when it comes to perf from doing so.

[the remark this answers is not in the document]
SparkyPotato: `Parser` and `Index` each just have a single `Vec` (`Index` also preallocates it with the required capacity), so I'm not sure that's worth the added complexity.

SparkyPotato:
> I think for errors we can have a method that gives us the proper name. Also, considering we have other front-ends that don't use the same syntax (i.e. `vec4<f32>`) for types, we should avoid passing those through IR.

Done!

teoxoy:
Thanks! There seem to be a few more places where we still set the name:
- 5 in `Lowerer.construct`
- 1 in `ExpressionContext.resolve_type`

src/lib.rs
SparkyPotato: `ExpressionContext::resolve_type` also does require cloning, and I'm not sure how I can get around that.

teoxoy: `TypeResolution` has a `Clone` impl. See how it's used here: https://github.com/gfx-rs/naga/blob/1650581fe7a94ed40067313f3cd205414dd4febf/src/front/glsl/types.rs#L296

Consider adding a `Typifier.get_handle` fn; could be used by both frontends.

src/front/wgsl/mod.rs
```diff
@@ -169,53 +182,75 @@ pub enum Error<'a> {
...
+        expected: Range<u32>,
+        found: u32,
+    },
+    FunctionReturnsVoid(Span),
     Other,
 }
 
 impl<'a> Error<'a> {
-    fn as_parse_error(&self, source: &'a str) -> ParseError {
+    pub fn as_parse_error(&self, source: &'a str) -> ParseError {
```
teoxoy, line 207: Let's keep this private

src/front/wgsl/mod.rs
```diff
@@ -4723,28 +3796,2227 @@ impl Parser {
         }
     }
 
-    pub fn parse(&mut self, source: &str) -> Result<crate::Module, ParseError> {
+    pub fn parse<'a>(&mut self, source: &'a str) -> Result<TranslationUnit<'a>, ParseError> {
```
teoxoy, line 3797: Let's keep the fn signature the same considering this is part of our public API and there isn't much people can do with a `TranslationUnit<'a>`.

This means we don't have to expose `Lowerer` and we can undo the changes to `parse_str` & the benchmark file.

SparkyPotato:
@teoxoy I have done a few things:
1. Add `Typifier::get_handle` and used that in `ExpressionContext`.
2. Moved `ensure_type_exists` to `OutputContext`, and used that everywhere.
3. Created a sort of workaround for two-stage borrows to avoid the need to clone `TypeInner` - this does lead to code duplication, so let me know if you can think of a better solution.

teoxoy:
> 1. Added `Typifier::get_handle` and used that in `ExpressionContext`. The GLSL frontend has not been edited.

Feel free to change the GLSL frontend as well.

> 2. Moved `ensure_type_exists` to `OutputContext`, and used that everywhere.

We can make use of `ensure_type_exists` in a few more places:
- 2 in `ExpressionContext.create_zero_value_constant`
- 1 in `Lowerer.construct`

> 3. Created a sort of workaround for two-stage borrows to avoid the need to clone `TypeInner` - this does lead to code duplication, so let me know if you can think of a better solution.

Can't think of doing it any other way right now since we match on those `TypeInner`s but will keep it in mind.

Thanks for the updates!

teoxoy:
> The only difference in output is that I don't initialize `LocalVariable`s with constants since I'm not too sure if that added complexity is actually worth it. I can add it back if it's something wanted, though.

_from https://github.com/gfx-rs/naga/pull/2075#issuecomment-1304453265_

I also noticed this by looking at the test snapshots. I think it should be fine for now considering we'll have to deal with const expressions differently soon.

SparkyPotato:
All changes done.

teoxoy:
I'll do one last round of review tomorrow and take a look at all those issues as well.
Will probably edit the top comment so that they will all get automatically closed.

Considering the size and scope of the PR, we have a lot to go through.
Thanks for being so receptive!

SparkyPotato:
Great! Thanks for taking the time to review!

teoxoy:
Considering we are fixing #2105 with this PR, it would be great to add some tests for type aliases.

SparkyPotato:
> Considering we are fixing #2105 with this PR, it would be great to add some tests for type aliases.

Is there any specific place I should add them? Or should I create a new snapshot test file?

teoxoy:
A new snapshot test file would be best. Thanks!

src/front/wgsl/construction.rs
```diff
@@ -1,679 +1,635 @@
...
+    }
+
+    pub(super) fn const_construct(
+        &mut self,
+        span: Span,
+        constructor: &ConstructorType<'source>,
+        components: &[Handle<ast::Expression<'source>>],
+        mut ctx: OutputContext<'source, '_, '_>,
+    ) -> Result<crate::ConstantInner, Error<'source>> {
+        // TODO: Support zero values, splatting and inference.
```
teoxoy, line 552: What does this refer to?

teoxoy:
There were too many conversions between `Span` (`Range<usize>`) and `NagaSpan`. i.e. `Span.into()`/`NagaSpan.to_range().unwrap()`.

And even previsouly when I worked on the WGSL frontend, the `Span`(`Range<usize>`)/`NagaSpan` duality only got in the way.

I removed `Span`(`Range<usize>`) completely and the code is much nicer to work on now. (no more unnecessary `.clone()` calls either because ranges don't implement `Clone`).

src/front/wgsl/construction.rs
SparkyPotato: That would be #1956 and #1957.

src/front/wgsl/ast.rs
```diff
@@ -0,0 +1,344 @@
...
+
+#[derive(Debug)]
+pub struct Statement<'a> {
+    pub kind: StatementKind<'a>,
+    pub span: Span,
+}
+
+#[derive(Debug)]
+pub enum StatementKind<'a> {
+    VarDecl(Box<VarDecl<'a>>),
```
teoxoy, line 195: I think this box can be removed.

src/front/wgsl/ast.rs
```diff
@@ -0,0 +1,344 @@
...
+    Image {
+        dim: crate::ImageDimension,
+        arrayed: bool,
+        class: crate::ImageClass,
+    },
+    Sampler {
+        comparison: bool,
+    },
+    BindingArray {
+        base: Box<Type<'a>>,
```
teoxoy, line 176: I think we can replace the 3 boxes in `TypeKind` with `Handle`s (and an additional `Arena`).

src/front/wgsl/ast.rs
teoxoy: Also, let's rename the variant and type to `LocalDecl` (similar to our `GlobalDecl`).

SparkyPotato: Without that, the size of a statement is severely bloated.

Performance doesn't really change anyways, even with our relatively declaration-heavy tests.

src/front/wgsl/ast.rs
SparkyPotato: I'm not sure this is worth it either, as type nesting is rare, and is also usually quite shallow.

src/front/wgsl/ast.rs
teoxoy: Hmm, 56 vs 80.

But if we put `ast::Type` in an `Arena` and share `Handle`s in the AST, it gets us back down to 56.

SparkyPotato: I still think that pessimising the 'normal' path (non-nested types) to arena pushes won't be worth it compared to the relatively rare nested types and local declaration.

Though, as is usual with most micro-optimizations like this, I really doubt it's gonna make any difference that isn't just normal benchmark noise.

src/front/wgsl/ast.rs
teoxoy: I agree that most types won't be nested but we are dealing with user input here, so we need to expect the worst scenario where we'll have a bunch of `Box`es spread out in memory.

src/front/wgsl/ast.rs
teoxoy: I tried this change locally and got 2-3% better perf (compared with this branch).
It's not groundbreaking but it allows us to still be performant with arbitrary nesting while also keeping things consistent.

jimblandy: Arenas are the way to go here.

SparkyPotato:
Looks great!

I also think we should use something like `aho_corasick` or `daachorse` for the reserved identifier detection since we'll go to `O(ident_length)` instead of `O(reserved_list_length)`.

teoxoy:
> I also think we should use something like `aho_corasick` or `daachorse` for the reserved identifier detection since we'll go to `O(ident_length)` instead of `O(reserved_list_length)`.

That's a good idea!

The speed of `daachorse` compared with `aho_corasick` looks substantial (according to the benchmarks `daachorse` has in their readme) but we could check out both and see which one is faster for us. 
Let's experiment with this in a separate PR though (this change should be quite self-contained).

SparkyPotato:
Is most of the reviewing done? Can we start looking at separating things again, because `wgsl/mod.rs` is a gargantuan mess, to say the least?

I was thinking of a few things:
- Rename the main `Parser` to `Frontend`, which caches the lexer, parser, and lowerer (A small API change). 
- Move each pass into its own individual module, as I had done.

teoxoy:
> Is most of the reviewing done? 

I still need to take a look at some things but on my side mostly yes.
@jimblandy just started to take a look.

> I was thinking of a few things:
> - Rename the main `Parser` to `Frontend`, which caches the lexer, parser, and lowerer (A small API change). 
> - Move each pass into its own individual module, as I had done.

Looking at the diff, most of the `Lowerer` code looks new (at least to git) and I think we can move it in a `lowerer.rs` file but let's keep the `Parser` in `mod.rs` since the diff for it is more or less accurate. If we want to move other things around, we can do so after this PR lands since we'll squash it and git won't recognize the file moves.

Regarding the `Parser` name, let's keep it as it is for now (at least in our public API) since we use the same terminology for the other frontends. If we want to change it in our public API, I'd say we do it for all frontends which might as well be a separate change.

SparkyPotato:
If we're going to move things around later, might as well move everything into different files in a later PR, and the same for the `Frontend` rename.

SparkyPotato:
Is there anything left to do?

teoxoy (approved):
From my side this is good to go. There are still some small things that can be improved but I think we can do that in following PRs.

Solid work!

jimblandy:
I'm getting the following errors when I try `make validate-spv`:

```
Validating spv/access.spvasm
error: line 145: Structure id 12 decorated as Block for variable in Uniform storage class must follow standard uniform buffer layout rules: member 0 is a matrix with stride 8 not satisfying alignment to 16
  %Baz = OpTypeStruct %mat3v2float

make: *** [Makefile:34: validate-spv] Error 1
```

The CI doesn't seem upset, though - am I doing something wrong?

jimblandy:
It seems like CI is using `focal/universe amd64 spirv-tools amd64 2020.1-2`, and I'm using `SPIRV-Tools v2022.4-dev v2022.2-85-ga98f05d0`.

src/span.rs
```diff
@@ -85,6 +93,15 @@ impl From<Range<usize>> for Span {
     }
 }
 
+impl std::ops::Index<Span> for str {
```
jimblandy, line 96: sweet

src/front/wgsl/mod.rs
```diff
@@ -30,10 +29,9 @@ use codespan_reporting::{
         termcolor::{ColorChoice, NoColor, StandardStream},
     },
 };
-use std::{borrow::Cow, convert::TryFrom, ops};
+use std::{borrow::Cow, convert::TryFrom, ops::Range};
 use thiserror::Error;
 
-type Span = ops::Range<usize>;
```
jimblandy, line 36: very nice

src/front/wgsl/mod.rs
```diff
@@ -328,241 +334,291 @@ impl<'a> Error<'a> {
...
+                            ),
+                        ]
+                    })
+                    .collect(),
+                notes: vec![],
+            },
+            Error::ConstExprUnsupported(span) => ParseError {
+                message: "this constant expression is not supported".to_string(),
+                labels: vec![(span, "expression is not supported".into())],
+                notes: vec!["this should be fixed in a future version of Naga".into()],
```
jimblandy, line 588: Let's link to an issue here. (If we don't have an issue, let's file one.)

src/front/wgsl/mod.rs
```diff
@@ -328,241 +334,291 @@ impl<'a> Error<'a> {
...
+                    }
+                    .into(),
+                )],
+                notes: vec![if uint {
+                    format!("suffix the integer with a `u`: '{}u'", &source[span])
+                } else {
+                    let span = span.to_range().unwrap();
+                    format!(
+                        "remove the `u` suffix: '{}'",
+                        &source[span.start..span.end - 1]
```
jimblandy, line 607: There's no way this could end up falling in the midst of a multibyte UTF-8 character, right?

src/front/wgsl/index.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    pub fn visit_ordered(&self) -> impl Iterator<Item = Handle<ast::GlobalDecl<'a>>> + '_ {
+        self.dependency_order.iter().copied()
+    }
+}
+
+pub struct ResolvedDependency<'a> {
+    pub decl: Handle<ast::GlobalDecl<'a>>,
+    pub usage: Span,
+}
```
jimblandy: It seems like this doesn't need to be `pub`.

```suggestion
struct ResolvedDependency<'a> {
    decl: Handle<ast::GlobalDecl<'a>>,
    usage: Span,
}
```

src/front/wgsl/index.rs
```diff
@@ -0,0 +1,143 @@
...
+    module: &'temp ast::TranslationUnit<'source>,
+    visited: Vec<bool>,
+    temp_visited: Vec<bool>,
+    path: Vec<ResolvedDependency<'source>>,
+    out: Vec<Handle<ast::GlobalDecl<'source>>>,
+}
+
+impl<'a> DependencySolver<'a, '_> {
+    fn solve(mut self) -> Result<Vec<Handle<ast::GlobalDecl<'a>>>, Error<'a>> {
+        for id in self.module.decls.iter().map(|x| x.0) {
```
jimblandy, line 97: The way Naga generally does this is clearer, I think:

```suggestion
        for (id, _decl) in self.module.decls.iter() {
```

src/front/wgsl/index.rs
```diff
@@ -0,0 +1,143 @@
...
+                            usage: dep.usage,
+                        })
+                    } else {
+                        let start_at = self
+                            .path
+                            .iter()
+                            .rev()
+                            .enumerate()
+                            .find(|&(_, dep)| dep.decl == dep_id)
+                            .map(|x| x.0)
```
jimblandy, line 145:
```suggestion
                            .find_map(|(i, dep)| (dep.decl == dep_id).then_some(i))
```

jimblandy (commented):
Okay, I've finally got some time to start looking at this. I pushed a bunch of comments and docs I wrote in the process of reading; if you could give that a look and make sure I haven't really misunderstood something, that'd be great.

Some initial comments:

jimblandy:
I have more time blocked out for reviewing this tomorrow.

src/front/wgsl/index.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        self.temp_visited[id_usize] = false;
+        self.visited[id_usize] = true;
+        self.out.push(id);
+
+        Ok(())
+    }
+}
+
+const fn decl_ident<'a>(decl: &ast::GlobalDecl<'a>) -> ast::Ident<'a> {
```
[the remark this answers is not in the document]
SparkyPotato: It would, but that wouldn't work when we implement `static_assert`.

src/front/wgsl/mod.rs
SparkyPotato: Yeah the last character is always guaranteed to be a `u`.

SparkyPotato:
> I pushed a bunch of comments and docs I wrote in the process of reading; if you could give that a look and make sure I haven't really misunderstood something, that'd be great.

I made a slight change to the docs on `Error::Redefinition`, since it's used also for the redefinition of locals.

jimblandy:
This program doesn't seem to work:

```
fn f() -> i32 {
   return g(1);
}

fn g(i: i32) -> i32 {
   return i;
}
```

I get the following output:

```
error: no definition in scope for identifier: 'g'
  ┌─ new.wgsl:2:11
  │
2 │    return g(1);
  │           ^ unknown identifier
```

jimblandy:
The `ast::Expression::Call` case needs to record the reference in `ctx.unresolved`, the way `parse_function_statement` does.

src/front/wgsl/mod.rs
```diff
@@ -1559,1180 +1697,348 @@ impl Parser {
...
+                    to,
+                    ty_span: span,
                 }
-                crate::ConstantInner::Composite {
-                    ty: composite_ty,
-                    components,
+            }
+            // everything else must be handled later, since they can be hidden by user-defined functions.
+            _ => {
+                let arguments = self.parse_arguments(lexer, ctx.reborrow())?;
```
jimblandy, line 1873: This needs to record a dependency as well (and there should be a test for this):

```suggestion
                ctx.unresolved.insert(ast::Dependency {
                    ident: name,
                    usage: name_span,
                });

                let arguments = self.parse_arguments(lexer, ctx.reborrow())?;
```

jimblandy (commented):
Some more comments:

src/front/wgsl/mod.rs
```diff
@@ -3302,1438 +2532,2807 @@ impl Parser {
...
+        lexer: &mut Lexer<'a>,
+        mut ctx: ParseExpressionContext<'a, '_, '_>,
+    ) -> Result<Handle<ast::Type<'a>>, Error<'a>> {
+        self.push_rule_span(Rule::TypeDecl, lexer);
+
+        let (name, span) = lexer.next_ident_with_span()?;
+
+        let ty = match self.parse_type_decl_impl(lexer, name, ctx.reborrow())? {
+            Some(ty) => ty,
+            None => ast::Type::User(ast::Ident { name, span }),
```
jimblandy, line 2635: Another missing dependancy:

```suggestion
            None => {
                ctx.unresolved.insert(ast::Dependency {
                    ident: name,
                    usage: span,
                });
                ast::Type::User(ast::Ident { name, span })
            },
```

jimblandy (commented):
I think we need a bunch of new tests for forward references.

SparkyPotato:
Fixed most of the issues, and added a test (`module-scope.wgsl`) for all global declarations. Do tell me if I've missed anything.

jimblandy:
We don't support this yet:

```
type MyF32 = f32;
type MyVecF32 = vec3<MyF32>;
```

This gives an error:

```
error: unknown scalar type: 'MyF32'
  ┌─ new.wgsl:2:22
  │
2 │ type MyVecF32 = vec3<MyF32>;
  │                      ^^^^^ unknown scalar type
  │
  = note: Valid scalar types are f32, f64, i32, u32, bool
```

The spec says that the type parameter to `vec3` is a `type_specifier`, not `type_specifier_without_ident`.

Given that this is going to need to support forward references too, I think this is going to take some pretty big changes to `ast::Type`. The old code didn't support it either, so should we leave this for another PR?

SparkyPotato:
I'd say leaving it for a later PR would be a better idea.

src/front/wgsl/mod.rs
```diff
@@ -3980,760 +3002,2339 @@ impl Parser {
...
-        }
+        Ok(ctx.ensure_type_exists(inner))
     }
 
-    pub fn parse(&mut self, source: &str) -> Result<crate::Module, ParseError> {
-        self.reset();
+    fn constant(
+        &mut self,
+        expr: Handle<ast::Expression<'source>>,
+        arena: &Arena<ast::Expression<'source>>,
```
jimblandy, line 5241: Taking the arena explicitly here, while also taking the `OutputContext`, is delicate, and I think there's a bug. When I compile the following program:

```
fn f() -> i32 {
   let a = array<i32,u32(42)>(64);
   return a[13];
}
```

I end up with an array whose size is 64, not 42:

- The `ast::ConstructorType::Array` case of `Lowerer::constructor` carefully passes `ctx.global_expressions` as the `arena` argument to `self.constant`, so that the array's size is looked up in the global expression arena, as it should, yielding the `Construct` expression for `u32(42)`. This is all fine.
- But then the `ast::Expression::Construct` case of `Lowerer::constant_inner` simply calls `const_construct`, which doesn't let you specify any arena. Since the `OutputContext` does have a `read_expressions` arena, we switch back to the function's expression arena.
- Now the operand of that `u32` construction expression is looked up in the wrong arena, yielding `64` instead of `42`.

As a result, we end up with a malformed `Module`:

```
Module {
    types: {
        [1]: Type {
            name: None,
            inner: Scalar {
                kind: Sint,
                width: 4,
            },
        },
        [2]: Type {
            name: None,
            inner: Scalar {
                kind: Uint,
                width: 4,
            },
        },
        [3]: Type {
            name: None,
            inner: Array {
                base: [1],
                size: Constant(
                    [2],
                ),
                stride: 4,
            },
        },
    },
    constants: {
        [1]: Constant {
            name: None,
            specialization: None,
            inner: Scalar {
                width: 4,
                value: Sint(
                    64,
                ),
            },
        },
        [2]: Constant {
            name: None,
            specialization: None,
            inner: Composite {
                ty: [2],
                components: [
                    [1],
                ],
            },
        },
        [3]: Constant {
            name: None,
            specialization: None,
            inner: Scalar {
                width: 4,
                value: Sint(
                    13,
                ),
            },
        },
    },
    ...
```

There's a second bug, in that the construction expression isn't evaluated by the front end, but the arena-crossing leaves us with `u32(64)` instead of `u32(42)`.

We could fix this by introducing something like `global_expr_as_expr` on `OutputContext`, and then dropping the `arena` argument, so that the entire array expression is interpreted using the proper context. But I wonder if there isn't some not-to-horrible way to catch these kinds of bugs at compile time.

jimblandy (changes requested):
A bug, and a few more comments:

jimblandy:
I pushed another batch of documentation. If you could give them a quick look, that'd be great.

I think I'm almost done with the review. It's holidays here, but I can pick this up on Monday and try to finish it off.

src/front/wgsl/mod.rs
teoxoy: I remember seeing this kind of behavior while working on `const-expressions` (on top of this PR). IIRC I fixed it by always making sure the array length goes in/comes from `global_expressions`.

I was thinking at the time that it might be nice to know from which arena the handle is from (by tagging the arena and its handles with a phantom type or some id); then we could panic if we try to index an arena with a handle from a different arena. Actually thinking about this now, we could achieve the same thing by using a tuple  with the ast expression type and a marker type (`Global` and `Local`) for each of the 2 expression arenas.

Or just putting all the ast expressions of all functions in one big arena in the module and avoid the confusion altogether.

SparkyPotato: I fixed this by just giving `const_construct` the arena as well.

src/front/wgsl/mod.rs
```diff
@@ -3980,760 +3002,2339 @@ impl Parser {
...
-                if let Some(explicit_ty) = given_ty {
-                    let con = &module.constants[const_handle];
-                    let type_match = match con.inner {
-                        crate::ConstantInner::Scalar { width, value } => {
-                            module.types[explicit_ty].inner
-                                == crate::TypeInner::Scalar {
-                                    kind: value.scalar_kind(),
-                                    width,
-                                }
+    fn resolve_ast_type(
```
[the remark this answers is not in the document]
SparkyPotato: I renamed everything to just what they works on, removing the `parse_` and `lower_` prefixes.

src/front/wgsl/mod.rs
SparkyPotato: > Or just putting all the ast expressions of all functions in one big arena in the module and avoid the confusion altogether.

This seems like a great idea. 😛

SparkyPotato:
The docs look great!

src/front/wgsl/mod.rs
jimblandy: Just one big arena sounds cool to me. Are there any downsides?

jimblandy:
I have a patch ready that merges the expression arenas. It seems to work fine.

src/front/mod.rs
```diff
@@ -89,6 +89,19 @@ impl Typifier {
         self.resolutions[expr_handle.index()].inner_with(types)
     }
 
+    pub fn get_handle(
+        &self,
+        expr_handle: Handle<crate::Expression>,
+        types: &mut UniqueArena<crate::Type>,
+    ) -> Handle<crate::Type> {
```
jimblandy, line 96: Rather than doing this, I think we should use `TypeResolution` more places in `front::wgsl`, and just lean on the `ensure_type_exists` methods to convert those into `Handle` values when we really need to. Working on a patch for this now.

jimblandy:
Hmm. This drops the name:

```
type T = i32;
let y: T = 1;
```

SparkyPotato:
> Hmm. This drops the name from the type:
> 
> ```
> type T = i32;
> const y: T = 1;
> ```

Yeah, we discard type aliases as soon as they are resolved. Using type alias names when possible is quite hard, even rustc doesn't do it.

Solutions to fixing that include:
- Duplicating the type with a different name. This would require `TypeInner: Clone`, however.
- Creating a `TypeInner::Alias` variant. This would work but I don't think type aliases should be a part of the IR.

jimblandy:
@teoxoy Could you give [040b700](https://github.com/gfx-rs/naga/pull/2075/commits/040b70083aed43bed2d593b1107846a895500593) a review?

teoxoy (approved):

teoxoy:
LGTM, I think this should be ready to merge now.

jimblandy (approved):
All right, let's go.

vicary:
Thanks for the hard work guys!

When would be the next release date with this included? I really want to see this land in `wgpu` and Deno!

jimblandy:
@cwfitzgerald is planning the next release. Connor?

cwfitzgerald:
25th of this month!

vicary:
@cwfitzgerald That's exciting, thanks!

PR #2078, opened 2022-10-06
Fix repository link pointing to the 0.9 branch

SludgePhD:
0.10 is released, but following the *repository* link sends you to the v0.9 branch. This PR links to the repo instead, leaving navigating to the right commit or branch to the user.


cwfitzgerald (approved):
wgsl has settled enough, that the maintenance burden of this just isn't worth it anymore, thank you!

PR #2079, opened 2022-10-08
glsl-in: Fix position propagation in lowering

JCapucho:
When lowering `Select` expressions the position could be wrongfully updated from `AccessBase { constant_index: false }` to `AccessBase { constant_index: true }` this caused dynamic indexing in an array behind a structure to fail if it was stored in a constant.

Furthermore the position could also be updated from `Rhs` to `AccessBase`, this could cause issues because `AccessBase` doesn't load variables (which `Rhs` does), so accessing a member from a structure behind a pointer would return the wrong result.


teoxoy (approved):
Looks good!

PR #2085, opened 2022-10-14
build: move to the Rust 2021 edition

ErichDonGubler:
Since the MSRV of `naga` [is currently 1.56][msrv], I don't think there's a strong reason to stay with the 2018 edition, and there _are_ a [few good reasons][edition-guide] to move to the 2021 edition.

I did this migration mostly automatically, per [official Rust guidelines]:

```sh
$ cargo fix --edition --all-targets
$ sed -i Cargo.toml 's/2018/2021'
$ cargo fix --edition-idioms --allow-dirty # doesn't change anything
```

The only manual edit needed to stymie a new warning introduced was the removal of the `TryFrom` import from several modules, since it's now in the 2021 prelude.

[msrv]: https://github.com/gfx-rs/naga/blob/a7193d652e49180661a6ca047889e0685667acdb/.github/workflows/pipeline.yml#L14
[edition-guide]: https://doc.rust-lang.org/edition-guide/rust-2021/index.html
[official Rust guidelines]: https://doc.rust-lang.org/cargo/commands/cargo-fix.html#edition-migration

### DIFF
--- a/src/front/wgsl/number.rs
+++ b/src/front/wgsl/number.rs
@@ -24,12 +24,9 @@ impl Number {
     /// lossy, return an error.
     fn abstract_to_concrete(self) -> Result<Number, NumberError> {
         match self {
-            Number::AbstractInt(num) => {
-                use std::convert::TryFrom;
-                i32::try_from(num)
-                    .map(Number::I32)
-                    .map_err(|_| NumberError::NotRepresentable)
-            }
+            Number::AbstractInt(num) => i32::try_from(num)
+                .map(Number::I32)
+                .map_err(|_| NumberError::NotRepresentable),
             Number::AbstractFloat(num) => {
                 let num = num as f32;
                 if num.is_finite() {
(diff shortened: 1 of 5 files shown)


src/front/wgsl/number.rs
```diff
@@ -94,7 +91,7 @@ fn parse(input: &str) -> (Result<Number, NumberError>, &str) {
     /// returns `true` and consumes `X` bytes from the given byte buffer
     /// if the given `X` nr of patterns are found at the start of the buffer
     macro_rules! consume {
-        ($bytes:ident, $($($pattern:pat)|*),*) => {
+        ($bytes:ident, $($($pattern:pat_param)|*),*) => {
```
jimblandy, line 94: I think what we actually want is to just replace the whole `$(pattern:pat_param)|*` with `$pattern:pat` (with corresponding changes in the replacement), since the effect the author was going for was actually an or-pattern.

src/front/wgsl/number.rs
```diff
@@ -106,7 +103,7 @@ fn parse(input: &str) -> (Result<Number, NumberError>, &str) {
     /// if one of the given patterns are found at the start of the buffer
     /// returning the corresponding expr for the matched pattern
     macro_rules! consume_map {
-        ($bytes:ident, [$($($pattern:pat)|* => $to:expr),*]) => {
+        ($bytes:ident, [$($($pattern:pat_param)|* => $to:expr),*]) => {
```
jimblandy, line 106: Same here.

jimblandy (changes requested):
This looks great! But don't forget `cli/Cargo.toml`, and I'd like to see the macros addressed differently.

src/front/wgsl/number.rs
ErichDonGubler: Done; added e0dcc726 to address this.

ErichDonGubler:
@jimblandy: Just fixed `cli/Cargo.toml` with 00220ec4, which I'll `autosquash` right now. I believe I've addressed all feedback now. 😄

PR #2086, opened 2022-10-14
refactor: factor out `Validator::global_var_ty` in expr. validation

ErichDonGubler:


src/valid/expression.rs
```diff
@@ -1518,4 +1422,32 @@ impl super::Validator {
         };
         Ok(stages)
     }
+
+    fn global_var_ty(
+        module: &crate::Module,
+        function: &crate::Function,
+        ty: Handle<crate::Expression>,
```
jimblandy, line 1429: I think this argument should be named `expr`, not `ty`.

src/valid/expression.rs
```diff
@@ -1518,4 +1422,32 @@ impl super::Validator {
...
+    ) -> Result<Handle<crate::Type>, ExpressionError> {
+        use crate::TypeInner as Ti;
```
jimblandy, line 1431: Suggestion: I'd probably not bother with this, but I would use:
```
use crate::Expression as Ex;
```

jimblandy (approved):
Very nice!

Just two minor points, but otherwise this is good to go.

src/valid/expression.rs
ErichDonGubler: Ah, good catch. Not sure why I did that. 😓 Fixed with c692bd6a.

ErichDonGubler:
@jimblandy: I've used `fixup!` commits to address your feedback here, to optimize review cycles. The idea is to use `git rebase --autosquash --interactive ...` before merging, so history becomes clean again. I'll do that rebase now, since you've already approved this, and I don't want to block your ability to immediately merge.

jimblandy (approved):
Impeccable.

PR #2090, opened 2022-10-19
Add handle validation pass to `Validator`

### LINKED ISSUE #1692: Handle validation pass
We should consider introducing an early pass to validation that checks all handles are valid, allowing the rest of validation to assume that arena indexing is infallible.

We recently improved the validator's robustness when presented with modules with invalid `Handle`s. This was necessary, but it had an unpleasant impact on the validator code and some interfaces. Some operations became fallible (e.g. [`try_size`](https://github.com/gfx-rs/naga/blob/f36cfb93dcd54ba1afea2b990b14e84982aeed5c/src/proc/mod.rs#L89), and various error types acquired bad-handle variants (e.g. [`TypeLayoutError::BadHandle`](https://github.com/gfx-rs/naga/blob/f36cfb93dcd54ba1afea2b990b14e84982aeed5c/src/proc/layouter.rs#L41)), and bits and pieces of [error-mapping code](https://github.com/gfx-rs/naga/blob/f36cfb93dcd54ba1afea2b990b14e84982aeed5c/src/proc/layouter.rs#L98) needed to be introduced.

Specifically:
- It's not always obvious where the checks are needed. We found most of the current set by fuzzing.
- Back ends are permitted to assume validation has succeeded, so they either need to unwrap (setting an uncomfortable precedent) or call alternate functions that panic on bad handles (boilerplate, and unclear when to use which alternative)
- The checks are a distraction from the code's primary purpose.
- There's a lot of redundancy in the error types.

An alternative would be to have one pass that runs at the beginning of validation that walks all functions, statement trees, expressions, types, and constants, and checks all handles. Then subsequent validation code can assume handles are valid. It can use the `Index` implementations on arenas, simplify function return types and error enums, and omit error remapping code.

The handle validation pass need not return very detailed error information; handle validation errors are only of interest to front end developers, not WebGPU or wgpu users. Capturing a backtrace would be the most important thing.

ErichDonGubler:
Resolves #1692. Recommended review experience is commit-by-commit.

### DIFF
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -127,9 +127,20 @@ impl<T> Handle<T> {
 )]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct Range<T> {
-    inner: ops::Range<u32>,
+    pub(crate) inner: ops::Range<u32>,
     #[cfg_attr(any(feature = "serialize", feature = "deserialize"), serde(skip))]
     marker: PhantomData<T>,
+}
+
+impl<T> Range<T> {
+    #[cfg(feature = "validate")]
+    pub(crate) const fn erase_type(self) -> Range<()> {
+        let Self { inner, marker: _ } = self;
+        Range {
+            inner,
+            marker: PhantomData,
+        }
+    }
 }
 
 impl<T> Clone for Range<T> {
@@ -326,6 +337,15 @@ impl<T> Arena<T> {
         {
             let _ = handle;
             Span::default()
+        }
+    }
+
+    /// Assert that `handle` is valid for this arena.
+    pub fn check_contains_handle(&self, handle: Handle<T>) -> Result<(), BadHandle> {
+        if handle.index() < self.data.len() {
+            Ok(())
+        } else {
+            Err(BadHandle::new(handle))
         }
     }
 }
@@ -551,6 +571,15 @@ impl<T: Eq + hash::Hash> UniqueArena<T> {
         self.set
             .get_index(handle.index())
             .ok_or_else(|| BadHandle::new(handle))
+    }
+
+    /// Assert that `handle` is valid for this arena.
+    pub fn check_contains_handle(&self, handle: Handle<T>) -> Result<(), BadHandle> {
+        if handle.index() < self.set.len() {
+            Ok(())
+        } else {
+            Err(BadHandle::new(handle))
+        }
     }
 }
 
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -127,18 +127,36 @@ impl<T> Handle<T> {
 )]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct Range<T> {
-    pub(crate) inner: ops::Range<u32>,
+    inner: ops::Range<u32>,
     #[cfg_attr(any(feature = "serialize", feature = "deserialize"), serde(skip))]
     marker: PhantomData<T>,
 }
 
 impl<T> Range<T> {
-    #[cfg(feature = "validate")]
     pub(crate) const fn erase_type(self) -> Range<()> {
         let Self { inner, marker: _ } = self;
         Range {
             inner,
             marker: PhantomData,
+        }
+    }
+}
+
+// NOTE: Keep this diagnostic in sync with that of [`BadHandle`].
+#[derive(Clone, Debug, thiserror::Error)]
+#[error("Handle range {range:?} of {kind} is either not present, or inaccessible yet")]
+pub struct BadRangeError {
+    // This error is used for many `Handle` types, but there's no point in making this generic, so
+    // we just flatten them all to `Handle<()>` here.
+    kind: &'static str,
+    range: Range<()>,
+}
+
+impl BadRangeError {
+    pub fn new<T>(range: Range<T>) -> Self {
+        Self {
+            kind: std::any::type_name::<T>(),
+            range: range.erase_type(),
         }
     }
 }
@@ -346,6 +364,22 @@ impl<T> Arena<T> {
             Ok(())
         } else {
             Err(BadHandle::new(handle))
+        }
+    }
+
+    /// Assert that `range` is valid for this arena.
+    pub fn check_contains_range(&self, range: &Range<T>) -> Result<(), BadRangeError> {
+        // Since `range.inner` is a `Range<u32>`, we only need to
+        // check that the start precedes the end, and that the end is
+        // in range.
+        if range.inner.start > range.inner.end
+            || self
+                .check_contains_handle(Handle::new(range.inner.end.try_into().unwrap()))
+                .is_err()
+        {
+            Err(BadRangeError::new(range.clone()))
+        } else {
+            Ok(())
         }
     }
 }
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -15,6 +15,15 @@ pub struct BadHandle {
     pub index: usize,
 }
 
+impl BadHandle {
+    fn new<T>(handle: Handle<T>) -> Self {
+        Self {
+            kind: std::any::type_name::<T>(),
+            index: handle.index(),
+        }
+    }
+}
+
 /// A strongly typed reference to an arena item.
 ///
 /// A `Handle` value can be used as an index into an [`Arena`] or [`UniqueArena`].
@@ -282,10 +291,9 @@ impl<T> Arena<T> {
     }
 
     pub fn try_get(&self, handle: Handle<T>) -> Result<&T, BadHandle> {
-        self.data.get(handle.index()).ok_or_else(|| BadHandle {
-            kind: std::any::type_name::<T>(),
-            index: handle.index(),
-        })
+        self.data
+            .get(handle.index())
+            .ok_or_else(|| BadHandle::new(handle))
     }
 
     /// Get a mutable reference to an element in the arena.
@@ -540,10 +548,9 @@ impl<T: Eq + hash::Hash> UniqueArena<T> {
 
     /// Return this arena's value at `handle`, if that is a valid handle.
     pub fn get_handle(&self, handle: Handle<T>) -> Result<&T, BadHandle> {
-        self.set.get_index(handle.index()).ok_or_else(|| BadHandle {
-            kind: std::any::type_name::<T>(),
-            index: handle.index(),
-        })
+        self.set
+            .get_index(handle.index())
+            .ok_or_else(|| BadHandle::new(handle))
     }
 }
 
--- a/src/proc/mod.rs
+++ b/src/proc/mod.rs
@@ -101,7 +101,7 @@ impl super::TypeInner {
     ///
     /// # Panics
     ///
-    /// Panics if the `constants` doesn't contain a referenced handle. This may not happen in
+    /// Panics if `constants` doesn't contain a referenced handle. This may not happen in
     /// a properly validated IR module.
     pub fn size(&self, constants: &super::Arena<super::Constant>) -> u32 {
         match *self {
--- a/src/proc/mod.rs
+++ b/src/proc/mod.rs
@@ -98,11 +98,6 @@ impl super::TypeInner {
     }
 
     /// Get the size of this type.
-    ///
-    /// # Panics
-    ///
-    /// Panics if `constants` doesn't contain a referenced handle. This may not happen in
-    /// a properly validated IR module.
     pub fn size(&self, constants: &super::Arena<super::Constant>) -> u32 {
         match *self {
             Self::Scalar { kind: _, width } | Self::Atomic { kind: _, width } => width as u32,
--- a/src/proc/mod.rs
+++ b/src/proc/mod.rs
@@ -97,11 +97,14 @@ impl super::TypeInner {
         }
     }
 
-    pub fn try_size(
-        &self,
-        constants: &super::Arena<super::Constant>,
-    ) -> Result<u32, crate::arena::BadHandle> {
-        Ok(match *self {
+    /// Get the size of this type.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the `constants` doesn't contain a referenced handle. This may not happen in
+    /// a properly validated IR module.
+    pub fn size(&self, constants: &super::Arena<super::Constant>) -> u32 {
+        match *self {
             Self::Scalar { kind: _, width } | Self::Atomic { kind: _, width } => width as u32,
             Self::Vector {
                 size,
@@ -122,8 +125,7 @@ impl super::TypeInner {
             } => {
                 let count = match size {
                     super::ArraySize::Constant(handle) => {
-                        let constant = constants.try_get(handle)?;
-                        constant.to_array_length().unwrap_or(1)
+                        constants[handle].to_array_length().unwrap_or(1)
                     }
                     // A dynamically-sized array has to have at least one element
                     super::ArraySize::Dynamic => 1,
@@ -132,13 +134,7 @@ impl super::TypeInner {
             }
             Self::Struct { span, .. } => span,
             Self::Image { .. } | Self::Sampler { .. } | Self::BindingArray { .. } => 0,
-        })
-    }
-
-    /// Get the size of this type. Panics if the `constants` doesn't contain
-    /// a referenced handle. This may not happen in a properly validated IR module.
-    pub fn size(&self, constants: &super::Arena<super::Constant>) -> u32 {
-        self.try_size(constants).unwrap()
+        }
     }
 
     /// Return the canonical form of `self`, or `None` if it's already in
--- a/src/valid/expression.rs
+++ b/src/valid/expression.rs
@@ -1,4 +1,7 @@
 #[cfg(feature = "validate")]
+use std::ops::Index;
+
+#[cfg(feature = "validate")]
 use super::{
     compose::validate_compose, validate_atomic_compare_exchange_struct, FunctionInfo, ShaderStages,
     TypeFlags,
@@ -7,7 +10,7 @@ use super::{
 use crate::arena::UniqueArena;
 
 use crate::{
-    arena::{BadHandle, Handle},
+    arena::Handle,
     proc::{IndexableLengthError, ResolveError},
 };
 
@@ -18,10 +21,6 @@ pub enum ExpressionError {
     DoesntExist,
     #[error("Used by a statement before it was introduced into the scope by any of the dominating blocks")]
     NotInScope,
-    #[error("Depends on {0:?}, which has not been processed yet")]
-    ForwardDependency(Handle<crate::Expression>),
-    #[error(transparent)]
-    BadDependency(#[from] BadHandle),
     #[error("Base type {0:?} is not compatible with this expression")]
     InvalidBaseType(Handle<crate::Expression>),
     #[error("Accessing with index {0:?} can't be done")]
@@ -132,15 +131,18 @@ struct ExpressionTypeResolver<'a> {
 }
 
 #[cfg(feature = "validate")]
-impl<'a> ExpressionTypeResolver<'a> {
-    fn resolve(
-        &self,
-        handle: Handle<crate::Expression>,
-    ) -> Result<&'a crate::TypeInner, ExpressionError> {
+impl<'a> Index<Handle<crate::Expression>> for ExpressionTypeResolver<'a> {
+    type Output = crate::TypeInner;
+
+    #[allow(clippy::panic)]
+    fn index(&self, handle: Handle<crate::Expression>) -> &Self::Output {
         if handle < self.root {
-            Ok(self.info[handle].ty.inner_with(self.types))
+            self.info[handle].ty.inner_with(self.types)
         } else {
-            Err(ExpressionError::ForwardDependency(handle))
+            panic!(
+                "Depends on {:?}, which has not been processed yet",
+                self.root
+            )
         }
     }
 }
@@ -166,7 +168,7 @@ impl super::Validator {
 
         let stages = match *expression {
             E::Access { base, index } => {
-                let base_type = resolver.resolve(base)?;
+                let base_type = &resolver[base];
                 // See the documentation for `Expression::Access`.
                 let dynamic_indexing_restricted = match *base_type {
                     Ti::Vector { .. } => false,
@@ -179,7 +181,7 @@ impl super::Validator {
                         return Err(ExpressionError::InvalidBaseType(base));
                     }
                 };
-                match *resolver.resolve(index)? {
+                match resolver[index] {
                     //TODO: only allow one of these
                     Ti::Scalar {
                         kind: Sk::Sint | Sk::Uint,
@@ -257,7 +259,7 @@ impl super::Validator {
                     Ok(limit)
                 }
 
-                let limit = resolve_index_limit(module, base, resolver.resolve(base)?, true)?;
+                let limit = resolve_index_limit(module, base, &resolver[base], true)?;
                 if index >= limit {
                     return Err(ExpressionError::IndexOutOfBounds(
                         base,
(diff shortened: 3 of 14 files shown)


ErichDonGubler:
@jimblandy: I think that this is finally ready! 🙌🏻

ErichDonGubler:
@jimblandy: Just wanted to ping here again, and get confirmation on whether reviewing this is a priority for you or not. Whatever the case, thanks in advance! 🙂

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+                &crate::TypeInner::Pointer { base, .. } => {
+                    this_handle.check_dep(base)?;
+                }
+                &crate::TypeInner::Array { base, .. } => {
+                    this_handle.check_dep(base)?;
+                }
+                &crate::TypeInner::Struct { ref members, .. } => {
+                    this_handle.check_dep_iter(members.iter().map(|m| m.ty))?;
+                }
+                &crate::TypeInner::BindingArray { base, .. } => {
```
jimblandy, line 55: Everything that applies to `Array` sizes applies here as well, I think.

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+                validate_constant(init_expr)?;
+            }
+        }
+
+        let validate_expressions = |expressions: &Arena<crate::Expression>,
+                                    local_variables: &Arena<crate::LocalVariable>|
+         -> Result<_, InvalidHandleError> {
+            for (this_handle, expression) in expressions.iter() {
+                match expression {
+                    &crate::Expression::Access { base, .. }
```
jimblandy, line 99: We need to check the `index` of an `Access` expression as well.

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+                ty,
+                init,
+            } = global_variable;
+            validate_type(ty)?;
+            if let Some(init_expr) = init {
+                validate_constant(init_expr)?;
+            }
+        }
+
+        let validate_expressions = |expressions: &Arena<crate::Expression>,
```
jimblandy, line 94: Could we just inline this into `validate_function`?

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+                    } => {
+                        this_handle
+                            .check_dep(image)?
+                            .check_dep(coordinate)?
+                            .check_dep_opt(array_index)?
+                            .check_dep_opt(sample)?
+                            .check_dep_opt(level)?;
+                    }
+                    &crate::Expression::ImageQuery { image, query } => {
+                        this_handle.check_dep(image)?.check_dep_opt(match query {
```
jimblandy, line 162: The way this turned out, I think it'd just be clearer to have the `match` on the outside, and call `check_dep_opt` from the `ImageQuery::Size` arm. Listing out all the variant is good, though.

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+                }
+            }
+            Ok(())
+        };
+
+        let validate_function = |function: &_| {
+            let &crate::Function {
+                name: _,
+                arguments: _,
+                result: _,
```
jimblandy, line 233: The arguments and result have type handles.

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+                ref local_variables,
+                ref expressions,
+                named_expressions: _,
```
jimblandy, line 236: These have expression handles.

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+                body: _,
```
jimblandy, line 237: Since many kinds of statements have expression handles (`Statement::Emit`, `Statement::If`) and function handles (`Statement::Call`), we'll need a recursive function to walk the statement tree and check those handles.

src/valid/mod.rs
```diff
@@ -310,6 +313,8 @@ impl Validator {
         &mut self,
         module: &crate::Module,
     ) -> Result<ModuleInfo, WithSpan<ValidationError>> {
+        Self::validate_module_handles(module).map_err(|e| ValidationError::from(e).with_span())?;
```
jimblandy, line 316: I think this should go after the `reset` and `reset_types` calls, and be conditional on the `"validate"` feature. When that's turned off, panicking is just fine. :D

jimblandy (changes requested):
Okay, this looks great - just what we need. There are a few omissions that need to be addressed, and I had some suggestions.

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+            ref constants,
+            ref entry_points,
+            ref functions,
+            ref global_variables,
+            ref types,
+        } = module;
+
+        // NOTE: Types being first is important. All other forms of validation depend on this.
+        for (this_handle, type_) in types.iter() {
+            let &crate::Type { name: _, ref inner } = type_;
```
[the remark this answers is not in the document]
ErichDonGubler: I'm actually not sure about the balancing act WRT where to stop destructuring. I'm a fan of _destructuring all the things_, but it _does_ make quite a few humans recoil at the thought of “I have to change this _potentially every time_ I change this representation?” To me, though, it never _really_ causes significant problems; mostly unexpected compile failures that are easy, if not trivial, to resolve. LSP refactors + `rustc` make for a nice toolkit to quickly resolve compile errors, if even needed.

On the flip side, fuzzers are usually a (vital) last line of defense against open-ended input issues, but I also don't think the maintenance burden is a significant motivator for abdicating analytical defense-in-depth for this pass. However, I'm not a long-time maintainer of _any_ code like this, so I'll defer to your opinion here.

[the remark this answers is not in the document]
ErichDonGubler: Destructured s'more with d48c7133, BTW. 🙂

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+                | &crate::TypeInner::Matrix { .. }
+                | &crate::TypeInner::ValuePointer { .. }
+                | &crate::TypeInner::Atomic { .. }
+                | &crate::TypeInner::Image { .. }
+                | &crate::TypeInner::Sampler { .. } => (),
+                &crate::TypeInner::Pointer { base, .. } => {
+                    this_handle.check_dep(base)?;
+                }
+                &crate::TypeInner::Array { base, .. } => {
+                    this_handle.check_dep(base)?;
```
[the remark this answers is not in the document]
ErichDonGubler: RE: needing to do some low-key type checking here: it _does_ feel a little yucky to take on some overlap with type-checking, but I feel like that's the fundamental trade-off with having multiple passes, in general. I have implemented a fix with 468b72d5, taking the route of converting a crate-private error type to existing `ValidationError` variants (which is similar to your suggestion to implement `InvalidHandleError::InvalidArraySizeConstant`, but not quite the same); LMK what you think.

> The array size is a `Constant`, so that handle needs to be validated here.

Ah, so I missed the `BindingArray` ca—ah, and I see you've noted it already. 😅

src/valid/handles.rs
ErichDonGubler: Oh, oof, good eye. Fixed with 4c5e2099.

src/valid/handles.rs
ErichDonGubler: Updated as suggested with 555f4207.

src/valid/handles.rs
ErichDonGubler: Oh, oof, I feel silly for missing this. 😳 Resolved with 3fba607f05b3f958d45e6139453cbdb5240fd5e0.

src/valid/handles.rs
ErichDonGubler: Oof again. 😖 Resolved with 9ce5abba.

src/valid/handles.rs
ErichDonGubler: Yep, this was a pretty big omission on my part, sorry. 😓 Pushed 23637853 and 75654ac7. Whaddaya think?

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+                }
+                &crate::TypeInner::Struct { ref members, .. } => {
+                    this_handle.check_dep_iter(members.iter().map(|m| m.ty))?;
```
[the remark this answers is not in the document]
ErichDonGubler: I changed to destructure more; see also [this comment](https://github.com/gfx-rs/naga/pull/2090#discussion_r1047294339).

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+        let &crate::Module {
+            ref constants,
+            ref entry_points,
+            ref functions,
+            ref global_variables,
+            ref types,
+        } = module;
+
+        // NOTE: Types being first is important. All other forms of validation depend on this.
+        for (this_handle, type_) in types.iter() {
```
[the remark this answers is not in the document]
ErichDonGubler: Renamed as suggested with fe7e59cb.

src/valid/handles.rs
ErichDonGubler: Since you're using [a reference to another convo](https://github.com/gfx-rs/naga/pull/2090#discussion_r1044946235)...here's [a reference to my answer there](https://github.com/gfx-rs/naga/pull/2090#discussion_r1047294521). 😉

src/valid/mod.rs
ErichDonGubler: Changed as suggested in 4b1165fb. I feel silly for not considering the question of `cfg`'ing things, given the surrounding code. 😓

src/valid/handles.rs
ErichDonGubler: ~~Sure! Changed with 3412651ac5ccfbf5a1a5e5158cf0c00760fe1ff9.~~ EDIT: Reverted with 202d699a06c6607c57980d727a0bc299b8a77caf.

Actually, the inner `match` here is still something we'll want to re-use. We can inline the looping logic (which I did with 9ce5abba). After moving the closure to an associated function (see 23637853), I did end up re-using it with 75654ac7 (see also the convo we have about statements not getting checked [here](https://github.com/gfx-rs/naga/pull/2090#discussion_r1047296957)).

src/proc/mod.rs
```diff
@@ -97,11 +97,14 @@ impl super::TypeInner {
...
-    pub fn try_size(
-        &self,
-        constants: &super::Arena<super::Constant>,
-    ) -> Result<u32, crate::arena::BadHandle> {
-        Ok(match *self {
+    /// Get the size of this type.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the `constants` doesn't contain a referenced handle. This may not happen in
```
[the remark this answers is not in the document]
ErichDonGubler: Huh. I would have thought that it would be _less_ clear what error handling to expect from what is now the only public way to get the size of a type. 🤔

In general, I've found it more beneficial to document the conditions under which a public API could panic. You're saying that you don't think that that outweighs the maintenance cost, in this case?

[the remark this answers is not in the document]
ErichDonGubler: (OT, but still most relevant to this convo out of this whole PR: remove the `the` article with c533142a.)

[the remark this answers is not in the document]
ErichDonGubler: I've removed the `Panics` section with 65251d65.

ErichDonGubler:
Did some further `clippy` warning hunting with a02b02b8ca0a471c7ccc2c964a521920b5794b0f, 00f4f6b4, ce6f85e8, and 64932357.

src/valid/handles.rs
```diff
@@ -0,0 +1,337 @@
...
+    where
+        T: Eq + Hash,
+    {
+        arena.check_contains_handle(self)?;
+        Ok(self)
+    }
+
+    /// Check that `depends_on` was constructed before `self` by comparing handle indices.
+    ///
+    /// If `self` describes a valid value (i.e., it has been validated using
```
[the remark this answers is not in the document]
ErichDonGubler: Fixed as suggested with e7cc0df4.

ErichDonGubler:
@jimblandy: Once you've had a chance to look through the round of fixes I've pushed up here, LMK, and I'll squash things down into a sane set of commits again. Everything squashes down cleanly, so 😤.

src/valid/handles.rs
```diff
@@ -0,0 +1,637 @@
...
+    ForwardDependency(#[from] FwdDepError),
+    #[error(transparent)]
+    BadRange(#[from] BadRangeError),
+}
+
+#[derive(Clone, Debug)]
+pub(super) struct InvalidArraySizeError {
+    pub(super) ty: Handle<crate::Type>, // NOTE: type erased, probably wasn't `()` to begin with!
+    pub(super) ty_name: Option<String>,
+    pub(super) depends_on: Handle<crate::Constant>, // NOTE: type erased, probably wasn't `()` to begin with!
```
jimblandy, line 536: These comments about type-erased handles don't apply here.

src/valid/handles.rs
```diff
@@ -0,0 +1,637 @@
...
+}
+
+#[derive(Clone, Debug, thiserror::Error)]
+#[error("{subject:?} depends on {depends_on:?}, which has not been processed yet")]
+pub struct FwdDepError {
+    subject: Handle<()>,    // NOTE: type erased, probably wasn't `()` to begin with!
+    depends_on: Handle<()>, // NOTE: type erased, probably wasn't `()` to begin with!
```
jimblandy, line 543: Suggest:

```suggestion
    // This error is used for many `Handle` types, but there's no point in making this generic,
    // so we just flatten them all to `Handle<()>` here.
    subject: Handle<()>,
    depends_on: Handle<()>,
```

src/valid/handles.rs
```diff
@@ -0,0 +1,637 @@
...
+}
+
+// NOTE: Keep this diagnostic in sync with that of [`BadHandle`].
+#[derive(Clone, Debug, thiserror::Error)]
+#[error("Handle range {range:?} of {kind} is either not present, or inaccessible yet")]
+pub struct BadRangeError {
+    kind: &'static str,
+    range: Range<()>, // NOTE: type erased, probably wasn't `()` to begin with!
```
jimblandy, line 551: similar suggestion

jimblandy (changes requested):
When our CI runs:
```
cargo check -p naga --no-default-features
```
it gets a lot of warnings about unused functions, etc. in `src/valid/handles.rs`. We should put `#[cfg(feature = "validate")]` in the appropriate places (perhaps on the entire `valid::handles` module?) to make the non-validating configuration compile cleanly again.

src/proc/mod.rs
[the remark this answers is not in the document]
jimblandy: > In general, I've found it more beneficial to document the conditions under which a public API could panic. You're saying that you don't think that that outweighs the maintenance cost, in this case?

That's a fair question, and it's a point that comes up in many contexts, so let me explain my thinking:

The information content of a message is the -log₂ of the probability of its being received: for example, a message that is received with probability 0.25 conveys two bits of information. But this means that a message that is almost always received has a probability approaching 1, and thus conveys almost zero bits. Hearing "good morning" from someone who always says that is nice, but it doesn't tell you much.

Almost every public Naga API that operates on a module or a part of a module is going to access arenas, and thus will panic in the presence of invalid handles. If we were to place these panic notices correctly, almost every public API in Naga would have a "Panics" section saying the same thing. That would be a message with a probability approaching one, and thus one that conveys little information, and thus one that might be best omitted, since the reader has a limited attention span.

However, if we document this general approach in higher-level documentation, like the lib.rs comments on `Arena`s, that's being read by users who are getting an overview of Naga and are new to its API. Such readers are probably not ready to make many assumptions about Naga's behavior, so explaining that validation checks for bad handles, but subsequent stages assume these checks have been done and will panic (but not exhibit UB) if presented with bad handles, is not at all expected, and would convey many bits. So putting the message in a place where it has the most effect on the readers' understanding improves its information content.

The Rust standard library documentation for slice methods has a "Panics" section on every method that takes an index. For example, `split_at(&self, mid: usize)` says, "Panics if mid > len." I *think* this situation is different from Naga's, because there are various ways to respond to an index that is out of bounds. You could truncate (like the `windows` method when `size` is too large), it could be UB (like `split_unchecked`), etc. But in Naga, a bad handle is always meaningless. So there's really no other possible response.

In any case, if we're going to mention the panics, then I want to do it systematically - and that's not for this PR.

jimblandy (commented):
I'll need to get back to the rest of this tomorrow, but here are some comments so far.

ErichDonGubler:
> When our CI runs:
> 
> ```
> cargo check -p naga --no-default-features
> ```
> 
> it gets a lot of warnings about unused functions, etc. in `src/valid/handles.rs`. We should put `#[cfg(feature = "validate")]` in the appropriate places (perhaps on the entire `valid::handles` module?) to make the non-validating configuration compile cleanly again.

Silenced warnings with 2fab863a.

ErichDonGubler:
@jimblandy:

> When our CI runs … it gets a lot of warnings about unused functions, etc. in `src/valid/handles.rs`. We should put `#[cfg(feature = "validate")]` in the appropriate places (perhaps on the entire `valid::handles` module?) to make the non-validating configuration compile cleanly again.

Silenced warnings with 113957aa.

src/valid/handles.rs
```diff
@@ -0,0 +1,637 @@
...
+#[derive(Clone, Debug)]
+pub(super) struct InvalidArraySizeError {
+    pub(super) ty: Handle<crate::Type>, // NOTE: type erased, probably wasn't `()` to begin with!
+    pub(super) ty_name: Option<String>,
+    pub(super) depends_on: Handle<crate::Constant>, // NOTE: type erased, probably wasn't `()` to begin with!
+}
+
+#[derive(Clone, Debug, thiserror::Error)]
+#[error("{subject:?} depends on {depends_on:?}, which has not been processed yet")]
+pub struct FwdDepError {
```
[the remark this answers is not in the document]
ErichDonGubler: How's 7f0faf9d look?

src/valid/handles.rs
```diff
@@ -0,0 +1,637 @@
...
+    fn from(source: BadRangeError) -> Self {
+        Self::InvalidHandle(source.into())
+    }
+}
+
+impl From<InvalidArraySizeError> for HandleValidationError {
+    fn from(source: InvalidArraySizeError) -> Self {
+        Self::InvalidArraySize(source)
+    }
+}
```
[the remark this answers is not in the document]
ErichDonGubler: N.B. that the `From` conversions here may not be necessary with some well-placed `map_err`; I hadn't tried that, but that could cut the trait implementations down, in case that changes your calculus.

WRT moving to `ValidationError`, I'm not feeling picky, I guess. Moved to "just" return a `ValidationError` with df755c36.

src/valid/handles.rs
ErichDonGubler: Fixed with 53eca848.

src/valid/handles.rs
ErichDonGubler: Fixed with 53eca848.

src/valid/handles.rs
ErichDonGubler: Fixed with 53eca848.

ErichDonGubler:
I've autosquashed all reviewed content into a few (hopefully nice) commits. Now that I've rebased onto `master`, I'm getting unrelated [CI failures](https://github.com/gfx-rs/naga/actions/runs/3707003285/jobs/6282899141). I've submitted https://github.com/gfx-rs/naga/pull/2176 to resolve them.

src/valid/handles.rs
```diff
@@ -0,0 +1,631 @@
...
+                &crate::TypeInner::Struct {
+                    ref members,
+                    span: _,
+                } => {
+                    this_handle.check_dep_iter(members.iter().map(|m| m.ty))?;
+                }
+                &crate::TypeInner::BindingArray { base, size } => {
+                    this_handle.check_dep(base)?;
+                    validate_array_size(size)?;
+                }
```
jimblandy, line 97: Suggestion: combine this `match` arm with with `TypeInner::Array` via an `|` pattern.

src/valid/handles.rs
```diff
@@ -0,0 +1,631 @@
...
+                                source: TypeError::InvalidArraySizeConstant(constant),
+                            });
+                        }
+                    }
+                    crate::ArraySize::Dynamic => (),
+                };
+                Ok(this_handle)
+            };
+
+            match inner {
```
jimblandy, line 69: It's Naga style to match `*inner`, drop the references from all the patterns, and use `ref` as appropriate to avoid moves. This is Clippy lint [match_ref_pats](https://rust-lang.github.io/rust-clippy/master/index.html#match_ref_pats), so I don't know why Clippy isn't complaining about this.

src/valid/handles.rs
```diff
@@ -0,0 +1,631 @@
...
+        let validate_block = |block| Self::validate_block_handles(block, expressions, functions);
+        let validate_expr = |handle| Self::validate_expression_handle(handle, expressions);
+        let validate_expr_opt = |handle_opt| {
+            if let Some(handle) = handle_opt {
+                validate_expr(handle)?;
+            }
+            Ok(())
+        };
+
+        block.iter().try_for_each(|stmt| match stmt {
```
jimblandy, line 385: Similarly: please `match *stmt` and simplify all the patterns.

src/valid/handles.rs
```diff
@@ -0,0 +1,631 @@
...
+                handle.check_dep(pointer)?;
+            }
+            crate::Expression::ImageSample {
+                image,
+                sampler,
+                gather: _,
+                coordinate,
+                array_index,
+                offset,
+                level: _,
```
jimblandy, line 273: The `level` field is a `SampleLevel`, which can have handles.

jimblandy (changes requested):

src/valid/handles.rs
ErichDonGubler: If `clippy` doesn't or can't enforce this style…is it worth enforcing manually? 🤔

Changed as suggested with af211507.

src/valid/handles.rs
ErichDonGubler: Changed as suggested with 62f43a66.

src/valid/handles.rs
ErichDonGubler: Gonna need to make some pizza…with all this d'oh. 😩

Resolved with e1229f1b.

src/valid/handles.rs
ErichDonGubler: Changed as suggested with b43468a6.

src/arena.rs
```diff
@@ -118,11 +127,22 @@ impl<T> Handle<T> {
 )]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub struct Range<T> {
-    inner: ops::Range<u32>,
+    pub(crate) inner: ops::Range<u32>,
```
jimblandy, line 130: Since we've already added a `check_contains_handle` method to `Arena`, let's also add a `check_contains_range` method, let `valid::handles`' `Range::check_valid_for` method can just call that, and let this stay private.

src/valid/expression.rs
```diff
@@ -132,15 +131,18 @@ struct ExpressionTypeResolver<'a> {
...
+    type Output = crate::TypeInner;
+
+    #[allow(clippy::panic)]
+    fn index(&self, handle: Handle<crate::Expression>) -> &Self::Output {
         if handle < self.root {
-            Ok(self.info[handle].ty.inner_with(self.types))
+            self.info[handle].ty.inner_with(self.types)
         } else {
-            Err(ExpressionError::ForwardDependency(handle))
+            panic!(
```
jimblandy, line 142:
```suggestion
            // `Validator::validate_module_handles` should have caught this.
            panic!(
```

jimblandy (changes requested):
Looks great! Just two small changes, and then I think we're ready to go.

src/valid/handles.rs
jimblandy: It's all good, that's what reviews are for. (Pizza is tasty.)

jimblandy (approved):

PR #2099, opened 2022-10-23
Fix incorrect atomic bounds check on metal back-end

### LINKED ISSUE #1848: [msl-out] atomic functions try to take reference of rvalue
In some WGSL shader code in bevy-hikari:
```rust
struct VoxelBuffer {
    data: array<atomic<u32>, 16777216>;
}
...
[[group(3), binding(2)]]
var<storage, read_write> voxel_buffer: VoxelBuffer;
...
    var packed = pack4x8unorm(vec4<f32>(color.rgb, color.a / 255.0));

    let index = spatial_index(in.world_position.xyz);
    let voxel = &voxel_buffer.data[linear_index(index)];
    atomicMax(voxel, packed);
```
So that is basically an `atomixMac(ptr<private, atomic<u32>, read_write>, ref<function, u32, read_write>)`. The problem on metal is that this produces:
```msl
2022-04-17T15:18:10.341677Z ERROR wgpu::backend::direct: Handling wgpu errors as fatal by default    
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
      note: label = `opaque_mesh_pipeline`
    Internal error in FRAGMENT shader: Metal: program_source:302:15: warning: unused variable '_e89'
        float _e89 = compute_roughness(_e88);
              ^
program_source:374:58: error: cannot take the address of an rvalue of type 'metal::uint' (aka 'unsigned int')
    metal::uint _e199 = metal::atomic_fetch_max_explicit(&metal::uint(_e196) < 16777216 ? voxel_buffer.data.inner[_e196] : DefaultConstructible(), _e198, metal::memory_order_relaxed);
                                                         ^~~~~~~~~~~~~~~~~~~
```
So the _e196, which is presumably a uint, is wrapped in metal::uint() which makes it an rvalue, and then the reference of it is attempted to be taken. That is, I believe, not what is intended here.

wgpu-0.12.0
naga-0.8.5

raphlinus:
Generalize put_atomic_fetch to handle `exchange` as well, rather than special-cased code which didn't do the bounds check (the check handling as fixed in #1703 but only for the fetch cases, exchange was skipped).

Fixes #1848

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -1188,6 +1188,17 @@ impl<W: Write> Writer<W> {
         value: Handle<crate::Expression>,
         context: &ExpressionContext,
     ) -> BackendResult {
+        self.put_atomic_operation(pointer, "fetch_", key, value, context)
+    }
+
+    fn put_atomic_operation(
+        &mut self,
+        pointer: Handle<crate::Expression>,
+        key1: &str,
+        key2: &str,
+        value: Handle<crate::Expression>,
+        context: &ExpressionContext,
+    ) -> BackendResult {
         // If the pointer we're passing to the atomic operation needs to be conditional
         // for `ReadZeroSkipWrite`, the condition needs to *surround* the atomic op, and
         // the pointer operand should be unchecked.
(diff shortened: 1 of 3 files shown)


jimblandy (changes requested):
Thank you very much!

Could you also add a test case for this to `tests/in/bounds-check-zero-atomic.wgsl`?

raphlinus:
Test added. I also manually verified that it works in the compute_shader_101 testbed by doing cargo patch.

I'm also willing to implement `atomicCompareExchangeWeak`, but it's a bigger code change, as it needs to construct a compare_exchange_result. I'm not 100% sure whether returning false or true is a better choice - in the common case where there's a retry loop, the former will cause infinite loops, which is possibly more useful for development to identify a problem, but also more annoying for users that bump into it.

jimblandy:
> I'm also willing to implement atomicCompareExchangeWeak

Fantastic. But yes, let's definitely address that separately.

PR #2102, opened 2022-10-25
build(cli): use conventional `bin` path for `naga-cli`

ErichDonGubler:


cwfitzgerald:
Any particular reason for this?

ErichDonGubler:
@cwfitzgerald: Good question! Context: It was an easy change, so I thought filing a PR first would be more interesting for discussion than trying another avenue for consensus first. This is mostly a bikeshed/yak shave, and I'd feel no sadness if this got closed on basis of everybody being comfortable with current configuration, or just waited for discussion.

The reasoning for this change is philosophical. I think it's more consistent with DX in the overall Rust ecosystem; that is, using the `bin` folder embraces Cargo's convention-over-configuration approach more fully. It keeps the naming of `naga-cli`'s output local to the binary's entry point. As a Rustacean of several years, I generally expect the vast majority of use cases with a single `bin` target to be covered by `src/main.rs` and/or the `src/bin/` folder, since there's no need for granularity with settings like `required-features`.

teoxoy:
According to https://doc.rust-lang.org/cargo/guide/project-layout.html, `src/main.rs` should be the location of the default executable and work without the extra config in `Cargo.toml`. Maybe we should only remove the extra config in `Cargo.toml`?

cwfitzgerald:
src/main.rs will end up with an executable named `naga-cli`, not `naga`, which we want.

teoxoy:
Ah, I see. Because the name of the package is `naga-cli`.

I'm neutral on this change.

jimblandy (approved):
This seems to have no effect on the visible behavior, since the entry point file is renamed to `src/bin/naga.rs` - it just uses file names instead of `Config.toml` settings. Seems good to me.

PR #2103, opened 2022-10-26
Use `source` member name for source errors

ErichDonGubler:
This lets us omit a `#[source]` annotation in the case of named members. Most other popular crates with a `derive` interface (like [`snafu`](https://docs.rs/snafu/0.7.3/snafu/index.html), [`derive-more`](https://jeltef.github.io/derive_more/derive_more/error.html)) use this convention, so it's pretty portable, overall.

Also, minor "bonus" refactor re-rolling a `for` loop in the `wgsl` backend. 😅 Reviewing commit-by-commit makes this easy to see.


teoxoy (approved):
LGTM

PR #2111, opened 2022-11-03
Add support for zero-initializing workgroup memory

### LINKED ISSUE #53: [meta] shader transformations
This is a list of generic shader transformations that we need to do for WebGPU:
- [x] out-of-bound handling of accessing arrays in uniform and storage buffers. The index must be either clamped, or zeroes returned.
- [x] out-of-bound handling of accessing storage textures.
- [ ] zero-initialization of threadgroup memory in compute shaders

teoxoy:
closes #53

Possible future optimizations:
- spread out initialization across workgroup (most useful for arrays)
- detect user initialization to avoid double initialization


src/back/msl/writer.rs
```diff
@@ -3901,6 +3881,219 @@ impl<W: Write> Writer<W> {
...
+            } else {
+                write!(self.out, "{}", level)?;
+                access_stack.write(&mut self.out, &self.names)?;
+                writeln!(self.out, " = {{}};")?;
+            }
+
+            Ok(())
+        }
+
+        // TODO: we do extra work here, replace with a check of TypeFlags::CONSTRUCTIBLE
```
teoxoy, line 4072: I'd like some thoughts on this. Currently don't retain the `TypeFlags` after doing the validation.

src/back/msl/writer.rs
```diff
@@ -3901,6 +3881,218 @@ impl<W: Write> Writer<W> {
...
+                write!(self.out, "{}", level)?;
+                access_stack.write(&mut self.out, &self.names)?;
+                writeln!(self.out, " = {{}};")?;
+            }
+
+            Ok(())
+        }
+    }
+
+    // TODO: we do extra work here, replace with a check of TypeFlags::CONSTRUCTIBLE
```
teoxoy, line 4073: I'd like some thoughts on this. Currently don't retain the `TypeFlags` after doing the validation.

src/back/glsl/mod.rs
```diff
@@ -1487,6 +1492,41 @@ impl<'a, W: Write> Writer<'a, W> {
...
+        for (handle, var) in vars {
+            let name = &self.names[&NameKey::GlobalVariable(handle)];
+            write!(self.out, "{}{} = ", back::INDENT, name)?;
+            self.write_zero_init_value(var.ty)?;
+            writeln!(self.out, ";")?;
+            write_barrier = true;
+        }
+
+        if write_barrier {
+            self.write_barrier(crate::Barrier::WORK_GROUP, back::Level(1))?;
```
ErichDonGubler, line 1524: question: Where can I learn more about this?

TODO: try to find out about this myself first.

src/back/glsl/mod.rs
```diff
@@ -3562,6 +3592,19 @@ impl<'a, W: Write> Writer<'a, W> {
...
+        if flags.contains(crate::Barrier::STORAGE) {
+            writeln!(self.out, "{}memoryBarrierBuffer();", level)?;
+        }
+        if flags.contains(crate::Barrier::WORK_GROUP) {
+            writeln!(self.out, "{}memoryBarrierShared();", level)?;
+        }
+        writeln!(self.out, "{}barrier();", level)?;
+        Ok(())
+    }
+
```
ErichDonGubler, line 3607: nitpick(non-blocking): It would have been nice to have this refactored code be put into a separate commit.

src/back/spv/writer.rs
```diff
@@ -65,6 +65,7 @@ impl Writer {
             annotations: vec![],
             flags: options.flags,
             bounds_check_policies: options.bounds_check_policies,
+            zero_initialize_workgroup_memory: options.zero_initialize_workgroup_memory,
```
ErichDonGubler, line 68: thought(non-blocking): I feel like using pattern destructuring for data forwarding would be a really good fit for this sort of logic in `Self::new` and `Self::reset`. I'm the type that doesn't trust myself without that sort of safety net, tho, so YMMV. 😅

ErichDonGubler (commented):
Not finished reviewing yet, but so far, this LGTM. I'll see if I can finish this before end-of-week.

nitpick(non-blocking): 1k LoC in a single commit is a _lot_. I do think there's some opportunity for breaking this up, but fortunately, the candidates for splitting commits look easy (which tends to mean that the impact isn't high, thankfully). I'll try to call them out where I can.

question: Isn't this sort of new functionality the sort of thing we'd want a `CHANGELOG` entry for? I see that we don't currently have a PR template encouraging this, which might be something I can do a la `wgpu`.

note: RE: the `Path` warning: this warning has been fixed with 01972463, which you should catch with a rebase. 🙌🏻

jimblandy:
> Possible future optimizations:
> 
>     * spread out initialization across workgroup (most useful for arrays)
> 
>     * detect user initialization to avoid double initialization

These are good ideas - could you file an issue for each?

src/back/msl/writer.rs
```diff
@@ -3901,6 +3881,218 @@ impl<W: Write> Writer<W> {
...
+                        for i in 0..count {
+                            access_stack.enter(Access::Index(i), |access_stack| {
+                                self.write_workgroup_variable_initialization(
+                                    module,
+                                    base,
+                                    access_stack,
+                                    level,
+                                )
+                            })?;
+                        }
```
jimblandy, line 4044: We should just emit an actual loop in MSL, not unroll it.

jimblandy (changes requested):
I haven't looked at the SPIR-V back end yet, but I do have one change I want to see, so I'll just post what I have now.

I'd also be happy to see a `TypeFlags`-based revision, if you want.

jimblandy:
Sorry, some of those comments were intended as replies to Erich's questions, but GitHub didn't see it that way so they're kind of out-of-context.

src/back/glsl/mod.rs
```diff
@@ -1487,6 +1492,41 @@ impl<'a, W: Write> Writer<'a, W> {
         Ok(())
     }
 
+    fn write_workgroup_variables_initialization(
+        &mut self,
+        ctx: &back::FunctionCtx,
+    ) -> BackendResult {
+        if !self.options.zero_initialize_workgroup_memory
+            || !ctx.ty.is_compute_entry_point(self.module)
```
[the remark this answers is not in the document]
teoxoy: > comment: Huh, I actually don't know why whether this is a compute shader matters here. 🤔

Variables in the workgroup address space can only be used by compute entry points.

> it might be a bit clearer to pull this condition out and wrap its negation around the call to `write_workgroup_variables_initialization`.

Will do!

src/back/spv/writer.rs
teoxoy: I don't use destructuring much as I like seeing where the var came from at first glance.
@jimblandy thoughts?

teoxoy:
@ErichDonGubler

> nitpick(non-blocking): 1k LoC in a single commit is a lot. I do think there's some opportunity for breaking this up, but fortunately, the candidates for splitting commits look easy (which tends to mean that the impact isn't high, thankfully). I'll try to call them out where I can.

Most changes seem to come from the test snapshots. Will move the barrier code refactor to a separate commit.

> question: Isn't this sort of new functionality the sort of thing we'd want a `CHANGELOG` entry for? I see that we don't currently have a PR template encouraging this, which might be something I can do a la `wgpu`.

Tbh, I prefer adding the changelog entries when we do a release since I think it results in higher quality changelogs but what could also have the same effect is asking authors to add a changelog entry (same as wgpu) and doing a last pass before a release to make sure the changelog is cohesive.

src/back/msl/writer.rs
```diff
@@ -3901,6 +3881,218 @@ impl<W: Write> Writer<W> {
...
+            crate::TypeInner::Array { base, .. } => does_ty_contain_atomic(module, base),
+            crate::TypeInner::Struct { ref members, .. } => {
+                for member in members {
+                    if does_ty_contain_atomic(module, member.ty)? {
+                        return Ok(true);
+                    }
+                }
+                Ok(false)
+            }
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
teoxoy: We don't guarantee that the backends don't panic if they are given invalid `Module`s though.
We have a bunch of `unreachable!()`s and `unwrap()`s in the backends that will cause a panic if the `Module` is invalid.

Validation should be run at least once on a given `Module` and can be omitted later if it was already validated.

src/back/glsl/mod.rs
```diff
@@ -3499,7 +3529,7 @@ impl<'a, W: Write> Writer<'a, W> {
     fn write_zero_init_value(&mut self, ty: Handle<crate::Type>) -> BackendResult {
         let inner = &self.module.types[ty].inner;
         match *inner {
-            TypeInner::Scalar { kind, .. } => {
+            TypeInner::Scalar { kind, .. } | TypeInner::Atomic { kind, .. } => {
```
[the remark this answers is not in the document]
teoxoy: GLSL has no concept of atomics at the type level and should be treated as scalars.
This was needed because workgroup variables might contain atomics which should be initialized just as scalars.

jimblandy (commented):
This changed the API for backends, so it needs a CHANGELOG.md entry.

Don't all these writes need to be limited to a single invocation? Right now we're writing to all this storage from every invocation, and that's going to be a race and a waste.

src/back/spv/writer.rs
jimblandy: If I'm going to talk about things 'paying off', then I should say what the cost is: I think it takes a bit longer to read the destructuring version, there are more variables around that could be mixed with other things, and as Teo said, it's less obvious, when looking at a use, what its origin was.

That is, when I see `options.bounds_check_policies`, I know exactly where it came from, but when I see just `bounds_check_policies`, it could be an argument, it could have been computed, etc. and I have to guess or go find out.

jimblandy (changes requested):
I think the only remaining question here is the one I raised above: should we restrict initialization to a single invocation?

tests/out/glsl/workgroup-var-init.main.Compute.glsl
```diff
@@ -0,0 +1,26 @@
...
...0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u, 0u), 0, int[8](0, 0, 0, 0, 0, 0, 0, 0));
```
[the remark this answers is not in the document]
teoxoy: Yeah, I tried to find a way to zero init more succinctly in GLSL but couldn't find anything. Tint seems to do the same.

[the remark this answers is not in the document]
jimblandy: Oh well...

teoxoy:
> This changed the API for backends, so it needs a CHANGELOG.md entry.

Yeah, the question is if we want to require updates to the changelog as part of PRs or do it ourselves before a release.

> Don't all these writes need to be limited to a single invocation? Right now we're writing to all this storage from every invocation, and that's going to be a race and a waste.

Indeed, pushed a new commit.

jimblandy (changes requested):
Still need to process the SPIR-V changes, but I did find something in the MSL.

teoxoy:
@jimblandy I addressed the latest comments.

tests/in/padding.param.ron
```diff
@@ -18,5 +18,6 @@
 		inline_samplers: [],
 		spirv_cross_compatibility: false,
 		fake_missing_bindings: false,
+		zero_initialize_workgroup_memory: true,
```
jimblandy, line 21: If you prefer, rather than editing all these `.ron` files, you could use the [`#serde(default = "path")`](https://serde.rs/field-attrs.html#default--path) attribute. I think the value has to be a function name - maybe add it to `back/mod.rs`?

jimblandy (changes requested):
One problem in the validator changes. I'll look at this again first thing tomorrow morning.

src/valid/type.rs
```diff
@@ -463,7 +465,8 @@ impl super::Validator {
                     }
                 };
 
-                let base_mask = TypeFlags::COPY | TypeFlags::HOST_SHAREABLE;
+                let base_mask =
+                    TypeFlags::COPY | TypeFlags::HOST_SHAREABLE | TypeFlags::CONSTRUCTIBLE;
```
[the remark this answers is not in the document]
teoxoy: Good catch, I adopted the same style of just `base & mask` from the struct code. I think it makes it easier to reason about.

tests/in/padding.param.ron
teoxoy: We don't seem to do it for any of the other members, not sure why but wanted to keep things consistent.
I think if we want to introduce defaults, we should do so for all members and override the defaults in the ron files.

jimblandy: Maybe best left to a follow-up PR.

src/back/spv/writer.rs
```diff
@@ -1118,6 +1199,135 @@ impl Writer {
...
+                let pointer_type_id = self.get_uint3_pointer_type_id(class);
+
+                Instruction::variable(pointer_type_id, varying_id, class, None)
+                    .to_words(&mut self.logical_layout.declarations);
+
+                self.decorate(
+                    varying_id,
+                    spirv::Decoration::BuiltIn,
+                    &[spirv::BuiltIn::GlobalInvocationId as u32],
+                );
```
jimblandy: I understand why, but it's a shame that we have to replicate code like this. This is an example of a case where lowering the Naga IR to some intermediate representation before generating SPIR-V would simplify things. But that shouldn't be addressed in this PR.

jimblandy (approved):
Looks good!

PR #2112, opened 2022-11-03
Fix 1.65 clippy lints

teoxoy:


src/back/spv/image.rs
```diff
@@ -1017,7 +1017,7 @@ impl<'w> BlockContext<'w> {
                     Id::D3 => 3,
                 };
                 let extended_size_type_id = {
-                    let array_coords = if arrayed { 1 } else { 0 };
+                    let array_coords = usize::from(arrayed);
```
ErichDonGubler, line 1020: thought: TIL about `impl From<bool> for {integer}`. 👀

ErichDonGubler (approved):
LGTM! Tho I have no merging rights. @jimblandy?

src/back/spv/image.rs
cwfitzgerald: TIL also - I always just blasted to integer via `as`

PR #2115, opened 2022-11-04
Error if the uniformity requirements for a barrier aren't met

DJMcNab:
I think this is the only other case in `process_block` where a uniformity requirement can be created without being validated.

### DIFF
--- a/src/valid/analyzer.rs
+++ b/src/valid/analyzer.rs
@@ -780,13 +780,25 @@ impl FunctionInfo {
                         ExitFlags::empty()
                     },
                 },
-                S::Barrier(_) => FunctionUniformity {
-                    result: Uniformity {
-                        non_uniform_result: None,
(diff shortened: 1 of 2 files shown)


teoxoy:
Sorry for the late reply, we were putting off any uniformity changes/updates because the spec was still in flux.

I for one am not too familiar with the existing code and was planning to get more familiar with it once we update our uniformity checks to be in line with the spec.

I'll give this a proper review once I get to that (if nobody else does in the meantime).

jimblandy (approved):
This looks good - thank you!

jimblandy:
Context: we're going to have to entirely replace `valid/analyzer.rs`, because the spec has something more accurate (and more complex). But this is simply a straightforward extension of what we're already doing in `S::Emit`, so I think it's fine.

PR #2118, opened 2022-11-04
DO NOT MERGE: Test: can we skip an `rspirv` to fix other backends' negation?

ErichDonGubler:
Intended to support https://github.com/gfx-rs/naga/pull/2087. Do not merge!


ErichDonGubler:
Test successful! Looks like CI doesn't complain here.

PR #2132, opened 2022-11-20
Document `EarlyDepthTest` and `ConservativeDepth` syntax in WGSL

coreh:
👋 Hey there,

Was trying to figure out how to explicitly enable early depth tests in WebGPU/WGSL shaders, and noticed there was documentation on the syntax for `EarlyDepthTest` and `ConservativeDepth` options for GLSL/HLSL/SPIR-V, but not for WGSL. 

Had to do some digging through the parser code to figure out how to do it, so this PR adds documentation, so that hopefully others can find it more easily in the future.

```wgsl
@fragment
@early_depth_test
fn fragment(in: FragmentInput) -> @location(0) vec4<f32> {
  // ....
}
```

_OR_

```wgsl
@fragment
@early_depth_test(greater_equal/less_equal/unchanged) // pick one mode
fn fragment(in: FragmentInput) -> @location(0) vec4<f32> {
  // ....
}
```


teoxoy (approved):
LGTM, thanks!

Issue #2135, opened 2022-11-26
Issues with textureGather on texture_2d<u32>


JMS55:
Edit: Can't reproduce on naga master. However, now a different validation error is produced in wgpu:
```
Caused by:
    In Device::create_compute_pipeline
      note: label = `ambient_occlusion_denoise_pipeline`
    error matching shader requirements against the pipeline
    unable to filter the texture (ResourceBinding { group: 0, binding: 1 }) by the sampler (ResourceBinding { group: 1, binding: 0 })
    integer textures can't be sampled
```

Caused by the following match ignoring the gather field in naga: https://github.com/gfx-rs/naga/blob/df4e40327c280a038365fb3d4ec839353f06eed3/src/valid/analyzer.rs#L571-L594

---

According to the spec, this should be a valid operation. However, I get the following error in my project:

```
var depth_differences: texture_2d<u32>;

error: Entry point denoise at Compute is invalid
    ┌─ wgsl:106:18
    │
106 │     let edges0 = textureGather(0, depth_differences, point_clamp_sampler, uv);
    │                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ naga::Expression [17]
    │
    = Expression [17] is invalid
    = Unable to operate on image class Sampled { kind: Uint, multi: false }
```

Taking a look at naga's source code, it looks like this is an `InvalidImageClass` Err. There's only 3 places it's created, of which the following seems most likely:

https://github.com/gfx-rs/naga/blob/2a11d830bb8a4965758cab787196322de4f24e19/src/valid/expression.rs#L382-L389

It looks like originally this match statement assumes a float texture only, which is a valid assumption when this was originally used for only regular samples, but when gathers were added, this assumption was broken.

JMS55:
@teoxoy this commit https://github.com/gfx-rs/naga/pull/2138/commits/9ddf38e90d4fb522e806c0155a0f3cacd6b6763e turned out to be necessary for my shader https://github.com/JMS55/bevy/blob/gtao/crates/bevy_pbr/src/ssao/gtao.wgsl#L41. The tests I wrote just didn't catch it :/.

teoxoy:
What is the error exactly?

JMS55:
Ah sorry, the error is:
```
Caused by:
    In Device::create_compute_pipeline
      note: label = `ssao_denoise_pipeline`
[...]

PR #2138, opened 2022-12-01
Fix textureGather on texture_2d<u32/i32>

JMS55:
Fixes https://github.com/gfx-rs/naga/issues/2135.

I'm not sure if I did this correctly, so please double check if this is right. It's enough to let my shader pass validation, however.

### DIFF
--- a/src/valid/analyzer.rs
+++ b/src/valid/analyzer.rs
@@ -571,7 +571,7 @@ impl FunctionInfo {
             E::ImageSample {
                 image,
                 sampler,
-                gather: _,
+                gather,
                 coordinate,
                 array_index,
                 offset: _,
@@ -581,15 +581,17 @@ impl FunctionInfo {
                 let image_storage = GlobalOrArgument::from_expression(expression_arena, image)?;
                 let sampler_storage = GlobalOrArgument::from_expression(expression_arena, sampler)?;
 
-                match (image_storage, sampler_storage) {
-                    (GlobalOrArgument::Global(image), GlobalOrArgument::Global(sampler)) => {
-                        self.sampling_set.insert(SamplingKey { image, sampler });
-                    }
-                    _ => {
-                        self.sampling.insert(Sampling {
-                            image: image_storage,
-                            sampler: sampler_storage,
-                        });
+                if gather.is_none() {
+                    match (image_storage, sampler_storage) {
+                        (GlobalOrArgument::Global(image), GlobalOrArgument::Global(sampler)) => {
+                            self.sampling_set.insert(SamplingKey { image, sampler });
+                        }
+                        _ => {
+                            self.sampling.insert(Sampling {
+                                image: image_storage,
+                                sampler: sampler_storage,
+                            });
+                        }
                     }
                 }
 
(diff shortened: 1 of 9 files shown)


teoxoy:
I'd like to see some lines added in the `tests/in/image.wgsl` test snapshot that covers those overloads.

JMS55:
Added some tests.

src/valid/analyzer.rs
```diff
@@ -581,15 +581,17 @@ impl FunctionInfo {
...
+                    match (image_storage, sampler_storage) {
+                        (GlobalOrArgument::Global(image), GlobalOrArgument::Global(sampler)) => {
+                            self.sampling_set.insert(SamplingKey { image, sampler });
+                        }
+                        _ => {
+                            self.sampling.insert(Sampling {
+                                image: image_storage,
+                                sampler: sampler_storage,
+                            });
+                        }
```
teoxoy, line 594: I don't see why we'd ignore gather operations here.

tests/in/image.wgsl
```diff
@@ -71,9 +75,7 @@ fn queries() -> @builtin(position) vec4<f32> {
     let dim_3d_lod = textureDimensions(image_3d, 1);
     let dim_2s_ms = textureDimensions(image_aa);
 
-    let sum = dim_1d + dim_2d.y + dim_2d_lod.y + dim_2d_array.y + dim_2d_array_lod.y +
-        dim_cube.y + dim_cube_lod.y + dim_cube_array.y + dim_cube_array_lod.y +
-        dim_3d.z + dim_3d_lod.z;
+    let sum = dim_1d + dim_2d.y + dim_2d_lod.y + dim_2d_array.y + dim_2d_array_lod.y + dim_cube.y + dim_cube_lod.y + dim_cube_array.y + dim_cube_array_lod.y + dim_3d.z + dim_3d_lod.z;
```
teoxoy, line 78: This looks like a stylistic change. Let's stick to what we had before.

src/valid/expression.rs
```diff
@@ -384,6 +384,14 @@ impl super::Validator {
...
                         multi: false,
                     } => false,
+                    crate::ImageClass::Sampled {
+                        kind: crate::ScalarKind::Uint,
+                        multi: false,
+                    } => false,
+                    crate::ImageClass::Sampled {
+                        kind: crate::ScalarKind::Sint,
+                        multi: false,
+                    } => false,
```
teoxoy, line 394:
```suggestion
                    crate::ImageClass::Sampled {
                        kind: crate::ScalarKind::Uint | crate::ScalarKind::Sint,
                        multi: false,
                    } if gather.is_some() => false,
```

We should only allow these if it's a gather operation.

teoxoy (changes requested):

tests/in/image.wgsl
JMS55: wgsl-analyzer was auto formatting this, but I undid it.

src/valid/analyzer.rs
JMS55: Putting the gather in the sampling set lead to issues. I asked on matrix, and it was suggested that perhaps we shouldn't put it in the set. Doing that fixed my issues.

I'm not sure if it's actually the correct fix, but it seems to work for me. I don't have enough experience with naga to know what the correct fix would be, otherwise.

teoxoy: I undid this change locally and tests ran fine. Will push it.

PR #2145, opened 2022-12-05
[wgsl-in] Update texel formats

teoxoy:
see https://gpuweb.github.io/gpuweb/wgsl/#recursive-descent-syntax-texel_format


teoxoy:
The extra formats will be supported via an extension (see https://github.com/gfx-rs/naga/issues/2155) and capability (see https://github.com/gfx-rs/naga/issues/2161).

PR #2146, opened 2022-12-05
[wgsl-in] Remove non-32bit integers

teoxoy:
Validation already fails on any non-32bit integers but this change makes it clearer what the frontend actually supports.

Also commented out `f16`, we can uncomment it once we implement the extension.


jimblandy (approved):
Looks good to me.

I kind of wish we could just keep all these types in the front end, and reject them in validation. But they're not in the WGSL spec, so that doesn't make sense.

PR #2148, opened 2022-12-05
add Range::new_from_bounds

robtfm:
when working with naga IR i've found i need to be able to construct `Range`s (specifically for `Statement::Emit` which can't be constructed otherwise). it is possible currently, but only via serde hacking.

i added a `Range::new_from_bounds` method to allow construction.

first PR, apologies if i missed any standard process.

### DIFF
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -153,6 +153,15 @@ impl<T> Iterator for Range<T> {
     }
 }
 
+impl<T> Range<T> {
+    pub fn new_from_bounds(first: Handle<T>, last: Handle<T>) -> Self {
+        Self {
+            inner: (first.index() as u32)..(last.index() as u32 + 1),
+            marker: Default::default(),
+        }
+    }
+}
+
 /// An arena holding some kind of component (e.g., type, constant,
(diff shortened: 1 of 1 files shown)


teoxoy:
> i need to be able to construct `Range`s (specifically for `Statement::Emit` which can't be constructed otherwise).

What about using [`Arena.range_from`](https://docs.rs/naga/latest/naga/struct.Arena.html#method.range_from)? That's what we use in our front-ends.

robtfm:
thanks, i could work with that if i had to, but it's a bit awkward as i'm not building statements and expressions in tandem. 

my context is i am removing parts of naga modules, so i iterate through determining my filter for the various parts, then rebuild expressions and blocks (etc) separately afterwards. i guess i could rework it if you're opposed to adding this.

teoxoy:
I see, so you are using this [`serde_range`](https://github.com/robtfm/naga_oil/blob/52ff64518640ce7d421b0847e8bbdc1133aeb5f4/src/util.rs#L99) "hack" to do this.

teoxoy (approved):

Issue #1755, opened 2022-02-28
atomicCompareExchangeWeak not working


ria8651:
**Description**
When using atomicCompareExchangeWeak I get a validation error: `Result type for [33] doesn't match the statement`
```rust
let exchange = atomicCompareExchangeWeak(&n.lock, 0u, 1u);
```

**Expected vs observed behaviour**
I'd expect the behaviour explained in the spec.

**Platform**
 * macOS
 * wgpu 0.12 or wgpu master

kvark:
We could implement *something* in the backends, but it would be partially wasted work, since this isn't going to be the API of WGSL but rather something ad-hoc.

teoxoy:
Validation was updated by https://github.com/gfx-rs/naga/pull/2165 which also added support for SPIR-V codegen.
Let's track remaining work (adding support for the builting to the other remaining backends) in https://github.com/gfx-rs/naga/issues/1413.

PR #2165, opened 2022-12-09
Add support for WGSL's `atomicCompareExchangeWeak` with the `__atomic…

aweinstock314:
…_compare_exchange_result` struct, and add SPIR-V codegen for it.

Partially addresses https://github.com/gpuweb/gpuweb/pull/2113, #1755.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1401,11 +1401,7 @@ pub enum Expression {
     /// Result of calling another function.
     CallResult(Handle<Function>),
     /// Result of an atomic operation.
-    AtomicResult {
-        kind: ScalarKind,
-        width: Bytes,
-        comparison: Option<Handle<Type>>,
-    },
+    AtomicResult { ty: Handle<Type>, comparison: bool },
     /// Get the length of an array.
     /// The expression must resolve to a pointer to an array with a dynamic size.
     ///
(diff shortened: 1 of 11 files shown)


src/valid/function.rs
```diff
@@ -363,11 +363,17 @@ impl super::Validator {
                 .into_other());
         }
         match context.expressions[result] {
-            //TODO: support atomic result with comparison
+            //TODO: does the result of an atomicCompareExchange need additional validation, or does the existing validation for
+            // the struct type it returns suffice?
```
teoxoy, line 367: We should still validate that `comparison_ty.old_value.ty.kind == ptr_kind && comparison_ty.old_value.ty.width == ptr_width` (pseudocode).

src/lib.rs
```diff
@@ -1404,7 +1404,7 @@ pub enum Expression {
     AtomicResult {
         kind: ScalarKind,
         width: Bytes,
-        comparison: bool,
+        comparison: Option<Handle<Type>>,
     },
```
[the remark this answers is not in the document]
aweinstock314: Changed to `AtomicResult { ty: Handle<Type>, comparison: bool },`

src/valid/function.rs
aweinstock314: `validate_function` and `validate_expression` now both call into `validate_atomic_compare_exchange_struct`, which checks the struct members' names and types (including checking that `old_value` satisfies the same predicate on scalars that the functions previously checked directly for).

teoxoy (approved):
Good stuff!

PR #2172, opened 2022-12-13
Allow `u32` coordinates for `textureStore`/`textureLoad`

PENGUINLIONG:
As illustrated in https://www.w3.org/TR/WGSL/#texturestore:

> F is a [texel format](https://www.w3.org/TR/WGSL/#texel-format)
> C is [i32](https://www.w3.org/TR/WGSL/#i32), or [u32](https://www.w3.org/TR/WGSL/#u32)
> ...

So I guess `vec<u32>` should be a valid `textureStore` coordinate.

I saw that in `back::glsl::write_texture_coord` the coord is explicitly cast to `ivec` so I think that's good. Please remind me if anything should be changed accordingly in the codegens.


teoxoy (changes requested):
I looked over the backends and this shouldn't cause any issues.
Could you add some lines to the `tests/in/image.wgsl` file to test for those overloads?

Thanks!

teoxoy:
This PR would in theory allow `textureLoad` to also accept `u32` coordinates. Could you also add a test for that?
Also, make sure you push the changed test snapshots (that's why CI is failing).

teoxoy:
Looks like we need to check for `component_kind == crate::ScalarKind::Uint` and convert `array_index_i32_id` to a uint (similar to how we do for the else branch below).

https://github.com/gfx-rs/naga/blob/8e1b0529e62f39f2051ec813f99fd682d1a43b5f/src/back/spv/image.rs#L317-L336

PENGUINLIONG:
ah thanks for the hint ;)

PENGUINLIONG:
wow thats funny. i thought the glsl back is not affected.

PENGUINLIONG:
Oof. Can I rerun the DXC test?

src/back/glsl/mod.rs
```diff
@@ -3146,10 +3146,16 @@ impl<'a, W: Write> Writer<'a, W> {
             None => {
                 if tex_1d_hack {
                     write!(self.out, "ivec2(")?;
+                } else if vector_size == 1 {
+                    write!(self.out, "int(")?;
+                } else {
+                    write!(self.out, "ivec{}(", vector_size)?;
```
teoxoy, line 3152: Let's do this only if necessary (i.e. only if the `coordinate` expression's type is unsigned).

teoxoy (approved):
Looks good! Thanks!

PR #2177, opened 2022-12-17
Fix warnings when building tests without validation.

jimblandy:


src/valid/interface.rs
```diff
@@ -102,12 +102,14 @@ fn storage_usage(access: crate::StorageAccess) -> GlobalUse {
...
     output: bool,
-    flags: super::ValidationFlags,
     types: &'a UniqueArena<crate::Type>,
     type_info: &'a Vec<super::r#type::TypeInfo>,
     location_mask: &'a mut BitSet,
     built_ins: &'a mut crate::FastHashSet<crate::BuiltIn>,
     capabilities: Capabilities,
+
+    #[cfg_attr(not(feature = "validate"), allow(dead_code))]
+    flags: super::ValidationFlags,
```
teoxoy, line 112: Could we put this behind `#[cfg(feature = "validate")]`?

teoxoy (approved):
LGTM, just a non-blocking question.

PR #2180, opened 2022-12-19
Supporting gl_PointCoord

Neo-Zhixing:
Support `gl_PointCoord` in the glsl frontend.

### DIFF
--- a/src/back/hlsl/conv.rs
+++ b/src/back/hlsl/conv.rs
@@ -162,6 +162,7 @@ impl crate::BuiltIn {
             Self::VertexIndex => "SV_VertexID",
             // fragment
             Self::FragDepth => "SV_Depth",
+            Self::PointCoord => "VPOS",
             Self::FrontFacing => "SV_IsFrontFace",
             Self::PrimitiveIndex => "SV_PrimitiveID",
             Self::SampleIndex => "SV_SampleIndex",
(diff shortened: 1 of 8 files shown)


src/back/hlsl/conv.rs
```diff
@@ -162,6 +162,7 @@ impl crate::BuiltIn {
             Self::VertexIndex => "SV_VertexID",
             // fragment
             Self::FragDepth => "SV_Depth",
+            Self::PointCoord => "VPOS",
```
teoxoy, line 165: This is only available in dx9 / sm2 which we don't support. Let's error with "Unsupported builtin" instead (as we do for `ViewIndex`). And if we are here already, please also do the same for `PointSize`.

See also https://github.com/gpuweb/gpuweb/issues/332 for more context.

teoxoy (changes requested):
Thanks for the PR!
Have a small request for the HLSL backend.

src/back/hlsl/conv.rs
Neo-Zhixing: I don't think you can remove gl_PointSize that easily. In the `quad-vert` The hlsl backend emits `PSIZE` in the `gl_PerVertex` struct despite it not being used. I've only made it so that it errors with Unsupported builtin on gl_PointCoord.

teoxoy: I see, thanks for trying though! I opened https://github.com/gfx-rs/naga/issues/2182 to do it later.

teoxoy (approved):

PR #2189, opened 2022-12-30
[glsl-out] emit reflection info for non-struct uniforms

Rainb0wCodes:
Every global variable with `AddressSpace::Uniform | AddressSpace::Storage` will be written using `write_interface_block`, so this just exposes the reflection information that was previously private.


teoxoy (approved):
LGTM, thanks for the PR!

Issue #393, opened 2021-01-30
Image array support


kvark:
It would be nice to support this, since wgpu-core does, and one of the wgpu-rs examples uses this.
Example GLSL:
```
layout(set = 0, binding = 0) uniform texture2D u_Textures[2];
```
cc @cwfitzgerald

teoxoy:
Since https://github.com/gfx-rs/naga/pull/1845 got merged we now support this in all backends and the WGSL frontend.

Remaining work would be GLSL and SPIR-V frontends.

Patryk27:
fyi, I've got something working for the SPIR-V frontend - I'll try submitting the patch tomorrow 🙂

PR #2199, opened 2023-01-07
[spv-in] Support binding arrays

Patryk27:
This pull request extends https://github.com/gfx-rs/naga/pull/1845 by implementing support for binding arrays for the SPIR-V frontend 🙂 

In general, the changes here boil down to adjusting `parse_type_array()` so that it emits `crate::TypeInner::BindingArray` for texture-based arrays and sampler-based array -- as you'll see in the comments, the approach implemented here is kinda hacky, because - contrary to the wgls' frontend - when parsing `OpTypeArray` / `OpTypeRuntimeArray` we can't yet know whether it's going to be used through a binding or not.

I think that ideally instead of taking a guess in `parse_type_array()`, we'd adjust `parse_global_variable()` to do something like:

```rust
if let crate::TypeInner::Array { base, size, .. } = module.types[original_ty].inner {
    if dec.desc_set.is_none() || dec.desc_index.is_none() {
        let ty2 = crate::Type {
            name: None,
            inner: crate::TypeInner::BindingArray { base, size },
        };

        ty = module.types.insert(ty2, Default::default());
    }
}
```

... but that doesn't work, since it leaves the original type in `module.types`, causing [`validate_type()`](https://github.com/gfx-rs/naga/blob/5b4e94630be452cdbb662940c3cf637f9dccfade/src/valid/type.rs#L379) to fail later (because it always iterates through all of the types, even if they are not actually used anywhere, as would be the case here).

I'm currently writing a raytracer using rust-gpu and I've confirmed that a shader allowed by Naga with this patch actually works (on Linux + Nvidia, I haven't checked it on Windows or Mac).

Related to:
https://github.com/gfx-rs/naga/issues/393.


src/front/spv/mod.rs
```diff
@@ -4793,32 +4874,42 @@ impl<I: Iterator<Item = u32>> Parser<I> {
...
             class: crate::ImageClass::Storage { format, access: _ },
-        } = module.types[effective_ty].inner
+        } = module.types[ty].inner
         {
             // Storage image types in IR have to contain the access, but not in the SPIR-V.
             // The same image type in SPIR-V can be used (and has to be used) for multiple images.
             // So we copy the type out and apply the variable access decorations.
             let access = dec.flags.to_storage_access();
-            let ty = crate::Type {
+            let ty2 = crate::Type {
```
jimblandy, line 4901: Maybe 'adjusted_ty` or `fixed_access_ty` would be better names.

tests/in/spv/binding-arrays.static.spvasm
```diff
@@ -0,0 +1,78 @@
...
+;; `TypeInner::BindingArray` and support indexing it through `OpAccessChain`
+;; and `OpInBoundsAccessChain`.
+;;
+;; Code in here corresponds to, more or less:
+;;
+;; ```rust
+;; #[spirv(fragment)]
+;; pub fn main(
+;;     #[spirv(descriptor_set = 0, binding = 0)]
+;;     images: &[Image!(2D, type=f32, sampled); 256,
```
jimblandy, line 11: missing `]`?

tests/snapshots.rs
```diff
@@ -580,7 +580,7 @@ fn convert_spv(name: &str, adjust_coordinate_space: bool, targets: Targets) {
     check_targets(&module, name, targets);
     naga::valid::Validator::new(
         naga::valid::ValidationFlags::all(),
-        naga::valid::Capabilities::empty(),
+        naga::valid::Capabilities::all(),
```
jimblandy, line 583: I'm a little uncomfortable changing the conditions under which we're validating the other tests. Let's have another `bool` argument to `convert_spv` instead. Or, ideally, combine `adjust_coordinate_space` and the new bool into a `bitflags!` type, so that we use meaningful names in the call sites instead of `false, true`, etc.

jimblandy (changes requested):
This looks great - thank you! Just some minor changes:

tests/in/spv/binding-arrays.static.spvasm
Patryk27: true, true

tests/snapshots.rs
Patryk27: Actually, as it turns out, those new tests don't need any particular capabilities - I think they needed them for a while during development, when I accidentally broke uniformity analysis for some opcodes 🙃 

Anyway, I've rolled back to `Capabilities::empty()` and everything seems to be passing.

Patryk27:
Okie, ready for round 2 🙂 

Pipelines are failing, but it doesn't seem to be related to the changes here 👀

tests/snapshots.rs
```diff
@@ -605,12 +605,9 @@ fn convert_spv_all() {
         true,
         Targets::HLSL | Targets::WGSL | Targets::METAL,
     );
-    convert_spv(
-        "empty-global-name",
-        true,
-        Targets::HLSL | Targets::WGSL | Targets::METAL,
-    );
```
teoxoy, line 612: Was removing this intended?

Patryk27: Yeah, it seems to be duplicated 👀 (unless this duplication is part of the test, ofc. - but it doesn't seem to be)

Jerrody:
@teoxoy @jimblandy 
I think it can be merged!

jimblandy (approved):
Lovely!

Issue #1744, opened 2022-02-23
Implement WGSL uniformity analysis


jimblandy:
Naga should implement the uniformity analysis introduced in gpuweb/gpuweb@0561a0ee.

I think this would replace the uniformity analysis we have now. It could still be enabled by `ValidationFlags::CONTROL_FLOW_UNIFORMITY`.

The algorithm given in the spec looks, at first glance, like it should be straightforward to implement on Naga's IR, even though we don't have handles/indices for statements, only expressions.

PR #2201, opened 2023-01-10
Workgroup uniform load

### LINKED ISSUE #2191: implement `workgroupUniformLoad`
[spec](https://gpuweb.github.io/gpuweb/wgsl/#workgroupUniformLoad-builtin)

It's important to get the barriers right here.

DJMcNab:
Fixes #2191

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -3736,7 +3736,7 @@ impl<'a, W: Write> Writer<'a, W> {
         named: Handle<crate::Expression>,
         ctx: &back::FunctionCtx,
     ) -> BackendResult {
-        match ctx.info[handle].ty {
+        match ctx.info[named].ty {
             proc::TypeResolution::Handle(ty_handle) => match self.module.types[ty_handle].inner {
                 TypeInner::Struct { .. } => {
                     let ty_name = &self.names[&NameKey::Type(ty_handle)];
@@ -3751,7 +3751,7 @@ impl<'a, W: Write> Writer<'a, W> {
             }
         }
 
-        let base_ty_res = &ctx.info[handle].ty;
+        let base_ty_res = &ctx.info[named].ty;
         let resolved = base_ty_res.inner_with(&self.module.types);
 
         write!(self.out, " {name}")?;
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2129,13 +2129,13 @@ impl<'a, W: Write> Writer<'a, W> {
                 // GLSL doesn't have pointers, which means that this backend needs to ensure that
                 // the actual "loading" is happening between the two barriers.
                 // This is done in `Emit` by never emitting a variable name for pointer variables
-                self.write_barrier(crate::Barrier::WORK_GROUP, level);
+                self.write_barrier(crate::Barrier::WORK_GROUP, level)?;
 
                 let result_name = format!("{}{}", back::BAKE_PREFIX, result.index());
                 // Expressions cannot have side effects, so just writing the expression here is fine.
-                self.write_named_expr(pointer, result_name, result, ctx);
+                self.write_named_expr(pointer, result_name, result, ctx)?;
 
-                self.write_barrier(crate::Barrier::WORK_GROUP, level);
+                self.write_barrier(crate::Barrier::WORK_GROUP, level)?;
             }
             // Stores a value into an image.
             Statement::ImageStore {
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2132,6 +2132,7 @@ impl<'a, W: Write> Writer<'a, W> {
                 self.write_barrier(crate::Barrier::WORK_GROUP, level)?;
 
                 let result_name = format!("{}{}", back::BAKE_PREFIX, result.index());
+                write!(self.out, "{level}")?;
                 // Expressions cannot have side effects, so just writing the expression here is fine.
                 self.write_named_expr(pointer, result_name, result, ctx)?;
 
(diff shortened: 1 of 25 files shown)


src/lib.rs
```diff
@@ -1404,6 +1404,11 @@ pub enum Expression {
     CallResult(Handle<Function>),
     /// Result of an atomic operation.
     AtomicResult { ty: Handle<Type>, comparison: bool },
+    /// Result of a [`Statement::WorkgroupUniformLoad`]
+    ///
+    /// This expression is uniform by definition
+    /// (The pointer here is stored for type resolution only)
+    WorkgroupUniformLoadResult { pointer: Handle<Expression> },
```
teoxoy, line 1411:
```suggestion
    WorkgroupUniformLoadResult { ty: Handle<Type> },
```

I think we should keep this definition consistent with `AtomicResult`.

src/valid/function.rs
```diff
@@ -805,6 +812,40 @@ impl super::Validator {
...
+                            )
+                            .with_span_handle(result, context.expressions))
+                        }
+                    }
+                    match context.resolve_type(pointer, &self.valid_expression_set)? {
+                        &Ti::Pointer { base, .. }
+                            if self.types[base.index()]
+                                .flags
+                                // HACK: The set of types for SIZED|CONSTRUCTIBLE
+                                // is, as far as I can tell, exactly the types required (i.e. fixed-size data types without atomic types)
```
teoxoy, line 837: Not a hack, this is correct. `CONSTRUCTIBLE` actually implies `SIZED`, so it could be omitted.

src/valid/expression.rs
```diff
@@ -1424,6 +1424,8 @@ impl super::Validator {
                     return Err(ExpressionError::InvalidArrayType(expr));
                 }
             },
+            // Note: The pointer being valid is handled by validation of the statement
```
teoxoy, line 1427: Won't need this comment if we change the definition.

tests/out/wgsl/workgroupUniformLoad.wgsl
```diff
@@ -0,0 +1,16 @@
+let SIZE: u32 = 128u;
+
+@group(0) @binding(0) 
+var<storage> arr_i32_: array<i32,SIZE>;
+
+@compute @workgroup_size(4, 1, 1) 
+fn test_workgroupUniformLoad(@builtin(workgroup_id) workgroup_id: vec3<u32>) {
+    let x = (&arr_i32_[workgroup_id.x]);
+let _e5 = workgroupUniformLoad(x);
```
teoxoy, line 9: Should be properly indented.

src/back/glsl/mod.rs
```diff
@@ -2015,6 +2015,30 @@ impl<'a, W: Write> Writer<'a, W> {
 
                 writeln!(self.out, "{}barrier();", level)?;
             }
+            Statement::WorkgroupUniformLoad { pointer, result } => {
+                writeln!(self.out, "{}memoryBarrierShared();", level)?;
+                writeln!(self.out, "{}barrier();", level)?;
```
teoxoy, line 2020: We recently introduced a new function `write_barrier` for all backends that we could use.

teoxoy (changes requested):
Thanks for the PR!
Have a few comments but overall looks good.

jimblandy:
Tragically, this predates #2075 by three commits, so the WGSL front end changes need to be rebased.

tests/in/workgroupUniformLoad.wgsl
```diff
@@ -0,0 +1,13 @@
+let SIZE: u32 = 128u;
+
+@group(0) @binding(0)
+var<storage,read> arr_i32: array<i32, SIZE>;
+
+@compute @workgroup_size(4)
+fn test_workgroupUniformLoad(@builtin(workgroup_id) workgroup_id: vec3<u32>) {
+    let x = &arr_i32[workgroup_id.x];
+    let val = workgroupUniformLoad(x);
```
jimblandy, line 9: This test case is incorrect - the argument to `workgroupUniformLoad` must be uniform, but the pointer you've constructed here is definitely not uniform.

jimblandy (changes requested):

tests/in/workgroupUniformLoad.wgsl
DJMcNab: Can you clarify how this is non uniform?

The argument to the field access is uniform, the global is uniform, so both arguments to the indexing operation are uniform. I'm not sure which operation here introduces non-uniformity.

jimblandy: I'm sorry - the patch says `workgroup_id`, which is uniform, but I somehow read `local_invocation_id`. The test is fine.

jimblandy (changes requested):
This looks good; sorry about the misunderstanding. This just needs to be rebased over the #2075 changes, then.

src/valid/analyzer.rs
```diff
@@ -776,6 +783,32 @@ impl FunctionInfo {
                     },
                     exit: ExitFlags::empty(),
                 },
+                S::WorkGroupUniformLoad { pointer, .. } => {
+                    let _condition_nur = self.add_ref(pointer);
+
+                    // We choose not to check that this occurs in what we believe to be a uniform condition
+                    // because our definition includes some blocks which are actually uniform (by the WGSL spec)
```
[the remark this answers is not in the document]
jimblandy: If I'm understanding correctly, checking uniformity here would break Vello on wgpu only because Naga does not yet implement the uniformity analysis in the WGSL spec, but rather uses a previously developed, less permissive analysis, and we expect that when Naga does implement the specified analysis, checking it here will not break Vello.

If that's the case, then it seems fine to omit the check for now. I think the comment could be clarified:

> Don't check that this call occurs in uniform control flow until Naga implements WGSL's standard uniformity analysis (#1744). The uniformity analysis Naga uses now is less accurate than the one in the WGSL standard, causing Naga to reject correct uses of `workgroupUniformLoad` in some interesting programs.

teoxoy (approved):

jimblandy (approved):

PR #2205, opened 2023-01-11
glsl: include unused items

robtfm:
- added a `back::glsl::WriterFlags` flag `INCLUDE_UNUSED_ITEMS`.
when set, unused globals, functions using unused globals, and named constants are included in the output glsl.

- added a named copy of named constants in front::glsl into the module, to allow exporting them above. this adds a bit of unnecessary weight to the module, but hopefully it's not a problem?

motivation is for better glsl support in naga_oil, where i rely on modules to contain all members even if unused at point of construction.

### DIFF
--- a/src/front/glsl/variables.rs
+++ b/src/front/glsl/variables.rs
@@ -501,6 +501,20 @@ impl Parser {
                     mutable: false,
                 };
 
+                if let Some(name) = name.as_ref() {
+                    let constant = self.module.constants.try_get(init).unwrap();
+
+                    // add a named copy of the constant
+                    self.module.constants.fetch_or_append(
+                        Constant {
+                            name: Some(name.clone()),
+                            specialization: constant.specialization,
+                            inner: constant.inner.clone(),
+                        },
+                        meta,
+                    );
+                }
+
                 (GlobalOrConstant::Constant(init), lookup)
             }
             StorageQualifier::AddressSpace(mut space) => {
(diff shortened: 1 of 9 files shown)


src/front/glsl/variables.rs
```diff
@@ -501,6 +501,20 @@ impl Parser {
...
+                    self.module.constants.fetch_or_append(
+                        Constant {
+                            name: Some(name.clone()),
+                            specialization: constant.specialization,
+                            inner: constant.inner.clone(),
+                        },
+                        meta,
+                    );
+                }
+
```
teoxoy, line 517: Can we avoid this extra copy and name the actual constant generated by the front-end?

robtfm: we'd still need a copy per name, but yes. it'll be a little more involved though.

currently the original unnamed constant is added by `ConstantSolver::solve` (20 refs), from `Parser::solve_constant` (5 refs), from `ParsingContext::parse_init_declarator_list`.

`ParsingContext::parse_init_declarator_list` then calls `DeclarationContext::add_var` which calls `Parser::add_global_var` where i've added the copy.

the cleanest solution would be for the constant solver and `solve_constant` to take an optional name in as well. if you agree it's the right solution i'll try that and update the call sites.

JMS55: Btw, @robtfm your comment (the one above this one) seems to have gotten marked as spam.

robtfm: i marked it as spam ... the solution was much simpler than i was making out (it's in 830840d) as the consts are in an `Arena` which allows mutable access, not a `UniqueArena` which doesn't. i should have resolved the conversation as well, which i'll do now.

teoxoy (approved):
LGTM

PR #2207, opened 2023-01-14
[wgsl-in] Split into multiple files

SparkyPotato:
In #2075, we left all the changes in one file to make reviewing easier. However, this made `front/wgsl/mod.rs` huge (5,447 lines), and quite challenging to follow.

This PR splits each pass into its own module, bringing both `parse/mod.rs` and `lower/mod.rs` to a more manageable ~2,000 lines. Other than `use`s, everything has been copy-pasted.

Since the frontends now do more than parse, I also renamed them to `Frontend` instead of `Parser` to make the intent more clear: https://github.com/gfx-rs/naga/pull/2075#issuecomment-1321089842


jimblandy:
This looks great, but let's fix those doc links.

SparkyPotato:
Oops, missed that.

Rustdoc seems to be acting weird, however.
Linking `crate::front::wgsl::lower::ExpressionContext::local_table` directly errors, but if I `use crate::front::wgsl::lower::ExpressionContext`, it stops complaining.

jimblandy (changes requested):
This looks good, just some minor points.

Acknowledging Teo's suggestion in #2075 that all the front ends should use `Frontend` instead of `Parser`, I have to say I would also approve a PR that just left GLSL alone.

src/front/glsl/builtins.rs
```diff
@@ -1617,7 +1617,7 @@ impl MacroCall {
     /// finally returns the final expression with the correct result
     pub fn call(
         &self,
-        parser: &mut Parser,
+        parser: &mut Frontend,
```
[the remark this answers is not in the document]
SparkyPotato: A lot of F2 mashing later, I think I have got everything.

src/front/wgsl/error.rs
```diff
@@ -0,0 +1,684 @@
...
+impl std::error::Error for ParseError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum NumberType {
+    I32,
+}
```
jimblandy, line 100: I guess we don't need this any more. If you don't want to just delete this altogether as part of this patch, be sure to file a follow-up PR or at least an issue about it.

src/front/wgsl/parse/mod.rs
```diff
@@ -0,0 +1,2066 @@
...
+use crate::front::wgsl::parse::number::Number;
+use crate::front::SymbolTable;
+use crate::{Arena, FastHashSet, Handle, Span};
+
+pub mod ast;
+pub mod conv;
+pub mod lexer;
+pub mod number;
+
+/// State for constructing an AST expression.
```
jimblandy, line 12: It might be worth putting in a comment:

```suggestion
/// State for constructing an AST expression.
///
/// Not to be confused with `lower::ExpressionContext`.
```

src/front/wgsl/lower/mod.rs
```diff
@@ -0,0 +1,2368 @@
...
+        OutputContext {
+            ast_expressions: self.ast_expressions,
+            globals: self.globals,
+            types: self.types,
+            module: self.module,
+        }
+    }
+}
+
+pub struct ExpressionContext<'source, 'temp, 'out> {
```
jimblandy, line 131: Might be worth adding a comment:

```suggestion
/// State for lowering an `ast::Expression` to Naga IR.
///
/// Not to be confused with `parser::ExpressionContext`.
pub struct ExpressionContext<'source, 'temp, 'out> {
```

jimblandy (approved):
Looks good to me.  Just minor comments.

I don't know how @jcapucho is going to feel about the GLSL changes. If he's not happy with them, we can back them out.

jimblandy (approved):
Looks good - thanks!

jimblandy:
Clippy got clevererer.

jimblandy:
Filed #2229 for the Clippy problems.

Issue #2162, opened 2022-12-08
Add support for 16bit norm storage texture formats


teoxoy:
```
r16unorm
r16snorm
rg16unorm
rg16snorm
rgba16unorm
rgba16snorm
```

`wgpu` supports those via `Features::TEXTURE_FORMAT_16BIT_NORM`.

We should add a similar `Capability::16BIT_NORM_STORAGE_TEXTURES_FORMATS` and `naga_ext_16bit_norm_storage_texture_formats` (see https://github.com/gfx-rs/naga/issues/2155) to support those.

Might become an official extension, see https://github.com/gpuweb/gpuweb/issues/3333 / https://github.com/gpuweb/gpuweb/issues/3001.

teoxoy:
Support landed for those, I updated https://github.com/gfx-rs/naga/issues/2155 to include the extension.

Issue #2155, opened 2022-12-06
Naga specific WGSL extensions


teoxoy:
- `naga_ext_push_constants`
    - allows usage of `push_constant` Address Space
- `naga_ext_f64`
    - allows usage of `f64` type
- `naga_ext_early_depth_test`
    - allows usage of `@early_depth_test` attribute
- `naga_ext_multiview` (or `naga_ext_view_index`?)
    - allows usage of `@builtin(view_index)` builtin
- `naga_ext_primitive_index`
    - allows usage of `@builtin(primitive_index)` builtin
- `naga_ext_binding_array` (this will end up having 8 `Capabilities` see https://github.com/gfx-rs/naga/issues/2151)
    - allows usage of `binding_array` type

All of these have their own Capabilities therefore the validator is already aware of those non-standard items but I think we should still have our own set of extensions so that it's obvious when a shader uses those items.

Introducing a `strict_extensions: bool` field for the `Parser` will limit the extensions to only those defined by the spec (for now only `f16`).

Relevant errors should be different based on the `strict_extensions` field. Example: using an `f64` type with `strict_extensions: true` should yield an error that informs the user that there is no such type; on the other hand, if `strict_extensions` were `false`, the error should inform the user that the `naga_ext_f64` extension is required.

jimblandy:
>All of these have their own Capabilities therefore the validator is already aware of those non-standard items but I think we should still have our own set of extensions so that it's obvious when a shader uses those items.

So the idea is, to use some extension, you have to add the bit to `Capabilities`, and then also put `enable BLAH;` at the top? And the benefit would be that nobody would be writing Naga-specific shaders without them being clearly marked as such?

teoxoy:
`Capabilities` that are not part of the WebGPU/WGSL specs will be set automatically by wgpu depending on some flag (see https://github.com/gfx-rs/wgpu/issues/3274).
Users will only have to `enable naga_ext_<name>;` in their shaders to use a naga specific extension (given they set the "Strict WebGPU compliance mode" to `false`).

PR #2210, opened 2023-01-17
Support 16-bit unorm/snorm formats

fintelia:
This PR adds handling for 16-bit unorm and snorm formats.

After making a few tweaks to wgpu, I was able to use this patch to load and run several GLSL compute shaders operating on r16 unorm textures. I haven't tested the other frontends/backends but the code I added mirrors the existing cases, so there's a good chance they should behave properly


jimblandy:
This looks interesting, but before I review in detail:

We can't unconditionally extend WGSL, since that affects Firefox's implementation of the WebGPU standard, and we'd introduce incompatibilities between browsers. This probably needs to be conditional on some new flag in `naga::valid::Capabilities`.

teoxoy (approved):
LGTM, but could you also add them to the SPV front-end?

jimblandy (approved):
@teoxoy points out that #2161 covers making the non-WebGPU formats we already support in the WGSL front end conditional, and that's in milestone 4. So I think this is fine to merge as it is.

teoxoy:
Sorry, I approved it without reading the comments properly.

I thought https://github.com/gfx-rs/naga/issues/2161 would cover this and could be done in a separate PR since there are a bunch of other formats. But ideally this PR should close https://github.com/gfx-rs/naga/issues/2162 (we don't yet have the framework for naga specific extensions but we can move that point from the issue into https://github.com/gfx-rs/naga/issues/2155).

@fintelia adding the `Capability` would be appreciated since we have to wire it to wgpu as well.

teoxoy (commented):
Thanks for introducing the new capability!
Could you also add the new formats to the spv front-end?

PR #2212, opened 2023-01-17
[glsl-in] Use std430 layout when validating arrays within structs

### LINKED ISSUE #2211: [glsl-in] Arrays in structs produce array stride errors
This shader:

```glsl
#version 460 core

struct A {
    float b[2];
};

void main() {}
```

Produces this error:
```
error: 
  ┌─ shader.frag:4:5
  │
4 │     float b[2];
  │     ^^^^^^^^^^ naga::Type [3]

Type [3] '' is invalid: 
	Array stride 16 does not match the expected 4
```

fintelia:
Fixes #2211. However I don't know whether this change has broader consequences

### DIFF
--- a/src/front/glsl/parser/types.rs
+++ b/src/front/glsl/parser/types.rs
@@ -86,7 +86,7 @@ impl<'source> ParsingContext<'source> {
                 self.expect(parser, TokenValue::LeftBrace)?;
                 let mut members = Vec::new();
                 let span =
-                    self.parse_struct_declaration_list(parser, &mut members, StructLayout::Std140)?;
+                    self.parse_struct_declaration_list(parser, &mut members, StructLayout::Std430)?;
                 let end_meta = self.expect(parser, TokenValue::RightBrace)?.meta;
                 meta.subsume(end_meta);
                 let ty = parser.module.types.insert(


JCapucho:
This problem unfortunately is not that simple, the layout to use is only known when the struct is used (in glsl), for example take the following code:

```glsl
#version 460 core

struct A {
    float b[2];
};

layout(std140, set = 0, binding = 0) uniform Data {
    A a;
};

layout(std430, set = 0, binding = 1) buffer DataBuffer {
    A b;
};

void main() {}
```

The struct `A` has two different layouts! One for the uniform and another for the buffer, also note that in glsl uniforms by default can't use the `std430` layout (unless an extension is enabled), that is why right now it defaulted to std140, every struct I ran into was used in a uniform.

The solution here is to generate the structs with the correct layout as needed, unfortunately this also requires auto casting between the different layouts in some situations (for example a function picked to accept std430 structs and you pass it a std140 struct).

fintelia:
Unfortunately, naga's validation currently rejects any shaders that declare structs with layouts that do not match std430. In #2211 the struct is never even used, but the fact that it contains an array with elements requiring <=8 byte alignment causes it to fail validation.

teoxoy:
@fintelia the fix for #2211 is most likely going to be removing those lines from the validator since they are redundant. We defer erroring wrt alignment/stride (the errors assigned right below this section of code) until we see the data types being used either as uniform/storage.

https://github.com/gfx-rs/naga/blob/461fdda4250cdf9aa49de960afd0fa5042a51bc9/src/valid/type.rs#L383-L389

It's worth opening a new issue for the use-case @JCapucho brought up though (will do tomorrow if nobody else does in the meanwhile).

PR #2215, opened 2023-01-18
Remove overly restrictive array stride check

### LINKED ISSUE #2211: [glsl-in] Arrays in structs produce array stride errors

fintelia:
Fixes #2211 by removing the stride check, as suggested in https://github.com/gfx-rs/naga/pull/2212#issuecomment-1386177626. The alignments when used in uniforms and storage buffers are both validated below in the same function.

### DIFF
--- a/src/valid/type.rs
+++ b/src/valid/type.rs
@@ -380,14 +380,6 @@ impl super::Validator {
                 }
 
                 let base_layout = self.layouter[base];
-                let expected_stride = base_layout.to_stride();
-                if stride != expected_stride {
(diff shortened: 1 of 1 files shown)


teoxoy (approved):
Thanks!

Issue #2134, opened 2022-11-24
[wgsl] Update Built-in Functions


teoxoy:
### Add

Numeric Built-in Functions
- [ ] `acosh`
- [ ] `asinh`
- [ ] `atanh`
- [ ] `countLeadingZeros`
- [ ] `countTrailingZeros`
- [ ] `quantizeToF16`
- [ ] `refract`

Derivative Built-in Functions
- [ ] `dpdxCoarse`
- [ ] `dpdxFine`
- [ ] `dpdyCoarse`
- [ ] `dpdyFine`
- [ ] `fwidthCoarse`
- [ ] `fwidthFine`

Texture Built-in Functions
- [ ] `textureSampleBaseClampToEdge`

### Remove

- [ ] `isFinite`
- [ ] `isNormal`

_last sync: https://github.com/gpuweb/gpuweb/blob/6049e8fc99b8204e44b7ca14d7bbff02a5730dfb/wgsl/index.bs_

PR #2218, opened 2023-01-23
[wgsl] Partially Update Built-in Functions

evahop:
The following were low-hanging fruit from https://github.com/gfx-rs/naga/issues/2134#issue-1463112793.
I intended to do the remainder, but it may take me some time to reason about the different backends.

Adds
- acosh, asinh, atanh
- refract

Removes
- isFinite, isNormal


teoxoy (approved):
Looks good! Thank you for the PR!

PR #2223, opened 2023-01-26
[glsl] Inject default `gl_PointSize = 1.0` in vertex shaders if `FORCE_POINT_SIZE` option was set

REASY:
### Write `gl_PointSize = 1.0;` in case of vertex shader and set writer flag `FORCE_POINT_SIZE` 

According to https://registry.khronos.org/OpenGL/specs/es/3.2/GLSL_ES_Specification_3.20.html#built-in-language-variables
> The variable gl_PointSize is intended for a shader to write the size of the point to be rasterized. It is measured in pixels. If gl_PointSize is not written to, its value is undefined in subsequent pipe stages.

This is the first part of the changes to address https://github.com/gfx-rs/wgpu/issues/3179

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2023,17 +2023,10 @@ impl<'a, W: Write> Writer<'a, W> {
                                                         /* > We agreed that gl_ClipDistance and gl_CullDistance, and related language,
                                                             should not appear in the ESSL spec since it's not part of OpenGL ES 3.2.
                                                         */
-                                                        if let Version::Embedded {
-                                                            version,
-                                                            is_webgl: _,
-                                                        } = self.options.version
-                                                        {
-                                                            if version <= 320 {
-                                                                log::warn!(
+                                                        log::warn!(
                                                             "{:?} is not part of OpenGL ES <= 3.2",
-                                                            builtin);
-                                                            }
-                                                        }
+                                                            builtin
+                                                        );
                                                         continue;
                                                     }
                                                 }
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2019,6 +2019,21 @@ impl<'a, W: Write> Writer<'a, W> {
                                                 crate::BuiltIn::ClipDistance
                                                 | crate::BuiltIn::CullDistance => {
                                                     if self.options.version.is_es() {
+                                                        // According to https://github.com/KhronosGroup/GLSL/issues/132
+                                                        /* > We agreed that gl_ClipDistance and gl_CullDistance, and related language,
+                                                            should not appear in the ESSL spec since it's not part of OpenGL ES 3.2.
+                                                        */
+                                                        if let Version::Embedded {
+                                                            version,
+                                                            is_webgl: _,
+                                                        } = self.options.version
+                                                        {
+                                                            if version <= 320 {
+                                                                log::warn!(
+                                                            "{:?} is not part of OpenGL ES <= 3.2",
+                                                            builtin);
+                                                            }
+                                                        }
                                                         continue;
                                                     }
                                                 }
@@ -2087,7 +2102,7 @@ impl<'a, W: Write> Writer<'a, W> {
                                 .contains(WriterFlags::FORCE_POINT_SIZE)
                             && !has_point_size
                         {
-                            writeln!(self.out, "gl_PointSize = 1.0;",)?;
+                            writeln!(self.out, "gl_PointSize = 1.0;")?;
                             write!(self.out, "{level}")?;
                         }
                         writeln!(self.out, "return;")?;
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -218,6 +218,13 @@ bitflags::bitflags! {
         /// global variables that are not used in the specified entrypoint (including indirect use),
         /// all constant declarations, and functions that use excluded global variables.
         const INCLUDE_UNUSED_ITEMS = 0x4;
+        /// Emit `PointSize` output builtin to vertex shaders, which is
+        /// required for drawing with `PointList` topology.
+        ///
+        /// https://registry.khronos.org/OpenGL/specs/es/3.2/GLSL_ES_Specification_3.20.html#built-in-language-variables
+        /// The variable gl_PointSize is intended for a shader to write the size of the point to be rasterized. It is measured in pixels.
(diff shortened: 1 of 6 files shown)


teoxoy (changes requested):
I think this needs to be implemented in the same way it is for the spv backend.

https://github.com/gfx-rs/naga/blob/c7d02151f08d6285683795289b5725b827d836d1/src/back/spv/writer.rs#L524-L526

- rename `allow_point_size` to `force_point_size`
- check if we already have (and wrote/will write) the `BuiltIn::PointSize`

JCapucho:
It would also help to include a comment in the code explaining this for future reference with a mention of the section of where it happens in the specification (for glsl ES 3.1 this is in section 7.1.1)

src/back/glsl/mod.rs
```diff
@@ -2006,10 +2014,11 @@ impl<'a, W: Write> Writer<'a, W> {
                                         if let Some(crate::Binding::BuiltIn(builtin)) =
                                             member.binding
                                         {
+                                            has_point_size |= builtin == crate::BuiltIn::PointSize;
+
                                             match builtin {
                                                 crate::BuiltIn::ClipDistance
-                                                | crate::BuiltIn::CullDistance
-                                                | crate::BuiltIn::PointSize => {
```
REASY, line 2012: @teoxoy I removed `BuiltIn::PointSize` from the list to let it write that built-in. But I'm not sure it is going to work. Could you provide some context or example of input that will trigger this code to execute? I can then debug and get better understanding.

src/back/glsl/mod.rs
```diff
@@ -2057,20 +2066,30 @@ impl<'a, W: Write> Writer<'a, W> {
                             }
                         }
 
-                        if let back::FunctionType::EntryPoint(ep_index) = ctx.ty {
```
REASY, line 2060: We're already in EntryPoint, no need to pattern match again

src/back/glsl/mod.rs
teoxoy: I think it should be fine since `gl_PointSize` is available in gles (see https://docs.gl/), the others are not.

src/back/glsl/mod.rs
```diff
@@ -2005,10 +2013,11 @@ impl<'a, W: Write> Writer<'a, W> {
                                         if let Some(crate::Binding::BuiltIn(builtin)) =
                                             member.binding
                                         {
+                                            has_point_size |= builtin == crate::BuiltIn::PointSize;
```
teoxoy, line 2016: We also need this if the output is not a struct.

src/back/glsl/mod.rs
```diff
@@ -2056,20 +2065,30 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                        }
+
+                        if is_vertex_stage
+                            && self
+                                .options
+                                .writer_flags
+                                .contains(WriterFlags::FORCE_POINT_SIZE)
+                            && !has_point_size
+                        {
+                            writeln!(self.out, "gl_PointSize = 1.0;",)?;
```
REASY, line 2090: I do not see unit tests to easily test this. What would be suggested way to test the changes?

src/back/glsl/mod.rs
teoxoy: It might be worth turning this into an error instead of silently ignoring them, but that could be another PR.

src/back/glsl/mod.rs
teoxoy: Find an appropriate snapshot file or create a new one and set `writer_flags` like so:

https://github.com/gfx-rs/naga/blob/c7d02151f08d6285683795289b5725b827d836d1/tests/in/interpolate.param.ron#L11

src/back/glsl/mod.rs
REASY: @teoxoy ClipDistance and CullDistance are mentioned in https://registry.khronos.org/OpenGL/specs/es/3.2/GLSL_ES_Specification_3.20.htm in 14.2.5. Mapping Variables. What is the reason it is skipped in case of ES in the code?

image

teoxoy: Ah, I see. We need to check for a gles version >= 3.2 then. I think the check was incomplete since WebGL is based on gles 3.0 and most people using the backend target WebGL.

REASY: @teoxoy just found this https://github.com/KhronosGroup/GLSL/issues/132, so added warning log in case of use `ClipDistance` and `CullDistance` in unsupported version

teoxoy: I see, so the GLSL ES spec is wrong here and no gles version supports them.

src/back/glsl/mod.rs
```diff
@@ -2005,11 +2013,27 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                                                            version,
+                                                            is_webgl: _,
+                                                        } = self.options.version
+                                                        {
+                                                            if version <= 320 {
+                                                                log::warn!(
+                                                            "{:?} is not part of OpenGL ES <= 3.2",
+                                                            builtin);
+                                                            }
+                                                        }
```
teoxoy, line 2036: I think we should log the warning regardless of version since we found out that no gles version supports those.

src/back/glsl/mod.rs
```diff
@@ -2005,11 +2013,20 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                                                | crate::BuiltIn::CullDistance => {
                                                     if self.options.version.is_es() {
+                                                        // According to https://github.com/KhronosGroup/GLSL/issues/132
+                                                        /* > We agreed that gl_ClipDistance and gl_CullDistance, and related language,
+                                                            should not appear in the ESSL spec since it's not part of OpenGL ES 3.2.
+                                                        */
+                                                        log::warn!(
+                                                            "{:?} is not part of OpenGL ES <= 3.2",
+                                                            builtin
+                                                        );
```
teoxoy, line 2029:
```suggestion
                                                        // Note that gl_ClipDistance and gl_CullDistance are listed in the GLSL ES 3.2 spec but shouldn't
                                                        // See https://github.com/KhronosGroup/GLSL/issues/132#issuecomment-685818465
                                                        log::warn!(
                                                            "{:?} is not part of GLSL ES",
                                                            builtin
                                                        );
```

teoxoy (approved):
LGTM, thanks!

Issue #2123, opened 2022-11-08
HLSL back end may duplicate `unpack2x16float` arguments


jimblandy:
#2002 added support for WGSL's `unpack2x16float` function to the HLSL back end, but the code there may result in code for the argument expression being duplicated. The HLSL back end code reads:

```
  Function::Unpack2x16float => {
      write!(self.out, "float2(f16tof32(")?;
      self.write_expr(module, arg, func_ctx)?;
      write!(self.out, "), f16tof32((")?;
      self.write_expr(module, arg, func_ctx)?;
      write!(self.out, ") >> 16))")?;
  }
```

If `arg` is a baked expression, then this will be fine, but there's no way for the analyzer to know that the HLSL back end intends to use this expression twice, so its reference count could be one. The result would be that a call like:

    unpack2x16float(some expensive expr)

would generate HLSL like:

    float2(f16tof32(some expensive expr), f16to32((some expensive expr) >> 16))

The impact of this is limited, though: Naga IR ensures that the duplicated expression will never have function calls or side effects. And the result may be cleaned up by the HLSL compiler anyway. But it seemed unintentional.

The fix would probably be a pre-pass over the expressions in the HLSL back end to bump the reference count of expression that appear as the operand of `unpack2x64float`, or otherwise indicate that they need to be baked.

cc @expenses

teoxoy:
I'm thinking injecting a function to avoid this issue would be nicer. Thoughts?

teoxoy:
We seem to be already baking expressions for `dot` in the msl and glsl backends.

But I still feel like polyfilling those via functions in the output would be nicer. Have we considered this previously?

teoxoy:
Fixed by https://github.com/gfx-rs/naga/pull/2226

PR #2226, opened 2023-01-28
Add countLeadingZeros

evahop:
Adds `countLeadingZeros` from https://github.com/gfx-rs/naga/issues/2134#issue-1463112793

For negative expressions return `0`, otherwise return `31 - msb`.

This assumes `findUMsb`, `findMSB`, and `firstbithigh` all return an index from the `lsb`,
but I don't have the means to verify this is actually the case atm.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1114,14 +1114,13 @@ impl<'a, W: Write> Writer<'a, W> {
     fn update_expressions_to_bake(&mut self, func: &crate::Function, info: &valid::FunctionInfo) {
         use crate::Expression;
         self.need_bake_expressions.clear();
-        for expr in func.expressions.iter() {
-            let expr_info = &info[expr.0];
-            let min_ref_count = func.expressions[expr.0].bake_ref_count();
+        for (fun_handle, expr) in func.expressions.iter() {
+            let expr_info = &info[fun_handle];
+            let min_ref_count = func.expressions[fun_handle].bake_ref_count();
             if min_ref_count <= expr_info.ref_count {
-                self.need_bake_expressions.insert(expr.0);
+                self.need_bake_expressions.insert(fun_handle);
             }
 
-            let (fun_handle, expr) = expr;
             match *expr {
                 Expression::Math {
                     fun: crate::MathFunction::Dot,
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2949,9 +2949,7 @@ impl<'a, W: Write> Writer<'a, W> {
                                 write!(self.out, ") + 0.5) * {LOG2_E})), ")?;
 
                                 if let crate::ScalarKind::Uint = kind {
-                                    write!(self.out, "mix(uvec{s}(0u), uvec{s}(32u), equal(")?;
-                                    self.write_expr(arg, ctx)?;
-                                    write!(self.out, ", uvec{s}(0u))), lessThanEqual(")?;
+                                    write!(self.out, "uvec{s}(32u), lessThanEqual(")?;
                                     self.write_expr(arg, ctx)?;
                                     write!(self.out, ", uvec{s}(0u)))")?;
                                 } else {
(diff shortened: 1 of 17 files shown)


teoxoy (changes requested):
Behavior looks good!
But the GLSL and HLSL backends write the expr multiple times which won't always work.

The expression needs to either be baked or the function needs to be polyfilled.
I'd go with baking for now since we have the mechanism already implemented for GLSL.

https://github.com/gfx-rs/naga/blob/c7d02151f08d6285683795289b5725b827d836d1/src/back/glsl/mod.rs#L1767-L1768

I feel like polyfilling would be a nicer solution but we'd ideally make the transition all at once.
See also https://github.com/gfx-rs/naga/issues/2123 for more details.

evahop:
> The expression needs to either be baked or the function needs to be polyfilled. I'd go with baking for now since we have the mechanism already implemented for GLSL.

Okay, I see the issue.
I'm just now noticing `FindMSB` also has a minimum requirement of `400` while the `README.md` pegs it at `330+`.
Would it be preferable to replace it with something along the lines of [es.math.clz32.js](https://github.com/zloirock/core-js/blob/5385428316f590f4845006e2b15e24db3be4d7e9/packages/core-js/modules/es.math.clz32.js)?

evahop:
I meant `uints` in the last commit but I assume it's going to get squashed either way.
Otherwise that should do it, it should address the issue in https://github.com/gfx-rs/naga/issues/2123#issue-1441029855,
and I can follow up with a PR if the `findMSB`/`findLSB` need replacing.

evahop:
Had some downtime so just went ahead and did it for the scope of this PR and won’t likely have time to address it later.

evahop:
I see what you’re pointing out. You can rebase and open another issue. I tried applying the same polyfill  to `findMSB` just now, using the `((clz + 1) % 33) …` method on the `mls` backend, and I’m not sure if its producing the expected behavior, I need to mess with it some more.

src/back/glsl/mod.rs
```diff
@@ -2941,33 +2932,55 @@ impl<'a, W: Write> Writer<'a, W> {
...
                                 if let crate::ScalarKind::Uint = kind {
-                                    write!(self.out, "uvec{s}(ivec{s}(31) - findMSB(")?;
+                                    write!(self.out, "uvec{s}(")?;
+                                } else {
+                                    write!(self.out, "ivec{s}(")?;
+                                }
+
+                                write!(self.out, "vec{s}(31.0) - floor(log(vec{s}(")?;
+                                self.write_expr(arg, ctx)?;
+                                write!(self.out, ") + 0.5) * {LOG2_E})), ")?;
```
teoxoy: Can we use `log2(x)` instead of `log(x) * LOG2E`?

teoxoy:
> If I’m not mistaken, seems like `clz` will always produce a `0` for negative numbers, while `findMSB` returns an index of the first set `0`.

This sounds correct, so the msl polyfill is not correct for negative numbers.

src/back/glsl/mod.rs
```diff
@@ -2930,6 +2931,56 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                                    self.write_expr(arg, ctx)?;
+                                    write!(self.out, ", ivec{s}(0))))")?;
+                                }
+                            }
+                            crate::TypeInner::Scalar { kind, .. } => {
+                                write!(self.out, "(")?;
+                                self.write_expr(arg, ctx)?;
+
+                                if let crate::ScalarKind::Uint = kind {
+                                    write!(self.out, " == 0u ? 32u : uint(")?;
```
teoxoy, line 2968: For `uint`, I don't think we need an extra `if`. `floor(log2(0.5)) = -1`

src/back/glsl/mod.rs
evahop: Yes, feel free to make whatever changes. I just transcribed it verbatim at the time

PR #2229, opened 2023-01-31
Address Clippy 1.67 lints.

jimblandy:
* Inline identifiers into format strings.

[Since Rust 1.58], Rust format strings have been able to "capture
arguments simply by writing {ident} in the string." Clippy 1.67 made
the corresponding warning, `uninlined_format_args`, warn-by-default.
Inlined arguments seem more readable, so Naga should adopt them.

[Since Rust 1.58]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1580-2022-01-13

* Allow `clippy::needless_borrowed_reference`.

In Clippy 1.67, the `needless_borrowed_reference` lint [was enhanced]
to look into struct and tuple patterns, so that a line like this:

    for &(ref module, ref info) in inputs.iter()

where `inputs.iter()` is yielding `&(Module, ModuleInfo)` pairs,
elicits a warning. Clippy suggests, instead:

    for (module, info) in inputs.iter()

but this is at odds with Naga's preference that `match` patterns
should have the same type as the expression being matched, for which
we have enabled the `pattern_type_mismatch` lint since
9e5cc4c9 (2021-3-12).

[was enhanced]: https://github.com/rust-lang/rust-clippy/pull/9855

### DIFF
--- a/benches/criterion.rs
+++ b/benches/criterion.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::needless_borrowed_reference)]
+
 use criterion::*;
 use std::{fs, path::PathBuf, slice};
 
--- a/cli/src/bin/naga.rs
+++ b/cli/src/bin/naga.rs
@@ -99,8 +99,7 @@ impl FromStr for BoundsCheckPolicyArg {
             "unchecked" => BoundsCheckPolicy::Unchecked,
             _ => {
                 return Err(format!(
-                    "Invalid value for --index-bounds-check-policy: {}",
-                    s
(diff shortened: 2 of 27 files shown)


jimblandy:
We're not the only people who aren't too enthusiastic about the new `needless_borrowed_reference` behavior:
https://github.com/rust-lang/rust-clippy/issues/10245

teoxoy (approved):
It's so nice to see all of those args now being inlined. No more having to look for the right argument based on its position.

PR #2230, opened 2023-01-31
spv-out: support version 1.4

kvark:
Looks like 1.4 expects all global variables to be a part of the interface:
> Starting with version 1.4, the interface’s [storage classes](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Storage_Class) are all storage classes used in declaring all global variables referenced by the entry point’s call tree.


src/back/spv/writer.rs
```diff
@@ -585,6 +585,12 @@ impl Writer {
             }
 
             let mut gv = self.global_variables[handle.index()].clone();
+            if let Some(ref mut iface) = interface {
+                // Have to include global variables in the interface
+                if self.physical_layout.version >= 0x10400 {
```
teoxoy, line 590: Would be nice to do the version conversion later and keep the tuple, but can be another PR.

https://github.com/gfx-rs/naga/blob/c7d02151f08d6285683795289b5725b827d836d1/src/back/spv/writer.rs#L48

teoxoy (changes requested):

src/back/spv/writer.rs
kvark: possible, but not very important. We are already checking for the raw version in another place, so the PR is consistent with the code.

tests/snapshots.rs
```diff
@@ -561,6 +561,7 @@ fn convert_wgsl() {
             "workgroup-var-init",
             Targets::WGSL | Targets::GLSL | Targets::SPIRV | Targets::HLSL | Targets::METAL,
         ),
+        ("sprite", Targets::SPIRV),
```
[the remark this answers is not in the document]
kvark: trying to use git on windows, and all the generated files are marked as modified after cargo test :(
added the output now

kvark:
@teoxoy unfortunately the test validation is blocked on a thing. Our `make validate-spv` has "vulkan1.0" environment hard-coded currently, so it doesn't pass validation. We need to select the environment based on the SPV version that we generate. We could do this, for example, by changing the naming scheme ("xxx.spv14asm"), or by having sub-folders inside, e.g "tests/out/spv/1.4".

teoxoy:
I think it would be feasible to read the version from the file and map it to a vulkan version but that sounds quite involved.
We can go with a sub-folder for now - up to you.

kvark:
I think we should just merge this with some sort of a workaround, not blocking for a general solution (be it subfolders or not). Even without a test would be fine, since it's not a regression.

kvark:
@teoxoy PR now comes with the general solution that extracts SPV version from the comment

teoxoy (approved):
> PR now comes with the general solution that extracts SPV version from the comment

That's great, thanks for addressing it here!

PR #2232, opened 2023-01-31
valid: Improve forward declaration validation

JCapucho:
This PR fixes an issue with the dependency code and adds dependency validation to function calls.

The handle validator had an issue that caused it to try to build a handle with a zero based index while type erasing handles, this would crash because the index must be a `NonZeroU32` and it would also not be correct since the handle would be off by one from the original handle.

It also fixes the error message for the forward dependency not containing the subject's kind.

Lastly it adds a dependency check for function calls, to prevent forward declarations to pass to further stages and cause crashes (mitigates #2225)

### DIFF
--- a/src/valid/handles.rs
+++ b/src/valid/handles.rs
@@ -532,8 +532,8 @@ pub enum InvalidHandleError {
 
 #[derive(Clone, Debug, thiserror::Error)]
 #[error(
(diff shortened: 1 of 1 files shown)


jimblandy (approved):
Good catch - thanks very much.

PR #2233, opened 2023-01-31
Correctness and diagnostic fixes

### LINKED ISSUE #2169: [wgsl-in] Invalid assignment diagnostic doesn't extend to derived expressions
Given this code:
```rs
fn main() {
    let a = 1;
    a = 2;
}
```
Naga errors with:
```rs
error: invalid left-hand side of assignment
  ┌─ wgsl:3:2
  │
3 │     a = 2;
  │     ^ cannot assign to this expression
  │
  = note: 'a' is an immutable binding
  = note: consider declaring it with `var` instead of `let`
```
This lets us know exactly the issue and how to fix it. However, when given:
```rs
fn main() {
    let a = array(1, 2);
    a[0] = 2;
}
```
We error with:
```rs
error: invalid left-hand side of assignment
  ┌─ wgsl:3:2
  │
3 │     a[0] = 2;
  │     ^^^^ cannot assign to this expression
```
The error is the same as the above, however, we do not catch the fact that `a[0]` is derived from a `let`, and thus, don't tell the user how to fix it.

When validating assignments, we should recursively check if any of the arguments to a specific instruction is present in `ctx.named_expressions`.

### LINKED ISSUE #2209: [wgsl-in] panic on malformed input
The following program causes Naga to panic:

```
fn S()3-array<f32,0f>()=v}
```

The crash:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/front/wgsl/mod.rs:1388:83
```

### LINKED ISSUE #2216: [wgsl-in] `Parser::block` does not check for leading `{`
Given
```rs
fn S() 2}
```

`Parser::block` ignores the first token, `2`, and ends the block on seeing the `}`, essentially transforming the incorrect code into
```rs
fn S() {}
```

SparkyPotato:
- fixes #2216 
- fixes #2209 
- fixes #2169 

Also checks for the `;` after `break`, `continue`, and `discard`, as per the spec.

The diagnostic for #2209 is a bit weird:
```
error: type `array<f32, 0f>` is not constructible
  ┌─ wgsl:2:6
  │
2 │     -array<f32, 0f>() = v;
  │      ^^^^^^^^^^^^^^ type is not constructible
```
But it does point the user to where the problem is and doesn't crash.

The invalid assignment diagnostic is also improved overall:
```
error: invalid left-hand side of assignment
  ┌─ wgsl:2:9
  │
2 │     let a = array(1, 2);
  │         ^ this is an immutable binding
3 │     a[0] = 2;
  │     ^^^^ cannot assign to this expression
  │
  = note: consider declaring 'a' with `var` instead of `let`
```

### DIFF
--- a/src/front/wgsl/error.rs
+++ b/src/front/wgsl/error.rs
@@ -134,7 +134,7 @@ pub enum NumberError {
 pub enum InvalidAssignmentType {
     Other,
     Swizzle,
-    ImmutableBinding,
+    ImmutableBinding(Span),
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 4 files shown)


teoxoy (approved):
This looks great! Just need to resolve the conflict.

PR #2235, opened 2023-02-01
[msl-out] Update firstLeadingBit for signed integers

evahop:
Update `firstLeadingBit` to return `-1` when input is `0` or `-1`,
and return the index of the first set `0` bit for all other negative values.


src/back/msl/writer.rs
```diff
@@ -1721,9 +1721,56 @@ impl<W: Write> Writer<W> {
...
+                    match *inner {
+                        crate::TypeInner::Vector {
+                            size,
+                            kind: crate::ScalarKind::Sint,
+                            ..
+                        } => {
+                            let s = back::vector_size_str(size);
+                            write!(self.out, "{NAMESPACE}::select(int{s}(")?;
+                            write!(self.out, "{NAMESPACE}::floor(")?;
+                            write!(self.out, "{NAMESPACE}::log2(float{s}(")?;
```
jimblandy, line 1741: I would expect using a floating-point logarithm here to be much slower than using `metal::clz`. Could you see if that can work?

src/back/msl/writer.rs
```diff
@@ -1721,9 +1721,56 @@ impl<W: Write> Writer<W> {
...
+                            write!(self.out, "{NAMESPACE}::select(")?;
+                            self.put_expression(arg, context, false)?;
+                            write!(self.out, ", ~")?;
+                            self.put_expression(arg, context, false)?;
+                            write!(self.out, ", ")?;
+                            self.put_expression(arg, context, false)?;
+                            write!(self.out, " < 0))))), int{s}(-1), ")?;
+                            self.put_expression(arg, context, false)?;
+                            write!(self.out, " == 0 || ")?;
+                            self.put_expression(arg, context, false)?;
```
jimblandy, line 1751: One problem with using `arg` repeatedly here is that it will get repeated in the source code. If it happens to be long, that could be awkward. (If it contains further `firstLeadingBit` calls, the expansion could be exponential - unlikely, but still unpleasant.)

Instead, if we do need to use `arg` repeatedly like this, we should recognize `findMsb` in `Writer::update_expressions_to_bake`, as we do `crate::MathFunction::Dot`.

jimblandy (changes requested):

src/back/msl/writer.rs
```diff
@@ -1721,9 +1721,56 @@ impl<W: Write> Writer<W> {
                     self.put_expression(arg, context, true)?;
                     write!(self.out, ") + 1) % 33) - 1)")?;
                 } else if fun == Mf::FindMsb {
-                    write!(self.out, "((({NAMESPACE}::clz(")?;
-                    self.put_expression(arg, context, true)?;
-                    write!(self.out, ") + 1) % 33) - 1)")?
+                    let inner = context.resolve_type(arg);
+
+                    if let Some(crate::ScalarKind::Uint) = inner.scalar_kind() {
+                        write!(self.out, "((({NAMESPACE}::clz(")?;
```
jimblandy, line 1727: Not your bug, but this looks really wrong; filed as #2236.

src/back/msl/writer.rs
evahop: This is handled further along in the file

src/back/msl/writer.rs
evahop: Yeah small oversight on my part, I can fix this.

src/back/msl/writer.rs
evahop: Okay, I've simplified things, and hopefully addressed this issue.

src/back/msl/writer.rs
```diff
@@ -2275,42 +2315,42 @@ impl<W: Write> Writer<W> {
     ) {
         use crate::Expression;
         self.need_bake_expressions.clear();
-        for expr in func.expressions.iter() {
+        for (fun_handle, expr) in func.expressions.iter() {
```
jimblandy: I see that this is what the old `if let` called it, but let's just rename this `expr_handle`, since we no longer know at this point what kind of expression it is.

jimblandy (approved):
Looks good. I'm just going to rename something, and then I'll squash and merge this.

jimblandy:
Because I added a commit renaming a variable, GitHub stuck that "Co-authored" comment in there and I didn't notice it - but for the record, this PR is entirely @evahop's work.

(Force-pushing to master is a pain for others, and we've got it disabled.)

PR #2237, opened 2023-02-02
[msl-out] Replace `per_stage_map` with `per_entry_point_map`

armansito:
The existing `per_stage_map` field of MSL backend options specifies resource binding maps that apply to all entry points of each stage type. It is useful to have the ability to provide a separate binding index map for each entry point, especially when the same shader module defines multiple entry points of the same stage kind.

This patch introduces a new `per_entry_point_map` option. When fields are missing in the per-entry-point map, the code falls back to the existing `per_stage_map` option as before.

### DIFF
--- a/src/back/msl/mod.rs
+++ b/src/back/msl/mod.rs
@@ -27,10 +27,7 @@ holding the result.
 */
 
 use crate::{arena::Handle, proc::index, valid::ModuleInfo};
-use std::{
-    fmt::{Error as FmtError, Write},
-    ops,
-};
+use std::fmt::{Error as FmtError, Write};
 
 mod keywords;
 pub mod sampler;
@@ -69,7 +66,7 @@ pub type BindingMap = std::collections::BTreeMap<crate::ResourceBinding, BindTar
 #[cfg_attr(feature = "serialize", derive(serde::Serialize))]
 #[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
 #[cfg_attr(any(feature = "serialize", feature = "deserialize"), serde(default))]
-pub struct PerStageResources {
+pub struct EntryPointResources {
     pub resources: BindingMap,
 
     pub push_constant_buffer: Option<Slot>,
@@ -80,26 +77,7 @@ pub struct PerStageResources {
     pub sizes_buffer: Option<Slot>,
 }
 
-#[derive(Clone, Debug, Default, Hash, Eq, PartialEq)]
-#[cfg_attr(feature = "serialize", derive(serde::Serialize))]
-#[cfg_attr(feature = "deserialize", derive(serde::Deserialize))]
-#[cfg_attr(any(feature = "serialize", feature = "deserialize"), serde(default))]
-pub struct PerStageMap {
-    pub vs: PerStageResources,
-    pub fs: PerStageResources,
-    pub cs: PerStageResources,
-}
-
-impl ops::Index<crate::ShaderStage> for PerStageMap {
-    type Output = PerStageResources;
-    fn index(&self, stage: crate::ShaderStage) -> &PerStageResources {
-        match stage {
-            crate::ShaderStage::Vertex => &self.vs,
-            crate::ShaderStage::Fragment => &self.fs,
-            crate::ShaderStage::Compute => &self.cs,
-        }
-    }
-}
+pub type EntryPointResourceMap = std::collections::BTreeMap<String, EntryPointResources>;
 
 enum ResolvedBinding {
     BuiltIn(crate::BuiltIn),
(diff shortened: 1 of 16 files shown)


teoxoy:
I'm inclined to say that we replace `per_stage_map` with the new `per_entry_point_map`. We'll need this for wgpu soon enough too.

@jimblandy @kvark thoughts?

armansito:
I left the `per_stage_map` in place since I thought that might make it easier to transition the code that depends on it but I'm happy to remove it entirely in favor of `per_entry_point_map` if folks are supportive.

src/back/msl/mod.rs
```diff
@@ -298,10 +305,16 @@ impl Options {
...
+        ep: &crate::EntryPoint,
         res_binding: &crate::ResourceBinding,
     ) -> Result<ResolvedBinding, EntryPointError> {
-        match self.per_stage_map[stage].resources.get(res_binding) {
+        let target = self
+            .per_entry_point_map
+            .as_ref()
+            .and_then(|map| map.get(&ep.name))
+            .and_then(|res| res.resources.get(res_binding))
+            .or_else(|| self.per_stage_map[ep.stage].resources.get(res_binding));
```
teoxoy, line 316: Can we introduce a new fn on `Options` with the signature below and use it everywhere?
```rust
fn get_ep_resources<'ep>(&self, ep: &'ep EntryPoint) -> &PerStageResources
```

teoxoy (changes requested):
Thanks for the PR!

Let's postpone the removal of `per_stage_map`.
We should also rename `PerStageResources` to `EntryPointResources` when we do that to minimize breaking changes.

src/back/msl/mod.rs
armansito: Done. I changed `resolve_sizes_buffer` and `resolve_push_constants` to use the new helper. `resolve_resource_binding` is slightly different in that it falls back to `per_stage_map` only if the specific `res_binding` lookup fails (this allows the per stage map to define bindings that may not be present in the `per_entry_point_map`; which is why I left `resolve_resource_binding` as it is.

I could change the logic so that if `per_entry_point_map` is defined then `per_stage_map` gets ignored, which would be completely reasonable. I wanted to allow some flexibility but if `per_stage_map` will go away, then maybe allowing their combined use doesn't add much value.

tests/out/msl/resource-binding-map.msl
```diff
@@ -0,0 +1,70 @@
...
+, metal::sampler s [[sampler(1)]]
+, constant metal::float2& uniformOne [[buffer(0)]]
+) {
+    metal::float2 _e3 = uniformOne;
+    metal::float4 _e4 = t.sample(s, _e3);
+    return entry_point_twoOutput { _e4 };
+}
+
+
+struct entry_point_threeOutput {
```
armansito, line 50: `entry_point_three` is a case where `t` and `s` are defined in `per_stage_map` while `uniformOne` and `uniformTwo` are defined `per_entry_point_map`.

kvark:
@teoxoy sounds reasonable!

teoxoy:
@armansito since we got a +1 on removing `per_stage_map` let's do it in this PR - sorry for the back and forth

We can keep `get_ep_resources` if it makes code nicer and let's also rename `PerStageResources` to `EntryPointResources`.

Thank you!

armansito:
> 



> @armansito since we got a +1 on removing `per_stage_map` let's do it in this PR - sorry for the back and forth
> 
> We can keep `get_ep_resources` if it makes code nicer and let's also rename `PerStageResources` to `EntryPointResources`.
> 
> Thank you!

No problem! I just force pushed a new change (I erased the previous two commits since they didn't no longer made sense). PTAL!

tests/snapshots.rs
```diff
@@ -90,8 +90,11 @@ struct Parameters {
 #[allow(unused_variables)]
 fn check_targets(module: &naga::Module, name: &str, targets: Targets) {
     let root = env!("CARGO_MANIFEST_DIR");
-    let params = match fs::read_to_string(format!("{root}/{BASE_DIR_IN}/{name}.param.ron")) {
-        Ok(string) => ron::de::from_str(&string).expect("Couldn't parse param file"),
+    let filepath = format!("{root}/{BASE_DIR_IN}/{name}.param.ron");
+    let params = match fs::read_to_string(&filepath) {
+        Ok(string) => {
+            ron::de::from_str(&string).expect(&format!("Couldn't parse param file: {}", filepath))
```
armansito, line 96: Printing the failing file name made debugging a lot easier.

teoxoy: I can imagine 👍

teoxoy (approved):
Looks solid! Thanks!

PR #2239, opened 2023-02-02
spv-in: Don't apply interpolation to fragment shaders outputs

JCapucho:
SPIR-V doesn't allow the `Flat`, `NoPerspective`, `Sample` or `Centroid` decorations on fragment shaders outputs ([VUID-StandaloneSpirv-Flat-06201](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-StandaloneSpirv-Flat-06201)), but the spirv frontend was applying default interpolation to all outputs unconditionally.

This wasn't an issue for most shaders since they output floats and the default values for them don't interfere with SPIR-V semantics, but if the shader returned a uint or int the interpolation would be set to `Flat` which as stated above is disallowed.

This commit fixes the issue by only running the default interpolation code when constructing the entry point and if the stage/IO allow it.

### DIFF
--- a/src/front/spv/function.rs
+++ b/src/front/spv/function.rs
@@ -387,7 +387,18 @@ impl<I: Iterator<Item = u32>> super::Frontend<I> {
                                     None => continue,
                                     _ => {}
                                 }
-                                members.push(sm.clone());
+                                let mut sm = sm.clone();
+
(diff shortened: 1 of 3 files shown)


teoxoy (approved):
LGTM, we already do it properly for frag inputs.

PR #2240, opened 2023-02-03
fuzz: fix build failure

manunio:
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55648


teoxoy:
We should improve our CI to at least run `cargo check` on the `naga-fuzz` crate.

teoxoy:
Thank you for the PR!

PR #2243, opened 2023-02-05
Add countTrailingZeros

gents83:
Adds countTrailingZeros from #2134 

Use lsb or ctz.

It uses:
- HLSL: firstbitlow
- GLSL: findLSB
- MSL: ctz
- SPV: spirv::GLOp::FindILsb

PS: It's my first integration in Naga, so please double check it carefully :)


src/back/glsl/mod.rs
```diff
@@ -2960,6 +2961,78 @@ impl<'a, W: Write> Writer<'a, W> {
...
+                                    let s = back::vector_size_str(size);
+
+                                    if let crate::ScalarKind::Uint = kind {
+                                        write!(self.out, "uvec{s}(findLSB(")?;
+                                        self.write_expr(arg, ctx)?;
+                                        write!(self.out, ") - ivec{s}(1))")?;
+                                    } else {
+                                        write!(self.out, "mix(findLSB(")?;
+                                        self.write_expr(arg, ctx)?;
+                                        write!(self.out, ") - ivec{s}(1), ivec{s}(0), lessThan(")?;
```
evahop, line 2977: `findLSB` doesn't return `0` when the negative bit is set. So these `lessThan` checks are no longer needed.

gents83: True - We can use directly lsb

evahop:
@gents83
Please disregard my review, I did not mean to start one if i did.
The behavior of [countTrailingZeros](https://www.w3.org/TR/WGSL/#countTrailingZeros-builtin) is different than that of `findLSB`, `firstbitlow` etc. So the backends need to produce something equivalent. The majority of which produce -1 for zero where countTrailingZeros should return 32. I think you where much closer in your initial attempts.

I've updated a mistake from a previous comment to avoid confusion,
but something along the lines of `x == 0 ? 32 : findLSB - 1` seems right to me.

In the case of `findLSB` it's not supported bellow `400 & es 310` so there's a version check to ensure it's supported.

evahop:
Actually that's not quite right either, since `findLSB` index starts at zero. Subtracting `1` puts us back at `-1`.

gents83:
It should be lsb except when zero, right?

0...0001 = 0
0...0010 = 1
0...0100 = 2
....
1...0000 = 31
0...0000 = 32

Sounds correct?

teoxoy:
Sounds correct, so `GLSL`, `SPV` and `HLSL` need the following transform:

```
min(uint(findLSB(n)), 32)
```

`min` should be faster than branching here.

evahop:
I think we still need to branch, zero is going to produce `min(-1, 32)`.

gents83:
@evahop & @teoxoy  it should be fine now 
PS: sorry for the multiple commits, for some reasons I wasn't able to have makefile running correctly locally

src/back/hlsl/writer.rs
```diff
@@ -124,7 +124,8 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     | crate::MathFunction::Unpack2x16float => {
                         self.need_bake_expressions.insert(arg);
                     }
-                    crate::MathFunction::CountLeadingZeros => {
+                    crate::MathFunction::CountTrailingZeros
+                    | crate::MathFunction::CountLeadingZeros => {
```
evahop, line 128: You only have one instance of `self.write_expr(module, arg, func_ctx)?;` so baking isn't necessary here. This is further explained in https://github.com/gfx-rs/naga/pull/2226#pullrequestreview-1275659048.

src/back/hlsl/writer.rs
```diff
@@ -124,7 +124,8 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     | crate::MathFunction::Unpack2x16float => {
                         self.need_bake_expressions.insert(arg);
                     }
-                    crate::MathFunction::CountLeadingZeros => {
+                    crate::MathFunction::CountTrailingZeros
```
evahop, line 127: You only have one instance of `self.write_expr(module, arg, func_ctx)?;` so baking isn't necessary here. This is further explained in https://github.com/gfx-rs/naga/pull/2226#pullrequestreview-1275659048.

gents83: Ok clear! Tnx :)

src/back/spv/block.rs
```diff
@@ -888,6 +888,127 @@ impl<'w> BlockContext<'w> {
...
+                            &[arg_id],
+                        ));
+
+                        let cast_id = self.gen_id();
+                        block.body.push(Instruction::unary(
+                            spirv::Op::Bitcast,
+                            uint_type_id,
+                            cast_id,
+                            id,
+                        ));
```
teoxoy, line 990: No need to `Bitcast` here since `UMin` interprets both operands as unsigned.

src/back/spv/block.rs
```diff
@@ -888,6 +888,127 @@ impl<'w> BlockContext<'w> {
...
+
+                        let min_id = self.gen_id();
+                        block.body.push(Instruction::ext_inst(
+                            self.writer.gl450_ext_inst_id,
+                            spirv::GLOp::UMin,
+                            uint_type_id,
```
teoxoy, line 996: I think we can use `result_type_id` here directly, removing the need for the bitcast below.

src/back/spv/block.rs
```diff
@@ -888,6 +888,127 @@ impl<'w> BlockContext<'w> {
...
+                        let mut arg_id = arg0_id;
+                        if let Some(crate::ScalarKind::Uint) = arg_scalar_kind {
+                            arg_id = self.gen_id();
+                            block.body.push(Instruction::unary(
+                                spirv::Op::Bitcast,
+                                int_type_id,
+                                arg_id,
+                                arg0_id,
+                            ));
+                        }
```
teoxoy, line 974: I don't think we need this bitcast.

teoxoy (changes requested):
Thanks for the PR, the logic looks good!
Left a few comments regarding a few bitcasts.

teoxoy:
Also, @evahop, thanks for helping out with this PR!

src/back/spv/block.rs
```diff
@@ -888,6 +888,127 @@ impl<'w> BlockContext<'w> {
...
+                                    width,
+                                    pointer_space: None,
+                                }));
+
+                                self.temp_list.clear();
+                                self.temp_list
+                                    .resize(size as _, self.writer.get_constant_scalar(int, width));
+
+                                let id = self.gen_id();
+                                block.body.push(Instruction::constant_composite(
```
evahop, line 909: @teoxoy This isn't necessarily a urgent matter, but is `constant_composite` appropriate in these instances, and if so should be be caching them?

teoxoy: I think it's appropriate, not sure how else you'd achieve this result.
We are caching constant scalars, so it would be ideal to cache composites as well.

gents83:
Tnx for precious suggestions!

src/back/spv/block.rs
evahop: Appropriate in the sense that there's also a `composite_construct` being used to inline vectors. So would these ideally be updated as well or does it only make sense for instances where `get_constant_scalar` is used to compose the vector?

teoxoy: Ideally we would put the resulting constant composites together with the other constants.
I see you already opened a PR https://github.com/gfx-rs/naga/pull/2257 👍

src/back/spv/block.rs
```diff
@@ -888,6 +888,93 @@ impl<'w> BlockContext<'w> {
...
+                        let min_id = self.gen_id();
+                        block.body.push(Instruction::ext_inst(
+                            self.writer.gl450_ext_inst_id,
+                            spirv::GLOp::UMin,
+                            result_type_id,
+                            min_id,
+                            &[uint_id, id],
+                        ));
+
+                        return Ok(());
```
teoxoy, line 976:
```suggestion
                        MathOp::Custom(Instruction::ext_inst(
                            self.writer.gl450_ext_inst_id,
                            spirv::GLOp::UMin,
                            result_type_id,
                            id,
                            &[uint_id, lsb_id],
                        ))
```
I realized that this might not be working properly. `id` needs to be the result and `MathOp::Custom` returned since we later do `self.cached[expr_handle] = id;`.

src/back/spv/block.rs
```diff
@@ -888,6 +888,93 @@ impl<'w> BlockContext<'w> {
...
+                            _ => unreachable!(),
+                        };
+
+                        block.body.push(Instruction::ext_inst(
+                            self.writer.gl450_ext_inst_id,
+                            spirv::GLOp::FindILsb,
+                            int_type_id,
+                            id,
+                            &[arg0_id],
+                        ));
```
teoxoy, line 965:
```suggestion
                        let lsb_id = self.gen_id();
                        block.body.push(Instruction::ext_inst(
                            self.writer.gl450_ext_inst_id,
                            spirv::GLOp::FindILsb,
                            result_type_id,
                            lsb_id,
                            &[arg0_id],
                        ));
```
Need a new `lsb_id` to put our result in and we can actually reuse `result_type_id` here since the result of the lsb and the result of `countTrailingZeros` are compatible.

teoxoy (changes requested):
HLSL backend looks good now.
I noticed that the spv backend is not generating what it should.
I also opened https://github.com/gfx-rs/naga/pull/2258 to fix the `countLeadingZeros` impl which is not working properly for the same reasons.

teoxoy:
@gents83 - just a heads up that we landed https://github.com/gfx-rs/naga/pull/2257, you can now use `get_constant_composite`

gents83:
> @gents83 - just a heads up that we landed #2257, you can now use `get_constant_composite`

I cannot update my repo without discarding all changes :( Is it ok if I do another pull request right after this one is committed?

teoxoy (approved):
Looks good!

PR #2244, opened 2023-02-06
[glsl-in]: Document some fields of `naga::front::glsl::context::Context`.

jimblandy:


JCapucho:
@jimblandy needs a rebase (I tried to do it but don't have permissions)

Issue #2242, opened 2023-02-03
textureNumLevels() returns i32 in naga, but spec says u32


JMS55:
https://gpuweb.github.io/gpuweb/wgsl/#texturenumlevels

Issue #2010, opened 2022-07-16
WGSL texture query expressions return signed integers; should return unsigned


norepimorphism:
See here:
https://github.com/gfx-rs/naga/blob/17dcf39a1b2e4282943580302f510955704650f5/src/proc/typifier.rs#L533-L562

The current working and editor's drafts of the WGSL spec indicate that [`textureDimensions`](https://gpuweb.github.io/gpuweb/wgsl/#texturedimensions), [`textureNumLayers`](https://gpuweb.github.io/gpuweb/wgsl/#texturenumlayers), [`textureNumSamples`](https://gpuweb.github.io/gpuweb/wgsl/#texturenumsamples), and [`textureNumLevels`](https://gpuweb.github.io/gpuweb/wgsl/#texturenumlevels) should return unsigned coordinates.

norepimorphism:
I'll make a PR.

norepimorphism:
I'm curious how this is done for MSL. I see in [the MSL spec](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) that `<texture>::{get_width, get_height, get_num_mip_levels}` return a `uint` (p. 189). However, the code excerpt above from `typifier.rs` seems to suggest that all image queries return `Sint` regardless of the function signature defined by the frontend language's spec.

JCapucho:
Naga's IR is based on wgsl semantics, and since the `typifier` is part of the frontend it also follows wgsl semantics (note it doesn't follow each frontend language's spec), so it's the job of the backends to convert the IR into their respective languages while fixing the differences

PR #2245, opened 2023-02-07
Fix texture built-ins where u32 was expected

evahop:
fixes https://github.com/gfx-rs/naga/issues/2242
fixes https://github.com/gfx-rs/naga/issues/2010

- The Typifier was updated to expect `uint`
- Both `glsl` and `spv` frontends where updated to cast the result to `sint`.
- Both `glsl` and `spv` backends where updated to cast the result to `uint`.
- Remove cast in `msl` backend.
- Updates `wgsl` tests with incorrect types.


src/back/glsl/mod.rs
```diff
@@ -2534,6 +2534,13 @@ impl<'a, W: Write> Writer<'a, W> {
                     crate::ImageDimension::D3 => 3,
                     crate::ImageDimension::Cube => 2,
                 };
+
+                if components == 1 {
+                    write!(self.out, "uint(")?;
+                } else {
+                    write!(self.out, "uvec{components}(")?;
```
evahop, line 2541: The `spv` backend includes `array_coords` in determining the vector's size.
Not sure if this is need here as well.

tests/out/glsl/variations-glsl.main.Fragment.glsl
```diff
@@ -9,7 +9,7 @@ uniform highp samplerCube _group_0_binding_0_fs;
 void main_1() {
     ivec2 sizeCube = ivec2(0);
     float a = 0.0;
-    sizeCube = textureSize(_group_0_binding_0_fs, 0).xy;
+    sizeCube = ivec2(uvec2(textureSize(_group_0_binding_0_fs, 0).xy));
```
evahop, line 12: Because `glsl` and `spv` functions always return `sint` and, `wgsl` always returns `uints`, we get these redundant casts. Not sure if there's a way to avoid this.

src/back/glsl/mod.rs
JCapucho: If the image is arrayed this would require adding one more component to the resulting vector, but this is only for queries of the size, all other queries return a scalar and shouldn't be splatted.

JCapucho (changes requested):
Everything seems to be good implementation wise except for that comment that I left on the glsl backend code.

src/back/glsl/mod.rs
evahop: Okay, should be corrected, but feel free to make any changes if needed.

JCapucho:
@evahop Could you please rebase the branch, so that we can get the CI fixes

JCapucho (approved):
It looks good to me, but I want someone who knows the msl backend to also review since I'm not familiar with it.

PR #2250, opened 2023-02-14
[glsl/hlsl-out] Write sizes of arrays behind pointers in function arguments

### LINKED ISSUE #2248: [glsl-out] Array types used as function arguments get dropped
This shader:
```glsl
#version 460 core

void f(out vec4 foo[2]) {
    foo[1] = vec4(1.0);
}

void main() {}
```
When passed through naga (`cargo run -- shader.comp output-shader.comp`), produces this output:
```glsl
#version 310 es

precision highp float;
precision highp int;

void f(inout vec4 foo) {
    foo[1] = vec4(1.0);
    return;
}

void main_1() {
    return;
}

void main() {
    main_1();
    return;
}
```
Which doesn't compile because the argument type of `f` has changed from an array of `vec4` to a single `vec4`:
```
$ glslangValidator output-shader.comp 
output-shader.comp
ERROR: 0:10: 'assign' :  cannot convert from ' const 4-component vector of float' to ' temp highp float'
ERROR: 0:10: '' : compilation terminated 
ERROR: 2 compilation errors.  No code generated.
```

pluiedev:
Arrays can be put behind pointers in inout and out parameters in GLSL and HLSL, whose dimensions must be specified to let array access compile. So, we specify their dimensions.

Fixes #2248

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1439,8 +1439,16 @@ impl<'a, W: Write> Writer<'a, W> {
             write!(this.out, " {}", &this.names[&ctx.argument_key(i as u32)])?;
 
             // Write array size
-            if let TypeInner::Array { base, size, .. } = this.module.types[arg.ty].inner {
-                this.write_array_size(base, size)?;
+            match this.module.types[arg.ty].inner {
+                TypeInner::Array { base, size, .. } => {
+                    this.write_array_size(base, size)?;
+                }
+                TypeInner::Pointer { base, .. } => {
+                    if let TypeInner::Array { base, size, .. } = this.module.types[base].inner {
+                        this.write_array_size(base, size)?;
+                    }
+                }
+                _ => {}
             }
 
             Ok(())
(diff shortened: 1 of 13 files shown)


jimblandy (changes requested):
This looks fine, but it needs test coverage. There should be something in `tests/in/access.wgsl` that could be adapted to cover this.

pluiedev:
I'm trying to write a test now, but dealing with pointers to arrays is a bit annoying in WGSL... gimme a bit

pluiedev:
Yikes.. looks like HLSL has the same bug??!
```hlsl

void assign_array_through_ptr_fn(inout float4 foo_2)
{
    foo_2 = Constructarray2_float4_((1.0).xxxx, (2.0).xxxx);
    return;
}

[numthreads(1, 1, 1)]
void assign_through_ptr(uint3 __global_invocation_id : SV_DispatchThreadID)
{
    if (all(__global_invocation_id == uint3(0u, 0u, 0u))) {
        val = (uint)0;
    }
    GroupMemoryBarrierWithGroupSync();
    float4 arr[2] = (float4[2])0;

    arr = Constructarray2_float4_((6.0).xxxx, (7.0).xxxx);
    assign_through_ptr_fn(val);
    assign_array_through_ptr_fn(arr);
    return;
}
````

jimblandy (approved):
Thanks!

jimblandy (approved):

PR #2253, opened 2023-02-15
Require LF for snapshots

grovesNL:
Depending on how you have your gitconfig set up, running the snapshot tests can cause a diff across every snapshot on Windows where line endings might switch between LF and CRLF. Instead we can force LF for everything in the `tests/out` directory which should avoid this problem.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+tests/out/**/* text eol=lf


teoxoy:
Is this needed because the snapshots are always generated with LF line endings?

grovesNL:
Yeah exactly. It's been a problem since we first introduced them and I think others have mentioned running into this before.

teoxoy (approved):
LGTM

Issue #1161, opened 2021-08-03
Change frexp and modf to work with structures


kvark:
See https://github.com/gpuweb/gpuweb/pull/1973
I think this change needs to be done at IR level.

kvark:
Similar work needs to be done for atomic compare+exchange - https://github.com/gpuweb/gpuweb/pull/2113
I think all of this should wait for #1419 first.

kvark:
This isn't urgent for the release.

cwfitzgerald:
Currently bumping up against this when targeting WebGPU, as wgpu's water example uses modf.

kvark:
This should be simpler now that #2256 is in, which contains infra to create IR structure types.

PR #2256, opened 2023-02-17
Ray Query support

kvark:
Path support:
- [x] WGSL-in
- [x] validation
- [x] SPV-out
- [x] MSL-out
- [ ] HLSL-out (not planned for this PR)

Note: PR also adds basic infra for specially generated types in IR, which may be handy for #1161


src/valid/interface.rs
```diff
@@ -440,7 +440,9 @@ impl super::Validator {
                 match types[var.ty].inner {
                     crate::TypeInner::Image { .. }
                     | crate::TypeInner::Sampler { .. }
-                    | crate::TypeInner::BindingArray { .. } => {}
+                    | crate::TypeInner::BindingArray { .. }
+                    | crate::TypeInner::AccelerationStructure
+                    | crate::TypeInner::RayQuery => {}
```
jimblandy, line 445: Can `RayQuery` handles really be global? I thought they were always local variables.

src/valid/expression.rs
```diff
@@ -1427,6 +1427,7 @@ impl super::Validator {
                     return Err(ExpressionError::InvalidArrayType(expr));
                 }
             },
+            E::RayQueryProceedResult | E::RayQueryGetIntersection { .. } => ShaderStages::all(),
```
jimblandy, line 1430: Can we fill in the type checks here?

src/front/wgsl/lower/mod.rs
```diff
@@ -2378,4 +2434,28 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
+        let span = ctx.ast_expressions.get_span(expr);
+        let pointer = self.expression(expr, ctx.reborrow())?;
+
+        ctx.grow_types(pointer)?;
+        match *ctx.resolved_inner(pointer) {
+            crate::TypeInner::Pointer { base, .. } => match ctx.module.types[base].inner {
+                crate::TypeInner::RayQuery => Ok(pointer),
+                ref other => {
+                    log::error!("Pointer type to {:?} passed to ray query op", other);
+                    Err(Error::InvalidAtomicPointer(span))
```
jimblandy, line 2452: You probably don't want to use `InvalidAtomicPointer` here.

src/back/spv/writer.rs
```diff
@@ -961,9 +973,9 @@ impl Writer {
 
                     self.write_type_declaration_local(id, local);
 
-                    // If it's an image type, request SPIR-V capabilities here, so
-                    // write_type_declaration_local can stay infallible.
-                    self.request_image_capabilities(&ty.inner)?;
+                    // If it's an type that needs SPIR-V capabilities, request them now,
```
jimblandy, line 976:
```suggestion
                    // If it's a type that needs SPIR-V capabilities, request them now,
```

jimblandy (changes requested):
This is interesting. I don't fully understand the ray tracing API, but I've tried to follow what's going on as best I could. Just a few points to address.

src/lib.rs
```diff
@@ -1743,6 +1778,19 @@ pub struct EntryPoint {
...
+#[derive(Debug, Default)]
+#[cfg_attr(feature = "clone", derive(Clone))]
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "deserialize", derive(Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+pub struct SpecialTypes {
+    /// Type for `RayDesc`.
+    ray_desc: Option<Handle<Type>>,
+    /// Type for `RayIntersection`.
+    ray_intersection: Option<Handle<Type>>,
```
jimblandy, line 1791: Should these fields be `pub` too? Generally we want to support third-party producers and consumers of `Module`.

jimblandy (changes requested):

jimblandy:
@kvark By the way, I went ahead and pushed a commit that expands on the documentation, so be sure to pull before working on the branch. If the docs aren't accurate, feel free to just take the commit out.

src/valid/function.rs
```diff
@@ -801,6 +801,9 @@ impl super::Validator {
                 } => {
                     self.validate_atomic(pointer, fun, value, result, context)?;
                 }
+                S::RayQuery { query: _, fun: _ } => {
+                    //TODO
```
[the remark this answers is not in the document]
kvark: I had actually implemented validation in #2282 follow-up, now cherry-picking it here

src/front/wgsl/lower/mod.rs
```diff
@@ -644,14 +637,21 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
+        for decl_handle in self.index.visit_ordered() {
+            let span = tu.decls.get_span(decl_handle);
+            let decl = &tu.decls[decl_handle];
+
+            //TODO: find a nicer way?
+            if let Some(dep) = decl.dependencies.iter().find(|dep| dep.ident == "RayDesc") {
+                let ty_handle = ctx.module.generate_ray_desc_type();
+                ctx.globals
+                    .insert(dep.ident, LoweredGlobalDecl::Type(ty_handle));
+            }
```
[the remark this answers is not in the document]
kvark: Adding a comment. The issue can be seen on the `ray-query.wgsl` test if you comment this block out.

src/back/mod.rs
```diff
@@ -218,3 +218,26 @@ impl crate::Statement {
...
+bitflags::bitflags! {
+    /// Ray flags.
+    #[derive(Default)]
+    pub struct RayFlag: u32 {
+        const OPAQUE = 0x01;
+        const NO_OPAQUE = 0x02;
+        const TERMINATE_ON_FIRST_HIT = 0x04;
+        const SKIP_CLOSEST_HIT_SHADER = 0x08;
+        const CULL_FRONT_FACING = 0x10;
+        const CULL_BACK_FACING = 0x20;
```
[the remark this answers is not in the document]
kvark: Nicely spotted! Unsure how I got this wrong 😅

src/valid/interface.rs
kvark: I didn't test this comprehensively. I assume they can be private globals too.

kvark:
@jimblandy thanks for the review! Everything should be addressed now.
I'd also like to hear your thoughts on naming, i.e. `ray_desc` vs `RayDesc`.

jimblandy (approved):
Looks good to me. Give my `RayDesc` fix a look, and then feel free to merge if it seems okay.

kvark:
Much nicer than my version, thank you!

PR #2257, opened 2023-02-20
[spv-out] Cache constant composites

evahop:
Updates `cached_constants` to include composites.


teoxoy (approved):
This looks great! Thanks!

PR #2258, opened 2023-02-20
[spv-out] fix `countLeadingZeros` impl

teoxoy:
- `id` should have had the result of the last operation
- `MathOp::Custom` should have been returned since we later execute: `self.cached[expr_handle] = id;`
- removed unnecessary bitcast

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -951,33 +951,22 @@ impl<'w> BlockContext<'w> {
                             _ => unreachable!(),
                         };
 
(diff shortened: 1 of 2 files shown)


evahop:
Sorry about that. This was done in tandem with the `glsl` backend. Looks like aspects of that ended up over here without too much consideration.

teoxoy:
No worries, I haven't noticed it either.
SPV validation tests also didn't catch it because we never used the returned value.

jimblandy (approved):
Yes, looks better now.

PR #2259, opened 2023-02-20
Fix zero initialization of workgroup memory

teoxoy:
Use local (not global) invocation id - duh 🙄.
Credits go to @nical who found out the issue!


src/back/hlsl/writer.rs
```diff
@@ -1176,10 +1176,7 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                         if !func.arguments.is_empty() {
                             write!(self.out, ", ")?;
                         }
-                        write!(
-                            self.out,
-                            "uint3 __global_invocation_id : SV_DispatchThreadID"
-                        )?;
+                        write!(self.out, "uint3 __global_invocation_id : SV_GroupThreadID")?;
```
jimblandy, line 1179: We probably want to change the HLSL variable name as well. It is no longer a "global invocation id".

Pre-existing, but: What happens if the entry point already has an argument that is the `BuiltIn::LocalInvocationId`? Can HLSL handle having two `SV_GroupThreadId` arguments on the same function?

src/back/msl/writer.rs
```diff
@@ -3695,7 +3695,7 @@ impl<W: Write> Writer<W> {
                 };
                 writeln!(
                     self.out,
-                    "{separator} {NAMESPACE}::uint3 __global_invocation_id [[thread_position_in_grid]]"
+                    "{separator} {NAMESPACE}::uint3 __global_invocation_id [[thread_position_in_threadgroup]]"
```
jimblandy, line 3698: Similarly, let's rename this `__local_invocation_id`.

src/back/msl/writer.rs
```diff
@@ -3659,7 +3659,7 @@ impl<W: Write> Writer<W> {
                     &self.names[name_key]
                 };
 
-                if binding == &crate::Binding::BuiltIn(crate::BuiltIn::GlobalInvocationId) {
+                if binding == &crate::Binding::BuiltIn(crate::BuiltIn::LocalInvocationId) {
```
jimblandy, line 3662: Let's rename the Rust variable to `local_invocation_id`.

jimblandy (changes requested):
Sorry I missed this in the first review.

src/back/hlsl/writer.rs
teoxoy: > Pre-existing, but: What happens if the entry point already has an argument that is the `BuiltIn::LocalInvocationId`? Can HLSL handle having two `SV_GroupThreadId` arguments on the same function?

Yeah, only msl and spv seem to complain.

jimblandy (approved):

src/back/spv/writer.rs
```diff
@@ -430,7 +430,7 @@ impl Writer {
                         constituent_ids.push(id);
 
                         if binding == &crate::Binding::BuiltIn(crate::BuiltIn::GlobalInvocationId) {
```
DJMcNab, line 432: I'm digging into this again for gfx-rs/wgpu#4592. Why was this one left as Global?

teoxoy: It seems I missed this one 😅

PR #2262, opened 2023-02-26
enforce discard only in fragment

Uriopass:
Fixes https://github.com/gfx-rs/naga/issues/2263

as per [wgsl spec](https://www.w3.org/TR/WGSL/#discard-statement) "The discard statement must only be used in a fragment shader stage."


teoxoy (approved):
Looks good! Thank you for the PR!

PR #2264, opened 2023-03-03
Validate vertex entry point returns the position built-in

teoxoy:
fixes https://github.com/gfx-rs/wgpu/issues/3550


src/valid/interface.rs
```diff
@@ -353,7 +355,6 @@ impl VaryingContext<'_> {
                 .map_err(|e| e.with_span_context(span_context)),
             None => {
                 match self.types[ty].inner {
-                    //TODO: check the member types
```
jimblandy, line 356: Thanks for removing this. It's covered in `validate_impl` now, right?

src/valid/interface.rs
```diff
@@ -369,7 +370,6 @@ impl VaryingContext<'_> {
                                     #[cfg(not(feature = "validate"))]
                                     let _ = index;
                                 }
-                                // TODO: shouldn't this be validate?
```
jimblandy, line 372: I think this TODO meant, shouldn't this be conditional on the `"validate"` feature. I think we shouldn't be calling this function at all when `"validate"` isn't enabled. But that's not something for this PR; filed as #2302.

src/valid/interface.rs
```diff
@@ -545,6 +545,33 @@ impl super::Validator {
             return Err(EntryPointError::UnexpectedWorkgroupSize.with_span());
         }
 
+        #[cfg(feature = "validate")]
+        if ep.stage == crate::ShaderStage::Vertex {
+            let has_position = match ep.function.result {
```
jimblandy, line 550: If you check for this just after calling `VaryingContext::validate` on `ep.function.result`, then `crate::BuiltIn::Position { invariant: false }` will be in `result_built_ins` if it was present anywhere on the entry point's return value. You'll have to arrange to check even when `ep.function.result` is `None`, but even so, checking that might be simpler than doing a scan here.

tests/snapshots.rs
```diff
@@ -117,7 +117,7 @@ fn check_targets(module: &naga::Module, name: &str, targets: Targets) {
 
     let info = naga::valid::Validator::new(naga::valid::ValidationFlags::all(), capabilities)
         .validate(module)
-        .expect("Naga module validation failed");
+        .expect(&format!("Naga module validation failed on test '{name}'"));
```
jimblandy, line 120: :heart:

tests/out/wgsl/binop-frag.wgsl
jimblandy, line 1: Was this just a random file abandoned in the output directory?

jimblandy (changes requested):
I was wondering if it was right to enforce this on all modules, not just those derived from WGSL. But:

- GLSL says the vertex position is undefined if the vertex shader doesn't assign to `gl_Position`, so it seems like any vertex shader that doesn't do so should probably be flagged by Naga. 

- I can't find anything in the Vulkan or SPIR-V specifications that justifies enforcing this rule on SPIR-V input, but I also don't see the harm.

So, I guess we might as well go with this and see if anyone complains. Thanks for updating all the tests.

I think there's a better way to do the check.

src/valid/interface.rs
teoxoy: Ah, I see. I thought it meant using the validate fn.

src/valid/interface.rs
teoxoy: Yep

tests/out/glsl/multiview.main.Fragment.glsl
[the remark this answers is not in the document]
teoxoy: This is an odd one, the input test file for this should have been `tests/in/multiview.wgsl` which never wrote to the position.

src/valid/interface.rs
teoxoy: Good idea 👍

jimblandy (approved):
Thanks!

PR #2265, opened 2023-03-03
Split image bounds check policy

teoxoy:
Split image bounds check policy since we shouldn't need bounds checks for writes.
See https://github.com/gpuweb/gpuweb/issues/3893#issuecomment-1453847182.

### DIFF
--- a/src/proc/index.rs
+++ b/src/proc/index.rs
@@ -109,14 +109,24 @@ pub struct BoundsCheckPolicies {
     /// How should the generated code handle image texel references that are out
     /// of range?
     ///
-    /// This controls the behavior of [`ImageLoad`] expressions and
-    /// [`ImageStore`] statements when a coordinate, texture array index, level
-    /// of detail, or multisampled sample number is out of range.
+    /// This controls the behavior of [`ImageLoad`] expressions when a coordinate,
+    /// texture array index, level of detail, or multisampled sample number is out of range.
     ///
     /// [`ImageLoad`]: crate::Expression::ImageLoad
+    #[cfg_attr(feature = "deserialize", serde(default))]
+    pub image_load: BoundsCheckPolicy,
+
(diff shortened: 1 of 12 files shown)


jimblandy:
Since all the backends ignore writes anyway, the only thing one can do with an image write policy is ask for `Restrict`, which seems like something nobody actually wants.

Wouldn't it make more sense to just rename `BoundsCheckPolicy::image` to `image_load`, and remove bounds checking on `ImageStore` altogether?

teoxoy:
I wanted to first see the guarantee landed in the spec. Put up: https://github.com/gpuweb/gpuweb/pull/4194

src/proc/index.rs
```diff
@@ -109,14 +109,24 @@ pub struct BoundsCheckPolicies {
     /// How should the generated code handle image texel references that are out
```
jimblandy, line 109:
```suggestion
    /// How should the generated code handle image texel loads that are out
```

src/proc/index.rs
```diff
@@ -109,14 +109,24 @@ pub struct BoundsCheckPolicies {
...
-    /// [`ImageStore`] statements when a coordinate, texture array index, level
-    /// of detail, or multisampled sample number is out of range.
+    /// This controls the behavior of [`ImageLoad`] expressions when a coordinate,
+    /// texture array index, level of detail, or multisampled sample number is out of range.
     ///
     /// [`ImageLoad`]: crate::Expression::ImageLoad
+    #[cfg_attr(feature = "deserialize", serde(default))]
+    pub image_load: BoundsCheckPolicy,
+
+    /// How should the generated code handle image texel references that are out
```
jimblandy, line 119:
```suggestion
    /// How should the generated code handle image texel stores that are out
```

jimblandy (approved):
Looks good to me. We can just drop image_store if that comes to pass.

PR #2273, opened 2023-03-10
[glsl-in] Switch implicit type conversion

evahop:
GLSL allows for implicitly converting `int` -> `uint` in switch cases.


JCapucho (approved):
Looks good, nice work :)

PR #2277, opened 2023-03-14
Add partial derivative builtins

evahop:
Adds the following from https://github.com/gfx-rs/naga/issues/2134
- dpdxCoarse
- dpdxFine
- dpdyCoarse
- dpdyFine
- fwidthCoarse
- fwidthFine

Feel free to make any changes btw.

### DIFF
--- a/src/back/dot/mod.rs
+++ b/src/back/dot/mod.rs
@@ -505,7 +505,11 @@ fn write_function_expressions(
                 edges.insert("reject", reject);
                 ("Select".into(), 3)
             }
-            E::Derivative { axis, expr } => {
+            E::Derivative {
+                axis,
+                ctrl: _,
+                expr,
+            } => {
                 edges.insert("", expr);
                 (format!("d{axis:?}").into(), 8)
             }
--- a/src/back/dot/mod.rs
+++ b/src/back/dot/mod.rs
@@ -507,11 +507,11 @@ fn write_function_expressions(
             }
             E::Derivative {
                 axis,
-                ctrl: _,
+                ctrl,
                 expr,
             } => {
                 edges.insert("", expr);
-                (format!("d{axis:?}").into(), 8)
+                (format!("d{axis:?}{ctrl:?}").into(), 8)
             }
             E::Relational { fun, argument } => {
                 edges.insert("arg", argument);
--- a/src/back/dot/mod.rs
+++ b/src/back/dot/mod.rs
@@ -505,11 +505,7 @@ fn write_function_expressions(
                 edges.insert("reject", reject);
                 ("Select".into(), 3)
             }
-            E::Derivative {
-                axis,
-                ctrl,
-                expr,
-            } => {
+            E::Derivative { axis, ctrl, expr } => {
                 edges.insert("", expr);
                 (format!("d{axis:?}{ctrl:?}").into(), 8)
             }
--- a/src/back/hlsl/writer.rs
+++ b/src/back/hlsl/writer.rs
@@ -133,6 +133,13 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     _ => {}
                 }
             }
+
+            if let Expression::Derivative { axis, ctrl, expr } = *expr {
+                use crate::{DerivativeAxis as Axis, DerivativeControl as Ctrl};
+                if axis == Axis::Width && (ctrl == Ctrl::Coarse || ctrl == Ctrl::Fine) {
+                    self.need_bake_expressions.insert(expr);
+                }
+            }
         }
     }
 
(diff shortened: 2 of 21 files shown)


evahop:
Clippy is freaking out for me locally, so I'm just force pushing now that I've appeased the CI

src/lib.rs
```diff
@@ -976,15 +976,26 @@ pub enum AtomicFunction {
...
 #[cfg_attr(feature = "serialize", derive(Serialize))]
 #[cfg_attr(feature = "deserialize", derive(Deserialize))]
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
 pub enum DerivativeAxis {
-    X,
-    Y,
-    Width,
+    X(DerivativeControl),
+    Y(DerivativeControl),
+    Width(DerivativeControl),
```
teoxoy, line 998: I think it would be better to have `DerivativeControl` as a member of `Expression::Derivative`.

teoxoy (approved):
Thanks for the PR!
Overall it looks good!

One nit: Please try to to avoid using `_` for match arms especially when that coincides with one remaining enum variant.

teoxoy (changes requested):
*misclicked*

src/lib.rs
evahop: This does seem kind of clunky now, will do!

evahop:
> Thanks for the PR! Overall it looks good!
> 
> One nit: Please try to to avoid using `_` for match arms especially when that coincides with one remaining enum variant.

Sounds good, does this also apply to spreading the pattern if there's only one remaining field `{ x , .. }` vs `{ x, y: _ }`?

src/front/spv/mod.rs
```diff
@@ -1123,7 +1123,7 @@ impl<I: Iterator<Item = u32>> Frontend<I> {
         block: &mut crate::Block,
         block_id: spirv::Word,
         body_idx: usize,
-        axis: crate::DerivativeAxis,
+        (axis, ctrl): (crate::DerivativeAxis, crate::DerivativeControl),
```
evahop, line 1126: I got hit with `too many args` from clippy. Not sure if this is a proper solution or not

teoxoy:
> Sounds good, does this also apply to resting the pattern if there's only one remaining field `{ x , .. }` vs `{ x, y: _ }`?

I think that's fine and depends on context; sometimes it might be worth explicitly ignoring members, sometimes we might not care at all and use the rest operator.

src/back/dot/mod.rs
```diff
@@ -505,7 +505,11 @@ fn write_function_expressions(
                 edges.insert("reject", reject);
                 ("Select".into(), 3)
             }
-            E::Derivative { axis, expr } => {
+            E::Derivative {
+                axis,
+                ctrl: _,
```
teoxoy, line 510: I think we should emit this as well.

src/back/hlsl/writer.rs
```diff
@@ -2801,13 +2801,20 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                        Ctrl::Coarse => "ddx_coarse",
+                        Ctrl::Fine => "ddx_fine",
+                        Ctrl::None => "ddx",
+                    },
+                    Axis::Y => match ctrl {
+                        Ctrl::Coarse => "ddy_coarse",
+                        Ctrl::Fine => "ddy_fine",
+                        Ctrl::None => "ddy",
+                    },
+                    Axis::Width => "fwidth",
```
teoxoy, line 2817: We probably have to polyfill this on HLSL for the Coarse/Fine variants.

src/back/spv/block.rs
```diff
@@ -1334,15 +1334,35 @@ impl<'w> BlockContext<'w> {
...
+                    Axis::Y => match ctrl {
+                        Ctrl::Coarse => spirv::Op::DPdyCoarse,
+                        Ctrl::Fine => spirv::Op::DPdyFine,
+                        Ctrl::None => spirv::Op::DPdy,
+                    },
+                    Axis::Width => match ctrl {
+                        Ctrl::Coarse => spirv::Op::FwidthCoarse,
+                        Ctrl::Fine => spirv::Op::FwidthFine,
+                        Ctrl::None => spirv::Op::Fwidth,
+                    },
```
teoxoy, line 1365: It would be nicer if we only had one match here, testing for both the axis and ctrl at the same time (i.e. `match (axis, ctrl)`). This applies to a few other places in the other backends as well.

teoxoy (changes requested):
A rebase should take care of the linting issues; since we landed https://github.com/gfx-rs/naga/pull/2279.

src/back/spv/block.rs
evahop: Appreciate the feedback, will get to these shortly

src/back/dot/mod.rs
evahop: I'm not familiar with `dot` at all so I'd just be guessing here. If you wouldn't mind filling in this bit, everything else has been updated. 🤘

src/back/dot/mod.rs
```diff
@@ -505,7 +505,11 @@ fn write_function_expressions(
...
+                expr,
+            } => {
                 edges.insert("", expr);
                 (format!("d{axis:?}").into(), 8)
```
teoxoy, line 514:
```suggestion
                ctrl,
                expr,
            } => {
                edges.insert("", expr);
                (format!("d{axis:?}{ctrl:?}").into(), 8)
```

src/back/dot/mod.rs
```diff
@@ -505,9 +505,13 @@ fn write_function_expressions(
                 edges.insert("reject", reject);
                 ("Select".into(), 3)
             }
-            E::Derivative { axis, expr } => {
+            E::Derivative {
+                axis,
+                ctrl,
+                expr,
+            } => {
```
teoxoy, line 512:
```suggestion
            E::Derivative { axis, ctrl, expr } => {
```

teoxoy (approved):
Looks great - thank you for the improvements!

PR #1674, opened 2022-01-18
msl: inject point size if needed

kvark:
Fixes #1673

PR #2280, opened 2023-03-15
Rename `allow_point_size` to `allow_and_force_point_size`

teoxoy:
Follow-up to https://github.com/gfx-rs/naga/pull/1674 where the behavior changed but the name and docs haven't.


jimblandy (approved):
Looks good to me.

PR #2279, opened 2023-03-15
Enable match related lints

teoxoy:

PR #2281, opened 2023-03-15
fix(hlsl): emit constructor functions for arrays

### LINKED ISSUE #2184: [hlsl-out] Arrays in Structs Are Incorrectly Loaded When Loaded as A Whole Struct
## WGSL

```rust
struct Ah {
    inner: array<f32, 2>,
};
@group(0) @binding(0)
var<storage> ah: Ah;

@compute @workgroup_size(1)
fn cs_main() {
    _ = ah;
}
```

## HLSL

```hlsl
struct Ah {
    float inner[2];
};

ByteAddressBuffer ah : register(t0);

Ah ConstructAh(float arg0[2]) {
    Ah ret = (Ah)0;
    ret.inner = arg0;
    return ret;
}

[numthreads(1, 1, 1)]
void cs_main()
{
    Ah unnamed = ConstructAh({asfloat(ah.Load(0+0)), asfloat(ah.Load(0+4))});
}
```

## Solution

As far as I can tell, the only real solution is to make the array part of the initialization a temporary. I wish there was a way to initialize an array inline, but I don't think you can. I'm not entirely sure how to do it but this code is the code that needs to be assigned to a temporary. https://github.com/gfx-rs/naga/blob/2754a0136ecd4d371af5a85fc715c9f8f4c64fc1/src/back/hlsl/storage.rs#L145-L156. 

```hlsl
[numthreads(1, 1, 1)]
void cs_main()
{
    float temp[2] = {asfloat(ah.Load(0+0)), asfloat(ah.Load(0+4))};
    Ah unnamed = ConstructAh(temp);
}
```

ErichDonGubler:
Fixes #2184.


ErichDonGubler:
Looks like CI is only failing on https://github.com/gfx-rs/naga/pull/2279. 😮‍💨👍🏻

teoxoy:
> Looks like CI is only failing on #2279. 😮‍💨👍🏻

A rebase should get it working.

teoxoy (commented):
LGTM, but let's add/extend one of the test snapshots with the code that was previously not getting translated correctly

ErichDonGubler:
@teoxoy: Added a new test! I broke out a commit for adding a deliberately failing test case in HLSL, and then fixing it in the subsequent commit. Not sure if that's how you'd like to commit it, but I'm open to changing it if you want. 😊

teoxoy (approved):
Good stuff!

PR #2282, opened 2023-03-16
Support array bindings of buffers

### LINKED ISSUE #1864: Storage buffers in binding arrays aren't writable
Since indexing a `BufferArray` never produces a pointer, it's impossible to write to the contents of buffers accessed through a buffer array.

Permitting pointers to buffer arrays seems like a bad idea. Perhaps indexing a `BufferArray` whose base type is not `Image` or `Sampler` should just always produce a pointer to the buffer's value, as if it were an ordinary `var<storage>` or `var<uniform>`, even when the `base` operand to `Access` or `AccessIndex` is not a pointer.

cc @cwfitzgerald

kvark:
Built upon #2256 , which should land first. I can also move this one out (so far just one commit).
Only WGSL -> IR -> SPV path supported right now.
Closes #1864

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -196,9 +196,8 @@ impl<'w> BlockContext<'w> {
     fn is_intermediate(&self, expr_handle: Handle<crate::Expression>) -> bool {
         match self.ir_function.expressions[expr_handle] {
             crate::Expression::GlobalVariable(handle) => {
-                let ty = self.ir_module.global_variables[handle].ty;
-                match self.ir_module.types[ty].inner {
-                    crate::TypeInner::BindingArray { .. } => false,
+                match self.ir_module.global_variables[handle].space {
+                    crate::AddressSpace::Handle => false,
                     _ => true,
                 }
             }
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -3,8 +3,9 @@ Implementations for `BlockContext` methods.
 */
 
 use super::{
-    index::BoundsCheckResult, make_local, selection::Selection, Block, BlockContext, Dimension,
-    Error, Instruction, LocalType, LookupType, LoopContext, ResultMember, Writer, WriterFlags,
+    helpers, index::BoundsCheckResult, make_local, selection::Selection, Block, BlockContext,
+    Dimension, Error, Instruction, LocalType, LookupType, LoopContext, ResultMember, Writer,
+    WriterFlags,
 };
 use crate::{arena::Handle, proc::TypeResolution};
 use spirv::Word;
@@ -220,7 +221,6 @@ impl<'w> BlockContext<'w> {
         block: &mut Block,
     ) -> Result<(), Error> {
         let result_type_id = self.get_expression_type_id(&self.fun_info[expr_handle].ty);
-
         let id = match self.ir_function.expressions[expr_handle] {
             crate::Expression::Access { base, index: _ } if self.is_intermediate(base) => {
                 // See `is_intermediate`; we'll handle this later in
@@ -236,9 +236,15 @@ impl<'w> BlockContext<'w> {
                     crate::TypeInner::BindingArray {
                         base: binding_type, ..
                     } => {
+                        let space = match self.ir_function.expressions[base] {
+                            crate::Expression::GlobalVariable(gvar) => {
+                                self.ir_module.global_variables[gvar].space
+                            }
+                            _ => unreachable!(),
+                        };
                         let binding_array_false_pointer = LookupType::Local(LocalType::Pointer {
                             base: binding_type,
-                            class: spirv::StorageClass::UniformConstant,
+                            class: helpers::map_storage_class(space),
                         });
 
                         let result_id = match self.write_expression_pointer(
@@ -264,15 +270,6 @@ impl<'w> BlockContext<'w> {
                             None,
                         ));
 
-                        if self.fun_info[index].uniformity.non_uniform_result.is_some() {
-                            self.writer.require_any(
-                                "NonUniformEXT",
-                                &[spirv::Capability::ShaderNonUniform],
-                            )?;
-                            self.writer.use_extension("SPV_EXT_descriptor_indexing");
-                            self.writer
-                                .decorate(load_id, spirv::Decoration::NonUniform, &[]);
-                        }
                         load_id
                     }
                     ref other => {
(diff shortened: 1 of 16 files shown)


src/proc/typifier.rs
```diff
@@ -298,6 +298,7 @@ impl<'a> ResolveContext<'a> {
                             width,
                             space,
                         },
+                        Ti::BindingArray { base, .. } => Ti::Pointer { base, space },
```
jimblandy, line 301: There should never be a `Pointer` to a `BindingArray`, right? Why isn't the non-pointer `BindingArray` case that's already here good enough?

(Similar question about the similar change below.)

jimblandy (commented):
Just a few comments so far:

src/proc/typifier.rs
kvark: For all globals in non-`Handle` address space the resolved type is a pointer. See the typifier code.

src/valid/type.rs
```diff
@@ -630,7 +630,12 @@ impl super::Validator {
                 self.require_type_capability(Capabilities::RAY_QUERY)?;
                 TypeInfo::new(TypeFlags::DATA | TypeFlags::SIZED, Alignment::ONE)
             }
-            Ti::BindingArray { .. } => TypeInfo::new(TypeFlags::empty(), Alignment::ONE),
+            Ti::BindingArray { base, size: _ } => {
+                if base >= handle {
+                    return Err(TypeError::UnresolvedBase(base));
+                }
```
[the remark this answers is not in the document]
kvark: We are still doing these checks for `Ti::Pointer` and `Ti::Array`, so this one is there for consistency. If we want to remove them all, that's fine by me, but should be done separately.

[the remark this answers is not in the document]
jimblandy: We do want to remove them all, since they're redundant. There's no reason to make more work for the person who has to clean that up. Filed as #2303.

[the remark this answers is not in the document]
jimblandy: This doesn't need to be removed in this PR if you don't want. It's not like we'll be able to miss it when we fix #2303.

src/valid/type.rs
```diff
@@ -630,7 +630,12 @@ impl super::Validator {
...
+                self.types[base.index()].clone()
```
jimblandy, line 637: The old code wasn't right either, but it at least had the merit of at least not promising things about the type that weren't true, like `TypeFlags::ARGUMENT`. Can we make this accurate?

kvark: good point, thanks for your diligence! take a look at the new commit please

jimblandy (commented):
I see why `TypeFlags::DATA` is needed, but I'd also like us to pass this test, in the `wgsl-errors.rs` style:

```
#[test]
fn binding_array_local() {
    check_validation! {
        "fn f() { var x: binding_array<i32, 4>; }":
        Err(_)
    }
}
```

kvark:
@jimblandy take a look at the new commit. We aren't putting DATA on the binding arrays any more.

jimblandy:
How about this one:

```

#[test]
fn binding_array_private() {
    check_validation! {
        "var<private> x: binding_array<i32, 4>;":
        Err(_)
    }
}
```

src/valid/interface.rs
```diff
@@ -400,7 +400,11 @@ impl super::Validator {
         use super::TypeFlags;
 
         log::debug!("var {:?}", var);
-        let type_info = &self.types[var.ty.index()];
+        let inner_ty = match types[var.ty].inner {
+            crate::TypeInner::BindingArray { base, .. } => base,
```
jimblandy, line 404: Suggest:

```suggestion
            // A binding array is (mostly) supposed to behave the same as a
            // series of individually bound resources, so we can (mostly)
            // validate a `binding_array<T>` as if it were just a plain `T`.
            crate::TypeInner::BindingArray { base, .. } => base,
```

jimblandy (changes requested):
I think that test case I posted above shows that `validate_global_var` can't quite treat a `binding_array<T>` just like a `T`, although there's good reason to think that's usually the right thing.

src/back/spv/mod.rs
```diff
@@ -291,6 +291,7 @@ enum LocalType {
     PointerToBindingArray {
```
jimblandy, line 291: Pre-existing, but suggest:

```suggestion
    /// Equivalent to a [`LocalType::Pointer`] whose `base` is a Naga IR [`BindingArray`]. SPIR-V
    /// permits duplicated `OpTypePointer` ids, so it's fine to have two different [`LocalType`]
    /// representations for pointer types.
    ///
    /// [`BindingArray`]: crate::TypeInner::BindingArray
    PointerToBindingArray {
```

src/back/spv/block.rs
```diff
@@ -196,9 +197,8 @@ impl<'w> BlockContext<'w> {
     fn is_intermediate(&self, expr_handle: Handle<crate::Expression>) -> bool {
         match self.ir_function.expressions[expr_handle] {
             crate::Expression::GlobalVariable(handle) => {
-                let ty = self.ir_module.global_variables[handle].ty;
-                match self.ir_module.types[ty].inner {
-                    crate::TypeInner::BindingArray { .. } => false,
+                match self.ir_module.global_variables[handle].space {
+                    crate::AddressSpace::Handle => false,
```
jimblandy, line 201: Much nicer, thanks.

jimblandy:
Is this comment in `back/spv/block.rs` still accurate?

```
    /// Some cases we need to generate a different return type than what the IR gives us.
    /// This is because pointers to binding arrays don't exist in the IR, but we need to
    /// create them to create an access chain in SPIRV.
```

jimblandy (changes requested):
Other than these issues, and the additional test I mentioned above, this looks good to go.

src/valid/interface.rs
kvark: nice, thanks!

kvark:
@jimblandy thanks for suggestions! New validation error is added, tests improved, and documentation patches applied.

src/valid/type.rs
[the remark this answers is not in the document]
teoxoy: CI seems to now be failing due to this since we already merged https://github.com/gfx-rs/naga/pull/2308.

jimblandy (approved):
Thank you!

PR #2285, opened 2023-03-18
Update name of clippy::derived_hash_with_manual_eq

dneto0:


dneto0:
Ha it seems what exists is "working as intended"?  The CI bot says:

error: unknown lint: `clippy::derived_hash_with_manual_eq`
   --> src/back/msl/sampler.rs:159:9
    |
159 | #[allow(clippy::derived_hash_with_manual_eq)]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D unknown-lints` implied by `-D warnings`


But this was the one thing I needed to do to build Naga on my mac.

teoxoy:
We can't yet incorporate this change/rename since our MSRV is at 1.63 and clippy will start to warn users that are still using versions below rustc 1.68. We also recently updated our CI to run clippy on MSRV https://github.com/gfx-rs/naga/commit/6fcd04ec0644ab01475b59717fd0f2c5e7cafa95.

Newer versions of clippy will indeed complain about it and I'm not sure what we can do about it.

One thing worth noting is that the lint's behavior changed and so it was renamed. The fix here would be to remove the lint attribute altogether since clippy won't warn anymore in this case.

dneto0:
Ok. I have no idea what I'm doing with this.  :-)
I'll close this because your normal processes for upgrading tooling will address this.

PR #2286, opened 2023-03-19
[spirv-out] Fix adding illegal decorators on fragment outputs.

### LINKED ISSUE #2270: Outputting integral type from fragment shader causes Vulkan validation error
Validation layer trips off since Naga emits `flat` qualifier for integer outputs of fragment shader.
Note that wgpu correctly forces output to be an integer type when writing to a "true" (non-snorm/unorm) target.

Happens with a shader like this:
```
@fragment
fn fs_main_outline_mask(in: VertexOut) -> @location(0) vec2<u32> {
    return batch.outline_mask;
}
```

image
More details in this comment https://github.com/rerun-io/rerun/pull/1532#issuecomment-1461824202

Not familiar with Naga's code, but it looks like there's a check missing here for fragment outputs https://github.com/gfx-rs/naga/blob/master/src/back/spv/writer.rs#L1415

Wumpf:
Furthermore, fix allowing to add `Centroid` and `Sample` decorator to vertex inputs.
Fixes #2270

Testing:

Verified this against the scenario in #2270 on a Windows machine and ran existing tests. I expected `interpolate.wgsl` to produce a different spirv because of my change on  `Centroid` and `Sample`, but it seems those are already suppressed elsewhere as well (?).
Added test to verify how fragment outputs are passed out depending on type.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1407,7 +1407,7 @@ impl Writer {
             } => {
                 self.decorate(id, Decoration::Location, &[location]);
 
-                let no_decorations = 
+                let no_decorations =
                     // VUID-StandaloneSpirv-Flat-06202
                     // > The Flat, NoPerspective, Sample, and Centroid decorations
                     // > must not be used on variables with the Input storage class in a vertex shader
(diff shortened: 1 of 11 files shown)


teoxoy (approved):
Looks great!

teoxoy:
> I expected `interpolate.wgsl` to produce a different spirv because of my change on  `Centroid` and `Sample`, but it seems those are already suppressed elsewhere as well (?).

The struct in the file is only used as an output for the vertex stage and input for the fragment stage; this PR should have no effect on the SPIR-V output.

PR #2289, opened 2023-03-22
Ray Tracing Support (glsl->spv)

daniel-keitel:
Support for SPV_KHR_ray_tracing (includeng new shader stages)
from glsl to spir-v.

Requires: #2256

I will wait until Ray Query Support is merged, before I continue.


src/lib.rs
```diff
@@ -861,6 +869,8 @@ pub struct GlobalVariable {
     pub space: AddressSpace,
     /// For resources, defines the binding point.
     pub binding: Option<ResourceBinding>,
+    /// For some ray-tracing storage qualifiers, define the location
+    pub location: Option<u32>,
```
kvark, line 873: does this need to work differently from the existing mechanism of passing locations? Currently, `struct Binding` has a location, and it's carried for entry point inputs/outputs. The incoming ray payload should probably just be an argument to the entry point.

kvark: Moreover, we should probably consider just merging the `IncompingRayPayload` address space into `Input`, since it's non-ambiguous and likely has the same constraints.

daniel-keitel: All new storage classes are somewhere between  `StorageQualifier::Output` and `StorageQualifier::AddressSpace`.
They are read and write, have a location, no interpolation, no binding.
Additionally storage classes are mapped from AddressSpaces.
So `StorageQualifier::AddressSpace` seemed to be the better fit.
A new StorageQualifier may even be best.

cwfitzgerald:

PR #2291, opened 2023-03-24
[spv-out] Fix invalid spirv being generated from integer dot products

PyryM:
Short: this fixes integer dot product SPIR-V generation by moving a constant declaration to the right place.

Longer:
The code generation for integer dot products would (only sometimes, I am not sure why the tests do not catch this) produce SPIR-V that is technically invalid because it tries to declare constant null/0 inside of a function:

```
; Function 21
%21 = OpFunction %void None %22
%17 = OpLabel
// [...]
// vvv generated integer dot product
%38 = OpConstantNull %uint
%39 = OpCompositeExtract %uint %33 0
%40 = OpCompositeExtract %uint %36 0
%41 = OpIMul %uint %39 %40
%42 = OpIAdd %uint %38 %41
// [...]
```

Running against the SPIR-V validator (or a sufficiently pedantic driver like those for the Intel Arc Alchemist) causes a validation failure along the lines of:
```
error: line 55: ConstantNull cannot appear in a function declaration
  %38 = OpConstantNull %uint
```

Mysterious crashes from this had been noticed before on some hardware (e.g., https://github.com/gfx-rs/wgpu/issues/2694).

The fix is to use the already-provided helper function for declaring constant null/0.

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -1627,12 +1627,7 @@ impl<'w> BlockContext<'w> {
         size: u32,
         block: &mut Block,
     ) {
-        let const_null = self.gen_id();
-        block
-            .body
-            .push(Instruction::constant_null(result_type_id, const_null));
-
-        let mut partial_sum = const_null;
+        let mut partial_sum = self.writer.write_constant_null(result_type_id);
         let last_component = size - 1;
         for index in 0..=last_component {
             // compute the product of the current components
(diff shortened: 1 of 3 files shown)


teoxoy:
One thing before I merge this:

> Running against the SPIR-V validator (or a sufficiently pedantic driver like those for the Intel Arc Alchemist) causes a validation failure along the lines of:
> ```
> error: line 55: ConstantNull cannot appear in a function declaration
>   %38 = OpConstantNull %uint
> ```

How did you get `spirv-val` to produce this error?

I tried the command below (the one CI uses as well) and can't get `spirv-val` to error.
```
cat ./tests/out/spv/functions.spvasm | spirv-as --target-env spv1.0 -o - | spirv-val
```

PyryM:
Actually, running `tests/in/functions.wgsl` through CLI Naga will also result in the invalid spir-v being generated; there must be some difference between how the CLI (and also wgpu/wgpu_native) processes wgsl->spv and how the snapshot/test outputs are generated.

PR #2294, opened 2023-03-29
fix(hlsl-out): use `Interlocked<op>` intrinsic for atomic integers

### LINKED ISSUE #2284: hlsl-out miscompilation of atomic workgroup variable
Some atomic operations on workgroup-shared variables produce invalid HLSL. Similar to #2137, but without arrays.

This is using `naga v0.11.0`.

With this compute shader (WGSL):

```wgsl
var<workgroup> foo: atomic<u32>;

@compute
@workgroup_size(2)
fn test_miscompilation(@builtin(local_invocation_id) id: vec3<u32>) {
    if id.x == 0u {
        atomicStore(&foo, 0u);
    }

    workgroupBarrier();

    atomicAdd(&foo, 1u);
}
```

Naga generates the following HLSL:

```hlsl
groupshared uint foo;

[numthreads(2, 1, 1)]
void test_miscompilation(uint3 id : SV_GroupThreadID, uint3 __global_invocation_id : SV_DispatchThreadID)
{
    if (all(__global_invocation_id == uint3(0u, 0u, 0u))) {
        foo = (uint)0;
    }
    GroupMemoryBarrierWithGroupSync();
    if ((id.x == 0u)) {
        foo = 0u;
    }
    GroupMemoryBarrierWithGroupSync();
    uint _e8; foo.InterlockedAdd(0, 1u, _e8);
    return;
}
```

... which, when compiled with `dxc -T cs_6_7`, produces the following error:

```
test.hlsl:15:19: error: invalid format for vector swizzle 'InterlockedAdd'
    uint _e8; foo.InterlockedAdd(0, 1u, _e8);
```

ErichDonGubler:
Fixes #2284.

---

We currently assume that we are using raw `RWByteAddressBuffer` methods for all atomic operations (`<pointer>.Interlocked<op>(<raw_byte_offset>, …)`), which is only true when we use `var<storage, read_write>` globals. For `var<workgroup>` globals, we need `Interlocked<op>(<pointer>, …)`, using the original expression as the first argument.

Fix this by branching on the `pointer`'s address space in `Atomic` statements, and implementing the workgroup address space case with intrinsics.


src/back/hlsl/writer.rs
```diff
@@ -1856,14 +1856,66 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                                func_ctx,
+                                module.global_variables[*handle].ty,
+                            ),
+                            crate::Expression::Access { base, .. }
+                            | crate::Expression::AccessIndex { base, .. } => {
+                                is_atomic_scalar_expr(module, func_ctx, *base)
+                            }
+                            _ => false,
+                        }
+                    }
```
ErichDonGubler, line 1895: todo: What expressions and types have been missed that should be permitted?

src/back/hlsl/writer.rs
```diff
@@ -1856,14 +1856,66 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                let pointer_is_atomic_scalar = {
+                    fn is_atomic_scalar_ty<'a>(
+                        module: &'a Module,
+                        func_ctx: &'a back::FunctionCtx,
+                        ty: Handle<crate::Type>,
+                    ) -> bool {
+                        match &module.types[ty].inner {
+                            TypeInner::Atomic {
+                                // TODO: floats and booleans aren't valid to interlock, so uh…problem?
+                                kind: crate::ScalarKind::Sint | crate::ScalarKind::Uint,
```
ErichDonGubler, line 1868: todo: We actually _should_ be handling floats and booleans here, I'm guessing, but support doesn't seem straightforward to me in my (very limited) search:

* Floats need a specific shader model version or above, not sure which yet. I've seen mention in shader model 6 docs, but that might not be the minimum version.
* No idea about boolean, but I haven't seen it mentioned in any docs I've seen so far.

ErichDonGubler:
@jimblandy and I are planning on pair-reviewing this draft Soon™. 😊

tests/in/atomicSimple.wgsl
```diff
@@ -0,0 +1,49 @@
+var<workgroup> foo: atomic<u32>;
+var<workgroup> baz: array<atomic<u32>, 2>;
+
+@compute
+@workgroup_size(2)
+fn corge(@builtin(local_invocation_id) id: vec3<u32>) {
```
ErichDonGubler, line 6: nitpick: `corge` is random gibberish that served its purpose while I was still spending brain cycles understanding other things. Maybe something more sensible (even if bland, like `cs_main`) should be used here?

tests/out/glsl/atomicSimple.corge.Compute.glsl
```diff
@@ -0,0 +1,57 @@
...
+shared uint foo;
+
+shared uint baz[2];
+
+
+void main() {
+    if (gl_LocalInvocationID == uvec3(0u)) {
+        foo = 0u;
+        baz = uint[2](0u, 0u);
+    }
```
ErichDonGubler, line 17: question: I'm actually puzzled, but have a hypothesis as to why this code gets emitted, though there's no apparent equivalent in WGSL source. This appears to be automatically generated init., so variables have a well-known value. Does invocation `0` always run first? Is that right?

tests/out/hlsl/atomicSimple.hlsl
```diff
@@ -0,0 +1,40 @@
+
+groupshared uint foo;
+groupshared uint baz[2];
+
+[numthreads(2, 1, 1)]
+void corge(uint3 id : SV_GroupThreadID, uint3 __local_invocation_id : SV_GroupThreadID)
+{
+    if (all(__local_invocation_id == uint3(0u, 0u, 0u))) {
```
ErichDonGubler, line 8: question: What is `all(…)` here? 🤔

tests/out/glsl/atomicSimple.corge.Compute.glsl
ErichDonGubler: Bringing back in context from sync discussion with @jimblandy: This is actually a _tie-breaker_ for the workgroup, where we elect one and only one thread/invocation in the thread group/workgroup to set the group's state properly.

I'm satisfied with what I've learned here. Thanks, Jim!

tests/out/hlsl/atomicSimple.hlsl
ErichDonGubler: Responding to my own question, looks like this is the [`all(…)` intrinsic](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-all), which can be used on scalar, vectors, or matrices to determine that all components are non-zero. That…seems unnecessary here, but doesn't hurt anything, I suppose! Definitely nothing that should block here.

tests/out/glsl/atomicSimple.corge.Compute.glsl
ErichDonGubler: Interestingly, Raph Levien [notes](https://stackoverflow.com/questions/39393560/glsl-memorybarriershared-usefulness#comment119242328_39395947) that in OpenGL 4.6 and GL ES 3.2, the `memoryBarrierShared` might not be necessary here. IDK what our deps are, so 🤷🏻‍♂️. Just interesting trivia, I guess?

tests/snapshots.rs
```diff
@@ -505,6 +505,10 @@ fn convert_wgsl() {
                 | Targets::IR
                 | Targets::ANALYSIS,
         ),
+        (
+            "atomicSimple",
```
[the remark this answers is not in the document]
ErichDonGubler: Idea: `atomicOps`

src/back/hlsl/writer.rs
jimblandy: `ScalarKind` is indeed too loose for this particular operation. We forbid types other than i32 and u32 [here](https://github.com/gfx-rs/naga/blob/52043be00489d6dedb08c36bb97d018d11cb1e13/src/valid/type.rs#L306-L309).

tests/out/glsl/atomicSimple.corge.Compute.glsl
```diff
@@ -0,0 +1,57 @@
...
+    barrier();
+    uvec3 id = gl_LocalInvocationID;
+    foo = 1u;
+    baz[0] = 1u;
+    memoryBarrierShared();
+    barrier();
+    uint unnamed = foo;
+    uint unnamed_1 = baz[0];
+    memoryBarrierShared();
+    barrier();
```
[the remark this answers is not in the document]
ErichDonGubler: In the same link I discovered [elsewhere](https://github.com/gfx-rs/naga/pull/2294/files#r1153751514), I found an answer. Yippee!

tests/snapshots.rs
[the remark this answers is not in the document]
ErichDonGubler: Using `atomicOps` for now. Leaving this open to bikeshedding, but I'm not gonna spend more time on it without prompting.

src/back/hlsl/writer.rs
ErichDonGubler: Gotcha. Should be resolvable, then. I removed my `TODO` comment, and I think otherwise this code should be correct for the question I had.

tests/in/atomicSimple.wgsl
ErichDonGubler: Renamed to `cs_main`.

tests/out/glsl/atomicSimple.corge.Compute.glsl
```diff
@@ -0,0 +1,57 @@
...
+    }
+    memoryBarrierShared();
+    barrier();
+    uvec3 id = gl_LocalInvocationID;
+    foo = 1u;
+    baz[0] = 1u;
+    memoryBarrierShared();
+    barrier();
+    uint unnamed = foo;
+    uint unnamed_1 = baz[0];
```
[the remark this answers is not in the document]
ErichDonGubler: Ahhh, crap, just did a switch to `baz[1]`, and it looks like byte offsets are still being used. Now I get to figure out how to fix it…hopefully not by duplicating too much of `Writer::fill_access_chain`. 😓

ErichDonGubler:
@jimblandy: I've pushed some significant updates to this during my afternoon yesterday. Most of the logic is the same, just renamed or shuffled around a bit.

src/back/hlsl/writer.rs
```diff
@@ -1856,14 +1856,54 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                            &crate::Expression::GlobalVariable(handle) => {
+                                is_atomic_scalar_ty(module, module.global_variables[handle].ty)
+                            }
+                            &crate::Expression::Access { base, .. }
+                            | &crate::Expression::AccessIndex { base, .. } => {
+                                is_atomic_scalar_expr(module, func_ctx, base)
+                            }
+                            _ => false,
+                        }
+                    }
```
ErichDonGubler, line 1888: question: Should the helper functions inside the block returning accessing_byte_addr_buf's value be moved to somewhere else?

src/back/hlsl/writer.rs
```diff
@@ -1856,14 +1856,54 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                    ) -> bool {
+                        match &func_ctx.expressions[expr] {
+                            &crate::Expression::GlobalVariable(handle) => {
+                                is_atomic_scalar_ty(module, module.global_variables[handle].ty)
+                            }
+                            &crate::Expression::Access { base, .. }
+                            | &crate::Expression::AccessIndex { base, .. } => {
+                                is_atomic_scalar_expr(module, func_ctx, base)
+                            }
+                            _ => false,
```
ErichDonGubler, line 1886: question: RE: the wildcard pattern lines a la `_ => false`: Are there any expression or type constructs that we need in this PR before merging?

src/back/hlsl/writer.rs
ErichDonGubler: This has been invalidated by recent changes. Dropping.

ErichDonGubler:
@jimblandy: Just pushed up newest changes branching on address space. TIL a bunch, thanks for the help! ❤️

tests/out/hlsl/atomicSimple.hlsl
teoxoy: `all` is required here since the `==` operator returns a boolean vector and the if statement expects a boolean (see https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators#comparison-operators).

src/back/hlsl/writer.rs
```diff
@@ -1856,14 +1856,44 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                    match pointer_ty {
+                        &TypeInner::Pointer { space, .. } => space,
+                        &TypeInner::ValuePointer { space, .. } => space,
+                        ref other => {
+                            return Err(Error::Custom(format!(
+                                "unexpected resolved type for atomic pointer operand: {other:?}"
+                            )))
+                        }
+                    }
+                };
```
teoxoy, line 1873:
```suggestion
                let pointer_space = func_ctx.info[pointer].ty
                    .inner_with(&module.types)
                    .pointer_space();
```

We have a few helper functions that we could use here.

teoxoy (commented):
The new test file looks much more complete than what we had before!

https://github.com/gfx-rs/naga/blob/da8e911d9d207a7571241e875f7ac066856e8b3f/tests/in/access.wgsl#L154-L169

I think we could remove this but the `atomicExchange` function seems to be missing from the new test file.

ErichDonGubler:
@teoxoy:

> The new test file looks much more complete than what we had before!

😊

> I think we could remove this [([`access.wgsl:154-169`](https://github.com/gfx-rs/naga/blob/da8e911d9d207a7571241e875f7ac066856e8b3f/tests/in/access.wgsl#L154-L169))] but the `atomicExchange` function seems to be missing from the new test file.

Agreed. Added coverage for `atomicExchange` with 340a6982 and b14bd408, and removed the overlapping contents of `access.wgsl` with f8c7cb12 (and threw in e3101219 for fun). Will autosquash momentarily so this branch's contents are still nice commits.

src/back/hlsl/writer.rs
ErichDonGubler: 👀 Onice, implemented more-or-less as suggested with 57a9728b.

tests/out/hlsl/atomicOps.yaml
jimblandy, line 1: What is this file doing here?

jimblandy (approved):
Looks good. Some minor cleanups needed.

tests/out/hlsl/atomicOps.hlsl
```diff
@@ -0,0 +1,112 @@
...
+    }
+    GroupMemoryBarrierWithGroupSync();
+    storage_atomic_scalar.Store(0, asuint(1u));
+    storage_atomic_arr.Store(4, asuint(1));
+    storage_struct.Store(0, asuint(1u));
+    storage_struct.Store(4+4, asuint(1));
+    workgroup_atomic_scalar = 1u;
+    workgroup_atomic_arr[1] = 1;
+    workgroup_struct.atomic_scalar = 1u;
+    workgroup_struct.atomic_arr[1] = 1;
```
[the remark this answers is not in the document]
jimblandy: Connor found the following note in the [Direct3D functional spec](https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#7.14%20Shader%20Memory%20Consistency%20Model):

> atomic operations on UAV memory is atomic both with respect to other atomic operations as well as loads and stores.

Elsewhere, it also says:

> If a load or store operation spans more than 32-bits, the individual 32-bit operations are atomic, but not the whole.

I think this means that there's no need for special operations here: the proper translation of WGSL `atomicLoad` is indeed an ordinary HLSL `Load`.

PR #2296, opened 2023-03-31
Add several `clippy` lints

ErichDonGubler:
I noticed that I had let `dbg!(…)` sneak into commits in #2294 by accident. Clippy can help us with this! Set `deny(clippy::dbg_macro)`, so CI catches this for future contributions.

---

We do this for `panic`king, but I'm not sure if `deny` or `warn` is more appropriate for a debugging tool like `dbg!(…)`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -215,7 +215,17 @@ tree.
     clippy::rest_pat_in_fully_bound_structs,
     clippy::match_wildcard_for_single_variants
 )]
-#![cfg_attr(not(test), deny(clippy::panic))]
+#![deny(clippy::exit)]
+#![cfg_attr(
+    not(test),
+    warn(
+        clippy::dbg_macro,
+        clippy::panic,
+        clippy::print_stderr,
+        clippy::print_stdout,
+        clippy::todo
+    )
+)]
 
 mod arena;
 pub mod back;


teoxoy (commented):
I'm thinking `warn` might be more appropriate.
While we are at it, I think we should also warn on:
- [clippy::print_stderr](https://rust-lang.github.io/rust-clippy/master/index.html#print_stderr)
- [clippy::print_stdout](https://rust-lang.github.io/rust-clippy/master/index.html#print_stdout)
- [clippy::todo](https://rust-lang.github.io/rust-clippy/master/index.html#todo)
- [clippy::unimplemented](https://rust-lang.github.io/rust-clippy/master/index.html#unimplemented)
- [clippy::exit](https://rust-lang.github.io/rust-clippy/master/index.html#exit) (and deny this one)

ErichDonGubler:
@teoxoy: I'm not sure which compilation flag set you're suggesting those go under. I've push some changes with where I would assume they'd be best placed, LMK what you think.

RE: `clippy::unimplemented`: This actually causes warnings with current code. I think most instances are cases where we handle run-time issues that the type system doesn't currently catch? In those cases, I think migrating to `panic!(...)` would be acceptable, if the `not implemented: ` prepend is acceptable.

PR #2297, opened 2023-03-31
build: move from `make` to `cargo xtask` workflows

ErichDonGubler:
At @jimblandy's suggestion, I'm posting this WIP early. YMMV, and contributions are welcome to validate that all platforms still work correctly! It's mostly done, and bugs should be easy to fix, if they exist.

---

Re-implement `naga` development workflows using [`cargo xtask`]. Convert `make` logic and shader test configuration as file with Bash variables into an `xtask` crate and YAML files, respectively. 

Pros: 

* We now have a _portable_ workflow everywhere, which means Windows folks and people who don't install `make` don't have to suffer. 😮‍💨
* Workflow logic is now relatively easy to inspect and change. Whew! 💁🏻‍♂️💦
* Contributors can use their existing Rust knowledge to contribute to developer experience. 🎉
* `cargo xtask` is a relatively well-known convention for workflows in the ecosystem.
* We can do fancy things like allow folks to run at different log levels for workflows, depending on their tastes.

Cons:

* There's now a non-trivial compile step to project workflow. Incremental rebuilds seem to be pretty short, though!
* Code is much more verbose than the (very) terse `make` implementation.

[`cargo xtask`]: https://github.com/matklad/cargo-xtask


tests/out/hlsl/array-in-ctor.yaml
```diff
@@ -0,0 +1,5 @@
+vertex: []
+fragment: []
+compute:
```
ErichDonGubler, line 3: suggestion: Stylistically, it's entirely valid to omit `vertex` and `fragment`. Shall we strip out empty shader model categories?

xtask/.gitignore
```diff
@@ -0,0 +1,2 @@
+Cargo.lock
```
ErichDonGubler, line 1: thought: We _may_ want to keep the `Cargo.lock`, but I feel like it's annoying to do so, with no real gain here.

xtask/src/hlsl.rs
```diff
@@ -0,0 +1,85 @@
...
+    Empty,
+}
+
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct Config {
+    pub vertex: Option<Vec<ConfigItem>>,
+    pub fragment: Option<Vec<ConfigItem>>,
+    pub compute: Option<Vec<ConfigItem>>,
+}
```
ErichDonGubler, line 36: todo: Auto-generate YAML from `tests/snapshots.rs`, instead of Bash variables lists.

ErichDonGubler:
In case anybody is interested, here's the Nushell script I've been using to convert the Bash variables files in HLSL tests over to YAML:

```nu
fd config tests\ | lines | each {|file| open $file | lines | parse '{type}=({items})' | update items {|it| echo $it.items | str trim | split row ' ' | each {|test| $test | parse '{entry_point}:{target_profile}' } | flatten } | { path: $file config: $in } } | each {|config| { path: $config.path vertex: ($config.config | where type == "vertex" | reject type | first | $in.items) fragment: ($config.config | where type == "fragment" | reject type | first | $in.items) compute: ($config.config | where type == "compute" | reject type | first | $in.items) } } | each {|config| ($config | reject path | to yaml) | save -f ($config.path | path parse -e hlsl.config | upsert extension yaml| format '{prefix}{parent}\{stem}.{extension}') }
```

xtask/.gitignore
ErichDonGubler: Actually, this may be beneficial specifically for [caching with `rust-cache`](https://github.com/Swatinem/rust-cache#cache-effectiveness):

> It is also most effective for repositories with a `Cargo.lock` file. Library repositories with only a `Cargo.toml` file have limited benefits, as cargo will _always_ use the most up-to-date dependency versions, which may not be cached.

xtask/src/main.rs
```diff
@@ -0,0 +1,336 @@
...
+                        },
+                    )
+                }
+                ValidateSubcommand::Wgsl => visit_files(
+                    snapshots_base_out,
+                    "wgsl/*.wgsl",
+                    &mut found_err,
+                    |file, _| {
+                        // TODO: strip prefix to match old Makefile output?
+                        log::info!("Validating {}", file.display());
```
ErichDonGubler, line 223: N.B. that this output is technically different from the original `Makefile`; it inclues the path prefix of the glob search path, instead of "just" printing the base name. I don't think anybody will care, tho!

tests/out/hlsl/array-in-ctor.yaml
ErichDonGubler: Done, as of 7fac4866.

ErichDonGubler: Hmm, using `serde_with` to achieve this adds more than 20 build steps, and significantly adds to compile times and compile lengths by bringing in its own proc-macros and the `darling` ecosystem. Not sure if this is worth it. 🤔

tests/out/hlsl/access.ron
ErichDonGubler, line 1: - [ ] todo: Validate that we've actually ported all HLSL configs over faithfully.

ErichDonGubler:
I consider this ready to review, but not ready to merge until the checkboxes I've set for myself are done. I don't expect any feedback to significantly change for having checked them later.

tests/out/hlsl/access.ron
ErichDonGubler: This all looks correct to me. Resolving.

Makefile
```diff
@@ -1,131 +0,0 @@
...
-	xcrun -sdk macosx metal -c $< -mmacosx-version-min=10.11
-
-%.metallib: %.air
-	xcrun -sdk macosx metallib $< -o $@
-
-%.dot: $(SNAPSHOTS_BASE_IN)/%.wgsl $(wildcard src/*.rs src/front/wgsl/*.rs src/back/dot/*.rs bin/naga.rs)
-	cargo run --features wgsl-in,dot-out -- $< $@
-
-%.png: %.dot
-	dot -Tpng $< -o $@
```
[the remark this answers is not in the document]
ErichDonGubler: CC @jimblandy, @cwfitzgerald.

ErichDonGubler:
I've finished with all checkboxes I think I can resolve by myself. There's only one conversation truly outstanding, ATOW, will ping folks there.

Makefile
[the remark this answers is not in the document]
cwfitzgerald: I'm indifferent.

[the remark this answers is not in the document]
ErichDonGubler: Since nobody has felt strongly about this here or [in Matrix](https://matrix.to/#/!fjjkgHFcwtkREywzfk:matrix.org/$8fUNjQx5Pm-Pgz6fQH2CGRKeS8RmW-BOMlaQ4_qQQeE?via=matrix.org&via=mozilla.org&via=anontier.nl), I'm gonna mark this as `Resolved`/"won't fix".

ErichDonGubler:
This PR should basically be ready for full review and merging. 😤😀

xtask/src/main.rs
```diff
@@ -0,0 +1,328 @@
...
+                    .with_context(|| format!("failed to remove {criterion_artifact_dir:?}"))?;
+            }
+            EasyCommand::simple("cargo", ["bench"]).success()
+        }
+        Subcommand::Validate(cmd) => {
+            let mut found_err = false;
+            let ack_visiting = |path: &Path| log::info!("Validating {}", path.display());
+            match cmd {
+                ValidateSubcommand::Spirv => {
+                    visit_files(
```
teoxoy, line 99:
```suggestion
                    let spirv_as = "spirv-as";
                    let spirv_val = "spirv-val";
                    which(spirv_as)?;
                    which(spirv_val)?;
                    visit_files(
```

xtask/hlsl-snapshots/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+    Parse { source: DeRonErr },
+    Empty,
+}
+
+impl Display for BadRonParseKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            BadRonParseKind::Read { source } => Display::fmt(source, f),
+            BadRonParseKind::Parse { source } => Display::fmt(source, f),
+            BadRonParseKind::Empty => todo!(),
```
teoxoy, line 33: There is a remaining `todo!()` here.

xtask/src/main.rs
```diff
@@ -0,0 +1,328 @@
...
+                |file, _| {
+                    remove_file(file).with_context(|| format!("failed to remove file {file:?}"))
+                },
+            );
+            ensure!(
+                !found_err,
+                "failed to clean one or more files, see above output for more details"
+            );
+            Ok(())
+        }
```
teoxoy, line 84: Not sure if we need this, feels like something that was added for testing. Not a blocker though, just wanted to start the conversation.

xtask/src/main.rs
```diff
@@ -0,0 +1,328 @@
...
+                                cmd.stdin(Stdio::from(file))
+                                    .stdout(Stdio::piped())
+                                    .arg("--target-env")
+                                    .arg(format!("spv{version}"))
+                                    .args(["-", "-o", "-"])
+                            });
+                            let child = spirv_as_cmd
+                                .spawn()
+                                .with_context(|| format!("failed to spawn {cmd:?}"))?;
+                            EasyCommand::new("spirv-val", |cmd| cmd.stdin(child.stdout.unwrap()))
```
teoxoy, line 135:
```suggestion
                            EasyCommand::new(spirv_val, |cmd| cmd.stdin(child.stdout.unwrap()))
```

xtask/src/glob.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use anyhow::Context;
+use glob::glob;
+
+use crate::result::LogIfError;
+
+pub(crate) fn visit_files(
+    path: impl AsRef<Path>,
+    glob_expr: &str,
+    found_err: &mut bool,
```
teoxoy, line 11: I feel like we could do something else here where we'd avoid having this out param.

And/or have it be a `Status` enum instead to make things clearer (especially for closures).

teoxoy (changes requested):
Looks solid! Much easier to follow than the `Makefile`.

xtask/src/main.rs
ErichDonGubler: D'oh, good catch! Implemented a fix with 3b31bab5 that's very similar to yours.

xtask/src/main.rs
```diff
@@ -0,0 +1,328 @@
...
+                            };
+                            let expected_header_prefix = "; Version: ";
+                            let Some(version) =
+                                second_line.strip_prefix(expected_header_prefix) else {
+                                    bail!(
+                                        "no {expected_header_prefix:?} header found in {path:?}"
+                                    );
+                                };
+                            let file = open_file(path)?;
+                            let mut spirv_as_cmd = EasyCommand::new("spirv-as", |cmd| {
```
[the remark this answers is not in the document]
ErichDonGubler: See resolution at https://github.com/gfx-rs/naga/pull/2297#discussion_r1221405497.

xtask/hlsl-snapshots/src/lib.rs
ErichDonGubler: D'oh, fixed with 4c482134.

xtask/src/main.rs
ErichDonGubler: Let's see! Removed with 971ee01e. I've thrown a reversion into my [`ErichDonGubler/naga:xtask-clean`](https://github.com/ErichDonGubler/naga/tree/xtask-clean) branch, in case it's needed sometime soon. Otherwise, this content should always (?) be accessible in the history of this PR, if somebody actually wanted this.

xtask/src/glob.rs
ErichDonGubler: Hmm…yeah, this isn't a great API right now. Changed the `bool` to be a more explicit `enum ErrorStatus` with d5db0e13, and I think that that's a good change, overall. I experimented with trying to return things by value in 224da534, but I'm not persuaded that it's an improvement overall without more work.

ErichDonGubler:
Resolved some `clippy` lints with 5a659c89. Hmm…maybe we should be running `clippy` on `xtask` too? 🤔

ErichDonGubler:
Odd, I'm force-pushing some changes to my PR branch, but it's not updating here. 🤔 Trying to figure out if there's something I've done wrong here, but in the meantime, the commit hashes I've provided for `fixup!`s don't get link-ified.

ErichDonGubler:
Ah, looks like there's some significant issues with pull requests on [GitHub Status](https://www.githubstatus.com/):

image

ErichDonGubler:
@teoxoy: Once you've reviewed, or you have more feedback, most of the `fixup!`s here should squash down neatly with `git rebase -i --autosquash`. I've already done it locally, and can push when you're ready.

teoxoy (approved):
LGTM

ErichDonGubler:
@teoxoy: Done with 81124084, already auto-squashed.

teoxoy (approved):

PR #2298, opened 2023-04-02
Allow `array_index` to be unsigned

daxpedda:
Following the spec, functions with the parameter `array_index`, e.g. in combination with `texture_2d_array`, should be able to take a `u32` as well:
https://www.w3.org/TR/WGSL/#textureload
https://www.w3.org/TR/WGSL/#texturesample
https://www.w3.org/TR/WGSL/#texturestore

(I have no clue how to link to a more permanent version of the spec)


daxpedda:
Alright, this is far outside my comfort zone. Any help would be appreciated.

I assume that either some code was relying on this value being a `i32` and has to be adjusted to take that into account or SPIR-V just doesn't allow a `i32` in this case.

teoxoy:
SPIR-V requires a vector containing the coordinates and the array index. The type of the vector has to match with the type of the scalars that compose it and it's here that we assume the `array_index` is signed:

https://github.com/gfx-rs/naga/blob/7c00548075f51dc8c80ab8abdb4de6569ea59d46/src/back/spv/image.rs#L317-L352

src/back/spv/image.rs
```diff
@@ -314,41 +314,76 @@ impl<'w> BlockContext<'w> {
...
+
+                let reconciled_id = self.gen_id();
+                block.body.push(Instruction::unary(
+                    spirv::Op::ConvertUToF,
+                    component_type_id,
+                    reconciled_id,
+                    array_index_id,
+                ));
+                reconciled_id
+            }
```
teoxoy, line 382: I think we can reduce the repetition in this `match` by returning an `Option` with the spirv operation and writing the conversion after.

daxpedda: Done, much better indeed.

daxpedda:
> Could you also add a few more tests for this?

Could you point me in any specific direction, or do you have something in mind?

tests/in/binding-arrays.wgsl
```diff
@@ -63,6 +63,7 @@ fn main(fragment_in: FragmentIn) -> @location(0) vec4<f32> {
     v4 += textureLoad(texture_array_unbounded[non_uniform_index], pix, 0);
 
     u1 += textureNumLayers(texture_array_2darray[0]);
+    u1 += textureNumLayers(texture_array_2darray[0u]);
```
teoxoy, line 66: This is indexing a `binding_array` which I think is not what you intended.

teoxoy (commented):
Thanks for addressing all the comments - looks great!

> > Could you also add a few more tests for this?
>
> Could you point me in any specific direction, or do you have something in mind?

It seems the only test updated was `textureLoad`. I think we should at least test `textureStore` and `textureSample` as well. The other texture sampling functions use the same code paths so we might not need to test those.

tests/in/binding-arrays.wgsl
daxpedda: Done.

daxpedda:
Added a bunch more tests, hope that's not overkill.

teoxoy (approved):

cwfitzgerald:
We should boil this requirement up as a downlevel flag.

PR #2299, opened 2023-04-03
[wgsl] Use `alias` instead of `type` for type aliases

FL33TW00D:
This brings us in line with the spec: https://gpuweb.github.io/gpuweb/wgsl/#type-aliases

Currently Canary fails with `type` keyword.


ErichDonGubler (approved):
I don't have merge rights, but this LGTM. Let's see who else can approve this. 🙂

src/front/wgsl/parse/mod.rs
```diff
@@ -2213,7 +2213,7 @@ impl Parser {
                 let members = self.struct_body(lexer, ctx)?;
                 Some(ast::GlobalDeclKind::Struct(ast::Struct { name, members }))
             }
-            (Token::Word("type"), _) => {
+            (Token::Word("type"), _) | (Token::Word("alias"), _) => {
```
jimblandy, line 2216: I think it would be best to simply remove support for `type` altogether. Everyone will need to make that change soon eventually.

tests/in/boids.wgsl
```diff
@@ -1,4 +1,5 @@
 const NUM_PARTICLES: u32 = 1500u;
+alias Vec2 = vec2<f32>;
```
jimblandy, line 2: Again, much better to do this in `module-scope.wgsl` or `type-alias.wgsl`, which actually use the definition.

jimblandy (changes requested):
Thanks for fixing this! Just a few minor things.

src/front/wgsl/parse/mod.rs
FL33TW00D: Agreed!

teoxoy (approved):
LGTM

PR #2305, opened 2023-04-18
Remove front::wgsl::lower::ExpressionContext::reborrow.

jimblandy:
Rust's implicit reborrowing works fine.

This change would totally conflict with @teoxoy's const expression work, so we shouldn't worry about this until that's done. I just couldn't figure out what the `reborrow` functions were accomplishing, so I had to try it out.


jimblandy:
... and `RuntimeExpressionContext::reborrow`

PR #2308, opened 2023-04-20
refactor(valid)!: remove `TypeError::UnresolvedBase`

### LINKED ISSUE #2303: Is `naga::valid::TypeError::UnresolvedBase` still needed?
The `naga::valid::TypeError::UnresolvedBase` error is only raised by `Validator::validate_type`, which is only called after `Validator::validate_module_handles`, which checks that all [pointer](https://github.com/gfx-rs/naga/blob/f59668ccfaf7bdb3a7e43d84363a21c77357b2fe/src/valid/handles.rs#L84), [array](https://github.com/gfx-rs/naga/blob/f59668ccfaf7bdb3a7e43d84363a21c77357b2fe/src/valid/handles.rs#L92), and [struct member](https://github.com/gfx-rs/naga/blob/f59668ccfaf7bdb3a7e43d84363a21c77357b2fe/src/valid/handles.rs#L99) base type handles precede their uses.
So there should be no need for this error any more.

cc @ErichDonGubler

ErichDonGubler:
Resolves #2303.

### DIFF
--- a/src/valid/type.rs
+++ b/src/valid/type.rs
@@ -96,8 +96,6 @@ pub enum TypeError {
     InvalidWidth(crate::ScalarKind, crate::Bytes),
     #[error("The {0:?} scalar width {1} is not supported for an atomic")]
     InvalidAtomicWidth(crate::ScalarKind, crate::Bytes),
-    #[error("The base handle {0:?} can not be resolved")]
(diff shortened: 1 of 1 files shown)


teoxoy (approved):
Good stuff!

Issue #1924, opened 2022-05-14
Relax vector, matrix and array OOB validation


teoxoy:
Relax vector, matrix and array OOB validation by handling OOB accesses at runtime.

The following example should work; however, validation currently errors because the index is out of bounds.

```
@compute @workgroup_size(1)
fn main() {
    let idx = 9;
    let v = vec2<f32>();
    let m = mat2x2<f32>();
    let a = array<f32,1>();
    let vi = v[idx];
    let mi = m[idx];
    let ai = a[idx];
}
```

Related

- #1868
- #1923

PR #2309, opened 2023-04-21
Implement const-expressions (phase 2)

### LINKED ISSUE #97: Constant propagation
We should allow constant expressions for array sizes. This needs IR changes:
  - `enum ConstExpression` that takes a subset of `Expression` that's valid for constant propagation
  - the `Expression::Const(Handle<ConstExpression>)` to transition from regular expressions (instead of the current `Expression::Constant`)
  - separate `pub const_expressions: Arena<ConstExpression>` in a function

### LINKED ISSUE #471: Cannot define fractional constants
The following will fail to compile:
```wgsl
const w2: vec4<f32> = vec4<f32>(4.0/9.0, 0.0, 0.0, 0.0);
```

(Even if `const` is replaced with `var`, it will still fail)

### LINKED ISSUE #733: [wgsl-in] Support `const` operations in `const` variables
I'm not sure whether the official WGSL spec specifies anything about this, but I'm running into a pattern in my code where I'd like to do something like this:

```glsl
const phi_sun: f32 = PI * 2.0;
```

However, the compiler doesn't like this:

```glsl
error: expected separator, found '*'
   ┌─ wgsl:84:25
   │
84 │ const phi_sun: f32 = PI * 2.0;
   │                         ^ expected separator

error: aborting due to previous error
```

### LINKED ISSUE #1829: Implement creation-time expressions
https://gpuweb.github.io/gpuweb/wgsl/#creation-time-expr

Spec PR: https://github.com/gpuweb/gpuweb/pull/2668

### LINKED ISSUE #2106: Support Constants in Enum Cases
This is my shader file:

```

let TEXTURE_KIND_REGULAR:i32 = 0;
let TEXTURE_KIND_WARP:i32 = 1;
let TEXTURE_KIND_SKY:i32 = 2; 
.....

@fragment
fn main_fs(vertex: VertexOutput) -> FragmentOutput {
   
    var result: FragmentOutput;

       switch (push_constants.texture_kind) {
        case TEXTURE_KIND_REGULAR:
.....
```

That throws this error:
```

 case TEXTURE_KIND_REGULAR:
    │              ^^^^^^^^^^^^^^^^^^^^ expected unsigned/signed integer literal


    expected unsigned/signed integer literal, found 'TEXTURE_KIND_REGULAR'

```

Now.. i could be doing something wrong here but this is VERY confusing and there are not any good docs i can find about this.  

my 'let' vars should work for a switch case  --- no ?

### LINKED ISSUE #2313: [wgsl-in] 0.0 / 0.0 should not be allowed
Spec says that NaN is generally not defined. I think Tint is right here and Naga should catch this as well as Tint does:

```
Tint WGSL reader failure: :80:17 error: '0.0 / 0.0' cannot be represented as 'abstract-float'
    let a = 0.0 / 0.0;
```

### LINKED ISSUE #2430: workgroup_size:  expected 32-bit signed integer literal error
**Description**

The following compute shader code works in WebGPU:

```
const SIZE = 8;
@compute @workgroup_size(SIZE, SIZE, 1)
fn cs_main(@builtin(workgroup_id) workGroupId : vec3u, @builtin(local_invocation_id) localId: vec3u) {
    let x = localId.x + workGroupId.x * u32(SIZE);
    let y = localId.y + workGroupId.y * u32(SIZE); 
    ...
}  
```
But wgpu gives error:

```
@compute @workgroup_size(SIZE, SIZE, 1)
  │                      ^^^^ expected 32-bit signed integer literal
```

However, from WGSL [spec](https://www.w3.org/TR/WGSL/#attribute-workgroup_size):
workgroup_size: One, two or three parameters.
Each parameter [must](https://www.w3.org/TR/WGSL/#shader-creation-error) be a [const-expression](https://www.w3.org/TR/WGSL/#const-expressions) or an [override-expression](https://www.w3.org/TR/WGSL/#override-expressions). All parameters [must](https://www.w3.org/TR/WGSL/#shader-creation-error) be the same type, either [i32](https://www.w3.org/TR/WGSL/#i32) or [u32](https://www.w3.org/TR/WGSL/#u32).

It is clear that workgroup_size can be a const expression. Why wgpu requires it to be a 32-bit signed integer literal?

**Platform**
Windows 11, RTX 3060

teoxoy:
Second phase of https://github.com/gfx-rs/wgpu/issues/4484.

- fixes https://github.com/gfx-rs/naga/issues/97
- fixes https://github.com/gfx-rs/naga/issues/1829
- fixes https://github.com/gfx-rs/naga/issues/733
- fixes https://github.com/gfx-rs/naga/issues/471
- fixes https://github.com/gfx-rs/naga/issues/2430
- fixes https://github.com/gfx-rs/naga/issues/2106
- fixes https://github.com/gfx-rs/naga/issues/2313


teoxoy:
Allowing splats for evaluated const-expressions seems to surface an issue with DXC https://github.com/microsoft/DirectXShaderCompiler/issues/5082 ...

jimblandy:
@teoxoy Could you rebase this on current trunk?

Also, it's not a draft any more, right?

jimblandy:
Rebased on current trunk.

src/proc/constant_evaluator.rs
```diff
@@ -62,169 +75,109 @@ pub enum ConstantSolvingError {
...
+// Splat
+// Swizzle
+// Unary
+// Binary
+// Select
+// Relational
+// Math
+// As
+
+// TODO(teoxoy): consider accumulating this metadata instead of recursing through subexpressions
```
jimblandy, line 95: Please don't leave TODO comments in the code. Instead, file PRs for follow-up work, and link to them in the code if that's valuable.

src/valid/function.rs
```diff
@@ -957,7 +957,7 @@ impl super::Validator {
 
         if let Some(init) = var.init {
             let decl_ty = &gctx.types[var.ty].inner;
-            let init_ty = mod_info[init].inner_with(gctx.types);
+            let init_ty = fun_info[init].ty.inner_with(gctx.types);
```
jimblandy, line 960: `var.init` is in the constant expression arena - doesn't this look up its type as if it were a local expression?

src/valid/analyzer.rs
```diff
@@ -1021,6 +1021,12 @@ impl ModuleInfo {
             }
         }
 
+        for (_, expr) in fun.local_variables.iter() {
+            if let Some(init) = expr.init {
+                let _ = info.add_ref(init);
```
jimblandy, line 1026: Since `init` is in the constant expressions arena, this seems incorrect.

jimblandy:
Oh, I think I see - with the "use LocalVariable init" commit, `LocalVariable::init` is no longer a handle in `Module::const_expressions`. I'll verify this and resolve my comments above if it seems right.

jimblandy:
rebased

jimblandy:
Rebasing this on `compact` significantly reduces the size of the diff:

```
165 files changed, 9952 insertions(+), 10177 deletions(-)
158 files changed, 7043 insertions(+), 7488 deletions(-)
```

src/proc/mod.rs
```diff
@@ -588,29 +590,59 @@ impl GlobalCtx<'_> {
...
+    }
+
+    pub(crate) fn eval_expr_to_literal(
+        &self,
+        handle: crate::Handle<crate::Expression>,
+    ) -> Option<crate::Literal> {
+        self.eval_expr_to_literal_from(handle, self.const_expressions)
+    }
+
+    pub(crate) fn eval_expr_to_literal_from(
```
jimblandy, line 609: What do you think of this?

```
    pub(crate) fn eval_expr_to_literal_from(
        &self,
        handle: crate::Handle<crate::Expression>,
        arena: &crate::Arena<crate::Expression>,
    ) -> Option<crate::Literal> {
        match arena[handle] {
            crate::Expression::Constant(c) => {
                self.eval_expr_to_literal_from(self.constants[c].init, self.const_expressions)
            }
            crate::Expression::Literal(literal) => Some(literal),
            crate::Expression::ZeroValue(ty) => match self.types[ty].inner {
                crate::TypeInner::Scalar {
                    kind: crate::ScalarKind::Sint,
                    width: 4,
                } => Some(crate::Literal::I32(0)),
                crate::TypeInner::Scalar {
                    kind: crate::ScalarKind::Uint,
                    width: 4,
                } => Some(crate::Literal::U32(0)),
                crate::TypeInner::Scalar {
                    kind: crate::ScalarKind::Float,
                    width: 4,
                } => Some(crate::Literal::F32(0.0)),
                crate::TypeInner::Scalar {
                    kind: crate::ScalarKind::Float,
                    width: 8,
                } => Some(crate::Literal::F64(0.0)),
                crate::TypeInner::Scalar {
                    kind: crate::ScalarKind::Bool,
                    width: 1,
                } => Some(crate::Literal::Bool(false)),
                _ => None,
            },
            _ => None,
        }
    }
```

jimblandy: That is, just eliminate the internal function altogether in favor of recursion. The handle validation phase should have ensured that we have no loops.

src/front/wgsl/error.rs
```diff
@@ -694,15 +672,24 @@ impl<'a> Error<'a> {
                 labels: vec![],
                 notes: vec![],
             },
-            Error::ExpectedArraySize(span) => ParseError {
-                message: "array element count must resolve to an integer scalar (u32 or i32)"
-                    .to_string(),
-                labels: vec![(span, "must resolve to u32/i32".into())],
+            Error::ExpectedConstExprConcreteIntegerScalar(span) => ParseError {
+                message: "must be a const-expression that resolves to a concrete integer scalar (u32, i32)".to_string(),
```
jimblandy, line 676:
```suggestion
                message: "must be a const-expression that resolves to a concrete integer scalar (u32 or i32)".to_string(),
```

jimblandy (commented):
Some thoughts so far:

src/proc/mod.rs
teoxoy: The family of functions this fn is part of are called by the frontends.
I find that the invariants are clearer without recursion.

jimblandy:
I think there's some unfortunate terminology used in this PR.

A "constant" is no longer something that lives in an `Arena<Constant>`, or even in `Module::const_expressions`. At least as `ConstantEvaluator` uses the term, it's a subexpression consisting exclusively of `Literal`, `ZeroValue`, `Compose`, `Splat`, and `Constant` expressions, which can be held in any sort of `Arena<Expression>`.

When I see something like `ConstantEvaluator::register_constant`, I expect that to be something that adds an entry to `Module::constants` or `Module::const_expressions`, not something that appends an expression of a certain form to whatever arena I happen to be building at the moment.

What would be better? I think we should invent a new term, like "reduced" or "evaluated", or maybe keep the GLSL front end's term "solved", to refer to expressions that consist solely of the forms that can be produced by a successful constant evaluation. Then, say, `ConstantEvaluator::register_constant` could be named `register_reduced_expr` or `register_evaluated_expr` or `register_solved_expr`.

jimblandy:
This seems to cost about 7% in our benchmarks. That's well within what I'm willing to pay for the feature. We can file a followup if we care.

src/front/wgsl/lower/mod.rs
```diff
@@ -317,17 +319,70 @@ impl<'source, 'temp, 'out> ExpressionContext<'source, 'temp, 'out> {
...
+                                rctx.emitter.start(arena);
+                            }
+                            h
+                        },
+                    ),
+                };
+
+                let res = eval
+                    .try_eval_and_append(&expr, span)
+                    .map_err(|e| Error::ConstantEvaluatorError(e, span));
```
jimblandy, line 348: This `map_err` is meaningless because we're just going to drop the error on the floor below, right?

src/proc/constant_evaluator.rs
```diff
@@ -1,19 +1,30 @@
...
-    pub constants: &'a mut Arena<Constant>,
-    pub const_expressions: &'a mut Arena<Expression>,
+    pub constants: &'a Arena<Constant>,
+    pub expressions: &'a mut Arena<Expression>,
+    pub const_expressions: Option<&'a Arena<Expression>>,
+    pub append: Option<F>,
+}
+
+#[derive(Debug)]
+pub struct EvalData {
```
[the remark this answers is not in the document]
teoxoy: Oh wow, that's from an earlier version of the patch, I think it snuck back in when rebasing.

teoxoy:
I agree with the naming issue, I think we can go with `register_evaluated_expr`.

Maybe I should also review this patch since I worked on it for a while and things like this didn't stick out. There might be others as well.

jimblandy:
Here's a branch that gets rid of the closure argument to `ConstantEvaluator`: https://github.com/jimblandy/naga/tree/const-expressions-2-no-closure

See what you think.

src/proc/constant_evaluator.rs
```diff
@@ -604,15 +686,12 @@ impl<'a> ConstantSolver<'a> {
...
-    ) -> Result<Handle<Expression>, ConstantSolvingError> {
-        let left = self.eval_zero_value(left, span)?;
-        let right = self.eval_zero_value(right, span)?;
-
-        let expr = match (
-            &self.const_expressions[left],
-            &self.const_expressions[right],
-        ) {
+        span: Span,
+    ) -> Result<Handle<Expression>, ConstantEvaluatorError> {
```
jimblandy, line 690: GitHub won't let me put this comment on the right line, but:

The cases here for `BinaryOperator::Divide` and `Modulo` need to report divide by zero errors, not just flush them to zero.

teoxoy:
> Here's a branch that gets rid of the closure argument to `ConstantEvaluator`: https://github.com/jimblandy/naga/tree/const-expressions-2-no-closure
> 
> See what you think.

It looks good; feel free to push the commit!

jimblandy:
The commit I just pushed, 8c76859, crashes compiling wgpu's `examples/water/src/water.wgsl`:

```
thread 'main' panicked at 'index out of bounds: the len is 2 but the index is 2', src/proc/constant_evaluator.rs:317:32
stack backtrace:
...
   2: core::panicking::panic_bounds_check
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:84:5
   3: <usize as core::slice::index::SliceIndex<[T]>>::index
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/slice/index.rs:250:10
   4: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/slice/index.rs:18:9
   5: <alloc::vec::Vec<T,A> as core::ops::index::Index<I>>::index
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/alloc/src/vec/mod.rs:2628:9
   6: naga::proc::constant_evaluator::ConstantEvaluator::swizzle::{{closure}}
             at /home/jimb/rust/naga/src/proc/constant_evaluator.rs:317:32
   7: core::iter::adapters::map::map_fold::{{closure}}
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/iter/adapters/map.rs:84:28
   8: <core::iter::adapters::take::Take<I> as core::iter::traits::iterator::Iterator>::fold::ok::{{closure}}
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/iter/adapters/take.rs:109:30
   9: <core::iter::adapters::take::Take<I> as core::iter::traits::iterator::Iterator>::try_fold::check::{{closure}}
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/iter/adapters/take.rs:88:25
  10: core::iter::traits::iterator::Iterator::try_fold
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/iter/traits/iterator.rs:2238:21
  11: <core::iter::adapters::take::Take<I> as core::iter::traits::iterator::Iterator>::try_fold
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/iter/adapters/take.rs:97:13
  12: <core::iter::adapters::take::Take<I> as core::iter::traits::iterator::Iterator>::fold
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/iter/adapters/take.rs:112:9
  13: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/iter/adapters/map.rs:124:9
  14: core::iter::traits::iterator::Iterator::for_each
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/iter/traits/iterator.rs:831:9
  15: <alloc::vec::Vec<T,A> as alloc::vec::spec_extend::SpecExtend<T,I>>::spec_extend
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/alloc/src/vec/spec_extend.rs:40:17
  16: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter_nested::SpecFromIterNested<T,I>>::from_iter
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/alloc/src/vec/spec_from_iter_nested.rs:62:9
  17: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/alloc/src/vec/spec_from_iter.rs:33:9
  18: <alloc::vec::Vec<T> as core::iter::traits::collect::FromIterator<T>>::from_iter
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/alloc/src/vec/mod.rs:2649:9
  19: core::iter::traits::iterator::Iterator::collect
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/iter/traits/iterator.rs:1836:9
  20: naga::proc::constant_evaluator::ConstantEvaluator::swizzle
             at /home/jimb/rust/naga/src/proc/constant_evaluator.rs:314:34
```

jimblandy:
Reduced test case:

```
fn bad_compose(a: vec2<f32>, b: vec2<f32>) -> vec4<f32> {
   return vec4<f32>(a, b).wzyx;
}
```

It seems that the constant evaluator's code for `Compose` expressions doesn't actually understand all the cases that `Compose` can handle.

jimblandy:
Since the constant evaluator can be used by the front end, it can't assume that the expressions it's operating on are valid. It seems like almost any indexing operation will need to become a `get` and be checked.

jimblandy (commented):
I believe this patch will cause us to generate more ambitious SPIR-V constants, is that right?

If so, one thing we should check: Naga's `Composite` expression can concatenate vectors: you can build a `vec4` from two `vec2`s with a `Composite` expression. SPIR-V's `OpCompositeConstruct` will do this too - but `OpConstantComposite` does not. Or at least, the spec doesn't say it does. So we should make sure that this works.

edit: I was hoping that posting this comment as a "review" would give it a "resolve" button, but it seems like that didn't work. Oh well.

src/proc/constant_evaluator.rs
```diff
@@ -62,169 +82,107 @@ pub enum ConstantSolvingError {
...
+        expr: &Expression,
+        span: Span,
+    ) -> Result<Handle<Expression>, ConstantEvaluatorError> {
+        match *expr {
+            Expression::Literal(_) | Expression::ZeroValue(_) | Expression::Constant(_) => {
+                Ok(self.register_constant(expr.clone(), span))
+            }
+            Expression::Compose { ref components, .. } => {
+                for component in components {
+                    self.check_and_get(*component)?;
```
jimblandy, line 148: Since `check_and_get` can return a different handle, shouldn't this always be rebuilding the component vector?

jimblandy:
> It seems that the constant evaluator's code for `Compose` expressions doesn't actually understand all the cases that `Compose` can handle.

I have a work in progress patch for this up here: https://github.com/jimblandy/naga/tree/compose-flattens

It doesn't seem to fix the problem, but you can see the idea. In particular:

```
                // `Compose` expressions can concatenate vectors. This is easy
                // for backends to deal with, because all our target languages'
                // vector construction expressions do the same. But for the
                // constant evaluator itself, nearly any operation might need to
                // look inside a `Compose` expression to find the values. It
                // would be difficult for every such consumer to handle the case
                // of vector concatenation.
                //
                // Instead, the constant evaluator flattens out such
                // concatenation expressions here, so that a `Compose`
                // expression building a vector only ever has scalars as
                // operands.
```

jimblandy:
> It doesn't seem to fix the problem, but you can see the idea. In particular:

One reason this doesn't fix the problem is that we apply `try_eval_and_append` to all expressions, regardless of whether they're constants. So `ConstantEvaluator::swizzle` is going to see every `Swizzle` expression the WGSL front end ever generates, regardless of whether it's a constant or not.

Maybe we should call `ConstantEvaluator` something like `PeepholeOptimizer`, because it transforms all code, not just constants.

jimblandy:
> It seems that the constant evaluator's code for `Compose` expressions doesn't actually understand all the cases that `Compose` can handle.

This problem is also present in `ConstantEvaluator::access`.

`ConstantEvaluator::unary_op` is okay because it always recurses on its arguments.

`ConstantEvaluator::binary_op` is "okay" because it doesn't support the case where both operands are `Compose` expressions yet. When it does, it will need to flatten the arguments so that things like `vec4(v2,v2) + vec4(s,v3)` pair up the scalar operands correctly.

jimblandy:
Okay, https://github.com/gfx-rs/naga/pull/2485 addresses some of this??

teoxoy:
I also think only swizzles and accesses have this issue.

> I believe this patch will cause us to generate more ambitious SPIR-V constants, is that right?
> 
> If so, one thing we should check: Naga's `Composite` expression can concatenate vectors: you can build a `vec4` from two `vec2`s with a `Composite` expression. SPIR-V's `OpCompositeConstruct` will do this too - but `OpConstantComposite` does not. Or at least, the spec doesn't say it does. So we should make sure that this works.

Right, I think we need to flatten the components in the backend, will look into it.
HLSL, GLSL and MSL seem to be ok with these constructors.

teoxoy:
`spirv-val` is outputting:

```
error: line 58: OpConstantComposite Constituent <id> count does not match Result Type <id> '7[%v3float]'s vector component count.
  %31 = OpConstantComposite %v3float %26 %float_2
```

Will put up a fix for the SPIR-V backend.

jimblandy:
> Will put up a fix for the SPIR-V backend.

Oh, #2485 has a fix for this now. Sorry I didn't see this.

jimblandy:
Rebased on trunk with compaction.

src/proc/constant_evaluator.rs
teoxoy: Marking as resolved as we've taken the discussion to: https://github.com/gfx-rs/naga/pull/2493.

teoxoy:
Rebased on master and fixed-up ["[wgsl-in] eagerly evaluate const-expressions"](https://github.com/gfx-rs/naga/pull/2309/commits/9f2bab1a66e5accfe0037b575f9b4e15a66ea94c) with ["address some of the comments"](https://github.com/gfx-rs/naga/pull/2309/commits/9668a1be689a3da79da8c81d56074043943decb8).

src/proc/constant_evaluator.rs
teoxoy: Done in https://github.com/gfx-rs/naga/pull/2309/commits/0d536b998dbfe2eb5a14167a365ba4fbee25d02d

jimblandy (changes requested):
Just adding a review to make the PR status reflect our understanding. Please flag me when this is ready for a new review.

jimblandy:
@teoxoy Is it important to permit `Splat` and `ZeroValue` expressions in constant expressions? The evaluator would get simpler if it only had to deal with `Literal` and `Compose`.

teoxoy:
I pushed another commit e77c831fbbe83b4c0d27eb0e318b66426b6118a6 that avoids the issue below (with the modified test in that commit):
```
thread 'convert_wgsl' panicked at 'error: Cannot apply the binary op to the arguments
   ┌─ wgsl:59:43
   │
59 │ const TEST_CONSTANT_ALIAS_ADDITION: i32 = FOUR_ALIAS + FOUR_ALIAS;
   │                                           ^^^^^^^^^^^^^^^^^^^^^^^ see msg
```

teoxoy:
> @teoxoy Is it important to permit `Splat` and `ZeroValue` expressions in constant expressions? The evaluator would get simpler if it only had to deal with `Literal` and `Compose`.

There are some cases where we can avoid having to evaluate them before a given operation and for the other cases we can use `eval_zero_value_and_splat` to break them down into `Literal`s or `Compose`s.

jimblandy:
I noticed that `ConstantEvaluator::math` doesn't bother to call `eval_zero_value_and_splat`. But since there are so many missing math functions anyway, it seems like we should just treat that as just one aspect of the existing follow-up work.

I tried calling `eval_zero_and_splat` there, but of course it didn't make much difference. In fact all it did was bloat the module with unused `Compose` expressions equivalent to nearby splat expressions.

teoxoy:
I just pushed some changes to address that.

src/proc/constant_evaluator.rs
```diff
@@ -497,60 +499,181 @@ impl<'a> ConstantEvaluator<'a> {
...
+                    self.types,
+                ))
+                .collect();
+
+                let mid = components.len() / 2;
+                let (first, last) = components.split_at_mut(mid);
+                for (a, b) in first.iter_mut().zip(&*last) {
+                    *a = self.math_pow(*a, *b, span)?;
+                }
+                components.drain(mid..);
```
jimblandy, line 546: You can just use `Vec::truncate` here, I think.

src/proc/constant_evaluator.rs
teoxoy: Resolved in https://github.com/gfx-rs/naga/pull/2309/commits/e139963fde2a18dad342a19b7d0933df320aba86

jimblandy:
rebased on master

jimblandy:
I think this PR is going to make a real mess of our snapshot tests.
A lot of the snapshot tests are written like this:

```
fn negation_avoids_prefix_decrement() {
    let p0 = -1;
    let p1 = - -2;
    let p2 = -(-3);
    let p3 = -(- 4);
    let p4 = - - -5;
    let p5 = - - - - 6;
    let p6 = - - -(- -7);
    let p7 = (- - - - -8);
}
```

The intent of this test was to ensure that WGSL generation didn't accidentally squash pairs of unary negation operators into a single `--` prefix operator.

But with constant evaluation, the actual snapshot output for this is:

```
fn negation_avoids_prefix_decrement() {
    return;
}
```

All the expressions get evaluated eagerly to literals, which are `needs_pre_emit`, so we don't generate `let` declarations for them anymore, and they disappear entirely.

But the upshot is that the test is no longer covering what it was intended to.

jimblandy:
I dunno, maybe "make a real mess" is an exaggeration. It's not all of them. We just need to look at the snapshot output, fix up the ones that got subverted, and then rebase on top of that.

jimblandy:
This is weird:

```
     let _matrix = bar._matrix;
     let arr_1 = bar.arr;
-    let b = bar._matrix[3][0];
+    let b = bar._matrix[3u][0];
     let a_1 = bar.data[(arrayLength((&bar.data)) - 2u)].value;
     let c = qux;
```

jimblandy:
On master, the statements:

```
	let index = 3u;
	let b = bar._matrix[index].x;
```

produce the following IR:

```
                    [12]: Literal(
                        U32(
                            3,
                        ),
                    ),
                    [13]: GlobalVariable(
                        [2],
                    ),
                    [14]: AccessIndex {
                        base: [13],
                        index: 0,
                    },
                    [15]: AccessIndex {
                        base: [14],
                        index: 3,
                    },
                    [16]: AccessIndex {
                        base: [15],
                        index: 0,
                    },
                    [17]: Load {
                        pointer: [16],
                    },
```

On this PR's branch, they produce:

```
                    [12]: Literal(
                        U32(
                            3,
                        ),
                    ),
                    [13]: GlobalVariable(
                        [2],
                    ),
                    [14]: AccessIndex {
                        base: [13],
                        index: 0,
                    },
                    [15]: Access {
                        base: [14],
                        index: [12],
                    },
                    [16]: AccessIndex {
                        base: [15],
                        index: 0,
                    },
                    [17]: Load {
                        pointer: [16],
                    },
```

I would not have expected this branch to change an `AccessIndex` expression into an `Access` expression.

jimblandy:
This was introduced by "[wgsl-in] don't treat `let` declarations as `const` declarations".

I generally dislike implementations like this, especially the 'force_non_const` kludge, because it makes the data structures much harder to understand. If the thing is called a "constness tracker", then it should track constness and not do other odd jobs.

jimblandy:
@teoxoy Why was "[wgsl-in] don't treat let declarations as const declarations" necessary? Could we drop that change?

jimblandy:
This change is no good, in `tests/out/wgsl/atomicCompareExchange.wgsl`. You can't hoist a local variable's initialization up to the top level if that variable is local to a loop.

```diff
 @compute @workgroup_size(1, 1, 1) 
 fn test_atomic_compare_exchange_i32_() {
-    var i: u32;
+    var i: u32 = 0u;
     var old: i32;
-    var exchanged: bool;
+    var exchanged: bool = false;
 
-    i = 0u;
     loop {
         let _e2 = i;
         if (_e2 < SIZE) {
@@ -22,7 +21,6 @@ fn test_atomic_compare_exchange_i32_() {
             let _e6 = i;
             let _e8 = atomicLoad((&arr_i32_[_e6]));
             old = _e8;
-            exchanged = false;
             loop {
                 let _e12 = exchanged;
                 if !(_e12) {
```

jimblandy (changes requested):
See comment about `tests/out/wgsl/atomicCompareExchange.wgsl`.

teoxoy:
> This change is no good, in `tests/out/wgsl/atomicCompareExchange.wgsl`. You can't hoist a local variable's initialization up to the top level if that variable is local to a loop.
> 
> ```diff
>  @compute @workgroup_size(1, 1, 1) 
>  fn test_atomic_compare_exchange_i32_() {
> -    var i: u32;
> +    var i: u32 = 0u;
>      var old: i32;
> -    var exchanged: bool;
> +    var exchanged: bool = false;
>  
> -    i = 0u;
>      loop {
>          let _e2 = i;
>          if (_e2 < SIZE) {
> @@ -22,7 +21,6 @@ fn test_atomic_compare_exchange_i32_() {
>              let _e6 = i;
>              let _e8 = atomicLoad((&arr_i32_[_e6]));
>              old = _e8;
> -            exchanged = false;
>              loop {
>                  let _e12 = exchanged;
>                  if !(_e12) {
> ```

Good catch! I've pushed https://github.com/gfx-rs/naga/pull/2309/commits/4eeeb4921345b9ae7cd9baa26db03976d755ae51. It's also worth noting that we could have run into this pre https://github.com/gfx-rs/naga/pull/2075 since we were doing something similar https://github.com/gfx-rs/naga/pull/2075#issuecomment-1317151171.

Long-term we should consider moving local var declarations to the top in the SPIR-V backend instead since none of our other backends require this.

> All `OpVariable` instructions in a function must be in the first block in the function.

teoxoy:
> Why was "[wgsl-in] don't treat let declarations as const declarations" necessary? Could we drop that change?

I went through the spec recently and noticed we were not upholding its rules when it came to const-expressions.
Specifically:

> An expression is a const-expression if all its identifiers resolve to:
> - const-declarations, or
> - const-functions, or
> - type aliases, or
> - structure names.

The `AccessIndex` -> `Access` is also fine since we need to relax our errors on runtime expressions (https://github.com/gfx-rs/naga/issues/1924).

> I generally dislike implementations like this, especially the `force_non_const` kludge, because it makes the data structures much harder to understand. If the thing is called a "constness tracker", then it should track constness and not do other odd jobs.

The fact that our IR is not a direct representation of WGSL is the reason I couldn't come up with a better implementation. We translate `let` declarations to `Expression`s with an extra entry in `named_expressions` and not to a special kind of `Expression`. I'm open to suggestions though.

jimblandy:
> I went through the spec recently and noticed we were not upholding its rules when it came to const-expressions.

Okay, I see. This makes me a little more comfortable. I've pushed [53867cf](https://github.com/gfx-rs/naga/pull/2309/commits/53867cfab38647d30a148316dd5f916c64ade779) to at least explain the situation.

src/proc/constant_evaluator.rs
```diff
@@ -0,0 +1,1710 @@
...
+
+    fn register_evaluated_expr(
+        &mut self,
+        expr: Expression,
+        span: Span,
+    ) -> Result<Handle<Expression>, ConstantEvaluatorError> {
+        match expr {
+            Expression::Literal(literal) => crate::valid::validate_literal(literal)?,
+            _ => {}
+        }
```
jimblandy, line 1160: In the long run, I'm concerned that we're going to run into a steady stream of problems around error reporting, because it's a fundamental aspect of this design that constant evaluation takes place before other validation, which seems surprising to me. We have to be very careful to only const-evaluate valid expressions, lest we remove errors that need to be reported. And that means that we need to re-implement validation in the constant evaluator.

For now this is fine, but let's keep an eye on this.

jimblandy: All that said, this particular call to `validate_literal` is fine because it's checking for a condition that arises only through evaluation. Validating first wouldn't help.

teoxoy: The validation code is checking invariants that can't be checked by the type system given the current shape of the IR. Which more specifically for `Expression`s is checking the types of the operands.

The evaluator is operating on strongly typed data (`Literal`s) and is inherently checking the types involved in a given `Expression`.

I'm not sure if I'm missing something but I don't see this being a big issue.

src/lib.rs
jimblandy, line 1: Right now our documentation for `Module::const_expressions` is wrong. It says it can hold "constant expressions", which we define at the top of the file much more liberally than the rules currently enforced by `Validator::validate_const_expression` actually allow.

Shouldn't we be updating that to actually explain what backends can expect?

src/lib.rs
```diff
@@ -978,9 +978,7 @@ pub struct LocalVariable {
     pub name: Option<String>,
     /// The type of this variable.
     pub ty: Handle<Type>,
-    /// Initial value for this variable.
-    ///
-    /// Expression handle lives in const_expressions
+    /// Initial value for this variable. Must be a const-expression.
```
jimblandy, line 981: The term "const-expression" isn't defined anywhere in this file. I assume it should say "evaluated expression"?

There's no validation of this constraint.

Why did this move? It seems odd to require it to be an evaluated expression (I think that's the terminology we settled on, right?) but have it live in the function's expression arena.

src/lib.rs
teoxoy: When we add full support for override expressions, the `const_expressions` arena will hold the other variants of "constant expressions".

src/lib.rs
teoxoy: > The term "const-expression" isn't defined anywhere in this file. I assume it should say "evaluated expression"?

`const-expression` is what the spec calls them but we have defined them as "constant expressions" at the top of the file in https://github.com/gfx-rs/naga/commit/c7490b9832468ca7d1539978091c4769fe148a52; the "use LocalVariable init" commit originally predated that commit and updates like these get lost in rebases.

Similarly to the comment I left on the discussion above, when we add full support for override expressions, this will also be allowed to refer to override expressions which will include the other variants of "constant expressions".

> There's no validation of this constraint.

Added in https://github.com/gfx-rs/naga/pull/2309/commits/1c781270415a6ecfe808b98d6f48a525417343fd.

> Why did this move?

We would have had to copy the evaluated expressions into the `const_expressions` arena which didn't seem worth doing.

teoxoy:
If we want we can generally document that currently the IR only contains the evaluated subset of const-expressions. And delete that later.

jimblandy (approved):
I think this is ready to go.

Issue #1939, opened 2022-05-23
[hlsl-out] Incorrect behaviour with firstTrailingBit


hasali19:
This program incorrectly outputs `-1` with DirectX. I think this is because of the use of `firstbitlow`, since it works correctly with tint which uses its own polyfill for `firstbitlow` - perhaps naga needs to do the same?

WGSL:

```
@group(0)
@binding(1)
var<storage, read_write> k: i32;

@compute
@workgroup_size(1)
fn main() {
    k = max(~0, firstTrailingBit(1));
}
```

HLSL:

```hlsl
RWByteAddressBuffer k : register(u1);

[numthreads(1, 1, 1)]
void main()
{
    k.Store(0, asuint(max(~0, firstbitlow(1))));
    return;
}
```

evahop:
I currently don't have a method for testing `hlsl`, but this is perplexing. At first glance this would imply `firstbitlow` returns `-1` for a bit at the `0` index. That seems very counterintuitive.

teoxoy:
Looking at the [src of DXC](https://github.com/microsoft/DirectXShaderCompiler/blob/5874b72e81da082ab6ba585dc5f1592c3df27f61/utils/hct/gen_intrin_main.txt#L140), `firstbithigh` might be in the same boat but I'm not sure what the meaning of the syntax in that file is.

evahop:
I’m not sure of the syntax either but `firstbithigh` does appear to produce the same `umax` in the OP example. I will update this where needed.

PR #2315, opened 2023-04-24
[hlsl-out] Fix return type for firstbitlow/high

evahop:
Fixes https://github.com/gfx-rs/naga/issues/1939 where `firtbitlow` and `firstbithigh` always return a `uint`.


teoxoy (approved):
Looks great! Thank you!

Issue #1491, opened 2021-10-27
[wgsl-out] Missing `flat` qualifiers on vertex outputs or fragment inputs


cwfitzgerald:
Chrome Canary complains when I pass a naga-converted shader to it:

```
1 warning(s) generated while compiling the shader:
21:21 warning: integral user-defined vertex outputs must have a flat interpolation attribute
    [[location(3)]] member1: u32;
                    ^^^^^^^
```

Same for fragment intputs. We seem to just be missing `[[interpolate(flat)]]`.

teoxoy:
Opened https://github.com/gpuweb/gpuweb/issues/2760 to clarify the requirement (all user-defined IO vs user-defined vertex outputs and fragment inputs).

expenses:
I've been running into this problem with the shipped version of WebGPU in Chrome Canary. I'll try and cook up a fix.

expenses:
Here are examples of manual changes I've had to make:

```diff
diff --git a/single_view_fragment.spv.wgsl b/testing/single_view_fragment.spv.wgsl
index a410794..71e2045 100644
--- a/single_view_fragment.spv.wgsl
+++ b/testing/single_view_fragment.spv.wgsl
@@ -266,7 +266,7 @@ fn function() {
 }

 @fragment
-fn single_view__fragment(@location(0) param: vec3<f32>, @location(1) param_1: vec3<f32>, @location(2) param_2: vec2<f32>, @location(3) param_3: vec2<f32>, @location(5) param_4: u32, @builtin(front_facing) param_5: bool, @location(4) param_6: u32) -> @location(0) vec4<f32> {
+fn single_view__fragment(@location(0) param: vec3<f32>, @location(1) param_1: vec3<f32>, @location(2) param_2: vec2<f32>, @location(3) param_3: vec2<f32>, @location(5) @interpolate(flat) param_4: u32, @builtin(front_facing) param_5: bool, @location(4) @interpolate(flat) param_6: u32) -> @location(0) vec4<f32> {
     global_2 = param;
     global_3 = param_1;
     global_4 = param_2;
```

```diff
diff --git a/single_view_vertex.spv.wgsl b/testing/single_view_vertex.spv.wgsl
index 0afa948..5f013e1 100644
--- a/single_view_vertex.spv.wgsl
+++ b/testing/single_view_vertex.spv.wgsl
@@ -56,8 +56,8 @@ struct VertexOutput {
     @location(1) member_2: vec3<f32>,
[...]

PR #2318, opened 2023-04-26
[wgsl-out] Simplify map_binding_to_attribute to always return an interpolate attribute

expenses:
Effectively reverts https://github.com/gfx-rs/naga/pull/1103. Fixes https://github.com/gfx-rs/naga/issues/1491. @teoxoy @cwfitzgerald @kvark.


teoxoy:
Thanks for the PR! CI is failing since you need to also push the updated test snapshots.

PR #2319, opened 2023-04-26
Fold `Error` source chain `Display` output into shader errors

### LINKED ISSUE #4494: Consolidate diagnostic presentation logic into library error types
## Problem statement

Diagnostic information for type mismatch errors (and, from what I've seen, most validation errors) is fragmented to the point that it undermines all but immediate consumers of Naga CLI. Some quick examples:

* …from https://github.com/webgpu/webgpu-samples/blob/85dbf0dc496e75691adfc795830e9a046e46e6bc/src/sample/imageBlur/blur.wgsl#L39:

image

* …from https://github.com/webgpu/webgpu-samples/blob/85dbf0dc496e75691adfc795830e9a046e46e6bc/src/sample/particles/particle.wgsl#L109:

image

Ignore that we're rejecting perfectly fine code[^1] for a second. **Note how there are _three_ places we had to consult `naga-cli`'s output to consume all information provided[^2] for this message:**

1. The `log::error!(…)` output of `src/valid/{compose,expression}.rs`, to actually learn about the types that aren't matching.
2. A nicely formatted set of spans provided by `WithSpan::spans`, accessed in [`emit_annotated_error`](https://github.com/ErichDonGubler/naga/blob/67ea8f0c0684d80e1f54af9abd875034748d5e07/cli/src/bin/naga.rs#L551).
3. Finally, the `stderr` output of `naga-cli` emitted by [`print_err`](https://github.com/ErichDonGubler/naga/blob/67ea8f0c0684d80e1f54af9abd875034748d5e07/cli\src\bin\naga.rs#L164), which actually tells us the kind of expression that failed.

Because of this, **we are not creating a uniformly good experience** between users of Naga CLI and users of Naga as a library. Evidence of this can already be found in current Nightly builds of Firefox where _only (2) is actually presented in the JS console_ (😢 [related bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1829898)):

image

…where Firefox is missing critical information from (1) and (3) entirely. 😬❗

## Solution statement

A great example of how the above can all presented to a user can be found in [the front page of the `miette` project](https://github.com/zkat/miette/tree/2b4d67d7cd1ced4f0e81e74a0c636a8cfcf2074e/#example):

**My proposed solution is to expose an API to combine all error information into one or more `codespan_reporting::Diagnostic` (with spans, if available), which downstream can work with as they determine fit.** Concretely, the next short-term steps I see are:

- [x] Currently, both `naga-cli` and `wgpu-core` transform specific errors into `codespan_reporting::Diagnostic`s that are then rendered as separate terminal output. **`naga`-the-library should own the conversion of errors it returns into `Diagnostic`s**, and offer consumers like `wgpu-core` and `naga-cli` a chance to work with those instead.
- [ ] Instead of using `log::error!(…)` and the like as a separate fragmentary channel of error context, **`naga`'s front-end and validation errors should carry all relevant information in `Result::Err`s** returned from its method and function calls.

    The (non-exhaustive) list of known cases of this include:
    
	- [ ] Type mismatches
		- [ ] Binary operations

[^1]: For reference, resolving this is scoped to gfx-rs/wgpu#4400.
[^2]: There _is_ still something missing: we don't get any span help showing us which type corresponds to what locations in source, still. 😮‍💨 The `miette` example I provide later on should make this clear.

ErichDonGubler:
Partially resolves gfx-rs/wgpu#4494.

### DIFF
--- a/cli/src/bin/naga.rs
+++ b/cli/src/bin/naga.rs
@@ -564,7 +564,7 @@ pub fn emit_annotated_error<E: Error>(ann_err: &WithSpan<E>, filename: &str, sou
         .unwrap()
         // NOTE: `replace` is intended to correct the alignment of multi-line messages against the
         // `error: ` prefix, which is 7 characters.
-        .replace("\n", "\n       ");
+        .replace('\n', "\n       ");
     let diagnostic = Diagnostic::error().with_message(msg).with_labels(
         ann_err
             .spans()
--- a/cli/src/bin/naga.rs
+++ b/cli/src/bin/naga.rs
@@ -345,8 +345,9 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
             if let Some(input) = input_text {
                 let filename = input_path.file_name().and_then(std::ffi::OsStr::to_str);
                 emit_annotated_error(&error, filename.unwrap_or("input"), &input);
+            } else {
+                print_err(&mut stderr().lock(), &error);
             }
-            print_err(&mut stderr().lock(), &error);
             None
         }
     };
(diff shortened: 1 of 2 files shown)


cwfitzgerald:

PR #2332, opened 2023-05-10
Introduce `Expression::ZeroValue`.

jimblandy:
This is part of #2266. This change causes a lot of changes to the test output, so it's easier to review the rest of the constexpr work with this factored out.

This is all authored by @teoxoy; I just separated it out. I'll figure out how to fix up the commit author before we land this.

This needs to be rebased on #2331.


jimblandy:
@teoxoy, if you could look over the two commits here that are not part of #2331 and make sure I've done a reasonable job extracting `ZeroValue` from the rest, that'd be great.

teoxoy (approved):
Looks good!

jimblandy (commented):
Looks great.

PR #2333, opened 2023-05-10
Introduce `Expression::Literal`.

jimblandy:
This is part of #2266, which I'm trying to break down into smaller pieces that I can review. @teoxoy is the author; I'll figure out how to fix up the commit to say so tomorrow.

This introduces `Expression::Literal` to the IR, and supports it in the backends, but there's nothing in the front ends to introduce such expressions, so in its current state it has no effect on the snapshots.

This depends on #2331 and #2332.


jimblandy:
For the record, I've checked all the SPIR-V snapshots and they look fine.

I haven't actually reviewed the code changes yet.

src/back/spv/writer.rs
```diff
@@ -1812,13 +1782,17 @@ impl Writer {
             match constant.inner {
                 crate::ConstantInner::Composite { .. } => continue,
                 crate::ConstantInner::Scalar { width, ref value } => {
+                    let literal = crate::Literal::from_scalar(*value, width).ok_or_else(|| {
+                        eprintln!("JIMB: value = {value:?}, width = {width}");
```
teoxoy, line 1786: Looks like this made its way into the commit.

teoxoy (approved):
Looks good!

PR #2335, opened 2023-05-11
Introduce `GlobalCtx`

teoxoy:
Split from https://github.com/gfx-rs/naga/pull/2266


jimblandy (approved):
Filed #2336 with some naming followup.

PR #2338, opened 2023-05-11
Correct type of index and flush the emitter

teoxoy:
Fixes failing CI https://github.com/gfx-rs/naga/actions/runs/4950457380/jobs/8854096640


jimblandy (changes requested):
I know CI is on fire, but let's extract out that common code into its own function. This much duplication is not great.

jimblandy (approved):

Issue #2312, opened 2023-04-24
[wgsl-in] Redefining parameter is incorreclty allowed


Wumpf:
The following snippet redefines an incoming parameter which should not compile (and Tint will complain about it!):

```
fn fun(t: f32) -> f32 {
    let t = t + 1.0;
    return t;
}

```

PR #2342, opened 2023-05-15
[wgsl-in] Error on param redefinition

SparkyPotato:
Previously,
```rs
fn x(a: f32, a: vec2<f32>) {}
```
was silently accepted by naga, binding the second parameter to `a`.

Now, it errors with
```rs
error: redefinition of `a`
  ┌─ ..\naga tests\wgsl.wgsl:1:6
  │
1 │ fn x(a: f32, a: vec2<f32>) {}
  │      ^       ^ redefinition of `a`
  │      │
  │      previous definition of `a`
```

### DIFF
--- a/src/front/wgsl/parse/mod.rs
+++ b/src/front/wgsl/parse/mod.rs
@@ -2011,15 +2011,15 @@ impl Parser {
         ctx.local_table.push_scope();
 
         lexer.expect(Token::Paren('{'))?;
-        let mut statements = ast::Block::default();
+        let mut block = ast::Block::default();
         while !lexer.skip(Token::Paren('}')) {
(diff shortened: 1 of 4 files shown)


teoxoy:
Thanks for the PR; it looks good!

I think this would fix https://github.com/gfx-rs/naga/issues/2312 as well; would you mind adding one more test case for it?

SparkyPotato:
> Thanks for the PR; it looks good!
> 
> I think this would fix #2312 as well; would you mind adding one more test case for it?

The issue there was that we were pushing a scope for the arguments, and then another for the block. I  fixed that by not using `self.block`, and commenting the reason why so we don't end up with a regression during future refactoring.

teoxoy (approved):
Looks good, thanks!

PR #2347, opened 2023-05-18
Handle case insensitive FXC HLSL keywords.

PJB3005:
Honestly I realized while almost done with this that maybe trying to put maintenance effort into FXC rather than DXC wouldn't be worth it, but OH WELL. Feel free to deny this PR if we don't want to add bugfixes for FXC.

There are a few keywords like "pass" in HLSL that are actually case-insensitive for FXC. This can be disabled with strict mode, but even if you do that FXC will continue to give an error if you try to use them in identifiers (at all, with any casing).

This changes the namer code to escape these keywords even if the casing is different.

If you're wondering where I got the list from: I looked at the list of strings in D3DCompiler_47.dll. None of this is documented.

### DIFF
--- a/src/back/hlsl/keywords.rs
+++ b/src/back/hlsl/keywords.rs
@@ -4,9 +4,23 @@ HLSL Reserved Words
 - <https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-appendix-reserved-words>
 */
 
+// When compiling with FXC without strict mode, these keywords are actually case insensitive.
+// If you compile with strict mode and specify a different casing like "Pass" instead in an identifier, FXC will give this error:
+// "error X3086: alternate cases for 'pass' are deprecated in strict mode"
+// This behavior is not documented anywhere, but as far as I can tell this is the full list.
+pub const RESERVED_CASE_INSENSITIVE: &[&str] = &[
+    "asm",
+    "decl",
+    "pass",
+    "technique",
+    "Texture1D",
+    "Texture2D",
+    "Texture3D",
+    "TextureCube",
+];
+
 pub const RESERVED: &[&str] = &[
     "AppendStructuredBuffer",
-    "asm",
     "asm_fragment",
     "BlendState",
     "bool",
@@ -68,7 +82,6 @@ pub const RESERVED: &[&str] = &[
     "out",
     "OutputPatch",
     "packoffset",
-    "pass",
     "pixelfragment",
     "PixelShader",
     "point",
@@ -101,17 +114,13 @@ pub const RESERVED: &[&str] = &[
     "switch",
     "StructuredBuffer",
     "tbuffer",
-    "technique",
     "technique10",
     "technique11",
     "texture",
-    "Texture1D",
     "Texture1DArray",
-    "Texture2D",
     "Texture2DArray",
     "Texture2DMS",
     "Texture2DMSArray",
-    "Texture3D",
     "TextureCube",
     "TextureCubeArray",
     "true",
--- a/src/back/hlsl/keywords.rs
+++ b/src/back/hlsl/keywords.rs
@@ -121,7 +121,6 @@ pub const RESERVED: &[&str] = &[
     "Texture2DArray",
     "Texture2DMS",
     "Texture2DMSArray",
-    "TextureCube",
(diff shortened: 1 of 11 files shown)


src/back/hlsl/keywords.rs
```diff
@@ -101,17 +114,13 @@ pub const RESERVED: &[&str] = &[
...
     "technique11",
     "texture",
-    "Texture1D",
     "Texture1DArray",
-    "Texture2D",
     "Texture2DArray",
     "Texture2DMS",
     "Texture2DMSArray",
-    "Texture3D",
     "TextureCube",
```
teoxoy, line 124:
```suggestion
```

It's now in `RESERVED_CASE_INSENSITIVE`.

src/proc/namer.rs
```diff
@@ -102,7 +104,12 @@ impl Namer {
             }
             None => {
                 let mut suffixed = base.to_string();
-                if base.ends_with(char::is_numeric) || self.keywords.contains(base.as_ref()) {
+                if base.ends_with(char::is_numeric)
+                    || self.keywords.contains(base.as_ref())
+                    || self
+                        .keywords_case_insensitive
+                        .contains(&base.as_ref().into())
+                {
```
teoxoy, line 112: I think we can use [`eq_ignore_ascii_case`](https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case) instead of pulling in a new dep.

teoxoy (changes requested):
Thanks for the PR! I double checked FXC's behavior and this seems needed for the time being.

src/proc/namer.rs
PJB3005: I'm not sure how you'd use that with a hash set. Or would it be fine to just use a slice and iterate over it?

teoxoy: Right, I think what we need is just a wrapper type and a hash impl for it (similar to [this one](https://github.com/seanmonstar/unicase/blob/master/src/ascii.rs#L124-L131)).

```rust
/// A string wrapper type with an ascii case insensitive Eq and Hash impl
struct AsciiUniCase<S: AsRef<str>>(S);

impl<S: AsRef<str>> PartialEq<Self> for AsciiUniCase<S> {
    #[inline]
    fn eq(&self, other: &Self) -> bool {
        self.0.as_ref().eq_ignore_ascii_case(other.0.as_ref())
    }
}

impl<S: AsRef<str>> Eq for AsciiUniCase<S> {}

impl<S: AsRef<str>> Hash for AsciiUniCase<S> {
    #[inline]
    fn hash<H: Hasher>(&self, hasher: &mut H) {
        for byte in self.0.as_ref().as_bytes().iter().map(|b| b.to_ascii_lowercase()) {
            hasher.write_u8(byte);
        }
    }
}
```

plus a debug assert to make sure the reserved words are ascii.

```rust
debug_assert!(reserved_keywords_case_insensitive.iter().all(|s|s.is_ascii()));
```

src/proc/namer.rs
```diff
@@ -102,7 +104,12 @@ impl Namer {
             }
             None => {
                 let mut suffixed = base.to_string();
-                if base.ends_with(char::is_numeric) || self.keywords.contains(base.as_ref()) {
+                if base.ends_with(char::is_numeric)
+                    || self.keywords.contains(base.as_ref())
+                    || self
+                        .keywords_case_insensitive
+                        .contains(&AsciiUniCase(base.to_owned().into_owned()))
```
teoxoy, line 111: This will allocate unnecessarily, is there anything missing to get the suggestion working?

```suggestion
                        .contains(&AsciiUniCase(base))
```

PJB3005: I tried messing around with this (basically doing a `String` -> `&str` type conversion), but as far as I understand of Rust that would require unsafe code (that I'm not even sure would be stable). We'd need to implement `Borrow<T>` somehow which means we'd need to go `AsciiUnicase<String>` -> `&AsciiUnicase<str>`.

teoxoy: I think we can instead use a `&'static str` type since all those keywords are 'static (or `Cow<'static, str>` if we need to).
```rust
keywords_case_insensitive: FastHashSet<AsciiUniCase<&'static str>>,
```

PR #2353, opened 2023-05-25
Implement Pack/Unpack for HLSL

### LINKED ISSUE #1929: Missing extractBits and insertBits implementations for HLSL
Seems they were implemented for other backends in #1449 but not for HLSL.

Elabajaba:
~~These are currently untested, and I'm not sure how to do the signed versions.~~

edit: These have now all been tested against a re-implementation of the CTS tests [here](https://github.com/Elabajaba/wgpu/tree/mynagatest) (set `WGPU_BACKEND` env variable to dx12 and run the `naga-valid` example, it'll print out any failed tests).

Status:

- [x] extractBits
- [x] insertBits
- [x] pack4x8snorm
- [x] pack4x8unorm
- [x] pack2x16snorm
- [x] pack2x16unorm
- [x] pack2x16float
- [x] unpack4x8snorm
- [x] unpack4x8unorm
- [x] unpack2x16snorm
- [x] unpack2x16unorm
- [x] unpack2x16float (was already implemented)

Related: #1449 
Closes #1929


Elabajaba:
The HLSL output for what I've implemented is now valid, but I'm not sure how to test that the output is correct.

teoxoy:
Please add HLSL here

https://github.com/gfx-rs/naga/blob/907b7c7cda2447f2869820a27b8f331d7c512fab/tests/snapshots.rs#L448

so that we have and check the HLSL snapshots. CI will only make sure the output is valid HLSL but not run the file.

> but I'm not sure how to test that the output is correct.

We aren't currently testing behavior in naga, what I'd do is create a compute shader to test the behavior and open a PR on wgpu (that temporarily links to your rev of naga).

Elabajaba:
I hacked up a compute shader to test with (though I'm currently running it as an example and not a test, see the `naga-valid` example https://github.com/Elabajaba/wgpu/tree/mynagatest).

I've run into an issue where my `pack4x8snorm` and `pack2x16snorm` implementations are off by 1 compared to the spec when given -0.5 (for `pack2x16snorm([-0.5, -0.5])`, spec: `0xc001c001`, mine: `0xc000c000`), but I get the same off by 1 result when running it with Vulkan (Windows + AMD).

packing spec: https://github.com/gpuweb/cts/blob/main/src/unittests/conversion.spec.ts
unpacking spec: https://github.com/gpuweb/cts/blob/main/src/unittests/floating_point.spec.ts

teoxoy:
The issue is that the WGSL spec defines the behavior of `pack2x16snorm` as follows:

> Component e[i] of the input is converted to a 16-bit twos complement integer value ⌊ 0.5 + 32767 × min(1, max(-1, e[i])) ⌋ which is then placed in bits 16 × i through 16 × i + 15 of the result.

We need to floor instead of round which is the behavior of SPV and what the PR is currently implementing.

Having to polyfill this for SPV as well is unfortunate but I guess necessary since the rounding behavior might not always be the same across implementations of SPV.

> The fraction 0.5 rounds in a direction chosen by the implementation, presumably the direction that is fastest.

teoxoy:
I wonder how MSL behaves since they don't specify this in their spec.

src/back/hlsl/writer.rs
```diff
@@ -2675,13 +2731,135 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                                self.write_expr(module, arg, func_ctx)?;
+                                write!(
+                                    self.out,
+                                    "[0], -1.0, 1.0) * {scale}.0 ))) & 0xFFFF) | (((int(round(clamp("
+                                )?;
+                                self.write_expr(module, arg, func_ctx)?;
+                                write!(
+                                    self.out,
+                                    "[1], -1.0, 1.0) * {scale}.0)) & 0xFFFF) << 16))",
+                                )?;
```
teoxoy, line 2795: There are also `+ 0.5` missing here.

teoxoy:
I filed https://github.com/gpuweb/gpuweb/issues/4159 to hopefully iron out the behavior. Maybe we can wait a bit and see how to proceed.

Elabajaba:
I could've saved so much time had I known about https://github.com/gpuweb/gpuweb/pull/1189#issuecomment-721361704...

teoxoy:
I feel you 😅 - the spec repo usually contains a lot of info in PRs/issues since the decisions are taken based on what's in the APIs/shading languages it targets. I find myself using `git log -S` on the spec repo quite often.

Elabajaba:
`insertBits` is passing every test except the 32 count tests, and I'm not sure how to fix it.

```
insertBits(e: 0, newbits: 0b11111111111111111111111111111111, offset: 0, count: 32)
result:   0b00000000000000000000000000000000,
expected: 0b11111111111111111111111111111111
```

readable version of what I'm doing:
```
    let mask = ((((1u << count) - 1u) << offset) & 0xffffffffu);
    return = ((newbits << offset) & mask) | (e & ~mask);
```

magcius:
Could you do `let mask = (0xFFFFFFFFu >> (32 - count)) << offset` instead?

Elabajaba:
> Could you do `let mask = (0xFFFFFFFFu >> (32 - count)) << offset` instead?

Doesn't work when `count = 0` :(

Elabajaba:
The spec also calls for us to error on insertbits if we try and insert too much https://www.w3.org/TR/WGSL/#insertBits-builtin

If `count + offset` is greater than `w` (w is the bitwidth of `T`, where `T` is i32, u32, vecN\<i32>, or vecN\<u32>), then:

    It is a shader-creation error if count and offset are const-expressions.

    It is a pipeline-creation error if count and offset are override-expressions.

We don't currently do that for HLSL or Vulkan, and I don't know how to implement that.

teoxoy:
We can take care of that validation later once we got const-expressions fully working.

src/back/hlsl/writer.rs
```diff
@@ -2675,13 +2733,175 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                        }
+                    }
+                    Function::Pack2x16float => {
+                        write!(self.out, "f32tof16(")?;
+                        self.write_expr(module, arg, func_ctx)?;
+                        write!(self.out, "[0]) | (f32tof16(")?;
+                        self.write_expr(module, arg, func_ctx)?;
+                        write!(self.out, "[1]) << 16)")?;
+                    }
+                    Function::PackBits {
```
Elabajaba, line 2783: I was hoping to have less code duplication by putting the `Pack*` and `Unpack*` functions into PackBits/UnpackBits, but I didn't end up doing that as it seemed messier to have a bunch of `if`s scattered through the implementation vs just partially duplicating the code.

Should I move all the `Pack*` and `Unpack*` functions into their own implementations? (eg.                     `Function::Pack2x16unorm => {...}, Function::Pack4x8snorm => {...}, etc`)

teoxoy: That would be preferred since it would eliminate the need for all the remaining `if`s. But do keep the dims/scale separate (don't inline them).

src/back/hlsl/writer.rs
```diff
@@ -126,7 +126,17 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
                     crate::MathFunction::Asinh
                     | crate::MathFunction::Acosh
                     | crate::MathFunction::Atanh
-                    | crate::MathFunction::Unpack2x16float => {
+                    | crate::MathFunction::Unpack2x16float
+                    | crate::MathFunction::Unpack2x16snorm
+                    | crate::MathFunction::Unpack2x16unorm
+                    | crate::MathFunction::Unpack4x8snorm
+                    | crate::MathFunction::Unpack4x8unorm
+                    // TODO: These use multiple args, unsure how to bake them
```
teoxoy, line 134: Do you mean for insert/extract bits? If so, you'd have to insert the other `arg`s in `need_bake_expressions` as well.

teoxoy (changes requested):
I have not yet looked at the behavior but still left some comments.

Also, it seems the resolution of https://github.com/gpuweb/gpuweb/issues/4159 is that the rounding mode should have been unspecified so, we can use `round` (same as here https://github.com/gpuweb/gpuweb/pull/1189#issuecomment-721361704) instead of `floor(0.5 + x)` for the packing functions.

src/back/hlsl/writer.rs
Elabajaba: Do you mean changing eg. [this block](https://github.com/Elabajaba/naga/blob/217b14bdbb9439ddd10e746a433bbfd9448f23be/src/back/hlsl/writer.rs#L2691) from

```rust
Mf::Pack2x16unorm => Function::PackBits {
    signed: false,
    dims: 2,
    scale: 65535,
},
```

to

```rust
Mf::Pack2x16unorm => Function::Pack2x16unorm {
    dims: 2,
    scale: 65535,
},
```

src/back/hlsl/writer.rs
Elabajaba: I'm not sure why, but insert and extract bits both work without having their expressions baked.

Maybe because they're gated on arg1/2/3 being `Some()`?

teoxoy: I guess it works fine but the chain of expressions will be duplicated (written multiple times).

src/back/hlsl/writer.rs
```diff
@@ -2688,13 +2722,231 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                                16 => 0xffff,
+                                32 => 0xffffffff,
+                                64 => 0xffffffffffffffff,
+                                _ => {
+                                    return Err(Error::Unimplemented(format!(
+                                        "write_expr_math insert_bits for scalar_width {}",
+                                        scalar_width
+                                    )))
+                                }
+                            };
```
teoxoy, line 2796: I think we can just assume 32bit here, since that's all we support right now (same for `ExtractBits`).
Thanks for considering this though!

teoxoy (changes requested):
Looks great! Left a few minor comments.

teoxoy:
image

I wanted to point out that for `insertBits` and `extractBits` we are not capping the `offset` and `count`.
Considering the MSL and SPIR-V backends also require this it might be better for the WGSL frontend to inject those.

We'll also end up with a lot of duplication if we want to do it in-line as opposed to calling a function we injected previously (this is how we deal with constructors currently).

src/back/hlsl/writer.rs
Elabajaba: Should I leave `scalar_max` as a variable, or should I replace all the places where I use `scalar_max` with 0xFFFFFFFF?

src/back/hlsl/writer.rs
```diff
@@ -2688,13 +2722,231 @@ impl<'a, W: fmt::Write> super::Writer<'a, W> {
...
+                                self.write_expr(module, count, func_ctx)?;
+                                write!(self.out, " == 0 ? 0 : (")?;
+                                self.write_expr(module, arg, func_ctx)?;
+                                write!(self.out, " << ({scalar_width} - ")?;
+                                self.write_expr(module, count, func_ctx)?;
+                                write!(self.out, " - ")?;
+                                self.write_expr(module, offset, func_ctx)?;
+                                write!(self.out, ")) >> ({scalar_width} - ")?;
+                                self.write_expr(module, count, func_ctx)?;
+                                write!(self.out, "))")?;
```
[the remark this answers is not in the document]
Elabajaba: Other implementations use 2 code paths, but I just tested it and it looks like this one works for both?

src/back/hlsl/writer.rs
teoxoy: Let's keep them as variables.

src/back/hlsl/writer.rs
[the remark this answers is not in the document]
teoxoy: Maybe the unsigned version is slightly faster (an `&` vs two `-`)? I'd say we keep just the signed version.

teoxoy (approved):
Thank you for implementing this, it looks great!✨

Issue #2354, opened 2023-05-25
[spv-in] shaders that use `OpImageQuerySize` to read uint values and do comparisons with them don't pass validation


expenses:
After https://github.com/gfx-rs/naga/pull/2245 the following shader doesn't pass validation:
```hlsl
[[vk::binding(0), vk::image_format("rgba16f")]] RWTexture2D<float4> output_tex; 

[numthreads(8, 8, 1)]
void shader(
    uint3 id: SV_DispatchThreadID
) {
    uint width;
    uint height;
    output_tex.GetDimensions(width, height);

    if (id.x < width) {
        return;
    }

    output_tex[id.xy].x = id.z;
}
```
compiled with `dxc shader.hlsl -T cs_6_6 -E shader -spirv -HV 2021`
```
<snip>
         %20 = OpLoad %type_2d_image %output_tex
         %21 = OpImageQuerySize %v2uint %20
         %22 = OpCompositeExtract %uint %21 0
         %23 = OpCompositeExtract %uint %17 0
         %24 = OpULessThan %bool %23 %22
               OpSelectionMerge %25 None
               OpBranchConditional %24 %26 %25
         %26 = OpLabel
               OpBranch %18
         %25 = OpLabel
         %27 = OpCompositeExtract %uint %17 2
         %28 = OpConvertUToF %float %27
         %29 = OpVectorShuffle %v2uint %17 %17 0 1
         %30 = OpLoad %type_2d_image %output_tex
         %31 = OpImageRead %v4float %30 %29 None
         %32 = OpCompositeInsert %v4float %28 %31 0
         %33 = OpLoad %type_2d_image %output_tex
               OpImageWrite %33 %29 %32 None
               OpBranch %18
         %18 = OpLabel
               OpReturn
               OpFunctionEnd
```
results in this error:
```
[2023-05-25T21:51:46Z ERROR naga::valid::expression] Left: AccessIndex { base: [8], index: 0 } of type Scalar { kind: Uint, width: 4 }
[2023-05-25T21:51:46Z ERROR naga::valid::expression] Right: AccessIndex { base: [11], index: 0 } of type Scalar { kind: Sint, width: 4 }
Function [1] 'shader' is invalid: 
        Expression [14] is invalid
        Operation Less can't work with [13] and [12]
```
this is caused by this cast to an Sint:
[...]

PR #2357, opened 2023-05-26
[spv-in] Remove the cast from `parse_image_query_size`

expenses:
Would close https://github.com/gfx-rs/naga/issues/2354. See https://github.com/gfx-rs/naga/issues/2354#issuecomment-1563665811. This fixes the issue with my shaders (see https://github.com/gfx-rs/naga/issues/2354) but it might break others. Testing is probably important.


expenses:
Hmm, looks like the translation to glsl is invalid: https://github.com/gfx-rs/naga/pull/2357/commits/d2368a09b7546b1078df36cb42de00c99ca2724a#diff-932c2e40d7cfd7c549ead1917747c2d1ebf86471f59ca9a90f2a96e154ab063c.

evahop:
So I believe these casts existed because the wgsl equivalents will always be unsized. if spir-v can have both signed and unsigned integers this will still need to be casted for signed ints when targeting wgsl and unfortunately will be a little redundant for other targets unless we could be conditional based off the desired target.

evahop:
Looks like `parse_image_query_other` also contains the same cast.

teoxoy (changes requested):
Thanks for the PR! 

See https://github.com/gfx-rs/naga/issues/2354#issuecomment-1570374587 for the correct fix.

PR #2358, opened 2023-05-27
Bump `bitflags` to v2

daxpedda:
Changelog: https://github.com/bitflags/bitflags/blob/2.0.0/CHANGELOG.md#200.

This is a breaking change. The API that is implemented by the macro is now different.

I was a bit liberal with the derived `std` traits, which mimics the behavior of `bitflags` v1, let me know if we don't need all these traits. Potentially we could at least drop `PartialOrd` and `Ord`.


daxpedda:
CI currently fails on master with the same error as well.

jimblandy:
Why are the serializations changing?

Why do we have all these new `#[derive]` attributes?

daxpedda:
> Why are the serializations changing?

It was changed in v2. Now human-readable formats get a string, the name of the flag, JSON is a human-readable format, that's why it's affecting the tests. In `bincode` for example, it would still just use an integer.

See https://github.com/bitflags/bitflags/blob/2.0.0/CHANGELOG.md#changes-to-serde-serialization.

> Why do we have all these new `#[derive]` attributes?

In v1 the `bitflags!` macro derived them for us. This isn't the case anymore in v2, now derives need to be added manually.

This was done in a PR like this https://github.com/bitflags/bitflags/pull/282 and is unfortunately not highlighted in the changelog.

teoxoy:
@daxpedda I fixed the failing CI on master, could you rebase?

daxpedda:
Done.

daxpedda:
This error doesn't make a whole lot of sense to me, my PR should have no effect on this. But somehow CI on master succeeds, so what gives?

Should I just fix it?

tests/in/force_point_size_vertex_shader_webgl.param.ron
```diff
@@ -4,7 +4,7 @@
             version: 300,
             is_webgl: true
         ),
-        writer_flags: (bits: 16),
+        writer_flags: ("TEXTURE_SHADOW_LOD | INCLUDE_UNUSED_ITEMS | FORCE_POINT_SIZE"),
```
teoxoy, line 7:
```suggestion
        writer_flags: ("FORCE_POINT_SIZE"),
```

I think it's because of this.

daxpedda: Wow, not sure how I came up with this combination.
In any case, this should be `CLAMP_FRAG_DEPTH`, right?
https://github.com/daxpedda/naga/blob/cc985396dadc95039a0f3d5cfc818a6b4d5eba6b/src/back/spv/mod.rs#L604-L620

teoxoy: The `WriterFlags` you linked to is for the SPV backend, the one in question here is for the GLSL backend.

daxpedda: > The `WriterFlags` you linked to is for the SPV backend, the one in question here is for the GLSL backend.

Right ... thank you!

teoxoy:
I'm not sure how much value there is in deriving `PartialOrd, Ord` for these. @jimblandy thoughts?

Cargo.toml
```diff
@@ -45,7 +45,7 @@ harness = false
 
 [dependencies]
 arbitrary = { version = "1.0.2", features = ["derive"], optional = true }
-bitflags = "1.0.5"
+bitflags = "2.0.0"
```
teoxoy, line 48:
```suggestion
bitflags = "2"
```

daxpedda: Done.

ErichDonGubler:
@teoxoy:

> I'm not sure how much value there is in deriving `PartialOrd, Ord` for these. @jimblandy thoughts?

The only value I can think of for these (and `Hash`) is if somebody wanted to use these bitflags structures as an ordered set, or as tree keys. I don't think that that's surface we want to support right now, is it?

jimblandy:
We should derive only the traits we actually need. I believe they contribute to compile time, even if you don't use them, but more importantly, deriving them leads one to wonder what those traits are used for. Seeing that a trait is *not* derived is actually a helpful hint (of course, there could be impls elsewhere, so it's just a hint, but...).

teoxoy:
Some of these types are public and they previously had all of those traits automatically derived by bitflags. I imagine we always want `Clone, Copy, Debug`, I can see people using `PartialEq, Eq`. `PartialOrd, Ord, Hash` I'm not sure if are that useful for bitfields.

teoxoy:
Proposal: remove `PartialOrd, Ord, Hash` and if people want them back, we can add them back later

daxpedda:
Done.

daxpedda:
Some types require `Hash` as they are used in `HashMap`s: https://github.com/gfx-rs/naga/pull/2358/commits/89f1425dbe6c90e9b8bc61aaff8e242fbb00aff1.
`PartialOrd` and `Ord` could be removed by removing the requirements from types containing them.

teoxoy (approved):
LGTM, thanks!

PR #2361, opened 2023-05-30
[0.12] Support array bindings of buffers

kvark:
Backporting #2282 to crates


jimblandy (approved):
Verified as identical to 37b3c36a.

PR #2367, opened 2023-06-02
Complete HLSL reserved symbols

teoxoy:
- adds FXC & DXC intrinsics
- adds FXC & DXC vector and matrix shorthands
- adds DXC (reserved) keywords
- adds DXC resource and other types
- adds DXC scalar types


src/back/hlsl/keywords.rs
```diff
@@ -171,4 +174,730 @@ pub const RESERVED: &[&str] = &[
...
+
+// DXC scalar types, from https://github.com/microsoft/DirectXShaderCompiler/blob/18c9e114f9c314f93e68fbc72ce207d4ed2e65ae/tools/clang/lib/AST/ASTContextHLSL.cpp#L48-L254
+// + vector and matrix shorthands
+pub const TYPES: &[&str] = &{
+    const L: usize = 23 * (1 + 4 + 4 * 4);
+    let mut res = [""; L];
+    let mut c = 0;
+
+    /// For each scalar type, it will additionally generate vector and matrix shorthands
+    macro_rules! generate {
```
jimblandy, line 827: Would it be simpler to do this at run time in Rust code? If so, please consider using the `once_cell` crate and simplifying the code.

(It's okay to take on `once_cell` as a dependency, since it's been added to the standard library as of Rust 1.70, so we'll be able to take it out then.)

src/back/hlsl/keywords.rs
```diff
@@ -171,4 +174,730 @@ pub const RESERVED: &[&str] = &[
...
+        ($a:ident, $i:ident, [$($roots:literal),*], $x:tt) => {
```
jimblandy, line 828: If we do have to keep this macro - there's no need to pass `$a` and `$i` down to all the invocations. The `@inner push` case can simply refer to `res` and `c` directly, since they're visible at the point of the macro's definition.

jimblandy (approved):
Okay, I see - you want to use the macro to generate the names because you'd like to use `'static` as the lifetime in the namer.

Looks good to me - just consider the `$a` / `$i` simplification I suggested.

src/back/hlsl/keywords.rs
teoxoy: Opened https://github.com/gfx-rs/naga/pull/2381

PR #2372, opened 2023-06-06
spv-out: implement OpArrayLength on array buffer bindings

### LINKED ISSUE #2368: [spv-out] Cannot use arrayLength() on a runtime-sized array contained within a binding_array
Minimal repo:
```rust
struct IndexBuffer {
    buffer: array<u32>,
}

@group(0) @binding(0) var<storage> index_buffers: binding_array<IndexBuffer>;

let indices_count = arrayLength(&index_buffers[17u].buffer);
```

Naga returns an error here: https://github.com/gfx-rs/naga/blob/b7f4006e46313da063f8f2f930230767c9740239/src/back/spv/index.rs#L57

### LINKED ISSUE #4556: [spv-out] Cannot use arrayLength() on a runtime-sized array contained within a binding_array
Minimal repo:
```rust
struct IndexBuffer {
    buffer: array<u32>,
}

@group(0) @binding(0) var<storage> index_buffers: binding_array<IndexBuffer>;

let indices_count = arrayLength(&index_buffers[17u].buffer);
```

Naga returns an error here: https://github.com/gfx-rs/naga/blob/b7f4006e46313da063f8f2f930230767c9740239/src/back/spv/index.rs#L57

Naga seems to assume arrayLength() will reference a global directly, and doesn't account for indirect cases where the array comes from a binding array (either a static index into the binding array like 17u above, or dynamically with a variable).

kvark:
Fixes gfx-rs/wgpu#4556

### DIFF
--- a/src/back/spv/index.rs
+++ b/src/back/spv/index.rs
@@ -48,25 +48,39 @@ impl<'w> BlockContext<'w> {
         // inside a buffer that is itself an element in a buffer bindings array.
         // SPIR-V requires that runtime-sized arrays are wrapped in structs.
         // See `helpers::global_needs_wrapper` and its uses.
-        let (opt_array_index, global_handle, opt_last_member_index) = match self
+        let (opt_array_index_id, global_handle, opt_last_member_index) = match self
             .ir_function
             .expressions[array]
         {
-            // Note that SPIR-V forbids `OpArrayLength` on a variable pointer,
-            // so we aren't handling `crate::Expression::Access` here.
             crate::Expression::AccessIndex { base, index } => {
                 match self.ir_function.expressions[base] {
                     // The global variable is an array of buffer bindings of structs,
-                    // and we are accessing the last member.
+                    // we are accessing one of them with a static index,
+                    // and the last member of it.
                     crate::Expression::AccessIndex {
                         base: base_outer,
                         index: index_outer,
                     } => match self.ir_function.expressions[base_outer] {
                         crate::Expression::GlobalVariable(handle) => {
-                            (Some(index_outer), handle, Some(index))
+                            let index_id = self.get_index_constant(index_outer);
+                            (Some(index_id), handle, Some(index))
                         }
                         _ => return Err(Error::Validation("array length expression case-1a")),
                     },
+                    // The global variable is an array of buffer bindings of structs,
+                    // we are accessing one of them with a dynamic index,
+                    // and the last member of it.
+                    crate::Expression::Access {
+                        base: base_outer,
+                        index: index_outer,
+                    } => match self.ir_function.expressions[base_outer] {
+                        crate::Expression::GlobalVariable(handle) => {
+                            let index_id = self.cached[index_outer];
+                            (Some(index_id), handle, Some(index))
+                        }
+                        _ => return Err(Error::Validation("array length expression case-1b")),
+                    },
+                    // The global variable is a buffer, and we are accessing the last member.
                     crate::Expression::GlobalVariable(handle) => {
                         let global = &self.ir_module.global_variables[handle];
                         match self.ir_module.types[global.ty].inner {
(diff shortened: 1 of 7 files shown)


jimblandy:
Rebased to get rid of changes to irrelevant snapshot files

src/back/spv/index.rs
```diff
@@ -47,27 +47,76 @@ impl<'w> BlockContext<'w> {
         // only the latter, so this back end wraps bare runtime-sized arrays
         // in a made-up struct; see `helpers::global_needs_wrapper` and its uses.
         // This code must handle both cases.
-        let (structure_id, last_member_index) = match self.ir_function.expressions[array] {
+        let (opt_array_index, global_handle, opt_last_member_index) = match self
```
jimblandy, line 50: The comment above this needs to be updated.

src/back/spv/mod.rs
```diff
@@ -495,6 +495,11 @@ struct GlobalVariable {
     /// to refer to the global in the function body. This is the id of that access,
     /// updated for each function in `write_function`.
     access_id: Word,
+
+    /// The ID of a type that is the base for the array of bindings.
+    /// This is needed because the block encoder isn't expected to generate new types,
+    /// and there are cases where we need to get a pointer to one of those bases.
+    pointer_to_binding_array_element_type_id: Word,
```
jimblandy, line 502: Why can't the block encoder just call `get_type_id` and pass a `LookupType::Local(LocalType::PointerToBindingArray { .. })`? These `GlobalVariable` handles are really confusing, so I'd like to avoid accumulating more stuff here if we can help it.

jimblandy (changes requested):
I think the back/spv changes can be simplified; see the earlier comments.

src/back/spv/mod.rs
jimblandy: I see now that `PointerToBindingArray` isn't what's needed here, it's a pointer to an element of the binding array. But it still seems like there should always be a `Type` in the arena for that, since it must have been declared as a global var of type `BindingArray`, which refers to its `base` via a handle.

jimblandy: I believe you want the type of the inner `AccessIndex` expression's `base_outer` expression.

kvark: I tried to express this in the comment - "This is needed because the block encoder isn't expected to generate new types". I.e. architecturally we have the block encoder assuming that all types that would be needed are already there.
I don't want to change that architectural restriction, I think there is elegance to it.
So the code in the PR does this type creation ahead of time, and it needs a place to cache those results.
This is in line with existing code, I don't think we should require this PR to re-engineer this aspect.

kvark:
@jimblandy please take a look at the comments, PR is rebased

src/back/spv/index.rs
```diff
@@ -42,32 +42,81 @@ impl<'w> BlockContext<'w> {
...
+        // final member of a struct that is a global variable, or one of these
+        // inside a buffer that is itself an element in a buffer bindings array.
+        // SPIR-V requires that runtime-sized arrays are wrapped in structs.
+        // See `helpers::global_needs_wrapper` and its uses.
+        let (opt_array_index, global_handle, opt_last_member_index) = match self
+            .ir_function
+            .expressions[array]
+        {
+            // Note that SPIR-V forbids `OpArrayLength` on a variable pointer,
+            // so we aren't handling `crate::Expression::Access` here.
```
jimblandy, line 55: This is true, but irrelevant. `OpAccessChain` applied to a `OpVariable` doesn't produce a variable pointer. There should be no reason we can't support `Access` here. The following validates just fine:

```
                   OpCapability Shader
                   OpCapability Linkage
                   OpExtension "SPV_KHR_storage_buffer_storage_class"
                   OpMemoryModel Logical GLSL450

                   OpDecorate %struct Block
                   OpMemberDecorate %struct 0 Offset 0
                   OpMemberDecorate %struct 1 Offset 4

            %u32 = OpTypeInt 32 0
        %ptr_u32 = OpTypePointer StorageBuffer %u32
  %runtime_array = OpTypeRuntimeArray %u32
         %struct = OpTypeStruct %u32 %runtime_array
  %binding_array = OpTypeRuntimeArray %struct
         %ptr_ba = OpTypePointer StorageBuffer %binding_array

            %one = OpConstant %u32 1
                   
         %global = OpVariable %ptr_ba StorageBuffer

      %func_type = OpTypeFunction %u32 %u32 %u32
           %func = OpFunction %u32 None %func_type
         %index1 = OpFunctionParameter %u32
         %index2 = OpFunctionParameter %u32
       %prologue = OpLabel
            %ptr = OpAccessChain %ptr_u32 %global %index1 %one %index2
          %value = OpLoad %u32 %ptr
                   OpReturnValue %value
                   OpFunctionEnd
```

tests/in/binding-buffer-arrays.wgsl
```diff
@@ -23,5 +23,7 @@ fn main(fragment_in: FragmentIn) -> @location(0) u32 {
     u1 += storage_array[uniform_index].x;
     u1 += storage_array[non_uniform_index].x;
 
+    u1 += arrayLength(&storage_array[0].far);
```
jimblandy, line 26: In other words, we ought to be able to write:

```suggestion
    u1 += arrayLength(&storage_array[uniform_index].far);
```

jimblandy (changes requested):

src/back/spv/index.rs
kvark: Great point!
Impl is adjusted accordingly now.

kvark:
@jimblandy ready for round 3
Amazing review, as always, and thanks for the insights!

JMS55:
Thanks so much for the work on this! Any chance it can be rebased to naga main?

cwfitzgerald:

kvark:
Ported as https://github.com/gfx-rs/wgpu/pull/5428

PR #2375, opened 2023-06-07
Expose all ray-query flags in constants

kvark:


src/front/wgsl/parse/mod.rs
```diff
@@ -638,14 +638,62 @@ impl Parser {
...
+                ast::Expression::Literal(ast::Literal::Number(Number::U32(1)))
+            }
+            (Token::Word("RAY_QUERY_INTERSECTION_GENERATED"), _) => {
+                let _ = lexer.next();
+                ast::Expression::Literal(ast::Literal::Number(Number::U32(2)))
+            }
+            (Token::Word("RAY_QUERY_INTERSECTION_AABB"), _) => {
+                let _ = lexer.next();
+                ast::Expression::Literal(ast::Literal::Number(Number::U32(4)))
+            }
```
jimblandy, line 696: Which Vulkan flags do these correspond to? I wasn't able to find anything that seems to match this.

jimblandy (changes requested):
Since they're going to appear in the IR, these constants need to be a Rust bitflags in `lib.rs`. Please document which Vulkan flag set each one corresponds to.

src/front/wgsl/parse/mod.rs
kvark: Defined here - https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_ray_query.asciidoc

kvark:
Please take another look!

cwfitzgerald:

kvark:
Ported as https://github.com/gfx-rs/wgpu/pull/5429

Issue #2366, opened 2023-06-02
Add backend intrinsics/built-in functions to their respective reserved keywords list


teoxoy:
User defined symbols might collide with backend intrinsics, we need to rename them.

- [ ] HLSL
- [ ] MSL
- [ ] WGSL
- [ ] GLSL https://github.com/gfx-rs/naga/issues/2276

PR #2378, opened 2023-06-12
Omit non referenced expressions

teoxoy:


teoxoy:
Seems HLSL validation is failing on https://github.com/gfx-rs/naga/issues/2366 which should be fixed by https://github.com/gfx-rs/naga/pull/2367.

src/valid/analyzer.rs
```diff
@@ -675,12 +675,17 @@ impl FunctionInfo {
                 requirements: UniformityRequirements::empty(),
             },
             E::Math {
-                arg, arg1, arg2, ..
+                fun: _,
+                arg,
+                arg1,
+                arg2,
+                arg3,
```
jimblandy, line 682: Nice catch. I think I've seen several of these fixed recently. I did a quick look and I think this must the last - the other places that use `Math { .. }` are sites that can't produce results at all without the arguments, so `..` is probably okay there.

src/valid/analyzer.rs
```diff
@@ -868,7 +873,7 @@ impl FunctionInfo {
                 S::Loop {
                     ref body,
                     ref continuing,
-                    break_if: _,
+                    break_if,
```
jimblandy, line 876: Another nice catch.

jimblandy (approved):
Looks good!

PR #2379, opened 2023-06-13
add debug info for spv-out

wicast:
make the spv with debug info mapping to wgsl

pic

here is the capture file you can try
[wgsl_debug_info.zip](https://github.com/gfx-rs/naga/files/11730385/wgsl_debug_info.zip)

### DIFF
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -61,7 +61,7 @@ pub enum BlockExit {
 }
 
 #[derive(Debug)]
-pub struct DebugInfo<'a> {
+pub(crate) struct DebugInfoInner<'a> {
     pub source_code: &'a str,
     pub source_file_id: Word,
 }
@@ -1708,13 +1708,11 @@ impl<'w> BlockContext<'w> {
         naga_block: &crate::Block,
         exit: BlockExit,
         loop_context: LoopContext,
-        debug_info: Option<&DebugInfo>,
+        debug_info: Option<&DebugInfoInner>,
     ) -> Result<(), Error> {
         let mut block = Block::new(label_id);
-        let statements = naga_block.iter();
-        let mut spans = naga_block.span_iter();
-        for statement in statements {
-            if let (Some(debug_info), Some((_, span))) = (debug_info, spans.next()) {
+        for (statement, span) in naga_block.span_iter() {
+            if let Some(debug_info) = debug_info {
                 let loc: crate::SourceLocation = span.location(debug_info.source_code);
                 block.body.push(Instruction::line(
                     debug_info.source_file_id,
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -7,7 +7,7 @@ use super::{
     Dimension, Error, Instruction, LocalType, LookupType, LoopContext, ResultMember, Writer,
     WriterFlags,
 };
-use crate::{arena::Handle, proc::TypeResolution};
+use crate::{arena::Handle, proc::TypeResolution, Statement};
 use spirv::Word;
 
 fn get_dimension(type_inner: &crate::TypeInner) -> Dimension {
@@ -1712,7 +1712,16 @@ impl<'w> BlockContext<'w> {
     ) -> Result<(), Error> {
         let mut block = Block::new(label_id);
         for (statement, span) in naga_block.span_iter() {
-            if let Some(debug_info) = debug_info {
+            if let (Some(debug_info), false) = (
+                debug_info,
+                matches!(
+                    statement,
+                    &(Statement::Block(..)
+                        | Statement::If { .. }
+                        | Statement::Switch { .. }
+                        | Statement::Loop { .. })
+                ),
+            ) {
                 let loc: crate::SourceLocation = span.location(debug_info.source_code);
                 block.body.push(Instruction::line(
                     debug_info.source_file_id,
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -1717,8 +1717,10 @@ impl<'w> BlockContext<'w> {
                 matches!(
                     statement,
                     &(Statement::Block(..)
-                        | Statement::If { .. }
-                        | Statement::Switch { .. }
+                        | Statement::Break
+                        | Statement::Continue
+                        | Statement::Kill
+                        | Statement::Return { .. }
                         | Statement::Loop { .. })
                 ),
             ) {
@@ -1900,6 +1902,16 @@ impl<'w> BlockContext<'w> {
                     // SPIR-V requires the continuing to the `OpLoopMerge`,
                     // so we have to start a new block with it.
                     block = Block::new(preamble_id);
+                    // HACK the loop statement is begin with branch instruction,
+                    // so we need to put `OpLine` debug info before merge instruction
+                    if let Some(debug_info) = debug_info {
+                        let loc: crate::SourceLocation = span.location(debug_info.source_code);
+                        block.body.push(Instruction::line(
+                            debug_info.source_file_id,
+                            loc.line_number,
+                            loc.line_position,
+                        ))
+                    }
                     block.body.push(Instruction::loop_merge(
                         merge_id,
                         continuing_id,
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -60,6 +60,12 @@ pub enum BlockExit {
     },
 }
 
+#[derive(Debug)]
+pub struct DebugInfo<'a> {
+    pub source_code: &'a str,
+    pub source_file_id: Word,
+}
+
 impl Writer {
     // Flip Y coordinate to adjust for coordinate space difference
     // between SPIR-V and our IR.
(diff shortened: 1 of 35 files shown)


src/back/spv/writer.rs
```diff
@@ -1772,8 +1786,28 @@ impl Writer {
...
+
+            let debug_source_code = (|| {
+                Some((
+                    self.source_file_id?,
+                    ir_module.source_code.as_ref()?.as_str(),
+                ))
+            })();
+            self.debugs.push(Instruction::source(
+                spirv::SourceLanguage::GLSL,
+                450,
```
teoxoy, line 1808:
```suggestion
                spirv::SourceLanguage::Unknown,
                0,
```
Maybe we should use `Unknown` here instead.

src/back/spv/writer.rs
```diff
@@ -1772,8 +1786,28 @@ impl Writer {
             .to_words(&mut self.logical_layout.ext_inst_imports);
 
         if self.flags.contains(WriterFlags::DEBUG) {
-            self.debugs
-                .push(Instruction::source(spirv::SourceLanguage::GLSL, 450));
+            if let Some(file_name) = (|| {
+                Path::new(ir_module.file_path.as_ref()?)
+                    .file_name()?
+                    .to_str()
+            })() {
```
teoxoy, line 1793: Please avoid those immediately called closures - I don't see why they are needed.

src/back/spv/writer.rs
```diff
@@ -688,11 +691,22 @@ impl Writer {
...
 
+        let debug_info = (|| {
+            if !context.writer.flags.contains(WriterFlags::DEBUG) {
+                return None;
+            }
+            Some(DebugInfo {
+                source_code: ir_module.source_code.as_ref()?,
+                source_file_id: context.writer.source_file_id?,
+            })
+        })();
```
teoxoy, line 702: We can use [`then_some`](https://doc.rust-lang.org/std/primitive.bool.html#method.then_some) here.

src/back/spv/block.rs
```diff
@@ -1690,13 +1696,23 @@ impl<'w> BlockContext<'w> {
...
+        naga_block: &crate::Block,
         exit: BlockExit,
         loop_context: LoopContext,
+        debug_info: Option<&DebugInfo>,
     ) -> Result<(), Error> {
         let mut block = Block::new(label_id);
-
+        let statements = naga_block.iter();
+        let mut spans = naga_block.span_iter();
         for statement in statements {
```
teoxoy, line 1707:
```suggestion
        for (statement, span) in naga_block.span_iter() {
```

`span_iter` returns both

teoxoy (changes requested):
Looks great! I left a few minor comments.

Could you also add a test for this? So that we have this as part of our snapshots.

src/back/spv/writer.rs
wicast: This closure is a workround for the try block. I know this is a bit ugly, but I want to avoid nested `if let Some` checks for Option. I don't have any better idea yet.

src/lib.rs
```diff
@@ -1941,4 +1941,9 @@ pub struct Module {
     pub functions: Arena<Function>,
     /// Entry points.
     pub entry_points: Vec<EntryPoint>,
+
+    /// Source Code for debug
+    pub source_code: Option<String>,
+    /// File name for debug
+    pub file_path: Option<String>,
```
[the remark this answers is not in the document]
wicast: Although this pr is for spv, I think this can be useful if we do the debug info for other backends.
For backends that generate source code, we can track the original file and line number, then insert this information as comments.
Metal also has a debug symbol file format that might work like SPV, but the format has no document at all (I’m not sure if it’s a DWARF or Apple’s proprietary format).

src/back/spv/writer.rs
teoxoy: I think we can use `and_then` and `map`.

src/lib.rs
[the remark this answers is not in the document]
teoxoy: The idea was that every backend that supports this would take this info via their write functions, we wouldn't have to clone the source either (having the source have a lifetime in the module doesn't seem worth it).

> Metal also has a debug symbol file format that might work like SPV, but the format has no document at all (I’m not sure if it’s a DWARF or Apple’s proprietary format).

Are you referring to [metallibsym](https://developer.apple.com/documentation/metal/shader_libraries/generating_and_loading_a_metal_library_symbol_file?language=objc)?

We have some info about [metallib](https://github.com/gfx-rs/naga/wiki/Metallib-file-format) ([docs](https://developer.apple.com/documentation/metal/shader_libraries/compiling_shader_code_into_a_library_with_metal_s_command-line_tools?changes=latest_minor%20rel%3D&language=objc)) but not about the symbols file.

wicast:
I've add some test and re-write the spv writer.
It has some troubles with OpName tracking [see here](https://github.com/gfx-rs/naga/pull/2379/files#diff-37323fc67fde7849dec2889efe0ef3c8a716e9110747ebe0a08fa80c66ffdbdf), some of the local variables are missing,
anyway it's a usable version for debugging shaders.

tests/snapshots.rs
teoxoy, line 1: Do we need different code paths for "with debug" and without? You can set debug to true in the `.ron` file. Is there anything else we need to get it working?

Also, the new `spv_debug` folder is not covered by validation done in CI.

teoxoy (changes requested):

src/back/spv/block.rs
```diff
@@ -1699,13 +1705,30 @@ impl<'w> BlockContext<'w> {
...
-        for statement in statements {
+        for (statement, span) in naga_block.span_iter() {
+            if let (Some(debug_info), false) = (
+                debug_info,
+                matches!(
+                    statement,
+                    &(Statement::Block(..)
+                        | Statement::If { .. }
+                        | Statement::Switch { .. }
+                        | Statement::Loop { .. })
```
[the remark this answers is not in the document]
wicast: Oh...sorry for missing those exceptions.

> Maybe popping the last Line instruction if the instruction about to be pushed is a termination instruction? But the changes required to get that working are quite large...

After read the spec again, I think the only problem is `Loop` statement.
`If` and `Switch` is good to put `OpLine` right ahead because they start by `OpSelectionMerge`
`Loop` is start with a new `OpBranch` (which shouldn't put `OpLine` ahead) and then the`OpLoopMerge` (which we need to put `OpLine`).

tests/snapshots.rs
wicast: Sure, a bit complex I have done. Let me clean it up.

src/back/spv/writer.rs
```diff
@@ -1771,9 +1777,24 @@ impl Writer {
...
+                    source_code: debug_info.source_code,
+                    source_file_id,
+                })
+            }
+
+            self.debugs.push(Instruction::source(
+                spirv::SourceLanguage::Unknown,
+                450,
+                &debug_info_inner,
+            ));
```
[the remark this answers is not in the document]
wicast: Did you mean we don't need this for non-debug compile?

src/back/spv/block.rs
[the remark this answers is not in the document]
teoxoy: Right, we only need to omit `OpLine` before termination instructions.

src/back/spv/writer.rs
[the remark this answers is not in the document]
teoxoy: We can move it in the 2nd `if` where we check that we have `Some(debug_info)`, since I don't think it's useful to write `OpSource Unknown 0`.

src/back/spv/block.rs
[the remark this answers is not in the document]
wicast: > Right, we only need to omit `OpLine` before termination instructions.

sorry, I'am still a bit confused. Isn't that OpLine shouldn't precede termination instructions by the spec?

My current idea is keep exceptions like `Break`,`Continue` , plus `Loop`, and take the `Loop` as special case as inserting `OpLine` before `OpLoopMerge` emits.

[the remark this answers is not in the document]
teoxoy: That sounds right. I used **omit** in my previous message, I guess it can be easily mistaken for **emit** 😅.

teoxoy (approved):
LGTM, thanks!

src/back/spv/mod.rs
```diff
@@ -726,8 +736,17 @@ pub fn write_vec(
     options: &Options,
     pipeline_options: Option<&PipelineOptions>,
 ) -> Result<Vec<u32>, Error> {
-    let mut words = Vec::new();
+    let mut words: Vec<u32> = Vec::new();
     let mut w = Writer::new(options)?;
-    w.write(module, info, pipeline_options, &mut words)?;
+
+    if options.flags.contains(WriterFlags::DEBUG) {}
```
Gordon-F, line 742: @wicast This empty `if` looks like a bug

wicast: > @wicast This empty `if` looks like a bug

I've made another PR to fix it, alone with another debug option fix .

PR #2387, opened 2023-06-21
[v0.10] Backport #2290 ([spv-in] Convert conditional backedges to `break if`).

eddyb:
Forgot about this version initially, we still have a Rust-GPU user stuck on it, can this be backported?


jimblandy (approved):
Looks good to me.

PR #2389, opened 2023-06-21
[spv-out] Use `IndexSet` instead of `HashSet` for iterated sets (capabilities/extensions).

eddyb:
Naga currently doesn't pass its own CI because its tests hardcode a specific `hashbrown`+`fx` iteration order.

And then upstream upgraded `hashbrown` in `nightly-2023-06-16` which changed the order:
* https://github.com/rust-lang/rust/pull/104455

What's worse is I'm hitting this on a `v0.10` backport, where I would prefer not to backport this PR too:
* https://github.com/gfx-rs/naga/pull/2387

### DIFF
--- a/src/back/spv/mod.rs
+++ b/src/back/spv/mod.rs
@@ -595,10 +595,10 @@ pub struct Writer {
     ///
     /// If `capabilities_available` is `Some`, then this is always a subset of
     /// that.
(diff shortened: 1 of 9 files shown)


jimblandy (approved):
Looks good to me.

PR #2390, opened 2023-06-21
[v0.10] Backport #2389 ([spv-out] Use `IndexSet` instead of `HashSet` for iterated sets (capabilities/extensions)).

eddyb:
Nothing can land on CI on any branch without this fix (see #2389), and I want to land a backport (#2387).


jimblandy (approved):
Looks good to me, but see #2391.

PR #2393, opened 2023-06-26
Expose early depth test field

Joeoc2001:
For my use case, I need to be able to access the `conservative` field of the `EarlyDepthTest` struct. Every other top-level public struct has every field accessible, so this just brings `EarlyDepthTest` in line with the other components of a module.


teoxoy (approved):
LGTM

PR #2397, opened 2023-07-04
[glsl-out] textureSize level must be a signed integer.

nical:
This fixes shader compilation issues when writing wgsl code like `textureDimensions(texture, 0u)` with the GLSL backend.
[GLSL textureSize spec](https://registry.khronos.org/OpenGL-Refpages/gl4/html/textureSize.xhtml)


teoxoy (approved):
LGTM

PR #2398, opened 2023-07-07
Fix installing naga-cli from git

### LINKED ISSUE #2385: cargo install --git https://github.com/gfx-rs/naga naga-cli fails
Trying to execute
```
cargo install --git https://github.com/gfx-rs/naga naga-cli
```
fails with 
```
Updating git repository `https://github.com/gfx-rs/naga`
error: package `/home/dennis/.cargo/git/checkouts/naga-dbb2b19faed49210/f31093f/xtask/hlsl-snapshots/Cargo.toml` is a member of the wrong workspace
expected: /home/dennis/.cargo/git/checkouts/naga-dbb2b19faed49210/f31093f/Cargo.toml
actual:   /home/dennis/.cargo/git/checkouts/naga-dbb2b19faed49210/f31093f/xtask/Cargo.toml
```

Environment:
```
rustc 1.70.0 (90c541806 2023-05-31)

cargo 1.70.0 (ec8a8a0ca 2023-04-25)
release: 1.70.0
commit-hash: ec8a8a0cabb0e0cadef58902470f6c7ee7868bdc
commit-date: 2023-04-25
host: x86_64-unknown-linux-gnu
libgit2: 1.6.3 (sys:0.17.0 vendored)
libcurl: 8.0.1-DEV (sys:0.4.61+curl-8.0.1 vendored ssl:OpenSSL/1.1.1t)
os: Arch Linux Rolling Release [64-bit]
```

fornwall:
Move hl-snapshots out from the xtask workspace, since it's also used as a dev dependency in the naga crate itself for a test.

Fixes #2385, installing naga-cli from git.

To test installing from my fork before this PR is merged, run `cargo install naga-cli --git https://github.com/fornwall/naga.git --branch fix-hlsl-snapshots-workspace`.


ErichDonGubler (approved):
These changes LGTM, minus wanting to make sure everybody (CC @teoxoy, @jimblandy) felt comfortable adding another top-level directory to the repo. I think this may start to push the repo into having its own dedicated `crates` directory, but that can be a separate follow-up.

xtask/Cargo.toml
```diff
@@ -8,12 +8,11 @@ publish = false
...
+hlsl-snapshots = { path = "../hlsl-snapshots"}
 log = "0.4.17"
 pico-args = "0.5.0"
 shell-words = "1.1.0"
 which = "4.4.0"
 
 [workspace]
-
-members = [".", "./hlsl-snapshots"]
+members = ["."]
```
teoxoy, line 18: I think this can now be removed.

fornwall: Indeed - removed this in 6a31ddf26472ec7f689e95349b631c06a1156399 now, thanks!

fornwall:
> These changes LGTM, minus wanting to make sure everybody (CC @teoxoy, @jimblandy) felt comfortable adding another top-level directory to the repo. I think this may start to push the repo into having its own dedicated `crates` directory, but that can be a separate follow-up.

Thanks! Having something like

```
crates/cli
crates/fuzz
crates/hlsl-snapshots
crates/naga
crates/xtask
```

might make things more clear - I can propose that in a follow-up PR if you want?

xtask/Cargo.toml
teoxoy: Thanks! `[workspace]` could go as well, as it's now empty.

teoxoy:
I think a flatter directory structure is better considering we don't have that many other directories in root (besides the crates there are just 2 dot directories and a `benches`).

xtask/Cargo.toml
fornwall: > Thanks! [workspace] could go as well, as it's now empty.

@teoxoy Without having `[workspace]` in `xtask/Cargo.toml` we get this error:

```
error: current package believes it's in a workspace when it's not:
current:   [..]/naga/xtask/Cargo.toml
workspace: [..]/naga/Cargo.toml

this may be fixable by adding `xtask` to the `workspace.members` array of the manifest located at: /Users/fornwall/src/naga/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

So if we want to remove the workspace entry in `xtask` we need to add it to the root `Cargo.toml` workspace (or exclude it explicitly, with `workspace.exclude`). Let me know if you want that!

teoxoy: Ah, right. It's ok as it is since we need a separate lock file for it.

teoxoy (approved):

PR #2400, opened 2023-07-07
[glsl-in] Support commas in structure definitions

### LINKED ISSUE #2384: [glsl-in] Doesn't support commas in structure definitions
I have a structure like this:

```glsl
struct LightScatteringParams {
    float BetaRay, BetaMie, HGg, DistanceMul, BlendCoeff;
    vec3 SunDirection, SunColor;
};
```

It is relatively easy to reformat it to support naga, but naga should support this.

fornwall:
Fixes #2384.


teoxoy:
@JCapucho would you like to take a look at this?

src/front/glsl/parser/declarations.rs
```diff
@@ -628,35 +628,55 @@ impl<'source> ParsingContext<'source> {
...
-            let (name, end_meta) = self.expect_ident(frontend)?;
+            let (ty, mut meta) = self.parse_type_non_void(frontend)?;
 
-            meta.subsume(end_meta);
+            let (name, mut end_meta) = self.expect_ident(frontend)?;
+            let mut ty_copy = ty;
+            self.parse_array_specifier(frontend, &mut meta, &mut ty_copy)?;
 
-            self.parse_array_specifier(frontend, &mut meta, &mut ty)?;
+            let mut names_ty = vec![(name, ty_copy)];
```
JCapucho, line 637: I think we can do away with this allocation if we immediately add to the members vector. My idea would be to do something like:

```rust
let (base_ty, mut meta) = self.parse_type_non_void(frontend)?;

loop {
	let (name, name_meta) = self.expect_ident(frontend)?;
    let mut ty = base_ty;
    self.parse_array_specifier(frontend, &mut meta, &mut ty_copy)?;

	meta.subsume(name_meta);

	let info = offset::calculate_offset(
        ty,
        meta,
        layout,
        &mut frontend.module.types,
        &mut frontend.errors,
    );

    let member_alignment = info.align;
    span = member_alignment.round_up(span);
    align = member_alignment.max(align);

    members.push(StructMember {
        name: Some(name),
        ty: info.ty,
        binding: None,
        offset: span,
    });

    span += info.span;

	if colon { continue; }

	self.expect(frontend, TokenValue::Semicolon)?;

	break;
}
```

This way we don't need to allocate the `names_ty` and don't need to duplicate code for the first name.

JCapucho (changes requested):

JCapucho:
(I meant to write a comment in the request for changes, but pressed submit accidentally.)

I think the overall idea is right but we can improve the code to be more efficient. Also the test should be moved to `declarations.frag`.

fornwall:
> Also the test should be moved to declarations.frag.

Thanks, moved tests in https://github.com/gfx-rs/naga/pull/2400/commits/340e09cbf2e98f8b0c37e1e6bc4ba94ccb1bfee4.

src/front/glsl/parser/declarations.rs
fornwall: Thanks @JCapucho for great review feedback! Changed the code in https://github.com/gfx-rs/naga/pull/2400/commits/85896caecfff94237e78602ba3358da05698f28a, could you check?

JCapucho (approved):
Everything looks great now. Thank you.

PR #2404, opened 2023-07-10
[spv] Fix OpImageQueries to allow Uints

evahop:
This pull request conditionally casts Sints in OpImageQuery functions where previously it was assumed to always be Sint and simplifies the backend by removing unneeded casts.

fixes https://github.com/gfx-rs/naga/issues/2354

### DIFF
--- a/src/back/spv/image.rs
+++ b/src/back/spv/image.rs
@@ -1101,15 +1101,7 @@ impl<'w> BlockContext<'w> {
                 let query_id = self.gen_id();
                 block.body.push(Instruction::image_query(
                     spirv::Op::ImageQueryLevels,
-                    self.get_type_id(
-                        LocalType::Value {
-                            vector_size: None,
-                            kind: crate::ScalarKind::Uint,
-                            width: 4,
-                            pointer_space: None,
-                        }
-                        .into(),
-                    ),
+                    result_type_id,
                     query_id,
                     image_id,
                 ));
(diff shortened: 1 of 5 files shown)


src/back/spv/image.rs
```diff
@@ -1121,7 +1104,7 @@ impl<'w> BlockContext<'w> {
                     self.get_type_id(
                         LocalType::Value {
                             vector_size: None,
-                            kind: crate::ScalarKind::Sint,
+                            kind: crate::ScalarKind::Uint,
                             width: 4,
                             pointer_space: None,
                         }
```
teoxoy, line 1110: We can just use `result_type_id` as we previously did.

https://github.com/gfx-rs/naga/blob/24316fd4bc9fca3b0c9cdd858ea593114c208eba/src/back/spv/image.rs#L1099-L1104

src/back/spv/image.rs
```diff
@@ -1168,7 +1143,7 @@ impl<'w> BlockContext<'w> {
...
```
teoxoy, line 1149: Same here.

src/back/spv/image.rs
```diff
@@ -1196,7 +1163,7 @@ impl<'w> BlockContext<'w> {
...
```
teoxoy, line 1169: And here.

teoxoy (commented):
Thanks for the PR; behavior looks good!

Only left some nits that simplify the code.

src/back/spv/image.rs
evahop: thanks, didn't notice this

teoxoy (approved):
LGTM

Issue #2160, opened 2022-12-08
Validate access on `ImageClass::Storage` and `AddressSpace::Storage`


teoxoy:
WGSL only has write-only storage textures and never write-only storage buffers.

see https://gpuweb.github.io/gpuweb/wgsl/#texture-storage and https://gpuweb.github.io/gpuweb/wgsl/#storage-buffer

PR #2415, opened 2023-07-27
Validate storage buffer access

teoxoy:
Storage buffers only support `read` or `read_write` access modes (not `write`).

Related: https://github.com/gfx-rs/naga/issues/2160, https://github.com/gfx-rs/wgpu/pull/3984

### DIFF
--- a/src/valid/interface.rs
+++ b/src/valid/interface.rs
@@ -33,6 +33,8 @@ pub enum GlobalVariableError {
     ),
     #[error("Initializer doesn't match the variable type")]
     InitializerType,
+    #[error("Storage address space doesn't support write-only access")]
+    StorageAddressSpaceWriteOnlyNotSupported,
 }
 
 #[derive(Clone, Debug, thiserror::Error)]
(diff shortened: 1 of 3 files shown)


ErichDonGubler (approved):
A gray checkmark for you, sir!

teoxoy:
@JCapucho how do you feel about this change in the context of GLSL? Do people qualify their buffers as `writeonly`? I'd imagine it's quite niche.

JCapucho:
> @JCapucho how do you feel about this change in the context of GLSL? Do people qualify their buffers as `writeonly`? I'd imagine it's quite niche.

I did see one shader at one point that used it, but yes, it seems like a quite niche thing.

Also I did some little research and it seems that `writeonly` is more of an hint https://github.com/gpuweb/gpuweb/issues/935, I also took a glance at the vulkan and OpenGL specifications and it doesn't seem to be part of any validation for buffers.

I think this can be removed for now, and then emulated later in the glsl frontend.

PR #2417, opened 2023-07-30
derive PartialEq on Expression

robtfm:
in updating `naga_oil` i found i needed to have the `const_expressions` arena contain only unique expressions. this is needed because the equality test for `Const`s and `GlobalVariable`s does not inspect the `init` field, it just compares handles, so otherwise equal items with equivalent inits that are distinct handles do not test as equal. I rely on the `Arena::fetch_or_append` to deduplicate consts and globals which uses the derived equality.

adds the `expression-eq` feature to enable derive of `PartialEq` for `Expression`.

i note there are quite a lot of other `PartialEq` impls that are `cfg(test)`. I don't need any of them but perhaps the feature sohuld be more general and include them as well:
```rust
front::glsl::error::ErrorKind
front::glsl::error::Error

front::glsl::lex::LexerResultKind
front::glsl::lex::LexerResult

front::glsl::token::DirectiveKind
front::glsl::token::Directive
front::glsl::token::Token

valid::UniformityDisruptor
valid::compose::ComposeError
valid::expression::ExpressionError

valid::function::CallError
valid::function::AtomicError
valid::function::LocalVariableError
valid::function::FunctionError
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,6 @@ wgsl-out = []
 hlsl-out = []
 span = ["codespan-reporting", "termcolor"]
 validate = []
-expression-eq = []
 
 [[bench]]
 name = "criterion"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ wgsl-out = []
 hlsl-out = []
 span = ["codespan-reporting", "termcolor"]
 validate = []
+expression-eq = []
 
 [[bench]]
 name = "criterion"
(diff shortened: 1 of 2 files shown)


teoxoy:
Thanks for the PR!

I don't think we need to derive `PartialEq` for the other types in the glsl frontend and validation module.

Are the compilation times so drastic that we should gate this derive behind a feature flag?

teoxoy:
Let's always derive it.

PR #2420, opened 2023-08-01
Bump minimum supported Rust version (MSRV) to 1.65.

jimblandy:


jimblandy:
`wgpu`'s minimum supported Rust version is 1.65.

Firefox's is 1.66.

The CLI depends on `clap_builder` 4.3.19, whose MSRV is 1.64.

ErichDonGubler (approved):

teoxoy (commented):
> The CLI depends on `clap_builder` 4.3.19, whose MSRV is 1.64.

Is this the case? `clap_builder` is only used by `clap` which in turn is only used by `criterion`; which is a dev dependency. We also have MSRV checks in CI to make sure we don't depend on crates with a higher MSRV than the one we advertise.

PR #2421, opened 2023-08-02
[spv-out] Fix non-uniformity attributes

### LINKED ISSUE #2408: Artifacts when using binding_array with textures. #2282 Related
I originally created an issue on https://github.com/bevyengine/bevy/issues/9221 as that was where I first noticed the behaviour, but curiosity got the better of me, so while bisecting through naga and running the [texture-arrays](https://github.com/gfx-rs/wgpu/blob/trunk/examples/texture-arrays) example each step, I found that this https://github.com/gfx-rs/wgpu/commit/dcad7dfb commit resulted in the artifacts appearing. 

Okay, so it's in naga; more bisecting through naga and this commit https://github.com/gfx-rs/naga/commit/37b3c36a8f1aaa611e0dcfc4892333e83ac3a919 introduces the behaviour.

## Relevant system information:
OS: `NixOS 23.11.20230719.7f256d7 (Tapir)`
Display Server: `Wayland 1.22`
```ignore
AdapterInfo { name: "AMD Radeon RX 6800 XT (RADV NAVI21)", vendor: 4098, device: 29631, device_type: DiscreteGpu, driver: "radv", driver_info: "Mesa 23.1.3", backend: Vulkan } 
```

From what I can visibly observe, it seems like textures from other indices in the array are bleeding over into other textures. The artifact colors are *always* from another texture in the array.

I've attached a screenshot:
image
*Artifacts in the image, more prominent when resizing the window*

If there's any additional information I can provide to help track down the cause, i'm more than happy to pass it on!

cwfitzgerald:
Fixes #2408

This isn't the cleanest change, but I couldn't really figure out a better way to structure it.

### DIFF
--- a/tests/out/spv/binding-buffer-arrays.spvasm
+++ b/tests/out/spv/binding-buffer-arrays.spvasm
@@ -24,7 +24,7 @@ OpMemberDecorate %16 0 Offset 0
 OpDecorate %23 Location 0
 OpDecorate %23 Flat
 OpDecorate %26 Location 0
-OpDecorate %55 NonUniform
+OpDecorate %58 NonUniform
 %2 = OpTypeVoid
 %3 = OpTypeInt 32 0
 %4 = OpTypeStruct %3
--- a/src/back/spv/block.rs
+++ b/src/back/spv/block.rs
@@ -25,7 +25,7 @@ fn get_dimension(type_inner: &crate::TypeInner) -> Dimension {
 enum ExpressionPointer {
     /// The pointer to the expression's value is available, as the value of the
     /// expression with the given id.
-    Ready { pointer_id: Word },
+    Ready { pointer_id: Word, non_uniform: bool },
 
     /// The access expression must be conditional on the value of `condition`, a boolean
     /// expression that is true if all indices are in bounds. If `condition` is true, then
@@ -35,6 +35,7 @@ enum ExpressionPointer {
     Conditional {
         condition: Word,
         access: Instruction,
+        non_uniform: bool,
     },
 }
 
@@ -285,12 +286,15 @@ impl<'w> BlockContext<'w> {
                             class: helpers::map_storage_class(space),
                         });
 
-                        let result_id = match self.write_expression_pointer(
+                        let (result_id, non_uniform) = match self.write_expression_pointer(
                             expr_handle,
                             block,
                             Some(binding_array_false_pointer),
                         )? {
-                            ExpressionPointer::Ready { pointer_id } => pointer_id,
+                            ExpressionPointer::Ready {
+                                pointer_id,
+                                non_uniform,
+                            } => (pointer_id, non_uniform),
                             ExpressionPointer::Conditional { .. } => {
                                 return Err(Error::FeatureNotImplemented(
                                     "Texture array out-of-bounds handling",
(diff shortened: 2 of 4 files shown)


teoxoy:
There seems to be a lot of uncertainty around where the `NonUniform` decoration is required (https://github.com/KhronosGroup/SPIRV-Registry/issues/89, https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/3288#issuecomment-1484096764).

The SPIR-V Spec on [`NonUniform`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_decoration) doesn't say much.

> Apply only to an object. Asserts that the value backing the decorated <id> is not dynamically uniform. See the client API specification for more detail.

> Object: An instantiation of a non-void type, either as the Result <id> of an operation, or created through OpVariable.

The SPIR-V Spec on [Uniformity](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_uniformity) says it should be present on image operands of image operations otherwise they are assumed to be uniform.

> An <id> is defined to be [dynamically uniform](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#DynamicallyUniform) for a dynamic instance of an instruction if all invocations (in an invocation group) that execute the dynamic instance have the same value for that <id>. This is not something that is explicitly decorated, it is just a property that arises. This property is assumed to hold for operands of certain instructions, such as the Image operand of image instructions, unless that operand is decorated as NonUniform. Some implementations require more complex instruction expansions to handle non-dynamically uniform values in certain instructions, and thus it is mandatory for certain operands to be decorated as NonUniform if they are not guaranteed to be dynamically uniform.

The [VUID-RuntimeSpirv-NonUniform-06274](https://github.com/KhronosGroup/Vulkan-Docs/blob/42b2d689d26722ec4f61f557006056bbd2e48aed/appendices/spirvenv.adoc?plain=1#L840-L845) is more insightful, although I had to read it a few times to understand what (I think) it's trying to say. That is:
- image operands (for image operations) need the decoration; which matches what the SPIR-V spec was saying in the uniformity section; in our case the result of the `OpLoad`
- pointer operands (for load, store and atomic operations) need the decoration; in our case the result of `OpAccessChain`

> If an instruction loads from or stores to a resource (including atomics
> and image instructions) and the resource descriptor being accessed is
> not dynamically uniform, then the operand corresponding to that resource
> (e.g. the pointer or sampled image operand) must: be decorated with
> `NonUniform`

I've also looked at [glslang's source code](https://github.com/KhronosGroup/glslang/blob/bfd96d3044e1e2b0522a71166203ed4ddcf8d6e5/SPIRV/SpvBuilder.cpp#L3589-L3741) and it matches my understanding.

tests/out/spv/binding-buffer-arrays.spvasm
```diff
@@ -24,7 +24,7 @@ OpMemberDecorate %16 0 Offset 0
 OpDecorate %23 Location 0
 OpDecorate %23 Flat
 OpDecorate %26 Location 0
-OpDecorate %55 NonUniform
+OpDecorate %58 NonUniform
```
teoxoy, line 27: I think for buffers we need to keep the decoration on the access chain.

cwfitzgerald:
Well that's not needlessly complicated...

Thanks for doing that research!

teoxoy:
It was quite the rabbit hole 😅

PR #2422, opened 2023-08-04
[spv-out] Decorate binding array image/sample load result as `NonUniform` 

teoxoy:
See https://github.com/gfx-rs/naga/pull/2421#issuecomment-1665678646

Fixes https://github.com/gfx-rs/naga/issues/2408
Closes https://github.com/gfx-rs/naga/pull/2421

Commits are self-contained and don't require to be squashed.


cwfitzgerald (approved):
Very clean stuff!

PR #2426, opened 2023-08-08
Bump `indexmap` to v2

daxpedda:
Bumps `indexmap` to v2.
See the [changelog](https://github.com/bluss/indexmap/blob/master/RELEASES.md), I don't believe there are any relevant changes for `naga`.

Notably this updates `hashbrown` from v0.12 to v0.14 and `ahash` from v0.7 to v0.8.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,8 @@ dot-out = []
 glsl-in = ["pp-rs"]
 glsl-out = []
 msl-out = []
-serialize = ["serde", "bitflags/serde", "indexmap/serde-1"]
-deserialize = ["serde", "bitflags/serde", "indexmap/serde-1"]
+serialize = ["serde", "bitflags/serde", "indexmap/serde"]
+deserialize = ["serde", "bitflags/serde", "indexmap/serde"]
 arbitrary = ["dep:arbitrary", "bitflags/arbitrary", "indexmap/arbitrary"]
 spv-in = ["petgraph", "spirv"]
 spv-out = ["spirv"]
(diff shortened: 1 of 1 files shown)


daxpedda:
So the minimal versions check fails because `indexmap` v2.0.0 -> `arbitrary` v1.0.0 -> `derive_arbitrary` v1.0.0 depends on `syn` v1.0.0, which it doesn't actually support. It requires at least `syn` v1.0.6.

`derive_arbitrary` v1.3.0 doesn't have this issue, as it relies on `syn` v1.0.56.
`arbitrary` v1.3.0 would require at least `derive_arbitrary` v1.3.0, which if `indexmap` depends on could fix this issue.

So either we:
1. Make a PR to `indexmap` telling them to bump to `arbitrary` v1.3.
2. Or we could just add a line in our CI bumping `syn` to v1.0.6 after generating the minimal lockfile.

Happy to apply either solution.

ErichDonGubler:
I don't believe there are any blockers on the Firefox side for this
upgrade; Mike Hommey had a plan to force-downgrade `indexmap` 2 deps. to
v1, since they're source-compatible. I don't know if that plan was
executed, but I'm not concerned. CC @jimb.

teoxoy:
Looks like it https://searchfox.org/mozilla-central/source/build/rust/indexmap

It's missing feature flags but those shouldn't be too hard to add.

teoxoy:
@daxpedda 

> 1. Make a PR to `indexmap` telling them to bump to `arbitrary` v1.3.

I think this would be preferred.

fornwall:
> Make a PR to indexmap telling them to bump to arbitrary v1.3.

Created https://github.com/bluss/indexmap/pull/271

daxpedda:
As noted in bluss/indexmap#271, this was already broken on our side.
Fixed in #2429.

teoxoy:
If you rebase, I think we should be able to merge this one now.

PR #2429, opened 2023-08-10
Properly test `-Zminimal-versions`

daxpedda:
I noticed that using `--offline` here doesn't properly go all the way down to the actual minimal version of each dependency. Bit too lazy to dig into why exactly, and I'm not sure why `--offline` was used in the first place, but we should see CI fail in a moment.


teoxoy (approved):
LGTM, thanks!

PR #2435, opened 2023-08-13
[wgsl-in] Handle repeated or missing @workgroup_size

### LINKED ISSUE #2271: Bad error message for compute entry point without `@workgroup_size` attribute
Compiling the following shader module:

```
@compute fn f() {}
```

elicits the following error message from Naga:

```
    Entry point f at Compute is invalid
    Workgroup size is out of range
```

The workgroup size isn't out of range, it's missing altogether.

### LINKED ISSUE #2434: [wgsl-in] Fail if the @workgroup_size attribute is repeated
Follow up work on https://github.com/gfx-rs/naga/pull/2428:

Validate that the `@workgroup_size` attribute is not repeated, just as #2428 validates other attributes.

This will probably fix https://github.com/gfx-rs/naga/issues/2271, `Bad error message for compute entry point without @workgroup_size attribute`, as a side effect.

fornwall:
Fixes #2271 and also fixes #2434.

### DIFF
--- a/src/front/wgsl/error.rs
+++ b/src/front/wgsl/error.rs
@@ -242,6 +242,7 @@ pub enum Error<'a> {
     Other,
     ExpectedArraySize(Span),
     NonPositiveArrayLength(Span),
+    MissingWorkgroupSize(Span),
 }
 
 impl<'a> Error<'a> {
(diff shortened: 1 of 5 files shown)


fornwall:
Change for missing @workgroup_size (#2271):

With the below shader:
```wgsl
@compute
fn main() {}
```

Before:
Screenshot 2023-08-14 at 01 22 55

After:
Screenshot 2023-08-14 at 01 24 24

fornwall:
Change for repeated @workgroup_size (https://github.com/gfx-rs/naga/issues/2434):

For the below shader:
```wgsl
@compute @workgroup_size(1) @workgroup_size(1)
fn main() {}
```

Before:
Screenshot 2023-08-14 at 01 28 19

After:
Screenshot 2023-08-14 at 01 27 40

teoxoy (approved):
There is a bigger question if we should make it optional at the IR level but this looks good for now.

Thanks for the recent contributions! :)

PR #2440, opened 2023-08-16
[wgsl-in] Avoid splatting all binary operator expressions

### LINKED ISSUE #2439: [wgsl-in] Comparing scalar with vector type incorrectly passes validation
Spec says both sides of the comparision need to have the same type.
https://www.w3.org/TR/WGSL/#comparison-expr

The following snippet should therefore not pass validation.
```
const some_vec = vec3<f32>(1.0, 1.0, 1.0);

@fragment
fn main() -> @location(0) vec4<f32> {
    if (all(1.0 < some_vec)) {
        return vec4(0.0);
    }
    return vec4(1.0);
}
```

Note that Chrome will already correctly reject this, whereas Naga doesn't

fornwall:
Fixes #2439.

[Reference](https://www.w3.org/TR/WGSL/#vector-multi-component:~:text=Binary%20arithmetic%20expressions%20with%20mixed%20scalar%20and%20vector%20operands):
Screenshot 2023-08-16 at 12 08 10

### DIFF
--- a/src/front/wgsl/lower/mod.rs
+++ b/src/front/wgsl/lower/mod.rs
@@ -510,7 +510,13 @@ impl<'source, 'temp, 'out> ExpressionContext<'source, 'temp, 'out> {
         left: &mut Handle<crate::Expression>,
         right: &mut Handle<crate::Expression>,
     ) -> Result<(), Error<'source>> {
-        if op != crate::BinaryOperator::Multiply {
+        if matches!(
+            op,
+            crate::BinaryOperator::Add
+                | crate::BinaryOperator::Subtract
+                | crate::BinaryOperator::Divide
+                | crate::BinaryOperator::Modulo
+        ) {
             self.grow_types(*left)?.grow_types(*right)?;
 
             let left_size = match *self.resolved_inner(*left) {
--- a/src/front/wgsl/lower/mod.rs
+++ b/src/front/wgsl/lower/mod.rs
@@ -504,6 +504,12 @@ impl<'source, 'temp, 'out> ExpressionContext<'source, 'temp, 'out> {
     }
 
     /// Insert splats, if needed by the non-'*' operations.
+    ///
+    /// See the "Binary arithmetic expressions with mixed scalar and vector operands"
+    /// table in the WebGPU Shading Language specification for relevant operators.
+    ///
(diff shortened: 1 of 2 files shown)


src/front/wgsl/lower/mod.rs
```diff
@@ -510,7 +510,13 @@ impl<'source, 'temp, 'out> ExpressionContext<'source, 'temp, 'out> {
         left: &mut Handle<crate::Expression>,
         right: &mut Handle<crate::Expression>,
     ) -> Result<(), Error<'source>> {
-        if op != crate::BinaryOperator::Multiply {
+        if matches!(
+            op,
+            crate::BinaryOperator::Add
+                | crate::BinaryOperator::Subtract
+                | crate::BinaryOperator::Divide
+                | crate::BinaryOperator::Modulo
```
teoxoy, line 518: Looking at the table in the spec, `Multiply` should also work (not sure why it was previously the only operation ignored).

Note that the comment on the function also needs to be updated.

teoxoy (changes requested):

src/front/wgsl/lower/mod.rs
fornwall: @teoxoy `Multiply` is excluded here as (AFAIK) the backends are expected to handle vec*scalar operations.

We could insert splats here and it would still work, but would increase IR and output size needlessly, as in e.g.:

```diff
-    gl_Position = (((vec4(1.0) * _e4) * _e8) * _e12);
+    gl_Position = (((vec4(1.0) * vec4(_e4)) * vec4(_e9)) * vec4(_e14));
```

The [typifier](https://github.com/gfx-rs/naga/blob/master/src/proc/typifier.rs#L564-L612) handles vec*scalar (as well as matrix cases) for type checking.

I've added d848fdd6fce1b3062ca22122504ec949aec4cb92 to expand the function comment with this information, and also adding a reference to the specification.

Let me know what you think!

teoxoy: Ah, I see - thanks for the clarification!

teoxoy (approved):
Great stuff!

Wumpf:
awesome, thanks for picking this up!

PR #2443, opened 2023-08-18
[glsl-out] Use 'as f64' when writing out f32 literals

### LINKED ISSUE #2436: [glsl-out] Incorrect rounding for max-float constant expression
In v0.12 (prior to  #2266), this wgsl constant
```const f32max = 0x1.fffffep+127;```
would inline in glsl as `3.4028234663852886e38`.

However, in v0.13 this results into a glsl constant
```const float f32max = 3.4028235e38;```
When testing a provided max-float value (read from a texture in my case) for `>=f32max` this will now yield `false` in a WebGL setup.
Interestingly, this scenario works fine on Metal where this yields `constant float f32max = 340282350000000000000000000000000000000.0;`. I suspect that glsl applies different rounding here.

More backends and situations might be affected by this, but I was able to verify it only so far for glsl-out and rule it out for msl-out. It would be worth exploring the effect on other backends as well and find out if this has any other accuracy problems.

### LINKED ISSUE #4568: [glsl-out] Incorrect rounding for max-float constant expression
In v0.12 (prior to  gfx-rs/naga#2266), this wgsl constant
```const f32max = 0x1.fffffep+127;```
would inline in glsl as `3.4028234663852886e38`.

However, in v0.13 this results into a glsl constant
```const float f32max = 3.4028235e38;```
When testing a provided max-float value (read from a texture in my case) for `>=f32max` this will now yield `false` in a WebGL setup.
Interestingly, this scenario works fine on Metal where this yields `constant float f32max = 340282350000000000000000000000000000000.0;`. I suspect that glsl applies different rounding here.

More backends and situations might be affected by this, but I was able to verify it only so far for glsl-out and rule it out for msl-out. It would be worth exploring the effect on other backends as well and find out if this has any other accuracy problems.

fornwall:
Fixes gfx-rs/wgpu#4568.

See the linked issue for more information and discussion. While this restores to pre-#2266 behaviour, we're not sure that this is the best way forward.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -2215,7 +2215,9 @@ impl<'a, W: Write> Writer<'a, W> {
                     // Floats are written using `Debug` instead of `Display` because it always appends the
                     // decimal part even it's zero which is needed for a valid glsl float constant
                     crate::Literal::F64(value) => write!(self.out, "{:?}LF", value)?,
-                    crate::Literal::F32(value) => write!(self.out, "{:?}", value)?,
+                    // Use 'as f64' to work around precision issues in glsl shaders - see
+                    // https://github.com/gfx-rs/naga/issues/2436
+                    crate::Literal::F32(value) => write!(self.out, "{:?}", value as f64)?,
                     // Unsigned integers need a `u` at the end
                     //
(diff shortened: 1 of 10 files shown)


cwfitzgerald:

PR #2444, opened 2023-08-18
[wgsl-in] Only splat 'scalar op vec', not '_ op vec'

fornwall:
Change `binary_op_splat()` from splatting:

        vec op scalar
        _ op vec

to only splat:

        vec op scalar
        scalar op vec

This fixes some confusing error messages. Example function:

```wgsl
fn main(mat: mat3x3<f32>) {
    let vec = vec3<f32>(1.0, 1.0, 1.0);
    let result = mat / vec;
}
```

### DIFF
--- a/src/front/wgsl/lower/mod.rs
+++ b/src/front/wgsl/lower/mod.rs
@@ -525,13 +525,8 @@ impl<'source, 'temp, 'out> ExpressionContext<'source, 'temp, 'out> {
         ) {
             self.grow_types(*left)?.grow_types(*right)?;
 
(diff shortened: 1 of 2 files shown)


fornwall:
Error message before this change (note the `Scalar type Matrix` message):

Screenshot 2023-08-18 at 15 06 45

fornwall:
Error message after this change (in general formatting of validation errors needs improvement, but the underlying information is now correct):

Screenshot 2023-08-18 at 15 06 22

teoxoy (approved):
Makes sense, thanks!

PR #2445, opened 2023-08-19
[wgsl-in] Handle all(bool) and any(bool)

### LINKED ISSUE #1911: [wgsl-in] Error on provision of boolean scalars to all/any built-in functions
The specification provides two overloads for the `all` and `any` built-in functions, with one accepting boolean vectors and the second accepting boolean scalars. naga appears to only accept the former.

code example (analogous for `any`):
```
@compute @workgroup_size(1)
fn compute_main() {
	var var0 = all(true);
}
```

naga compilation error:
```
[2022-05-11T18:19:19Z ERROR naga::valid::expression] All/Any of type Scalar { kind: Bool, width: 1 }
error: 
  ┌─ all_any_scalar.wgsl:3:12
  │
3 │     var var0 = all(true);
  │               ^^^^^^^^^^ naga::Expression [2]

Entry point compute_main at Compute is invalid: 
	Expression [2] is invalid
	Relational argument [1] is not a boolean vector
```

fornwall:
In wgsl there are overloads for `all(bool)` and `any(bool)` which just returns the boolean unmodified.

As at least glsl [any](https://registry.khronos.org/OpenGL-Refpages/gl4/html/any.xhtml) and [all](https://registry.khronos.org/OpenGL-Refpages/gl4/html/all.xhtml) does not support that, let's remove them while lowering to get rid of them early.

Fixes #1911.

### DIFF
--- a/src/front/wgsl/lower/mod.rs
+++ b/src/front/wgsl/lower/mod.rs
@@ -1696,7 +1696,26 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
                     let argument = self.expression(args.next()?, ctx.reborrow())?;
                     args.finish()?;
 
-                    crate::Expression::Relational { fun, argument }
+                    // Check for no-op all(bool) and any(bool):
+                    let argument_unmodified = matches!(
+                        fun,
+                        crate::RelationalFunction::All | crate::RelationalFunction::Any
+                    ) && {
+                        ctx.grow_types(argument)?;
+                        matches!(
+                            ctx.resolved_inner(argument),
+                            &crate::TypeInner::Scalar {
+                                kind: crate::ScalarKind::Bool,
+                                ..
+                            }
+                        )
+                    };
+
+                    if argument_unmodified {
+                        return Ok(Some(argument));
+                    } else {
+                        crate::Expression::Relational { fun, argument }
+                    }
                 } else if let Some((axis, ctrl)) = conv::map_derivative(function.name) {
                     let mut args = ctx.prepare_args(arguments, 1, span);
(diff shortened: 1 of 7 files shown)


fornwall:
The build failure is #2447.

src/front/wgsl/lower/mod.rs
```diff
@@ -1696,7 +1696,20 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
...
                     args.finish()?;
 
-                    crate::Expression::Relational { fun, argument }
+                    // Check for no-op all(bool) and any(bool):
+                    let argument_unmodified = matches!(
+                        fun,
+                        crate::RelationalFunction::All | crate::RelationalFunction::Any
+                    ) && {
+                        ctx.grow_types(argument)?;
+                        matches!(ctx.resolved_inner(argument), &crate::TypeInner::Scalar { kind, .. } if kind == crate::ScalarKind::Bool)
```
teoxoy, line 1705:
```suggestion
                        matches!(ctx.resolved_inner(argument), &crate::TypeInner::Scalar { kind: crate::ScalarKind::Bool, .. })
```

teoxoy (approved):
Besides the suggestion, LGTM; I think it makes sense for this no-op to be handled by the frontend.

src/front/wgsl/lower/mod.rs
fornwall: Thanks!

fornwall:
> Besides the suggestion, LGTM; I think it makes sense for this no-op to be handled by the frontend.

Thanks - applied the suggestion and squashed the commits.

PR #2449, opened 2023-08-19
Fix validation and glsl parsing of ldexp

### LINKED ISSUE #1908: [wgsl-in] Error on usage of ldexp function
naga rejects (valid) usage of the `ldexp` built-in function. This happens with both the (scalar float, scalar int) and (vector float, vector int) overloads.

code example (analogous for vector arguments):
```
@compute @workgroup_size(1)
fn compute_main() {
	var var0 = ldexp(3.0, 3);
}
```

naga compilation error (identical message for vector arguments):
```
error: 
  ┌─ test.wgsl:3:12
  │
3 │     var var0 = ldexp(3.0, 3);
  │               ^^^^^^^^^^^^^^ naga::Expression [3]

Entry point compute_main at Compute is invalid: 
	Expression [3] is invalid
	Argument [1] to Ldexp as expression [2] has an invalid type.
```

fornwall:
Fixes #1908.


fornwall:
The build failure is #2447.

teoxoy (approved):
Good stuff!

PR #2450, opened 2023-08-21
Fix wgsl out struct layout

### LINKED ISSUE #4373: [wgsl-out] Align and size attributes not being output
Tested by passing a WGSL shader trough naga.

I think we'll need at least struct member alignment in the IR for this to work.

mikialex:
fix https://github.com/gfx-rs/wgpu/issues/4373 by explicitly adding a size mark on all struct fields.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -110,10 +110,10 @@ impl<W: Write> Writer<W> {
         for (handle, ty) in module.types.iter() {
             if let TypeInner::Struct {
                 ref members,
-                span: _,
+                span,
             } = ty.inner
             {
-                self.write_struct(module, handle, members)?;
+                self.write_struct(module, handle, members, span)?;
                 writeln!(self.out)?;
             }
         }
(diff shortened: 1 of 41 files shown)


mikialex:
a test failed due to the two struct field members having the same offset. I think this should be checked in validation and fixed in another pr.

src/back/wgsl/writer.rs
```diff
@@ -377,17 +373,28 @@ impl<W: Write> Writer<W> {
...
                 self.write_attributes(&map_binding_to_attribute(binding))?;
             }
+
+            let current_field_size = if index != members.len() - 1 {
+                let next_member = &members[index + 1];
+                next_member.offset - member.offset
+            } else {
+                struct_span - member.offset
+            };
+            write!(self.out, "@size({current_field_size}) ")?;
```
teoxoy, line 396: We can avoid writing the size explicitly if it's the same as the one inferred. See example below on how to get the size:

https://github.com/gfx-rs/naga/blob/0491d39232bafc30962f6ea240ab6ed457282a62/src/back/msl/writer.rs#L3178-L3179

teoxoy (changes requested):

cwfitzgerald:

PR #1441, opened 2021-09-30
Change frexp and modf to return structures

kvark:
Closes #1161
TODO:
- [x] IR
- [x] validation
- [x] typifier
- [ ] frontends (except WGSL)
- [ ] backends (except WGSL)

PR #2452, opened 2023-08-22
[wgsl-in] Handle modf and frexp

### LINKED ISSUE #594: Figure out modf story
See https://github.com/gpuweb/gpuweb/issues/1480
Currently, our backends are not producing good code for this.

### LINKED ISSUE #1161: Change frexp and modf to work with structures

fornwall:
- Fixes #1161
- Fixes #594 (this is outdated/a duplicate, and can be closed regardless of this PR)
- Replaces #1441 

Another approach than what is done here would be to use `TypeInner::Struct` to reuse struct handling logic. There are some changes necessary there to handle naming correctly, and e.g. avoid writing out the struct in wgsl-out (which already has the struct defined).

Let med know if you want that done instead (and if so, how to proceed there - perhaps adding something like a `enum StructType { UserDefined, ModfResult, FrexpResult }` (and perhaps `AtomicCompareExchangeWeaResult`?) to `TypeInner::Struct`)!

### DIFF
--- a/src/back/glsl/keywords.rs
+++ b/src/back/glsl/keywords.rs
@@ -477,4 +477,9 @@ pub const RESERVED_KEYWORDS: &[&str] = &[
     // entry point name (should not be shadowed)
     //
     "main",
+    // Naga utilities:
+    super::MODF_FUNCTION,
+    super::MODF_STRUCT,
+    super::FREXP_FUNCTION,
+    super::FREXP_STRUCT,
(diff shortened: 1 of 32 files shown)


fornwall:
Will replace this with a variant using `TypeInner::Struct`.

PR #2455, opened 2023-08-25
[spv-out] Always give structs with runtime arrays a Block decoration

### LINKED ISSUE #2322: Vulkan validation error: "OpTypeStruct containing an OpTypeRuntimeArray must be decorated with Block or BufferBlock"
I ran into an issue in https://github.com/webonnx/wonnx/issues/161 that causes a large amount of validation errors (due to the large number of shaders compiled by that project). This was surfaced when I updated the validation layers (I'm now at 1.3.243.0-1 of `vulkan-validation-layers` on Arch Linux), and I think the root cause is naga emitting invalid SPIR-V.

The validation errors all look like this:

```
[2023-04-18T00:42:44Z ERROR wgpu_hal::vulkan::instance] VALIDATION [UNASSIGNED-CoreValidation-Shader-InconsistentSpirv (0x6bbb14)]
        Validation Error: [ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object 0: handle = 0x7f1894544910, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x6bbb14 | SPIR-V module not valid: [VUID-StandaloneSpirv-OpTypeRuntimeArray-04680] Vulkan, OpTypeStruct containing an OpTypeRuntimeArray must be decorated with Block or BufferBlock.
      %ArrayVector = OpTypeStruct %_runtimearr_Vec4
    
[2023-04-18T00:42:44Z ERROR wgpu_hal::vulkan::instance]         objects: (type: DEVICE, hndl: 0x7f1894544910, name: ?)
```

And I think the reason we're seeing them is that every shader starts with a preamble like this:

```wgsl
type Scalar = f32;
type Vec3 = vec3<f32>;
type Vec4 = vec4<f32>;

struct Array {
	data: array<Scalar>
};

struct ArrayVector {
	data: array<Vec4>
};

type Mat3x3 = mat3x3<f32>;
type Mat4x4 = mat4x4<f32>;
type Mat4x3 = mat4x3<f32>;

struct ArrayMatrix {
	data: array<Mat4x4>
};

struct ArrayMatrix3 {
	data: array<Mat3x3>
};
```

TheoDulka:
Fixes #2322, and so will fix https://github.com/bevyengine/bevy/issues/8733 and fix https://github.com/webonnx/wonnx/issues/161

I'm sure the code can be polished and expanded on to take `BufferBlock`s into account, but I just based it off of https://vulkan.lunarg.com/doc/view/1.3.246.1/windows/1.3-extensions/vkspec.html#interfaces-resources-descset, table 21. Where uniform buffers, storage buffers, and inline uniform blocks all use the `Block` decoration.

### DIFF
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1016,11 +1016,26 @@ impl Writer {
                     ref members,
                     span: _,
                 } => {
+                    let mut has_runtime_array = false;
                     let mut member_ids = Vec::with_capacity(members.len());
                     for (index, member) in members.iter().enumerate() {
+                        let member_ty = &arena[member.ty];
+                        match member_ty.inner {
+                            crate::TypeInner::Array {
+                                base: _,
+                                size: crate::ArraySize::Dynamic,
+                                stride: _,
+                            } => {
+                                has_runtime_array = true;
+                            }
+                            _ => (),
+                        }
                         self.decorate_struct_member(id, index, member, arena)?;
                         let member_id = self.get_type_id(LookupType::Handle(member.ty));
                         member_ids.push(member_id);
+                    }
+                    if has_runtime_array {
+                        self.decorate(id, Decoration::Block, &[]);
                     }
                     Instruction::type_struct(id, member_ids.as_slice())
                 }
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1670,13 +1670,13 @@ impl Writer {
             // a runtime-sized array. In this case, we need to decorate it with
             // Block.
             if let crate::AddressSpace::Storage { .. } = global_variable.space {
-                let decorated_id = match ir_module.types[global_variable.ty].inner {
+                match ir_module.types[global_variable.ty].inner {
                     crate::TypeInner::BindingArray { base, .. } => {
-                        self.get_type_id(LookupType::Handle(base))
+                        let decorated_id = self.get_type_id(LookupType::Handle(base));
+                        self.decorate(decorated_id, Decoration::Block, &[]);
                     }
-                    _ => inner_type_id,
+                    _ => (),
                 };
-                self.decorate(decorated_id, Decoration::Block, &[]);
             }
             if substitute_inner_type_lookup.is_some() {
                 inner_type_id
--- a/src/back/spv/writer.rs
+++ b/src/back/spv/writer.rs
@@ -1667,8 +1667,9 @@ impl Writer {
         } else {
             // This is a global variable in the Storage address space. The only
             // way it could have `global_needs_wrapper() == false` is if it has
-            // a runtime-sized array. In this case, we need to decorate it with
-            // Block.
+            // a runtime-sized or binding array.
+            // Runtime-sized arrays were decorated when iterating through struct content.
+            // Now binding arrays require Block decorating.
             if let crate::AddressSpace::Storage { .. } = global_variable.space {
                 match ir_module.types[global_variable.ty].inner {
                     crate::TypeInner::BindingArray { base, .. } => {
--- a/tests/in/runtime-array-in-unused-struct.wgsl
+++ b/tests/in/runtime-array-in-unused-struct.wgsl
@@ -1,0 +1,12 @@
+struct DataStruct {
+    data: f32,
+    data_vec: vec4<f32>,
+}
+
+struct Struct {
+    data: array<DataStruct>,
+};
+
+struct PrimitiveStruct {
+    data: array<f32>,
+};
--- a/tests/out/spv/access.spvasm
+++ b/tests/out/spv/access.spvasm
@@ -73,6 +73,7 @@ OpMemberDecorate %20 2 Offset 96
 OpMemberDecorate %20 3 Offset 100
 OpMemberDecorate %20 4 Offset 144
 OpMemberDecorate %20 5 Offset 160
+OpDecorate %20 Block
 OpMemberDecorate %22 0 Offset 0
 OpMemberDecorate %22 0 ColMajor
 OpMemberDecorate %22 0 MatrixStride 8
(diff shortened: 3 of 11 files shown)


teoxoy (changes requested):
Looks good but I think we can now remove:

https://github.com/gfx-rs/naga/blob/0491d39232bafc30962f6ea240ab6ed457282a62/src/back/spv/writer.rs#L1657-L1669

Although:

image

Do the validation layers not require that the struct is used by a variable in the storage address space despite it seemingly being required by the VUID?

TheoDulka:
Removing the entire block decorating chunk would cause a problem, because this
https://github.com/gfx-rs/naga/blob/0491d39232bafc30962f6ea240ab6ed457282a62/tests/out/spv/binding-buffer-arrays.spvasm#L19 gets removed, leading to
```
error: line 35: Target of NonWritable decoration is invalid: must point to a storage image, uniform block, or storage buffer
  %8 = OpVariable %_ptr_StorageBuffer__arr__struct_6_uint_10 StorageBuffer
```
as an error.

Instead, I removed the extra decorating of the non-`BindingArray`s, which seems to work.

---

For your other comment, from what I can tell (minus `binding-buffer-arrays.wgsl`) the `Block` decorated structs are also in a `StorageBuffer`. An example in boids:

[`boids.spvasm`](https://github.com/gfx-rs/naga/blob/master/tests/out/spv/boids.spvasm#L23-L77)
```
OpName %9 "Particles"
...
OpDecorate %9 Block
...
%17 = OpTypePointer StorageBuffer %9
```
from
https://github.com/gfx-rs/naga/blob/0491d39232bafc30962f6ea240ab6ed457282a62/tests/in/boids.wgsl#L18-L20

teoxoy:
> Removing the entire block decorating chunk would cause a problem, because this
> 
> https://github.com/gfx-rs/naga/blob/0491d39232bafc30962f6ea240ab6ed457282a62/tests/out/spv/binding-buffer-arrays.spvasm#L19
> 
> gets removed, leading to
> ```
> error: line 35: Target of NonWritable decoration is invalid: must point to a storage image, uniform block, or storage buffer
>   %8 = OpVariable %_ptr_StorageBuffer__arr__struct_6_uint_10 StorageBuffer
> ```
> 
> as an error.
> 
> Instead, I removed the extra decorating of the non-`BindingArray`s, which seems to work.

I see, that makes sense; the comment was out of date. Can you update it or remove it?

> For your other comment, from what I can tell (minus `binding-buffer-arrays.wgsl`) the `Block` decorated structs are also in a `StorageBuffer`. An example in boids:
> 
> [`boids.spvasm`](https://github.com/gfx-rs/naga/blob/master/tests/out/spv/boids.spvasm#L23-L77)
> 
> ```
> OpName %9 "Particles"
> ...
> OpDecorate %9 Block
> ...
> %17 = OpTypePointer StorageBuffer %9
> ```
> 
> from
> 
> https://github.com/gfx-rs/naga/blob/0491d39232bafc30962f6ea240ab6ed457282a62/tests/in/boids.wgsl#L18-L20

Right but what I meant is if the struct doesn't get used at all; is it fine to always decorate them with `Block` even if unused?

In your example in https://github.com/gfx-rs/naga/issues/2322, the VUID gets triggered because we are missing the `Block` decoration but it might still get triggered even with the `Block` decoration since it's unused by a global variable in the `StorageBuffer` address space; at least that's how I'm interpreting the VUID.

teoxoy:
Wait, I don't understand why the VUID gets triggered, the validation code for this VUID is inside `ValidateVariable`

https://github.com/KhronosGroup/SPIRV-Tools/blob/9b923f7cc3dde6e1a4886b577677e52c3093ffcc/source/val/validate_memory.cpp#L758-L762

It only checks global variables. And we already decorate structs that have a runtime array if they are used by a global variable.

Can you share a WGSL shader that when ran through naga's SPV backend fails validation with the VUID? Or even better, add it as a snapshot test to the PR.

teoxoy:
My main question is why does the current code in master trigger the validation error at all?

The current code already writes the `Block` decoration on a struct containing the runtime array if used by a global variable.

Maybe I'm missing something and wanted to reproduce it locally.

TheoDulka:
Ahh I misunderstood, I see what you mean now. Yeah it does seem to only trigger when it's not used by a global variable. The bevy issue is from this bevy file [mesh_view_types.wgsl](https://github.com/bevyengine/bevy/blob/02b520b4e85fa95f1f4a85656a677e3fc2155f36/crates/bevy_pbr/src/render/mesh_view_types.wgsl#L91-L93), where the file doesn't use `PointLights` in a global variable, as it's used for importing the types. I assume it's the same case with the wonnx issue, `ArrayVector` might not be used. So I guess the validation checks if the struct has a `Block` decoration regardless of use in variables? The reason is beyond me.

I didn't realize my code just ended up moving the `Block` decorations, as the runtime arrays are already accounted for, but only when they're used in a global variable. When they aren't, that's when the validation error shows up. At this point my code handles runtime arrays in structs regardless of if the struct is used.

---

I added [runtime-array-in-unused-struct.wgsl](https://github.com/gfx-rs/naga/blob/adbb8063617c15b41dd585def84852c132920717/tests/in/runtime-array-in-unused-struct.wgsl) to the snapshots and changed the comment of the storage binding array decorating.

TheoDulka:
@teoxoy I found out why it doesn't matter if it's in a variable, because the error from the issues is spit out when checking the types, in `ValidateTypeStruct`

https://github.com/KhronosGroup/SPIRV-Tools/blob/9b923f7cc3dde6e1a4886b577677e52c3093ffcc/source/val/validate_type.cpp#L353-L360

The link you sent was for a different validation error when checking variables

> Wait, I don't understand why the VUID gets triggered, the validation code for this VUID is inside `ValidateVariable`
> 
> https://github.com/KhronosGroup/SPIRV-Tools/blob/9b923f7cc3dde6e1a4886b577677e52c3093ffcc/source/val/validate_memory.cpp#L758-L762

teoxoy:
Ah, I see! Thank you for tracking it down!

I think we can go ahead with this PR and omit unused structs in the future if it becomes an issue that we are decorating them despite them being unused.

Related:
https://github.com/KhronosGroup/SPIRV-Tools/pull/5094
https://github.com/KhronosGroup/glslang/issues/2439
https://github.com/KhronosGroup/glslang/pull/3130

teoxoy (approved):

Issue #2446, opened 2023-08-19
Support for the .glsl extension


eliemichel:
Using `shader.glsl` as input leads to `Unknown input file extension`. I believe using `.glsl` as an extension for GLSL files is not an exotic practice and should be supported. If the type of shader is needed (frag/vert/comp) this could be added as an command line argument.

teoxoy:
Sounds like a good addition, would you be interested in contributing it?

eliemichel:
Sure, here is a proposal: #2462

PR #2462, opened 2023-09-02
[cli] Add support for `.{vert,frag,comp}.glsl` files

eliemichel:
This is a response to #2446 which adds support for glsl files whose extension is either `.vert.glsl`, `.frag.glsl` or `.comp.glsl`.


teoxoy:
LGTM but CI is unhappy with the addition of the 2 new snapshots since we don't use the cli for those.

You can either remove the 2 snapshots or add support for the new extensions to the snapshot processing code.

PR #2463, opened 2023-09-05
[wgsl-in] Allow sign() to take int argument

### LINKED ISSUE #2458: `sign()` incorrectly does not accept i32 arguments (WGSL)
I believe the issue is right here: 

https://github.com/gfx-rs/naga/blob/6aca2e2b5ec1ac634e3e06e9a7d6113290070234/src/valid/expression.rs#L979

Sign can accept i32s per the WGSL spec: https://www.w3.org/TR/WGSL/#sign-builtin

fornwall:
Fixes #2458.

The main work here is to make it work on metal, which does not have an `int` overload of `sign()`. So we introduce a `naga_isign` polyfill function there.

Currently the metal backend always emits the `naga_isign` polyfill. What is a good way to only do that if necessary? We need to declare the function before it's used, so we can't do it after writing out the functions. Is there some mechanism to visit all expressions without duplicating a lot of logic?


ErichDonGubler (commented):
This mostly LGTM (noting that I don't have merge authority), minus nits and one question: would it be possible to omit this new polyfill unless needed? I'm not familiar with the matter, but I know we've thought about lazily emitting supporting polyfills WRT constructing array members of HLSL structures before.

☝🏻I think @jimblandy or @teoxoy would know the answer to this.

src/back/msl/writer.rs
```diff
@@ -34,6 +34,7 @@ const RAY_QUERY_FUN_MAP_INTERSECTION: &str = "_map_intersection_type";
 
 pub(crate) const MODF_FUNCTION: &str = "naga_modf";
 pub(crate) const FREXP_FUNCTION: &str = "naga_frexp";
+pub(crate) const NAGA_ISIGN_FUNCTION: &str = "naga_isign";
```
ErichDonGubler, line 37: nitpick: It would be more consistent with surrounding code to omit the `NAGA_*` prefix here:

```suggestion
pub(crate) const ISIGN_FUNCTION: &str = "naga_isign";
```

src/back/msl/writer.rs
```diff
@@ -4103,6 +4116,17 @@ impl<W: Write> Writer<W> {
         Ok(info)
     }
 
+    fn write_polyfills(&mut self) -> Result<(), Error> {
+        writeln!(self.out)?;
+        writeln!(
```
ErichDonGubler, line 4121: thought(non-blocking): It'd be nice to avoid two writes and roll the implicit `'\n'` into the next call. I'm not sure what other chunks of the codebase do off the top of my head, though.

src/valid/expression.rs
```diff
@@ -992,6 +991,22 @@ impl super::Validator {
                             _ => return Err(ExpressionError::InvalidArgumentType(fun, 0, arg)),
                         }
                     }
+                    Mf::Sign => {
+                        if arg1_ty.is_some() | arg2_ty.is_some() | arg3_ty.is_some() {
```
ErichDonGubler, line 995: question(blocking): Is there a reason we're using the bit-or operation (`|`) instead of the short-circuiting version (`||`)?

```suggestion
                        if arg1_ty.is_some() || arg2_ty.is_some() || arg3_ty.is_some() {
```

ErichDonGubler (commented):
This mostly LGTM (noting that I don't have merge authority), minus nits and one question: ~~would it be possible to omit this new polyfill unless needed? I'm not familiar with the matter, but I know we've thought about lazily emitting supporting polyfills WRT constructing array members of HLSL structures before.~~ EDIT: 😅 I see that you're also asking this question in the OP:

> Currently the metal backend always emits the naga_isign polyfill. What is a good way to only do that if necessary? We need to declare the function before it's used, so we can't do it after writing out the functions. Is there some mechanism to visit all expressions without duplicating a lot of logic?

☝🏻I think @jimblandy or @teoxoy would know the answer to this.

src/back/msl/writer.rs
fornwall: Indeed, thanks!

src/valid/expression.rs
fornwall: Thanks! There was no reason, created https://github.com/gfx-rs/naga/pull/2465 to fix this in other existing places in the same file.

src/back/msl/writer.rs
fornwall: Thanks! While it doesn't seem to be the existing practice, I've done the change to avoid the extra call now - let me know if I should revert that.

src/back/msl/writer.rs
```diff
@@ -1713,7 +1715,17 @@ impl<W: Write> Writer<W> {
                     Mf::Reflect => "reflect",
                     Mf::Refract => "refract",
                     // computational
-                    Mf::Sign => "sign",
+                    Mf::Sign => match arg_type {
```
jimblandy, line 1718: You should use `arg_type.scalar_kind()` here.

jimblandy (changes requested):
Rather than generating a definition for a polyfill, consider simply emitting the code inline. See the way the Metal backend handles `MathFunction::Dot` for an example:

- It calls `back::msl::Writer::put_dot_product` to emit the expression at the point of use.
- In `back::msl::Writer::update_expressions_to_bake`, we ensure that the operand is placed in a temporary variable first, so that we don't evaluate the argument expression more than once.

src/back/msl/writer.rs
fornwall: Thanks! Done in d16e605aa6f58043cef24b6cb8204bcf0dfb3cab

tests/in/math-functions.wgsl
```diff
@@ -8,6 +8,7 @@ fn main() {
     let d = radians(v);
     let e = saturate(v);
     let g = refract(v, v, f);
+    let h = sign(-1);
```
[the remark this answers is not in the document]
fornwall: Done in f6a2a918d48176cf255fd52b2ed85686950b07c3

fornwall:
> Rather than generating a definition for a polyfill, consider simply emitting the code inline. 

Thanks for the pointers @jimblandy! Changed to emitting inline in 32e96a560a39f8a90b18e8daa28d7ff7fc480d5e.

jimblandy (approved):
Thank you very much!

PR #2468, opened 2023-09-11
cli: Add `dump-before-validation` option.

jimblandy:
It's useful to be able to dump the IR before it's validated, so you can see what the validator is complaining about.

### DIFF
--- a/cli/src/bin/naga.rs
+++ b/cli/src/bin/naga.rs
@@ -83,6 +83,10 @@ struct Args {
     #[argh(switch)]
     version: bool,
 
+    /// dump IR before validation
+    #[argh(option)]
+    dump_before_validation: Option<String>,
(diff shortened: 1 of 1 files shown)


teoxoy:
From experience, I don't remember successful validation being a requirement to dump the module in a `.txt` file. I double checked the code and it seems to confirm this.

jimblandy:
oops, checking

jimblandy:
> From experience, I don't remember successful validation being a requirement to dump the module in a `.txt` file. I double checked the code and it seems to confirm this.

You're right - I always mis-read that code as returning early in case of error, but it doesn't if you've asked for output.

jimblandy:
This may not be the first time I've tried to write this PR :sweat_smile:

PR #2469, opened 2023-09-12
[wgsl-out] Don't include type parameter in splat expressions.

jimblandy:
Generate `vec4(1.0)` instead of `vec4<f32>(1.0)`, since WGSL requires us to infer the type parameter anyway.

This is factored out from `const-expressions-2`. Doing so removes a lot of extraneous changes from the snapshots.


teoxoy (approved):
LGTM

PR #2472, opened 2023-09-12
Implement module compaction

jimblandy:
Add a new Naga feature, `"compact"`, which adds a new function `naga::compact::compact`, which removes unused expressions, types, and constants from a `Module`.


jimblandy:
This should be easier to review commit by commit.

teoxoy:
Can we move the cli commits to another PR? They don't seem to be required for module compaction.

teoxoy (approved):
This is some nice machinery!

I was wondering if we can move the type/expression tracing in the analyzer so that we can always generate the usage metadata.
The backends will then use the metadata for naming expressions and (unnamed) types instead of having to mutate the IR.
Can be done later but do let me know what you think!

jimblandy:
> I was wondering if we can move the type/expression tracing in the analyzer so that we can always generate the usage metadata. The backends will then use the metadata for naming expressions and (unnamed) types instead of having to mutate the IR.

Oh, in the sense that the relocated handle indices are actually the names we want to use in output, so we can skip the compaction step. That makes sense.

I think there's some advantage to mutating the IR, though. Ideally the tests should use the backend code exactly the same way applications do. Mutation lets the backends remain ignorant of whether we performed compaction or not.

I think there are ideas to pursue here, but let's land this as-is for now.

teoxoy:
> Ideally the tests should use the backend code exactly the same way applications do.

I was proposing to always (without it being gated behind a feature flag) create the metadata and use it in the backends.
I was assuming the perf impact would be negligent but I guess we won't know unless we try.

PR #2474, opened 2023-09-13
Remove useless calls to `into_iter()`.

waywardmonkeys:
The `splice()` call takes `IntoIterator`, so calling `into_iter()` is not needed.


teoxoy (approved):
Makes sense!

PR #2477, opened 2023-09-14
clippy: use derived_hash_with_manual_eq.

waywardmonkeys:
This was renamed some time ago.


waywardmonkeys:
ah well, not that long ago.

PR #2480, opened 2023-09-15
Make `valid::expressions` module depend on "validate" feature.

jimblandy:
Rather than having `#[cfg(feature = "validate")]` attributes on almost every definition in `src/valid/expressions.rs`, move the definitions we need unconditionally out of the module, and just put the whole module under a `#[cfg]`.

### DIFF
--- a/src/valid/mod.rs
+++ b/src/valid/mod.rs
@@ -4,6 +4,7 @@ Shader validator.
 
 mod analyzer;
 mod compose;
+#[cfg(feature = "validate")]
 mod expression;
 mod function;
 mod handles;
@@ -12,7 +13,7 @@ mod r#type;
 
 use crate::{
     arena::Handle,
-    proc::{LayoutError, Layouter, TypeResolution},
+    proc::{IndexableLengthError, LayoutError, Layouter, ResolveError, TypeResolution},
     FastHashSet,
 };
 use bit_set::BitSet;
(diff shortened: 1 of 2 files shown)


src/valid/mod.rs
```diff
@@ -243,6 +243,121 @@ pub enum ValidationError {
     Corrupted,
 }
 
+#[derive(Clone, Debug, thiserror::Error)]
+#[cfg_attr(test, derive(PartialEq))]
+pub enum ExpressionError {
```
teoxoy, line 248: Do we have to move this out? Can we make it conditional on the feature flag instead?

jimblandy: The problem with making it conditional is that that changes the API of the validator. We want turning off `"validate"` to simply make producing `ModuleInfo` faster, without changing its API.

jimblandy: For example, clients that happen to be checking for particular `ExpressionError` variants shouldn't break.

Hmm. Maybe it's *preferable* for any users checking for `FunctionError::Expression` to get a compilation error if the `"validation"` feature is off, since that error will never be generated and users shouldn't be thinking it could happen?

API stability and deliberate API instability both seem to have arguments in their favor. Which do you think is more useful?

teoxoy: I think it's natural for feature flags to expose more of the API.

For users that turn on the flag, all enum variants of `ValidationError` will be present, for those who don't they may or may not be there but I don't think those users would exhaustively match the enum anyway.

cwfitzgerald:

jimblandy:
Since validation is no longer behind a feature, this PR is irrelevant.

PR #2481, opened 2023-09-15
[wgsl-in] make RuntimeExpressionContext::local_table a shared ref.

jimblandy:
The `front::wgsl::lowerer::RuntimeExpressionContext::local_table` field does not need to be a mutable reference, as expressions never introduce new local bindings.

### DIFF
--- a/src/front/wgsl/lower/mod.rs
+++ b/src/front/wgsl/lower/mod.rs
@@ -79,10 +79,17 @@ pub struct StatementContext<'source, 'temp, 'out> {
     /// `Handle`s we have built for them, owned by `Lowerer::lower`.
     globals: &'temp mut FastHashMap<&'source str, LoweredGlobalDecl>,
 
-    /// A map from `ast::Local` handles to the Naga expressions we've built for them.
+    /// A map from each `ast::Local` handle to the Naga expression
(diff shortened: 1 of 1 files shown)


teoxoy (approved):
Besides the suggestion, LGTM!

src/front/wgsl/lower/mod.rs
```diff
@@ -167,7 +174,16 @@ impl<'a, 'temp> StatementContext<'a, 'temp, '_> {
...
-    local_table: &'temp mut FastHashMap<Handle<ast::Local>, TypedExpression>,
+    /// A map from [`ast::Local`] handles to the Naga expressions we've built for them.
+    ///
+    /// This is always [`StatementContext::local_table`] for the
+    /// enclosing statement; see that documentation for details.
+    ///
+    /// [`ast::Local`]: super::parse::ast::Local
+    /// [`local_table`]: StatementContext::local_table
+    /// [`LocalVariable`]: crate::Expression::LocalVariable
+    /// [`FunctionArgument`]: crate::Expression::FunctionArgument
```
teoxoy, line 185:
```suggestion
```

I don't think these are needed.

PR #2484, opened 2023-09-17
Prevent using multiple push constant variables in one entry point.

andriyDev:
Previously, using multiple push constant variables was undefined behavior (in wgpu, this resulted in https://github.com/gfx-rs/wgpu/issues/4099). Now, multiple push constant variables per entry point is a validation error!

### DIFF
--- a/src/valid/interface.rs
+++ b/src/valid/interface.rs
@@ -90,6 +90,8 @@ pub enum EntryPointError {
     ForbiddenStageOperations,
     #[error("Global variable {0:?} is used incorrectly as {1:?}")]
     InvalidGlobalUsage(Handle<crate::GlobalVariable>, GlobalUse),
+    #[error("More than 1 push constant variable is used")]
+    MoreThanOnePushConstantUsed,
     #[error("Bindings for {0:?} conflict with other resource")]
     BindingCollision(Handle<crate::GlobalVariable>),
     #[error("Argument {0} varying error")]
@@ -699,6 +701,23 @@ impl super::Validator {
 
         for bg in self.bind_group_masks.iter_mut() {
             bg.clear();
+        }
+
+        #[cfg(feature = "validate")]
+        {
+            let used_push_constants = module
+                .global_variables
+                .iter()
+                .filter(|&(_, var)| var.space == crate::AddressSpace::PushConstant)
+                .map(|(handle, _)| handle)
+                .filter(|&handle| !info[handle].is_empty());
+            // Check if there is more than one push constant, and error if so.
+            // Use a loop for when returning multiple errors is supported.
+            #[allow(clippy::never_loop)]
+            for handle in used_push_constants.skip(1) {
+                return Err(EntryPointError::MoreThanOnePushConstantUsed
+                    .with_span_handle(handle, &module.global_variables));
+            }
         }
 
         #[cfg(feature = "validate")]


src/valid/interface.rs
```diff
@@ -701,6 +703,28 @@ impl super::Validator {
...
+        {
+            let mut used_push_constants = module
+                .global_variables
+                .iter()
+                .filter(|&(_, var)| var.space == crate::AddressSpace::PushConstant)
+                .map(|(handle, _)| handle)
+                .filter(|&handle| !info[handle].is_empty());
+            // Check if there is one push constant. Some iterators can restart iteration after
+            // `next` returns None, so don't continue if the iterator returns None.
+            if used_push_constants.next().is_some() {
```
teoxoy, line 716: Can we instead use [`.skip(1)`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip)? And replace the match below with just a for loop? In the future we'll want to keep accumulating as many errors as possible.

src/valid/interface.rs
```diff
@@ -90,6 +90,8 @@ pub enum EntryPointError {
     ForbiddenStageOperations,
     #[error("Global variable {0:?} is used incorrectly as {1:?}")]
     InvalidGlobalUsage(Handle<crate::GlobalVariable>, GlobalUse),
+    #[error("More than 1 push constant variable is used")]
+    MoreThanOnePushConstantUsed(Handle<crate::GlobalVariable>),
```
teoxoy, line 94:
```suggestion
    #[error("More than 1 push constant variable is used")]
    MoreThanOnePushConstantUsed,
```

andriyDev: Done.

src/valid/interface.rs
andriyDev: That is much cleaner, thank you!

src/valid/interface.rs
```diff
@@ -701,6 +703,21 @@ impl super::Validator {
...
+        #[cfg(feature = "validate")]
+        {
+            let used_push_constants = module
+                .global_variables
+                .iter()
+                .filter(|&(_, var)| var.space == crate::AddressSpace::PushConstant)
+                .map(|(handle, _)| handle)
+                .filter(|&handle| !info[handle].is_empty());
+            // Check if there is more than one push constant, and error if so.
+            for handle in used_push_constants.skip(1) {
```
andriyDev, line 715: There's a lint here that the loop never iterates. Should I add a deny for that lint here?

cwfitzgerald: Do if let Some(handle) = iter.next()

cwfitzgerald: Though honestly, it might be clearer to write the whole thing as a loop with early returns than an iterator.

andriyDev: Done.

src/valid/interface.rs
andriyDev: Ah actually, I couldn't use a loop else clippy gets upset. I just used `if let` to handle it.

src/valid/interface.rs
andriyDev: Oh looks like it doesn't like that either, because it prefers `.nth()`. This works now, but as @teoxoy mentioned, in the future we'll want to support accumulating errors.

src/valid/interface.rs
andriyDev: NEVERMIND, I just disabled the lint for the never loop temporarily and added a comment about it.

teoxoy (approved):
LGTM

PR #2485, opened 2023-09-18
ConstantEvaluator::swizzle: Handle vector concatenation and indexing

jimblandy:


src/proc/constant_evaluator.rs
```diff
@@ -305,20 +307,31 @@ impl ConstantEvaluator<'_> {
...
+            Expression::Compose { ty, ref components } => {
                 let dst_ty = get_dst_ty(ty)?;
 
-                let components = pattern
+                let mut flattened = [src_constant; 4]; // dummy value
+                let len = self
+                    .flatten_compose(ty, components)
+                    .zip(flattened.iter_mut())
+                    .map(|(component, elt)| *elt = component)
+                    .count();
```
teoxoy, line 318: Very interesting use of iterators - haven't seen/used them this way before.

src/proc/constant_evaluator.rs
```diff
@@ -827,6 +839,53 @@ impl ConstantEvaluator<'_> {
...
+            .flat_map(move |component| {
+                if let (
+                    true,
+                    &Expression::Compose {
+                        ty: _,
+                        components: ref subcomponents,
+                    },
+                ) = (is_vector, &self.expressions[*component])
+                {
+                    subcomponents
```
teoxoy, line 881: We might have to go one level deeper since you could have a `vec4(vec3(vec2(0, 0), 0), 0)`.

teoxoy (changes requested):
Thanks for fixing this! I have one concern.

src/proc/constant_evaluator.rs
jimblandy: Oh, good point!

jimblandy:
d2468a1 addresses the two-level nesting, but we run into the `OpConstantComposite` problem I mentioned in my [other review](https://github.com/gfx-rs/naga/pull/2309#pullrequestreview-1630016692). I'll work on that for a bit.

jimblandy:
This PR needs to be squashed.

jimblandy:
But it should be ready for review.

src/proc/constant_evaluator.rs
jimblandy: Yeah, `zip` has the perfect "shorter of the two" behavior. And `count` forces exhaustion of the iterator. So it's cute.

In general "cute" is not a good thing - but I think it's fine when its impact is strictly local.

teoxoy (approved):
Looks great! Thanks for also fixing the SPIR-V backend!

PR #2486, opened 2023-09-18
[msl-out] Add and fix min version checks

teoxoy:
Seems to have been added in https://github.com/gfx-rs/naga/pull/1711 but there is no mention of this requirement in the [MSL spec](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf). I also checked previous versions on the wayback machine.

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -4010,6 +4010,65 @@ impl<W: Write> Writer<W> {
                     }
                 }
 
+                // Check min MSL version for binding arrays
+                match var.space {
+                    crate::AddressSpace::Handle => match module.types[var.ty].inner {
+                        crate::TypeInner::BindingArray { base, .. } => {
+                            match module.types[base].inner {
+                                crate::TypeInner::Sampler { .. } => {
+                                    if options.lang_version < (2, 0) {
+                                        return Err(Error::UnsupportedArrayOf(
+                                            "samplers".to_string(),
+                                        ));
+                                    }
+                                }
+                                crate::TypeInner::Image { class, .. } => match class {
+                                    crate::ImageClass::Sampled { .. }
+                                    | crate::ImageClass::Depth { .. }
+                                    | crate::ImageClass::Storage {
+                                        access: crate::StorageAccess::LOAD,
+                                        ..
+                                    } => {
+                                        // Array of textures since:
+                                        // - iOS: Metal 1.2 (check depends on https://github.com/gfx-rs/naga/issues/2164)
+                                        // - macOS: Metal 2
+
+                                        if options.lang_version < (2, 0) {
+                                            return Err(Error::UnsupportedArrayOf(
+                                                "textures".to_string(),
+                                            ));
+                                        }
+                                    }
+                                    crate::ImageClass::Storage {
+                                        access: crate::StorageAccess::STORE,
+                                        ..
+                                    } => {
+                                        // Array of write-only textures since:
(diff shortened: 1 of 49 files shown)


tests/in/access.param.ron
```diff
@@ -5,7 +5,7 @@
 		adjust_coordinate_space: false,
 	),
 	msl: (
-		lang_version: (2, 0),
+		lang_version: (1, 2),
```
jimblandy, line 8: Here and elsewhere: why does this require 1.2?

jimblandy (changes requested):

tests/in/access.param.ron
teoxoy: - `access`
  We seem to be unable to use a `read_write` storage buffer in fragment stage prior to 1.2 ([see validation errors](https://github.com/gfx-rs/naga/actions/runs/6226374483/job/16898775262)).

  https://github.com/gfx-rs/naga/blob/54df410b949b36ef4fecc5a5e3949c77bb849b27/tests/in/access.wgsl#L145-L149

  This is not documented in the MSL spec but is mentioned [here](https://developer.apple.com/library/archive/documentation/Miscellaneous/Conceptual/MetalProgrammingGuide/WhatsNewiniOS10tvOS10andOSX1012/WhatsNewiniOS10tvOS10andOSX1012.html) (look for section "Function Buffer Read-Writes") that this became available in a feature sets that are also the first to support MSL 1.2.
- `bounds-check-image-restrict` & `bounds-check-image-rzsw`
  We seem to be unable to use `write` textures in fragment stage prior to 1.2 ([see validation errors](https://github.com/gfx-rs/naga/actions/runs/6226645516/job/16899569509)).
  This is also not documented in the MSL spec but is mentioned [here (same link as above)](https://developer.apple.com/library/archive/documentation/Miscellaneous/Conceptual/MetalProgrammingGuide/WhatsNewiniOS10tvOS10andOSX1012/WhatsNewiniOS10tvOS10andOSX1012.html) (look for section "Function Texture Read-Writes") that this became available in a feature set that is also the first to support MSL 1.2.
- `dualsource`
  We have validation for this and is also mentioned in the MSL spec.
  https://github.com/gfx-rs/naga/blob/54df410b949b36ef4fecc5a5e3949c77bb849b27/src/back/msl/mod.rs#L255

teoxoy: Added the undocumented checks in c96fcd3e0fadb49e5e58ef243d5c073f7328fa3f.

teoxoy:
That should be the last check (at least for functionality covered by our snapshots) for MSL version > 1.

teoxoy:
All commits are self-contained and passed CI individually but that's no longer obvious due to the rebase.

src/back/msl/writer.rs
```diff
@@ -3086,33 +3112,28 @@ impl<W: Write> Writer<W> {
...
+        for (_, ty) in module.types.iter() {
+            match ty.inner {
+                crate::TypeInner::AccelerationStructure | crate::TypeInner::RayQuery
+                    if options.lang_version < (2, 4) =>
+                {
+                    return Err(Error::UnsupportedRayTracing);
+                }
+                crate::TypeInner::RayQuery => {
+                    self.put_ray_query_type()?;
+                    break;
```
jimblandy, line 3124: So, I can see that this will only call `put_ray_query_type` if the version check has passed, and thus it's not necessary to check for `AccelerationStructure` any more. But this is kind of tangled; it worries me enough that I'll push a fixup.

jimblandy (approved):
Looks great.

PR #2493, opened 2023-09-20
ConstantEvaluator: Don't ignore the return value of check_and_get.

jimblandy:
This addresses [this review comment](https://github.com/gfx-rs/naga/pull/2309#discussion_r1328023956).


jimblandy:
rebased

teoxoy (commented):
I think why I haven't done this is because it should be fine that compose and splat reference constants.

I was using `check_and_get` more like just a `check` (no need to `get`).
Although `check_and_get` always copies the expressions from the const-expressions arena...

I will try to get the const tracking working tomorrow and see how it fits into the picture. Adding an additional function for just `check` and using it for compose and splat without having to copy additional expressions sounds like another solution.

Let me know what you think!

jimblandy:
> I think why I haven't done this is because it should be fine that compose and splat reference constants.

Oh. I just looked at what `check_and_get` was doing and assumed it shouldn't be copying stuff into our arena unless it mattered somehow. I was never clear on exactly why it mattered.

jimblandy:
> Adding an additional function for just check and using it for compose and splat without having to copy additional expressions sounds like another solution.

I think that would have been less confusing for me.

teoxoy:
Closing as it has been addressed by https://github.com/gfx-rs/naga/pull/2309/commits/0d536b998dbfe2eb5a14167a365ba4fbee25d02d.

PR #2495, opened 2023-09-20
Use fast hashing functions with `IndexMap` and `IndexSet` everywhere.

jimblandy:
Various spots in WGSL are using plain `IndexMap` and `IndexSet` from the `indexmap` crate, rather than a variant that uses the same faster hash function as the rest of Naga. Change these sites to all use the faster hash function.

### DIFF
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -5,8 +5,7 @@ use std::{cmp::Ordering, fmt, hash, marker::PhantomData, num::NonZeroU32, ops};
 /// the same size and representation as `Handle<T>`.
 type Index = NonZeroU32;
 
-use crate::Span;
-use indexmap::set::IndexSet;
+use crate::{FastIndexSet, Span};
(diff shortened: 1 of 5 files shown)


ErichDonGubler (approved):
LGTM from a code perspective, but I have a question that I think needs to be answered before merging: where are the benchmarks? How can I be sure this is actually making things faster?

jimblandy:
My intent here is not actually speed, but consistency. We have `FastHashMap` and `FastHashSet`, so it's odd to randomly use other hash functions depending on who wrote the code.

teoxoy (approved):

jimblandy:
> LGTM from a code perspective, but I have a question that I think needs to be answered before merging: where are the benchmarks? How can I be sure this is actually making things faster?

This is a good point. Filed #2499 (and naturally #2498).

PR #2497, opened 2023-09-20
Test that only constant expressions are hoisted to initializers.

jimblandy:
See comments in test for details.


teoxoy (approved):
Great stuff!

PR #2501, opened 2023-09-24
snapshots: Consolidate path handling.

jimblandy:
Add a new type, `Input`, representing a particular Naga input file, with methods to generate related paths - output files with a particular extension, parameter files, and the input file itself. Use this throughout `snapshots.rs` to generate paths.

Give `Input` utility methods for reading and writing files.

Each commit in this PR should pass CI independently. Some commits have a lot of pure renames, so reviewing commits separately is probably easiest.


teoxoy (approved):
Good stuff!

teoxoy:
Seems like this needs a manual rebase...

PR #2503, opened 2023-09-25
Add `Capabilities::ATOMIC_COMPARE_EXCHANGE_WEAK`.

jimblandy:
We haven't finished implementing WGSL's `atomicCompareExchangeWeak` on all platforms: wgpu#4364. It's valuable to have some configuration of validation that guarantees backends will succeed, and adding a `Capabilities` bit that is included by default but can be turned off accomplishes that.


jimblandy:
Yeah, I don't think this is a good idea after all.

Issue #2151, opened 2022-12-06
Add `binding_array` related `Capabilities`


teoxoy:
- add `Capabilities::TEXTURE_BINDING_ARRAY`
- add `Capabilities::BUFFER_BINDING_ARRAY`
- add `Capabilities::STORAGE_RESOURCE_BINDING_ARRAY` (maybe split this one)

PR #2505, opened 2023-09-25
Require new `Capabilities::BINDING_ARRAY` for `binding_array` type.

jimblandy:
Introduce a new `Capabilities::BINDING_ARRAY` flag, which enables use of the Naga `TypeInner::BindingArray` type.

Adjust tests that exercise binding arrays to pass this flag.

### DIFF
--- a/src/valid/mod.rs
+++ b/src/valid/mod.rs
@@ -114,6 +114,8 @@ bitflags::bitflags! {
         const RAY_QUERY = 0x1000;
         /// Support for generating two sources for blending from fragement shaders
         const DUAL_SOURCE_BLENDING = 0x2000;
+        /// Support for binding arrays.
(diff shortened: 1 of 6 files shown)


src/valid/mod.rs
```diff
@@ -114,6 +114,8 @@ bitflags::bitflags! {
         const RAY_QUERY = 0x1000;
         /// Support for generating two sources for blending from fragement shaders
         const DUAL_SOURCE_BLENDING = 0x2000;
+        /// Support for binding arrays.
+        const BINDING_ARRAY = 0x4000;
```
teoxoy, line 118: This will have to be split out eventually https://github.com/gfx-rs/naga/issues/2151. Should we do it now instead?

cwfitzgerald:

PR #2507, opened 2023-09-25
Disallow ptr to workgroup fn arguments

### LINKED ISSUE #2159: Disallow ptr to workgroup fn arguments
Remove `As::WorkGroup` from here:
https://github.com/gfx-rs/naga/blob/bf4e62b1ac78759705f544f71bf0b43b654a163c/src/valid/type.rs#L175

Additionally, I think we can remove this altogether since right below it, we check if the type can be used as an `ARGUMENT`
https://github.com/gfx-rs/naga/blob/bf4e62b1ac78759705f544f71bf0b43b654a163c/src/valid/function.rs#L888-L903

ref https://github.com/gpuweb/gpuweb/pull/3422

teoxoy:
Resolves https://github.com/gfx-rs/naga/issues/2159.

### DIFF
--- a/src/valid/function.rs
+++ b/src/valid/function.rs
@@ -1001,12 +1001,7 @@ impl super::Validator {
         #[cfg(feature = "validate")]
         for (index, argument) in fun.arguments.iter().enumerate() {
             match module.types[argument.ty].inner.pointer_space() {
-                Some(
-                    crate::AddressSpace::Private
-                    | crate::AddressSpace::Function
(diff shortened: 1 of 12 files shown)


jimblandy (changes requested):
This looks good. We need a test in `wgsl-errors.rs` that these pointers are successfully forbidden.

I guess we don't document pointer argument address space restrictions anywhere right now. Oh well.

jimblandy (approved):
Looks good to me!

Issue #2461, opened 2023-08-31
WGSL backend writes invalid WGSL in some cases


schell:
Given this SPV input
([fragment_equirectangular.spv.zip](https://github.com/gfx-rs/naga/files/12482159/fragment_equirectangular.spv.zip)) naga will write invalid wgsl output. You can see in this generated wgsl source that there is an undefined variable `inf`: 
```wgsl
var<private> global: vec3<f32>;
@group(0) @binding(2) 
var global_1: sampler;
@group(0) @binding(1) 
var global_2: texture_2d<f32>;
var<private> global_3: vec4<f32>;

fn function() {
    var phi_81_: bool;
    var phi_86_: vec3<f32>;

    let _e13 = global;
    let _e21 = (1.0 / sqrt(fma(_e13.z, _e13.z, fma(_e13.x, _e13.x, (_e13.y * _e13.y)))));
    if (bitcast<f32>((bitcast<u32>(_e21) & 2147483647u)) < inf) {
        phi_81_ = (_e21 > 0.0);
    } else {
        phi_81_ = false;
    }
    let _e28 = phi_81_;
    if _e28 {
        phi_86_ = (_e13 * _e21);
    } else {
        phi_86_ = vec3<f32>(0.0, 0.0, 0.0);
    }
    let _e31 = phi_86_;
    let _e40 = textureSample(global_2, global_1, vec2<f32>(fma(atan2(_e31.z, _e31.x), 0.1591, 0.5), fma(asin(_e31.y), 0.3183, 0.5)));
    global_3 = vec4<f32>(_e40.x, _e40.y, _e40.z, 1.0);
    return;
}

@fragment 
fn fragment_equirectangular(@location(0) param: vec3<f32>) -> @location(0) vec4<f32> {
    global = param;
    function();
    let _e3 = global_3;
    return _e3;
}
```
This source fails to parse - it errs with `InvalidHeader`.

Here's my test case:
```rust
        let opts = naga::front::spv::Options::default();
        let module =
            naga::front::spv::parse_u8_slice(include_bytes!("fragment_equirectangular.spv"), &opts)
            .unwrap();
        let mut validator =
            naga::valid::Validator::new(Default::default(), naga::valid::Capabilities::empty());
        let info = validator.validate(&module).unwrap();
        let wgsl =
            naga::back::wgsl::write_string(&module, &info, naga::back::wgsl::WriterFlags::empty())
            .unwrap();
        println!("wgsl: {wgsl}");
[...]

PR #2508, opened 2023-09-25
[valid] Disallow NaNs and infinities

teoxoy:
WGSL doesn't currently allow NaNs and infinities.

Resolves https://github.com/gfx-rs/naga/issues/2461.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -853,7 +853,9 @@ pub enum TypeInner {
 #[cfg_attr(feature = "deserialize", derive(Deserialize))]
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
 pub enum Literal {
+    /// May not be NaN or infinity.
     F64(f64),
+    /// May not be NaN or infinity.
     F32(f32),
     U32(u32),
     I32(i32),
--- a/src/valid/expression.rs
+++ b/src/valid/expression.rs
@@ -124,6 +124,29 @@ pub enum ExpressionError {
     InvalidWorkGroupUniformLoadResultType(Handle<crate::Type>),
     #[error("Shader requires capability {0:?}")]
     MissingCapabilities(super::Capabilities),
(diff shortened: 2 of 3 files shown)


jimblandy (approved):
We should also document that `Literal::F32` and `Literal::F64` must not be nan or infinity.

jimblandy (changes requested):
Sorry - could we also get some coverage in `tests/wgsl-errors.rs`?

teoxoy:
It's not possible to spell out NaNs or infinities in WGSL though.

I did run a manual test here: https://github.com/gfx-rs/naga/issues/2461#issuecomment-1734164109.

> We should also document that `Literal::F32` and `Literal::F64` must not be nan or infinity.

Should we? I keep going back and forth on what's best for IR documentation.
On the one hand we have code in the validator that enforces the invariants of the IR which is usually self-explanatory but sometimes quite involved.
On the other, it would be nice to see at a glance what those invariants are on the IR itself but I worry they will get out of sync or might be too lengthy.

In this case we'll also most likely get an extension + capability which will allow NaNs and infinities.
https://github.com/gpuweb/gpuweb/issues/2259 (SPIR-V has `SPV_KHR_float_controls`; haven't looked into the other languages).

jimblandy:
> It's not possible to spell out NaNs or infinities in WGSL though.

Oh, that's true. If this change were more involved I would want to imitate the unit tests from `analyzer.rs`, but it doesn't seem worth it. Maybe in the future we will be able to implement some coverage requirements!

> On the other, it would be nice to see at a glance what those invariants are on the IR itself but I worry they will get out of sync or might be too lengthy.

Yes, this is a risk. On the balance, though, I think it's really valuable to have the `lib.rs` docs be everything one needs to work with the IR. They're not there yet, but it's not too much trouble to avoid losing ground as we make new changes.

jimblandy (approved):

PR #2509, opened 2023-09-25
snapshots: Improve `convert_wgsl` and `convert_spv` legibility.

jimblandy:
Add a `use Targets::*` declaration to `convert_wgsl` and `convert_spv`, so that each test case fits on a single line, to improve legibility.

I think it's an improvement, but it's obviously a matter of taste. If someone objects, I'm fine with the status quo as well.

### DIFF
--- a/tests/snapshots.rs
+++ b/tests/snapshots.rs
@@ -603,180 +603,68 @@ fn write_output_wgsl(
 fn convert_wgsl() {
     let _ = env_logger::try_init();
 
+    use Targets::*;
     let inputs = [
-        // TODO: merge array-in-ctor and array-in-function-return-type tests after fix HLSL issue https://github.com/gfx-rs/naga/issues/1930
-        (
-            "array-in-ctor",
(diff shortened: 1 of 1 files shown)


teoxoy:
CI is unhappy because you can't "import" associated consts of a struct.

You'd probably have to do this for each target:
```rust
let X = Targets::X;
...
```

---

We could also add those consts to the struct instead:

```rust
const MAIN = Self::SPIRV.bits()
    | Self::METAL.bits()
    | Self::GLSL.bits()
    | Self::HLSL.bits()
    | Self::WGSL.bits();
const DEBUG = Self::IR.bits()
    | Self::ANALYSIS.bits()
    | Self::DOT.bits();
```

Since most snapshots use all the lang targets.

jimblandy:
Did I *really* not even test this before posting the PR? Apparently so.

jimblandy:
Maybe there are better ways to handle all this.

Issue #2234, opened 2023-02-01
with naga-10, from glsl to glsl target, compile error: names containing consecutive underscores are reserved


moyy:
with naga-10, compile glsl code to glsl target, when type name is ends with number, occus compile error

## compile error with glsl source code in vertex shader 

``` txt

Error: 0:38: 'M01__block_1Vetrtex' : identifiers containing two consecutive underscores (__) are reserved as possible future keywords

Error: 0:40: 'M02__block_2Vetrtex' : identifiers containing two consecutive underscores (__) are reserved as possible future keywords

```

## source code

``` glsl

#version 450

precision highp float;

layout(location=0) in vec2 position;

layout(set=0,binding=0) uniform M01 { 
    mat4 project;
    mat4 view;
};

layout(set=2,binding=0) uniform M02 {
    mat4 world;
};

void main() {
        gl_Position = project * view * world * vec4(position.x, position.y, 1.0, 1.0);
}
```

## specs of gles 3.0

By convention, all macro names containing two consecutive underscores ( __ ) are reserved for use by underlying software layers. Defining such a name in a shader does not itself result in an error, but may result in unintended behaviors that stem from having multiple definitions of the same name

moyy:
I found in version 300 es, consecutive underscores within #define should be treated as warning instead of error, don't know whether what I found is right.

As the spec of v3.0 said:

By convention, all macro names containing two consecutive underscores ( __ ) are reserved for use by underlying software layers. Defining such a name in a shader does not itself result in an error, but may result in unintended behaviors that stem from having multiple definitions of the same name

========================

source code:

``` glsl
layout(set = 2, binding = 0) uniform Transform {
    mat4 Model;
};
```

middle representation like this: 

``` glsl
struct Transform_2_0_ {
    // ...
}
```

last code in wgpu-0.14 output: 

``` glsl

struct Transform_2_0__block_1Vertex {
    // ...
}
```

now occus consecutive underscores, cause wasm compile error

error message as follow: identifiers containing two consecutive underscores are reserved as possible future keywords

moyy:
first underscore, I found in test:

![image](https://user-images.githubusercontent.com/25075465/215960745-5b58b194-4d7f-416b-b68f-fb5488de2313.png)


![image](https://user-images.githubusercontent.com/25075465/215960814-b10101c6-c81f-4d95-9923-8efda6f33b87.png)


last underscore in code:

[...]

PR #2510, opened 2023-09-26
Replace double underscore

### LINKED ISSUE #2412: [glsl-out] Rename identifiers containing two consecutive underscores since they are reserved
> In addition, all identifiers containing two consecutive underscores (__) are reserved for use by underlying software layers. Defining such a name in a shader does not itself result in an error, but may result in unintended behaviors that stem from having multiple definitions of the same name.

_from https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.html#keywords_

evahop:
Fixes #2412 and #2234.

Replaces consecutive underscores with a single underscore to comply with `glsl`.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -1129,7 +1129,7 @@ impl<'a, W: Write> Writer<'a, W> {
         let ty_name = &self.names[&NameKey::Type(global.ty)];
         let block_name = format!(
             "{}_block_{}{:?}",
-            ty_name,
+            ty_name.trim_end_matches('_'),
             self.block_id.generate(),
             self.entry_point.stage,
         );
--- a/src/front/type_gen.rs
+++ b/src/front/type_gen.rs
@@ -297,9 +297,7 @@ impl crate::Module {
                 );
 
                 crate::Type {
-                    name: Some(format!(
-                        "__atomic_compare_exchange_result<{kind:?},{width}>"
-                    )),
+                    name: Some(format!("_atomic_compare_exchange_result<{kind:?},{width}>")),
                     inner: crate::TypeInner::Struct {
                         members: vec![
                             crate::StructMember {
@@ -348,7 +346,7 @@ impl crate::Module {
                     (float_ty, width as u32)
                 };
 
-                let mut type_name = "__modf_result_".to_string();
+                let mut type_name = "_modf_result_".to_string();
                 if let Some(size) = size {
                     let _ = write!(type_name, "vec{}_", size as u8);
                 }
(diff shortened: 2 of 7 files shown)


src/front/type_gen.rs
```diff
@@ -426,7 +424,7 @@ impl crate::Module {
                     (float_ty, int_ty, width as u32)
                 };
 
-                let mut type_name = "__frexp_result_".to_string();
+                let mut type_name = "_frexp_result_".to_string();
```
teoxoy, line 427: We need to keep those as they are or else the user might be able to declare these types themselves and they will end up being equivalent.

src/back/glsl/mod.rs
```diff
@@ -1129,7 +1129,7 @@ impl<'a, W: Write> Writer<'a, W> {
         let ty_name = &self.names[&NameKey::Type(global.ty)];
         let block_name = format!(
             "{}_block_{}{:?}",
-            ty_name,
+            ty_name.trim_end_matches('_'),
```
teoxoy, line 1132:
```suggestion
            // avoid double underscores as they are reserved in GLSL
            ty_name.trim_end_matches('_'),
```

src/proc/namer.rs
```diff
@@ -46,6 +46,7 @@ impl Namer {
             .trim_end_matches(SEPARATOR);
 
         let base = if !string.is_empty()
+            && !string.contains("__")
```
teoxoy, line 49: Could you update the docs of this fn with the new rule?

teoxoy (changes requested):
Thanks for the PR!

src/front/type_gen.rs
evahop: Np, do they need to make it to the output of `glsl` also? Because I’m afraid the updates to namer will remove it regardless.

evahop: Also the update to `namer` has an effect on the other backends as well. 
Not sure if that was the desired effect or not (to have consistent names  across all of them). It just seemed like a convenient place to stick it. 

otherwise I could just remove it, and sanitize somewhere on the `glsl` backend only if that seems more appropriate.

There's still a handful of mixed usages of double and single underscore variables on some of the other backends not touched by this PR.

evahop: Np, do they need to make it to the output of glsl also? Because I’m afraid the updates to namer will remove it regardless.

Also the update to namer has an effect on the other backends as well. Not sure if that was a desirable outcome or not. It just seemed like a convenient place to stick it.

There's still a handful of mixed usages of double and single underscore-prefixes on the other backends not touched by this PR... so it's hard to say if it has a negative impact or not.

teoxoy: The namer is meant to be used by all textual backends so I guess it's fine. We can add per backend specific naming rules to it later if we decide it would be better.

> Np, do they need to make it to the output of glsl also? Because I’m afraid the updates to namer will remove it regardless.

Both the predeclared type and the user defined type will get renamed but since they will point to 2 different types in the arena they will get suffixed accordingly.

teoxoy (approved):

PR #2514, opened 2023-09-26
snapshots: convert_wgsl: Remove redundant cubeArrayShadow entry.

jimblandy:
The duplicate entry was introduced accidentally in #1845.


teoxoy:
Can we remove the duplicate one (the one before `binding-arrays`)?. To keep the blame accurate.

teoxoy (approved):

PR #2515, opened 2023-09-26
Disable uniformity analysis for the fragment stage

teoxoy:
Disables uniformity requirements for the following:
- derivative builtin functions
- `textureSample`
- `textureSampleBias`
- `textureSampleCompare`

These are the same builtins `derivative_uniformity` diagnostic targets.

### DIFF
--- a/src/valid/analyzer.rs
+++ b/src/valid/analyzer.rs
@@ -16,6 +16,10 @@ use std::ops;
 
 pub type NonUniformResult = Option<Handle<crate::Expression>>;
 
+// Remove this once we update our uniformity analysis and
+// add support for the `derivative_uniformity` diagnostic
+const DISABLE_UNIFORMITY_REQ_FOR_FRAGMENT_STAGE: bool = true;
+
 bitflags::bitflags! {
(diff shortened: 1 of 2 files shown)


src/valid/analyzer.rs
```diff
@@ -16,15 +16,19 @@ use std::ops;
 
 pub type NonUniformResult = Option<Handle<crate::Expression>>;
 
+// Remove this once we update our uniformity analysis and
+// add support for the `derivative_uniformity` diagnostic
+const DISABLE_UNIFORMITY_REQ_FOR_FRAGMENT_STAGE: bool = true;
```
jimblandy, line 21: This seems like more trouble than it's worth. But it's temporary, so let's go with it.

jimblandy (approved):

src/valid/analyzer.rs
teoxoy: Yeah, I wasn't sure how much of the uniformity analysis code was going to change to get us up to spec, so didn't want to spend too much time on this.

Issue #1464, opened 2021-10-16
Name clash when converting to Metal


almarklein:
Consider the following WGSL shader:
```wgsl
struct VertexInput {
    [[builtin(vertex_index)]] index : u32;
};
struct VertexOutput {
    [[location(0)]] color : vec4<f32>;
    [[builtin(position)]] pos: vec4<f32>;
};

[[stage(vertex)]]
fn vs_main(in: VertexInput) -> VertexOutput {
    var positions = array<vec2<f32>, 3>(vec2<f32>(0.0, -0.5), vec2<f32>(0.5, 0.5), vec2<f32>(-0.5, 0.7));
    let index = i32(in.index);
    let p: vec2<f32> = positions[index];

    var out: VertexOutput;
    out.pos = vec4<f32>(p, 0.0, 1.0);
    out.color = vec4<f32>(p, 0.5, 1.0);
    return out;
}

[[stage(fragment)]]
fn fs_main(in: VertexOutput) -> [[location(0)]] vec4<f32> {
    return in.color;
}
```
This validates successfully and runs fine on Vulkan and DX12. But when running on Metal, this produces an error:

> thread '<unnamed>' panicked at 'Internal { stage: VERTEX, error: "Metal: Compilation failed: \n\nprogram_source:xx:y: error: redefinition of 'index' with a different type: 'int' vs 'metal::uint' (aka 'unsigned int')\n int index = static_cast<int>(in.index);

When I use Naga from the command-line to convert to a metal shader I get:
```metal
// language: metal1.1
#include <metal_stdlib>
#include <simd/simd.h>

struct VertexInput {
    metal::uint index;
};
struct VertexOutput {
    metal::float4 color;
    metal::float4 pos;
};
struct type3 {
    metal::float2 inner[3];
};

struct vs_mainInput {
};
struct vs_mainOutput {
    metal::float4 color [[user(loc0), center_perspective]];
    metal::float4 pos [[position]];
};
vertex vs_mainOutput vs_main(
  metal::uint index [[vertex_id]]   // <------
) {
    const VertexInput in = { index };
    type3 positions;
    VertexOutput out;
    for(int _i=0; _i<3; ++_i) positions.inner[_i] = type3 {metal::float2(0.0, -0.5), metal::float2(0.5, 0.5), metal::float2(-0.5, 0.699999988079071)}.inner[_i];
    int index = static_cast<int>(in.index);   // <-----
    metal::float2 p = positions.inner[index];
[...]

PR #2521, opened 2023-09-28
[msl-out] Make varyings' struct members unique

### LINKED ISSUE #1490: [metal-out] Duplicate member in input structs
I don't have access to any macos hardware, but one of my users reported this error:
```
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    Internal error in VERTEX shader: Metal: Compilation failed: 

program_source:71:19: error: duplicate member 'position'
    metal::float2 position [[attribute(1)]];
                  ^
program_source:70:19: note: previous declaration is here
    metal::float2 position [[attribute(0)]];
                  ^
program_source:84:3: error: type 'vs_mainInput' is not valid for attribute 'stage_in'
  vs_mainInput varyings [[stage_in]]
  ^                       ~~~~
```

I'm suspecting that the shader that causes this looks like this:
```rust
struct Vertex {
    [[location(0)]] position: vec2<f32>;
};

struct NoteInstance{
    [[location(1)]] position: vec2<f32>;
    [[location(2)]] size: vec2<f32>;
    [[location(3)]] color: vec3<f32>;
    [[location(4)]] radius: f32;
};

struct VertexOutput {

};

[[stage(vertex)]]
fn vs_main(vertex: Vertex, note: NoteInstance) -> VertexOutput {
    var out: VertexOutput;
    return out;
}
```
Naga outputs this metal code for it:
```metal
struct vs_mainInput {
    metal::float2 position [[attribute(0)]];
    metal::float2 position [[attribute(1)]];
    metal::float2 size [[attribute(2)]];
    metal::float3 color [[attribute(3)]];
    float radius [[attribute(4)]];
};
```
As we can see there are two `position` fields in this output, and that causes an error

[full-shader-input-and-output.zip](https://github.com/gfx-rs/naga/files/7425749/full-shader-input-and-output.zip)

---
Kinda similar issue: https://github.com/gfx-rs/naga/issues/1464

evahop:
Fixes #1490

Naming conflicts can occur when multiple structs get flattened into one varyings struct.
This moves the `namer.call` invocation into the flattening routine so that each member gets a unique name.

### DIFF
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -3713,16 +3713,14 @@ impl<W: Write> Writer<W> {
                                 member.ty,
                                 member.binding.as_ref(),
                             ));
-                            let name_key = &NameKey::StructMember(arg.ty, member_index);
-                            flattened_member_names.insert(
-                                NameKey::StructMember(arg.ty, member_index),
-                                match member.binding {
-                                    Some(crate::Binding::Location { .. }) => {
-                                        varyings_namer.call(&self.names[name_key])
-                                    }
-                                    _ => self.namer.call(&self.names[name_key]),
-                                },
-                            );
+                            let name_key = NameKey::StructMember(arg.ty, member_index);
+                            let name = match member.binding {
+                                Some(crate::Binding::Location { .. }) => {
+                                    varyings_namer.call(&self.names[&name_key])
+                                }
+                                _ => self.namer.call(&self.names[&name_key]),
+                            };
+                            flattened_member_names.insert(name_key, name);
                         }
                     }
                     _ => flattened_arguments.push((
--- a/src/back/msl/writer.rs
+++ b/src/back/msl/writer.rs
@@ -3689,6 +3689,13 @@ impl<W: Write> Writer<W> {
                 }
             };
 
+            // Since `Namer.reset` wasn't expecting struct members to be
+            // suddenly injected into another namespace like this,
+            // `self.names` doesn't keep them distinct from other variables.
+            // Generate fresh names for these arguments, and remember the
+            // mapping.
+            let mut flattened_member_names = FastHashMap::default();
+
             // List all the Naga `EntryPoint`'s `Function`'s arguments,
             // flattening structs into their members. In Metal, we will pass
             // each of these values to the entry point as a separate argument—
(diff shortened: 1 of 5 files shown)


teoxoy (changes requested):
Thanks for the PR! It looks good but could you add a test for this? The one in the issue would work.

evahop:
Yes, I was going to add that there's also a `namer.namespace` method specifically for structs, It may be more appropriate, but I am not familiar enough with result types. I will try to use that instead, but no promises.

evahop:
Added tests, including one that confirms #1464 has also been fixed (not by me).
Abandoned  `namer.namespace` since `built-ins` don't necessarily end up in the struct, 
I just added a new `namer` solely for the locations.

evahop:
Added snapshot, including one that confirms #1464 has also been fixed (not by me).
Abandoned  `namer.namespace` and just ended up just added a new `namer` solely for the locations. It causes a bit of duplication, since I also stick them in `flattened_member_names` if you want.

evahop:
Added snapshot, which also confirms #1464 has been fixed (not by me).
Abandoned  `namer.namespace` and ended up just added a new `namer` solely for the locations. It causes a bit of duplication, since I also stick them in `flattened_member_names` if you want.

evahop:
Added the snapshot, which also confirms #1464 has been fixed (not by me).
Abandoned  `namer.namespace` and ended up just adding a new `namer` solely for the locations.

src/back/msl/writer.rs
```diff
@@ -3704,6 +3713,16 @@ impl<W: Write> Writer<W> {
...
+                            let name_key = &NameKey::StructMember(arg.ty, member_index);
+                            flattened_member_names.insert(
+                                NameKey::StructMember(arg.ty, member_index),
+                                match member.binding {
+                                    Some(crate::Binding::Location { .. }) => {
+                                        varyings_namer.call(&self.names[name_key])
+                                    }
+                                    _ => self.namer.call(&self.names[name_key]),
+                                },
+                            );
```
teoxoy, line 3725:
```suggestion
                            let name_key = NameKey::StructMember(arg.ty, member_index);
                            let name = match member.binding {
                                Some(crate::Binding::Location { .. }) => {
                                    varyings_namer.call(&self.names[&name_key])
                                }
                                _ => self.namer.call(&self.names[&name_key]),
                            };
                            flattened_member_names.insert(name_key, name);
```

teoxoy:
Seems like https://github.com/gfx-rs/naga/issues/1464 was fixed by https://github.com/gfx-rs/naga/pull/1766.

teoxoy (approved):
Good stuff!

PR #2522, opened 2023-09-28
Let `ConstantEvaluator` see through `Constant` exprs in `Splat` and `Compose` exprs.

jimblandy:
The `#[test]` tests may be redundant. Let me know what you think.


jimblandy:
I decided to open a new PR rather than resurrecting #2493.

teoxoy:
> The `#[test]` tests may be redundant. Let me know what you think.

I think we should rely on snapshots, they are a lot more maintainable.
But let's merge this as is, we can remove them later if they become too much of a burden.

teoxoy (approved):
Looks good!

PR #2525, opened 2023-10-02
Add `Rgb10a2Uint` format

teoxoy:
Also fixes a few usages of the wrong format `Rgb10a2` formats.


ErichDonGubler (approved):
LGTM[^1]! Nice to knock out yet another standards gap. 🙂 Guessing we need a Firefox bug for this, too?

[^1]: …though I don't have merge rights yet; @jimblandy, I guess we can do this, now that I have `wgpu` perms. already?)

PR #2531, opened 2023-10-03
[glsl/spv-out] Cull functions that should not be available for a given stage

### LINKED ISSUE #2524: Shader translation error, no function with name 'fwidth'
When I run a program that renders with wgpu using my intel chipset with mesa driver, on an optimus laptop, I'm getting a crash at the shader compilation stage for a shader that uses the fwidth built-in function.

```
[2023-10-01T15:17:12Z ERROR wgpu_hal::gles::egl] GLES: [ShaderCompiler/Error] ID 1 : 0:38(20): error: no function with name 'fwidth'
[2023-10-01T15:17:12Z INFO  wgpu_hal::gles::device] 	Compiled shader NativeShader(11)
[2023-10-01T15:17:12Z ERROR wgpu::backend::direct] Shader translation error for stage ShaderStages(NONE | VERTEX): 0:38(20): error: no function with name 'fwidth'

[2023-10-01T15:17:12Z ERROR wgpu::backend::direct] Please report it to https://github.com/gfx-rs/naga
[2023-10-01T15:17:12Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
```

According to the wgpu_hal logs, this is using OpenGL ES 3.2 with Mesa 23.1.8, and the shader language version is OpenGL ES GLSL ES 3.20.
The version of naga used is 0.12.2, pulled from wgpu 0.16.1

According to the [GLSL ES 3.20 spec](https://registry.khronos.org/OpenGL/specs/es/3.2/GLSL_ES_Specification_3.20.html#fragment-processing-functions), the fwidth function should be provided as a built-in function.

I'm not sure if this is an issue on naga's side or in mesa.

teoxoy:
fixes https://github.com/gfx-rs/naga/issues/2524

Done it for SPIR-V as well since the ability to separate entry points (https://github.com/gfx-rs/naga/pull/1329) was added to work around driver bugs and I can imagine them misbehaving in this case as well.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -827,6 +827,11 @@ impl<'a, W: Write> Writer<'a, W> {
 
             let fun_info = &self.info[handle];
 
+            // Skip functions that that are not compatible with this entry point's stage
+            if !fun_info.available_stages.contains(ep_info.available_stages) {
+                continue;
+            }
+
             // Write the function
             self.write_function(back::FunctionType::Function(handle), function, fun_info)?;
 
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -827,7 +827,15 @@ impl<'a, W: Write> Writer<'a, W> {
 
             let fun_info = &self.info[handle];
 
-            // Skip functions that that are not compatible with this entry point's stage
+            // Skip functions that that are not compatible with this entry point's stage.
+            //
+            // When validation is enabled, it rejects modules whose entry points try to call
+            // incompatible functions, so if we got this far, then any functions incompatible
+            // with our selected entry point must not be used.
+            //
+            // When validation is disabled, `fun_info.available_stages` is always just
+            // `ShaderStages::all()`, so this will write all functions in the module, and
+            // the downstream GLSL compiler will catch any problems.
             if !fun_info.available_stages.contains(ep_info.available_stages) {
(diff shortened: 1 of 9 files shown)


jimblandy (approved):
Looks good to me, but see if the comments I suggested are right.

src/back/glsl/mod.rs
```diff
@@ -827,6 +827,11 @@ impl<'a, W: Write> Writer<'a, W> {
 
             let fun_info = &self.info[handle];
 
+            // Skip functions that that are not compatible with this entry point's stage
```
jimblandy, line 830: Here and elsewhere:

```suggestion
            // Skip functions that that are not compatible with this entry point's stage.
            // When validation is enabled, we will reject programs that call this, so
            // this will only leave calls dangling when validation is disabled.
```

teoxoy: We will reject programs that call this but even with validation enabled, we will still write the function declaration.

jimblandy: Okay - how about this:

> When validation is disabled, `fun_info.available_stages` is always just `ShaderStages::all()`, so this will write all functions in the module, and the downstream GLSL compiler will catch any problems. When validation is enabled, it rejects modules whose entry points try to call incompatible functions, so if we got this far, then any functions incompatible with our selected entry point must not be used.

Issue #1894, opened 2022-05-09
[wgsl-in] external texture support


jinleili:
WGSL spec: https://gpuweb.github.io/gpuweb/wgsl/#external-texture-type

PR #2534, opened 2023-10-04
Add textureSampleBaseClampToEdge

evahop:
From gfx-rs/wgpu#4402

*For everything excluding spv:*
It creates a local variable containing the half-texel computed from each backend's respective image query function.
It then uses that half-texel to clamp the image coordinates to [half-texel, 1 - half-texel].

*For spv:*
Does the exact same thing only it's self contained.

The spec says at "base level" which I've interpreted to mean at level 0 everywhere.

The spv snapshot is a little noisy so the relevant bit begins on line `584`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1221,7 +1221,6 @@ pub enum SampleLevel {
         x: Handle<Expression>,
         y: Handle<Expression>,
     },
-    Base,
 }
 
 /// Type of an image query.
@@ -1415,6 +1414,7 @@ pub enum Expression {
         offset: Option<Handle<Expression>>,
         level: SampleLevel,
         depth_ref: Option<Handle<Expression>>,
+        clamp_to_edge: bool,
     },
 
     /// Load a texel from an image.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1221,6 +1221,7 @@ pub enum SampleLevel {
         x: Handle<Expression>,
         y: Handle<Expression>,
     },
+    Base,
 }
 
 /// Type of an image query.
--- a/src/back/dot/mod.rs
+++ b/src/back/dot/mod.rs
@@ -452,6 +452,7 @@ fn write_function_expressions(
                 offset: _,
                 level,
                 depth_ref,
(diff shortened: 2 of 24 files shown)


src/lib.rs
```diff
@@ -1214,6 +1214,7 @@ pub enum SampleLevel {
         x: Handle<Expression>,
         y: Handle<Expression>,
     },
+    Base,
```
teoxoy, line 1217: The spec refers to mip level 0 as the base level but we use `Zero` in the IR.
I'd add a `clamp_to_edge` bool on `Expression::ImageSample` instead.

teoxoy (commented):
I haven't yet looked in detail at the PR but wanted to ask if it would be possible to do the translation in the wgsl frontend instead given that none of the other shading languages have this function.

wgsl -> wgsl translation of this fn with a `texture_external` wouldn't work but I don't see the usefulness of wgsl -> wgsl translation in general.

@jimblandy what do you think?

evahop:
Removed `Base` on `SampleLevel` in favor of `clamp_to_edge`. I can update this to be frontend only once decided.

teoxoy:
Thanks!

> Not sure what triggered the review request, that wasn't intentional.

We now have a codeowners file with the naga team in it which causes github to automatically request reviews.

jimblandy:
@evahop Thanks for implementing this!

If you'd like to post work-in-progress PRs (which we're happy to see), consider marking them as "Draft" when you create the PR. That prevents reviews from being requested.

teoxoy:
@jimblandy and I talked about this and we are unsure where this translation belongs yet due to us not supporting `texture_external` yet (https://github.com/gfx-rs/naga/issues/1894). `textureSampleBaseClampToEdge` can be used with a `texture_2d` but its main use-case is with a `texture_external`.

The translation of `texture_external` & `textureSampleBaseClampToEdge` will most likely have to happen in the same place (either in the front-end or the back-ends).

I think we should ideally add support for `texture_external` first or at the same time with `textureSampleBaseClampToEdge`.

cwfitzgerald:

PR #2535, opened 2023-10-05
[wgsl] remove `outerProduct`

teoxoy:
This function has been removed from the WGSL spec (see https://github.com/gpuweb/gpuweb/pull/1290).


src/back/wgsl/writer.rs
```diff
@@ -1568,7 +1567,7 @@ impl<W: Write> Writer<W> {
                     Mf::Unpack2x16snorm => Function::Regular("unpack2x16snorm"),
                     Mf::Unpack2x16unorm => Function::Regular("unpack2x16unorm"),
                     Mf::Unpack2x16float => Function::Regular("unpack2x16float"),
-                    Mf::Inverse => {
+                    Mf::Inverse | Mf::Outer => {
```
ErichDonGubler, line 1570: question: Is there a reason we don't remove `Inverse` and `Outer` from `MathFunction` altogether? I'm assuming it's "only" for the benefit of this diagnostic here, and the fact that other front ends might emit these.

tests/in/glsl/math-functions.frag
```diff
@@ -44,7 +44,8 @@ void main() {
     vec4 minOut = min(a, b);
     vec4 reflectOut = reflect(a, b);
     vec3 crossOut = cross(a.xyz, b.xyz);
-    mat4 outerProductOut = outerProduct(a, b);
+    // TODO: support outerProduct function in wgsl output
```
ErichDonGubler, line 47: question: Would this `TODO`'d support mean we need to write a polyfill?

ErichDonGubler (approved):

tests/in/glsl/math-functions.frag
teoxoy: Yeah, similar to the inverse fn above.

src/back/wgsl/writer.rs
teoxoy: > I'm assuming it's "only" for the benefit of this diagnostic here, and the fact that other front ends might emit these.

Yeah, GLSL and SPIR-V have those so, ideally we'd polyfill them.

PR #2537, opened 2023-10-05
Properly recognize `Literal` expressions as non-dynamic indices.

jimblandy:
Restore `negative_index` test in `tests/wgsl-errors.rs`, as part of the `invalid_arrays` test function.


teoxoy (approved):
👍

PR #2542, opened 2023-10-06
Add StorageFormat::Bgra8Unorm

### LINKED ISSUE #2195: Support `bgra8unorm` texel format
The data written to this format has to be in RGBA order (look for "Texel Formats for Storage Textures" table [here](https://gpuweb.github.io/gpuweb/wgsl/#texel-formats)).

D3D12 swizzles the reads/writes automatically (see https://github.com/gpuweb/gpuweb/pull/3640#issuecomment-1341855296).
Metal also swizzles (see https://github.com/gpuweb/gpuweb/pull/3640#issuecomment-1352453687).

For SPIR-V, use the `Unknown` image format and request the `StorageImageWriteWithoutFormat` capability on writes.
I don't think Vulkan swizzles (or I can't find it does) so we will probably have to swizzle the writes ourselves.

nical:
Fixes #2195 
Blocks: https://github.com/gfx-rs/wgpu/pull/3634

I'm not sure what the best course is for the wgsl backend, so I defaulted to putting the "rgba8" swizzle which is the closest available format. I suppose that the backend would have to insert some extra code to convert when interacting with the storage texture (probably that's already done in other places?).

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -766,7 +766,7 @@ impl<'a, W: Write> Writer<'a, W> {
                             write!(self.out, "binding = {binding}")?;
                         }
                         if let Some((format, _)) = storage_format_access {
-                            let format_str = glsl_storage_format(format);
+                            let format_str = glsl_storage_format(format)?;
                             let separator = match layout_binding {
                                 Some(_) => ",",
                                 None => "",
@@ -1491,7 +1491,7 @@ impl<'a, W: Write> Writer<'a, W> {
                         ..
                     } = this.module.types[arg.ty].inner
                     {
-                        write!(this.out, "layout({}) ", glsl_storage_format(format))?;
+                        write!(this.out, "layout({}) ", glsl_storage_format(format)?)?;
                     }
 
                     // write the type
@@ -4213,10 +4213,10 @@ const fn glsl_dimension(dim: crate::ImageDimension) -> &'static str {
 }
 
 /// Helper function that returns the glsl storage format string of [`StorageFormat`](crate::StorageFormat)
-const fn glsl_storage_format(format: crate::StorageFormat) -> &'static str {
+fn glsl_storage_format(format: crate::StorageFormat) -> Result<&'static str, Error> {
     use crate::StorageFormat as Sf;
 
-    match format {
+    Ok(match format {
         Sf::R8Unorm => "r8",
         Sf::R8Snorm => "r8_snorm",
         Sf::R8Uint => "r8ui",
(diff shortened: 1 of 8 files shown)


nical:
Out of curiosity I looked at tint's solution and they do some substitution between bgra and rgba in their IR, see https://dawn.googlesource.com/tint/+/refs/heads/main/src/tint/lang/core/ir/transform/bgra8unorm_polyfill.cc

teoxoy:
> Out of curiosity I looked at tint's solution and they do some substitution between bgra and rgba in their IR, see https://dawn.googlesource.com/tint/+/refs/heads/main/src/tint/lang/core/ir/transform/bgra8unorm_polyfill.cc

That transform only runs for SPIR-V. The bgra8 -> rgba8 format replacement is not valid, they should use `Unknown` because SPIR-V's rgba8 format [is not compatible](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap52.html#spirvenv-image-formats) with the bgra8 one that we'll use on the API side.

The swizzle I think we probably need but we should probably test first.

> I'm not sure what the best course is for the wgsl backend, so I defaulted to putting the "rgba8" swizzle which is the closest available format. I suppose that the backend would have to insert some extra code to convert when interacting with the storage texture (probably that's already done in other places?).

I assume you mean the GLSL backend, we should probably fail with an unimplemented/unsupported error.

src/back/glsl/mod.rs
```diff
@@ -4238,6 +4238,7 @@ const fn glsl_storage_format(format: crate::StorageFormat) -> &'static str {
         Sf::Rgba8Snorm => "rgba8_snorm",
         Sf::Rgba8Uint => "rgba8ui",
         Sf::Rgba8Sint => "rgba8i",
+        Sf::Bgra8Unorm => "rgba8", // Note: there is no bgra8 or unknow qualifier to pick here.
```
teoxoy, line 4241: We should error here.

teoxoy (changes requested):

src/back/spv/writer.rs
```diff
@@ -1109,8 +1109,11 @@ impl Writer {
                 "64-bit integer storage image format",
                 &[spirv::Capability::Int64ImageEXT],
             ),
-            If::Unknown
-            | If::Rgba32f
+            If::Unknown => self.require_any(
+                "Unknown storage image format",
+                &[spirv::Capability::StorageImageWriteWithoutFormat],
```
teoxoy, line 1114: Can we instead do this in the `write_image_store` fn? It should be more accurate since the `Unknown` format is also used for sampled images.

```rust
match *self.fun_info[image].ty.inner_with(&self.ir_module.types) {
    crate::TypeInner::Image {
        class:
            crate::ImageClass::Storage {
                format: crate::StorageFormat::Bgra8Unorm,
                ..
            },
        ..
    } => self.writer.require_any(
        "Bgra8Unorm storage write",
        &[spirv::Capability::StorageImageWriteWithoutFormat],
    )?,
    _ => {}
}
```

teoxoy (approved):

PR #2548, opened 2023-10-09
[wgsl-out] Generate correct code for bit complement on integers.

jimblandy:
Remove incorrect special case for `UnaryOperator::Not` on vectors.

### DIFF
--- a/src/back/wgsl/writer.rs
+++ b/src/back/wgsl/writer.rs
@@ -1600,13 +1600,10 @@ impl<W: Write> Writer<W> {
                 let unary = match op {
                     crate::UnaryOperator::Negate => "-",
                     crate::UnaryOperator::Not => {
-                        match *func_ctx.resolve_type(expr, &module.types) {
-                            TypeInner::Scalar {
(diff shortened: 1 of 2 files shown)


src/back/wgsl/writer.rs
```diff
@@ -1604,8 +1604,7 @@ impl<W: Write> Writer<W> {
                             TypeInner::Scalar {
                                 kind: crate::ScalarKind::Bool,
                                 ..
-                            }
-                            | TypeInner::Vector { .. } => "!",
+                            } => "!",
```
teoxoy, line 1607: With this change boolean vectors are now using `~` instead of `!`. All backends use the code below (that we could also use here):
https://github.com/gfx-rs/naga/blob/3c7dbc4016b84a35c69b30305b12abaeefc21fd9/src/back/msl/writer.rs#L1539-L1543

Ideally we should just add another variant to `UnaryOperator` for `LogicalNot` and rename `Not` to `BitwiseNot`.

teoxoy (changes requested):

teoxoy (approved):

PR #2563, opened 2023-10-16
Add debugPrintf to wgsl

exrook:
# Moved to https://github.com/gfx-rs/wgpu/pull/4297

We can have printf in shaders :)

Supports wgsl-in, spv-in, spv-out, hlsl-out, glsl-out

Supporting glsl-in would need changes upstream in https://github.com/Kangz/glslpp-rs/ to parse the format strings.

msl has no support for this sort of feature as far as I can tell.

for more info see: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/debug_printf.md

cc: gfx-rs/wgpu#4410


exrook:
I'm disabling the test for hlsl-out since dxc [doesn't support printf when outputting dxil](https://github.com/microsoft/DirectXShaderCompiler/issues/357). I'll keep the hlsl-out feature in since it still does work with fxc and dxc when targeting spirv.

jimblandy:
If I understand our plan correctly, I think this needs a bit in `valid::Capabilities` as well.

jimblandy:
Tragically, Mesa doesn't support `SPV_KHR_non_semantic_info`, so I can't try this out.

exrook:
> Tragically, Mesa doesn't support `SPV_KHR_non_semantic_info`, so I can't try this out.

It's working on my machine with `mesa 23.2.1`  ;) , although I had to explicitly enable printf in the validation layer using the vulkan configurator in order to see output. I think it's possible to programmatically enable as well but I haven't looked into that yet.

I also have a local patch to wgpu that enables `VK_KHR_shader_non_semantic_info` but that doesn't seem to have any effect, my shader compiles & runs regardless of whether it's enabled or not.

> If I understand our plan correctly, I think this needs a bit in `valid::Capabilities` as well.

I'll try to figure this out, currently no other `Statements` interact with Capabilities so it's not immediately obvious to me where the check should end up.

PR #2564, opened 2023-10-16
[glsl-in, glsl-out] Fix warnings about redundant doc link targets.

jimblandy:


jimblandy:
Merging without review because this PR affects only comments.

PR #2565, opened 2023-10-18
[refactor] make use of `resolve_type` wherever possible

teoxoy:
Makes code nicer. The SPIR-V backend is missing a `resolve_type` fn but we can add that later.


jimblandy (approved):
Fantastic!

PR #2567, opened 2023-10-19
Make validation reject 64-bit floating-point literals.

jimblandy:
Make expression validation and constant expression validation reject `Literal` expressions containing `F64` literals unless the `FLOAT64` capability is enabled.


teoxoy (approved):
LGTM

PR #2571, opened 2023-10-19
[wgsl-in] consolidate type resolution logic in a few macros

teoxoy:


src/front/wgsl/lower/mod.rs
```diff
@@ -13,6 +13,32 @@ use crate::{Arena, FastHashMap, FastIndexMap, Handle, Span};
...
+pub(super) use resolve;
+
+macro_rules! resolve_inner {
+    ($ctx:ident, $expr:expr) => {{
+        $ctx.grow_types($expr)?;
+        $ctx.typifier()[$expr].inner_with(&$ctx.module.types)
+    }};
+}
+
+macro_rules! resolve_inner_x2 {
```
jimblandy, line 31: Suggest name: `resolve_inner_binary_op` or `resolve_inner_binary`.

src/front/wgsl/lower/mod.rs
```diff
@@ -1146,11 +1162,7 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
 
                     let initializer = match v.init {
                         Some(init) => {
-                            let initializer =
-                                self.expression(init, ctx.as_expression(block, &mut emitter))?;
-                            ctx.as_expression(block, &mut emitter)
-                                .grow_types(initializer)?;
```
jimblandy, line 1152: Was this call just completely extraneous?

jimblandy (approved):
This is a nice cleanup.

These need documentation, both for usage and their rationale for existence.

src/front/wgsl/lower/mod.rs
teoxoy: It was for the call a few lines below to `resolved_inner`.

PR #2573, opened 2023-10-20
[wgsl-in] Clean up swizzle/component access type.

jimblandy:
Rename `front::wgsl::lower::Composition` to `Components`. It doesn't represent a composition, it represents either a single component being selected from a vector or a swizzle.

Clean up code and helper function names.


teoxoy (approved):
Nice!

PR #2575, opened 2023-10-20
Don't use layout qualifiers to allow for GLSL 140 support

Zoxc:
This avoids emitting layout qualifiers for shader I/O to allow sufficient GLSL 140 support for the wgpu cube example. These must instead of bound by OpenGL code using the new `varying` field of `ReflectionInfo`.

Precision qualifiers are changed to only be emitted for ES.

### DIFF
--- a/src/back/glsl/mod.rs
+++ b/src/back/glsl/mod.rs
@@ -55,6 +55,7 @@ use std::{
     cmp::Ordering,
     fmt,
     fmt::{Error as FmtError, Write},
+    mem,
 };
 use thiserror::Error;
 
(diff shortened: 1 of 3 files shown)


teoxoy (changes requested):
Looks good but can you update the ron files for:

- `interpolate`
- `bounds-check-image-rzsw`
- `bounds-check-image-restrict`

to target ES? I think it wasn't meant for them to target desktop GL.

Zoxc:
They all seem to use non-ES features.

teoxoy:
Ah, I see.

teoxoy (approved):

PR #2577, opened 2023-10-21
[wgsl-in] Refactor construction lowering code.

jimblandy:
(Draft because it depends an a bunch of other stuff.)

- [wgsl-in] Unify ConcreteConstructor and ConcreteConstructorHandle.

  Replace the `ConcreteConstructor` and `ConcreteConstructorHandle`
  types in `front::wgsl::lower::construction` with a single type
  `Constructor` with a type parameter that determines how it refers to
  Naga types.

  In the single-argument vector construction case, construct `Compose`
  expressions when the lengths don't match, and leave the problem for
  validation to catch; `validate_compose` is a better place to invest
  time in improving diagnostics (and arguably producess more specific
  error messages already).

- [wgsl-in] Remove `Components` enum in favor of slice patterns.

  Remove `front::wgsl::lower::construction::Components`, and simply use
  slice patterns to decide which Naga code to generate. Supply the type
  of the first argument, if available, as an `Option`. This is slightly
  redundant, as the `Option` is `Some` if and only if the slice is
  non-empty, but in practice it is not too messy, and we can add match
  arms to check for misalignment.

### DIFF
--- a/src/front/wgsl/lower/construction.rs
+++ b/src/front/wgsl/lower/construction.rs
@@ -6,45 +6,57 @@ use crate::{Handle, Span};
 use crate::front::wgsl::error::Error;
 use crate::front::wgsl::lower::{ExpressionContext, Lowerer};
 
-enum ConcreteConstructorHandle {
-    PartialVector {
-        size: crate::VectorSize,
-    },
+/// A cooked form of `ast::ConstructorType` that uses Naga types whenever
+/// possible.
+enum Constructor<T> {
+    /// A vector construction whose component type is inferred from the
+    /// argument: `vec3(1.0)`.
+    PartialVector { size: crate::VectorSize },
+
+    /// A matrix construction whose component type is inferred from the
+    /// argument: `mat2x2(1,2,3,4)`.
     PartialMatrix {
         columns: crate::VectorSize,
         rows: crate::VectorSize,
     },
+
+    /// An array whose component type and size are inferred from the arguments:
+    /// `array(3,4,5)`.
     PartialArray,
-    Type(Handle<crate::Type>),
+
+    /// A known Naga type.
+    ///
+    /// When we match on this type, we need to see the `TypeInner` here, but at
+    /// the point that we build this value we'll still need mutable access to
+    /// the module later. To avoid borrowing from the module, the type parameter
+    /// `T` is `Handle<Type>` initially. Then we use `borrow_inner` to produce a
+    /// version holding a tuple `(Handle<Type>, &TypeInner)`.
+    Type(T),
 }
 
-impl ConcreteConstructorHandle {
-    fn borrow<'a>(&self, module: &'a crate::Module) -> ConcreteConstructor<'a> {
-        match *self {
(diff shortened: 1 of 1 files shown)


teoxoy (changes requested):
> * [wgsl-in] Unify ConcreteConstructor and ConcreteConstructorHandle.
>   Replace the `ConcreteConstructor` and `ConcreteConstructorHandle`
>   types in `front::wgsl::lower::construction` with a single type
>   `Constructor` with a type parameter that determines how it refers to
>   Naga types.

Nice refactor!

>   In the single-argument vector construction case, construct `Compose`
>   expressions when the lengths don't match, and leave the problem for
>   validation to catch; `validate_compose` is a better place to invest
>   time in improving diagnostics (and arguably producess more specific
>   error messages already).

This used to error with:

```
error: cannot cast a vec3<f32> to a vec2<f32>
   ┌─ wgsl:91:26
   │
91 │     let test = vec2<f32>(vec3<f32>());
   │                          ^^^^^^^^^^ cannot cast a vec3<f32> to a vec2<f32>
```

It now errors with:

```
Naga module validation failed on test 'access.wgsl': WithSpan { inner: Function { handle: [2], name: "test_matrix_within_array_within_struct_accesses", source: Expression { handle: [71], source: Compose(ComponentCount { given: 3, expected: 2 }) } }, spans: [(Span { start: 1431, end: 2264 }, "naga::Function [2]"), (Span { start: 1921, end: 1943 }, "naga::Expression [71]")] }
```

I think we should keep this as it was (at least for now) as there are other cases where you'd run into this same error and it would keep things consistent.

> * [wgsl-in] Remove `Components` enum in favor of slice patterns.
>   Remove `front::wgsl::lower::construction::Components`, and simply use
>   slice patterns to decide which Naga code to generate. Supply the type
>   of the first argument, if available, as an `Option`. This is slightly
>   redundant, as the `Option` is `Some` if and only if the slice is
>   non-empty, but in practice it is not too messy, and we can add match
>   arms to check for misalignment.

The nice thing about the `Components` enum was that it worked around rust's limitation on matching on `Vec`s. With this change, we are matching on slices and then have to copy the components.

I also found it easier to read the code before (even if it was a bit more lengthy) since there are now 3 things we are matching on inside a tuple + the components being a tuple themselves (containing the handle and the span).

jimblandy:
> The nice thing about the `Components` enum was that it worked around rust's limitation on matching on `Vec`s. With this change, we are matching on slices and then have to copy the components.
> 
> I also found it easier to read the code before (even if it was a bit more lengthy) since there are now 3 things we are matching on inside a tuple + the components being a tuple themselves (containing the handle and the span).

Okay - the intent was to make things clearer. I'll try putting it back. There was a reason I introduced the vector copies, related to my abstract types plan, but I'll see if I can keep them out. I do feel like there was room for improvement in the original `Components`, but maybe I didn't find the right balance.

jimblandy:
I ended up removing pretty much any changes to `Components`, so all that's left here is the stuff that I think you liked.

teoxoy (approved):
Thanks, looks good!

PR #2584, opened 2023-10-23
[wgsl-in] Fix expected min arg count of `textureLoad`

### LINKED ISSUE #2583: Wrong number of arguments error is wrong/misleading
**Repro steps**

```wgsl
@group(0) @binding(2) var fx_blend: texture_storage_2d<rgba8unorm, read_write>;

@fragment
fn fs_main(@builtin(position) pos: vec4<f32>) -> @location(0) vec4<f32> {
    return textureLoad(fx_blend, vec2<i32>(pos.xy), 0);
}
```

Expect behavior:

```
error: wrong number of arguments: expected 2, found 3
  ┌─ test.wgsl:5:12
  │
5 │     return textureLoad(fx_blend, vec2<i32>(pos.xy), 0);
  │            ^^^^^^^^^^^ wrong number of arguments
```

Observed behavior:

```
error: wrong number of arguments: expected 3, found 3
  ┌─ test.wgsl:5:12
  │
5 │     return textureLoad(fx_blend, vec2<i32>(pos.xy), 0);
  │            ^^^^^^^^^^^ wrong number of arguments
```

teoxoy:
fixes https://github.com/gfx-rs/naga/issues/2583

Also double-checked the min arg count of the other functions.

It was most likely incorrect due to the spec previously only having overloads with min 3 args but with https://github.com/gpuweb/gpuweb/pull/4326 (which we have parts of implemented already) new overloads with min 2 args have been added.

### DIFF
--- a/src/front/wgsl/lower/mod.rs
+++ b/src/front/wgsl/lower/mod.rs
@@ -2099,7 +2099,7 @@ impl<'source, 'temp> Lowerer<'source, 'temp> {
                             return Ok(None);
                         }
                         "textureLoad" => {
-                            let mut args = ctx.prepare_args(arguments, 3, span);
+                            let mut args = ctx.prepare_args(arguments, 2, span);
 
                             let image = args.next()?;
                             let image_span = ctx.ast_expressions.get_span(image);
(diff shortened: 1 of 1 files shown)


jimblandy (approved):
Looks good to me.

I haven't tried it, but it seems like it would be nicer to compute `min_args`'s final value in one place, as soon as we knew the image type, rather than having it be a side effect of random closures. But that problem was there already, so we don't need to address it here.

